PR #21, opened 2019-05-07
Support for integration tests and sample tests for method stub

fabsx00:
On each build, we want to ensure that all queries we showcase in the documentation still work. Furthermore, we want to provide a test-based spec for the query language, making it easy for people to pick up the language. This PR brings in the necessary boilerplate for this and some first tests for method stubs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ project/project
 project/target
 target
 fuzzyc2cpg
+.idea/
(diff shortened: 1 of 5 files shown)


fabsx00:
(We need to hold off on merging this until the `codepropertgraph` publication to mavencentral works again)

PR #48, opened 2019-05-12
Joern-Server

fabsx00:
This PR introduces a REST server that loads code property graphs and allows querying them. This makes it possible to query cpgs with any programming language. We also get around the long JVM startup time and CPG loading time in joern-query by turning it into a python script, which talks to the REST server.

Since both CPG creation/loading and querying can potentially take long, I've made this API asynchronous, that is, you ask the server to carry out operations and then come back to check for results.

This one is ready for review, although I would still like to add some additional tests before merging.


joern-server/src/main/scala/ScalatraBootstrap.scala
```diff
@@ -0,0 +1,16 @@
+import akka.actor.ActorSystem
+import io.shiftleft.joern.server._
+import org.scalatra._
+import javax.servlet.ServletContext
+
+class ScalatraBootstrap extends LifeCycle {
+
+  implicit val swagger = new JoernSwagger
+
+  val system = ActorSystem()
```
mpollmeier: here be dragons :fearful:

joern-server/src/main/scala/io/shiftleft/joern/server/JoernController.scala
```diff
@@ -0,0 +1,158 @@
...
+        logger.info(s"received query: $query")
+        runQuery(query)
+        Accepted()
+      }
+    }
+
+    def runQuery(query: String): AsyncResult =
+      new AsyncResult {
+        val is = Future {
+          Try({
```
mpollmeier: outer parens not needed

joern-server/src/main/scala/io/shiftleft/joern/server/JoernController.scala
```diff
@@ -0,0 +1,158 @@
...
+            val e = new ScriptEngineManager().getEngineByName("scala")
+            e.put("aCpg", cpg.get)
+            e.eval(s"""
+                import io.shiftleft.codepropertygraph.Cpg
+              | val cpg = aCpg.asInstanceOf[io.shiftleft.codepropertygraph.Cpg]
+              | $query
+            """.stripMargin).toString
+          }) match {
+            case Success(v)         => queryResult = Some(v)
+            case Failure(exception) => println(exception)
```
mpollmeier: this looks a little fishy: the exception message won't make it to the client I guess, and it'll just respond with 'ok' rather than a result?

joern-server/src/main/scala/io/shiftleft/joern/server/JoernController.scala
```diff
@@ -0,0 +1,158 @@
...
+          }
+
+        }
+      }
+  }
+
+  val queryResultBuilder = (apiOperation[String]("queryresult")
+    summary "query result"
```
mpollmeier: not sure if infix notation really helps readability here

joern-server/src/main/scala/io/shiftleft/joern/server/JoernController.scala
fabsx00: Fixed.

joern-server/src/main/scala/io/shiftleft/joern/server/JoernController.scala
fabsx00: This is run in the Future after the client has already received the OK, so, I can't really pass it back to the client via HTTP though. What I can do though is place it in queryResult as well. That way, the client sees it when picking up the result.

joern-server/src/main/scala/io/shiftleft/joern/server/JoernController.scala
```diff
@@ -0,0 +1,158 @@
...
+    tags "")
+
+  get("/queryresult", operation(queryResultBuilder)) {
+    if (queryResult.isDefined) {
+      val result = queryResult.get
```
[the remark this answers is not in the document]
fabsx00: Content-type was json, but you are right, it was possible that this would just return a non-json string. I've fixed this. We now return a JSON object of the form `{"response" : "string response", "isQueryCompleted" : true/false}`. The string response may still not be json - it is a string representation of whatever the query yields.

joern-server/src/main/scala/io/shiftleft/joern/server/JoernController.scala
```diff
@@ -0,0 +1,157 @@
...
+  val logger = LoggerFactory.getLogger(getClass)
+  protected implicit val jsonFormats: Formats = DefaultFormats
+  protected val applicationDescription = "Joern-Server REST API"
+  protected implicit def executor: ExecutionContext = system.dispatcher
+
+  private var cpg: Option[Cpg] = None
+  private var queryResult: Option[String] = None
+
+  before() {
+    contentType = formats("json")
```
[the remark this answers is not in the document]
fabsx00: yep.

PR #50, opened 2019-05-13
support for `sbt makeSite`

fabsx00:


fabsx00:
This turns out to not be required, so I'm closing this one.

PR #57, opened 2019-05-18
Joern CFG to dot script

b-64rry:
- Script extracts the CFG  of the Code Property Graph and writes it into a dot format:
we might want to generalize this for other edge types or extract only expressions (without identifiers, literals etc.)
 
- Minor change in JoernQuery.scala

### DIFF
--- a/joern-cli/src/main/scala/io/shiftleft/joern/JoernQuery.scala
+++ b/joern-cli/src/main/scala/io/shiftleft/joern/JoernQuery.scala
@@ -1,6 +1,6 @@
 package io.shiftleft.joern
 
-import better.files.File
+import java.io.FileReader
 import javax.script.ScriptEngineManager
 
 object JoernQuery extends App {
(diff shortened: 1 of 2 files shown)


joern-cli/src/main/scala/io/shiftleft/joern/JoernQuery.scala
```diff
@@ -19,10 +19,8 @@ object JoernQuery extends App {
     e.eval(cpgLoadingCode, context)
 
     if (config.isFile) {
-      val reader = File(config.query).fileReader
-      reader.foreach { r =>
-        println(e.eval(r, context))
-      }
+      val reader = new FileReader(config.query)
```
ml86: Why do we need to change the implementation here?

scripts/cfgToDot.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.tinkerpop.gremlin.structure.Direction
+import io.shiftleft.queryprimitives.steps.Implicits.JavaIteratorDeco
+import javax.script.ScriptEngineManager
+
+
+/** Some helper functions: adapted from ReachingDefPass.scala in codeproperty graph repo */
+def vertexToStr(vertex: Vertex): String = {
+  try {
+    val methodVertex = vertex.vertices(Direction.IN, "AST").nextChecked
+    val fileName = ExpandTo.methodToFile(methodVertex) match {
```
fabsx00: I never really got why we have this `ExpandTo` class. My guess is performance. If possible, let's not use it. I'd say though, let's just add a TODO for now to remind us that we should use the DSL instead `ExpandTo` - and merge this.

PR #71, opened 2019-06-04
ensure we're only using one log binding

mpollmeier:


mpollmeier:
True! Superseded by https://github.com/ShiftLeftSecurity/fuzzyc2cpg/pull/68

PR #93, opened 2019-08-28
WIP latest version - not compiling yet

mpollmeier:


mpollmeier:
@fabsx00 not sure what you had in mind re cpgqueryingtests - it's classes, e.g. `LanguageFrontend` are now in semanticcpg/src/test, so we can't use them any more..?

PR #101, opened 2019-09-11
Run data flow analysis as part of joern-parse

fabsx00:
Usually, people will not want to edit the semantics file. In this case, we should not apply semantics on each load (which can often take longer than the loading operation itself), but rather, we should apply semantics on parsing and store edges as part of `cpg.bin.zip`. This PR introduces this default behavior and allows `joern-parse` and `cpg2scpg` to be configured for deviating behavior.

### DIFF
--- a/joern-cli/src/main/scala/io/shiftleft/joern/Cpg2Scpg.scala
+++ b/joern-cli/src/main/scala/io/shiftleft/joern/Cpg2Scpg.scala
@@ -1,6 +1,8 @@
 package io.shiftleft.joern
 
 import io.shiftleft.SerializedCpg
+import io.shiftleft.dataflowengine.layers.dataflows.DataFlowRunner
+import io.shiftleft.dataflowengine.semanticsloader.SemanticsLoader
 import io.shiftleft.semanticcpg.layers.EnhancementRunner
 import org.slf4j.LoggerFactory
 
@@ -12,7 +14,7 @@ object Cpg2Scpg extends App {
 
   parseConfig.foreach { config =>
     try {
-      run(config.inputPath)
+      run(config.inputPath, config.dataFlow, config.semanticsFile)
     } catch {
       case exception: Exception =>
         logger.error("Failed to generate CPG.", exception)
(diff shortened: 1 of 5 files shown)


joern-cli/src/main/scala/io/shiftleft/joern/Cpg2Scpg.scala
```diff
@@ -21,24 +23,34 @@ object Cpg2Scpg extends App {
     System.exit(0)
   }
 
-  case class Config(inputPath: String)
+  case class Config(inputPath: String, dataFlow : Boolean, semanticsFile : String)
```
Ferada:
```suggestion
  case class Config(inputPath: String, dataFlow: Boolean, semanticsFile: String)
```

joern-cli/src/main/scala/io/shiftleft/joern/Cpg2Scpg.scala
```diff
@@ -21,24 +23,34 @@ object Cpg2Scpg extends App {
...
-    }.parse(args, Config(DEFAULT_CPG_IN_FILE))
+    }.parse(args, Config(DEFAULT_CPG_IN_FILE, true, CpgLoader.defaultSemanticsFile))
 
   /**
     * Load the CPG at `cpgFilename` and add enhancements,
     * turning the CPG into an SCPG.
     * @param cpgFilename the filename of the cpg
     * */
-  def run(cpgFilename: String): Unit = {
+  def run(cpgFilename: String, dataFlow : Boolean, semanticsFilename : String): Unit = {
```
Ferada:
```suggestion
  def run(cpgFilename: String, dataFlow: Boolean, semanticsFilename: String): Unit = {
```

joern-cli/src/main/scala/io/shiftleft/joern/Cpg2Scpg.scala
fabsx00: I want to run scalafmt in the next PR, because when I did, it also touched files that have nothing to do with this PR, making it hard to review.

Ferada: Fine.

PR #128, opened 2019-10-24
Add Travis CI Slack integration.

GlassAndOneHalf:


Ferada:
It says "secure", what can you all do with that token?

GlassAndOneHalf:
This is just the way that Travis identifies encrypted tokens. You can't do anything with the encrypted form of the token, only Travis knows how to decrypt it.

PR #133, opened 2019-10-25
Update install/build docs

fabsx00:
Fixes https://github.com/ShiftLeftSecurity/joern/issues/122

### DIFF
--- a/docs/content/essentials/installing/_index.md
+++ b/docs/content/essentials/installing/_index.md
@@ -23,25 +23,73 @@ your own JVM-based project, you only need joern-cli. If, however, you
 plan to access code property graphs using Python scripts or via
 command line tools, the joern-server and cpgclientlib are for you.
 
+
 ## Prerequisites
 
-* Python3
-  - Link: https://www.python.org/downloads/
-* Java runtime 8
+joern-cli and joern-server are Scala applications and should work on
+systems offering a Java virtual machine, e.g., Microsoft Windows, Mac
+OS X or Gnu/Linux. cpgclientlib is a Python 3 library.
+
+* Java 8
   - Link: http://openjdk.java.net/install/
-* Scala build tool (sbt)
-  - Link: https://www.scala-sbt.org/
 
-Python3 and Java 8 are present in the official package repositories of
-all major current Linux distributions and BSD flavors. For installing
-the the Scala build tool (sbt), please following the instructions at:
+* Python3 (only if you need cpgclientlib)
+  - Link: https://www.python.org/downloads/
 
-https://www.scala-sbt.org/download.html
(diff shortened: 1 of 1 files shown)


docs/content/importing/_index.md
```diff
@@ -0,0 +1,70 @@
++++
+title="Importing Code"
+weight=2
++++
+
+Code can be imported into joern in two ways. The program `joern-parse`
+can be used to create code property graphs and persist them in a
+format based on Google Protocol Buffer messages, or, `create-cpg` can
+be used to instruct joern-server to create a code property graph and
```
mpollmeier: not sure about the rest of these docs, but don't we normally write it as one word, i.e. `codepropertygraph`?

docs/content/installing/_index.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Configuring the JVM for Optimal Performance
+
+Code analysis can require lots of memory, and unfortunately, the JVM does not pick up the available amount of memory by itself. While tuning Java memory usage is a discipline in its own right, it is usually sufficient to specify the maximum available amount of heap memory using the JVM's -Xmx flag. The easiest way to achieve this globally is by setting the environment variable _JAVA_OPTS as follows:
+
+```
+export _JAVA_OPTS="-Xmx$NG"
+```
+
+where $N is the amount of memory in gigabytes. You can add this line to your shell startup script, e.g., ~/.bashrc or ~/.zshrc.
```
mpollmeier: this statement is correct but an example would make it foolproof

docs/content/importing/_index.md
fabsx00: `codepropertygraph` is the repo. Maybe it should be Code Property Graphs though (capital letters) since it's a name. I think I'll do that.

docs/content/installing/_index.md
fabsx00: Good point, will do.

docs/content/importing/_index.md
```diff
@@ -0,0 +1,70 @@
...
+can be used to create Code Property Graphs and persist them in a
+format based on Google Protocol Buffer messages, or, `create-cpg` can
+be used to instruct joern-server to create a code property graph and
+make it available via the REST API.
+
+## Creating Code Property Graphs with `joern-parse`
+
+Code property graphs can be created from C/C++ code using
+`joern-parse`. For example, to create a code property graph for the
+vulnerable sample program `tarpit-c`, you can issue the following
```
GlassAndOneHalf: It might be worth clarifying that `tarpit-c` is not a binary program but instead the source root.

fabsx00: Done

PR #135, opened 2019-10-25
Added ast-for-funcs script

max-leuthaeuser:
* added the ast-for-funcs script
* added test for it


scripts/ast-for-funcs/ast-for-funcs.scala
```diff
@@ -0,0 +1,27 @@
...
+  )
+
+Json.obj(
+  ("file", Json.fromString(cpg.file.name.l.head)),
+  ("functions", Json.fromValues(cpg.method.name.l.map { methodName =>
+    val method = cpg.method.name(methodName)
+    Json.obj(("function", Json.fromString(methodName)),
+      ("AST", Json.fromValues(method.astChildren.l.map(_.asJson))))
+  }))
+)
```
GlassAndOneHalf: I think it'd be nicer to define a case class for this, rather than writing it by hand:
```scala
import io.circe.generic.semiauto._

case class AstForFuncsFunction(function: String, AST: List[AstNode])
case class AstForFuncsResult(file: String, functions: List[AstForFuncsFunction])

implicit val encodeVertex: Encoder[AstNode] = ... // As before
implicit val encodeFuncResult: Encoder[AstForFuncsResult] = deriveEncoder
implicit val encodeFuncFunction: Encoder[AstForFuncsFunction] = deriveEncoder

AstForFuncResult(
  cpg.file.name.l.head,
  cpg.method.name.l.map { methodName =>
    val method = cpg.method.name(methodName)
    AstForFuncsFunction(method, method.astChildren.l)
  }
).asJson
```

scripts/ast-for-funcs/ast-for-funcs.scala
```diff
@@ -0,0 +1,27 @@
...
+      )
+    }))
+  )
+
+Json.obj(
+  ("file", Json.fromString(cpg.file.name.l.head)),
+  ("functions", Json.fromValues(cpg.method.name.l.map { methodName =>
+    val method = cpg.method.name(methodName)
+    Json.obj(("function", Json.fromString(methodName)),
+      ("AST", Json.fromValues(method.astChildren.l.map(_.asJson))))
```
fabsx00: Do we end up only with the flat list of nodes here or are edges also encoded in some way, e.g., via nesting?

max-leuthaeuser: Only nodes at the moment.

scripts/ast-for-funcs/ast-for-funcs.scala
max-leuthaeuser: Done. Thanks :+1:

scripts/ast-for-funcs/ast-for-funcs.scala
fabsx00: Alright. We're going to need edges, but feel free to merge this, and then we'll look into edges at a later point in time.

PR #141, opened 2019-10-26
Change shell to /bin/bash because of `[[`. Remove joern-query

fabsx00:
On a system where '/bin/sh' does not point to bash, we get
```
 ./joern
./joern: 3: ./joern: [[: not found
```
 Changing `./bin/sh` to `/bin/bash`. Not sure if there's a better fix, e.g., replacing `[[` with something else.

cc: @dukebarman, would be nice to know if this works on OSX


dukebarman:
Yes, it works.

PR #154, opened 2019-10-30
Add more color to query language docs

fabsx00:
Added some background info on the query language, especially about lazy evaluation.

### DIFF
--- a/docs/content/querying/_index.md
+++ b/docs/content/querying/_index.md
@@ -1,43 +1,48 @@
 +++
-Title="Querying Code Property Graphs"
+Title="Query Language"
 weight=5
 alwaysopen=true
 +++
 
 Joern/Ocular provides a domain specific language (DSL) for code
-analysis based on the Scala programming language. The great advantage
-of this approach is that all features of the host language are
-automatically inherited. Joern queries are simply lazily evaluated
-Scala expressions, and you can make use of the complete Scala
-programming language in Joern scripts.
+analysis based on [Scala](https://www.scala-lang.org/). It is an
+internal DSL, that is, it is a language defined via classes and
+methods as opposed to a formal grammar. The great advantage of this
+approach is that all features of the host language are automatically
+inherited. Joern queries are lazily evaluated Scala expressions, and
+you can make use of the complete Scala programming language in Joern
+scripts. In this section, we look into the basic concepts of our query
+language and illustrate them by example.
+
+In the following examples, we assume that a code property graph for
+the *tarpitc* program is loaded. For details on creating and loading
+CPGs, please check out [Importing Code](../importing) and [The Joern
+Shell](../shell).
 
 
-# Basic Concepts: Expressions, Labels, IDs
 
-To get a first idea of what information the Code Property Graph makes
-available, start the `joern` shell and type `cpg.`, then press the Tab
-key:
+# Nodes, Labels, and Expressions
 
-
-```bash
-joern> cpg.
-all             file            local           methodReturn    types
-argument        graph           member          namespace
-call            id              metaData        namespaceBlock
-close           identifier      method          parameter
-comment         literal         methodRef       typeDecl
-
-```
-
-The following query returns *all* nodes upon evaluation:
+To get a first idea of the data which the language makes accessible,
+let us begin querying for all nodes of the code property graph via
+`cpg.all`:
 
 ```bash
 cpg.all
 res8: NodeSteps[StoredNode] = io.shiftleft.semanticcpg.language.NodeSteps@4cce421e
 ```
-Notice that `cpg.all` is an **expression**, and this is expression is
-only **evaluated** if so requested, e.g., by converting the expression to a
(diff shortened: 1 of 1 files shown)


docs/content/querying/_index.md
```diff
@@ -1,63 +1,67 @@
...
-all             file            local           methodReturn    types
-argument        graph           member          namespace
-call            id              metaData        namespaceBlock
-close           identifier      method          parameter
-comment         literal         methodRef       typeDecl
-
-```
-
-The following query returns *all* nodes upon evaluation:
+To get a first idea of the data which the language makes accessible,
```
GlassAndOneHalf:
```suggestion
To begin exploring the data that the language exposes,
```

docs/content/querying/_index.md
```diff
@@ -1,63 +1,67 @@
...
+let us begin querying for all nodes of the code property graph via
```
GlassAndOneHalf:
```suggestion
let's create a query that selects all nodes in the code property graph with
```

docs/content/querying/_index.md
```diff
@@ -1,63 +1,67 @@
...
 cpg.all
 res8: NodeSteps[StoredNode] = io.shiftleft.semanticcpg.language.NodeSteps@4cce421e
 ```
-Notice that `cpg.all` is an **expression**, and this is expression is
-only **evaluated** if so requested, e.g., by converting the expression to a
-list as in the following query.
+
+Notice that the result is not a list of nodes but rather a variable type
+`NodeSteps[StoredNode]`. You can think of this as an *expression*
+which - only upon evaluation - yields all nodes [1]. The advantage of
```
GlassAndOneHalf:
```suggestion
which only upon evaluation will yield all nodes [1]. The advantage of
```

docs/content/querying/_index.md
```diff
@@ -1,63 +1,67 @@
...
+this two stage approach is that we can already perform calculations on
+expressions prior to costly evaluation. In particular, we can use the
+type system to perform validity checks and query optimization,
+features that are vital as we compose expression to form complex
```
GlassAndOneHalf:
```suggestion
features that are vital as we compose expressions to form complex
```

docs/content/querying/_index.md
```diff
@@ -90,39 +94,98 @@ method `toSet` is called. The resulting Scala set is converted to a
...
+[gremlin-scala](https://github.com/mpollmeier/gremlin-scala) on which
+our DSL is based. In Gremlin, `Steps` describe walks in a graph, which
+are combined via function composition to yield *traversals*. You can
+think of traversals as expressions that describe sets of nodes.
+
+
+
+## Expression Types and Starters
+
+For each node label, a corresponding *expression type* is defined in the
```
GlassAndOneHalf: I'm not 100% sure if this change makes sense, but just calling it a `type` seems more intuitive to me.

```suggestion
For each node label, a corresponding *type* is defined in the
```

docs/content/querying/_index.md
```diff
@@ -90,39 +94,98 @@ method `toSet` is called. The resulting Scala set is converted to a
...
+close           identifier      method          parameter
+comment         literal         methodRef       typeDecl
+
+```
+
+It is generally advisable to use specific starters as opposed to the
+generic `cpg.all`, as for the latter, it is unclear at compile time
+which types of nodes are processed. Using specific starters such as
+`.method` make it possible to use the Scala's type system for
+completion and to detect invalid queries even before evaluation.
```
GlassAndOneHalf:
```suggestion
auto-completion and to detect invalid queries before they are evaluated.
```

docs/content/querying/_index.md
```diff
@@ -90,39 +94,98 @@ method `toSet` is called. The resulting Scala set is converted to a
...
 ```
 
-Note that the type of the nodes returned by `cpg.id(10)` is unknown at
-compile time, and therefore, the returned type for `cpg.id(10).l`is
-`List[Nothing]`. In practice, you will know the types of nodes you are
-expecting from the previous query. For example, in the outlined
-scenario, we know that case are file nodes. An explicit cast can be
+Note that the type of the nodes returned by `cpg.id(34)` is unknown at
+compile time, and therefore, the returned type for `cpg.id(34).l`is
+`List[Nothing]`. In practice, the external tool will usually know the
```
GlassAndOneHalf: Not related to the content, but a returning a `List[Nothing]` from this seems super strange as it is a bottom type. I would expect `Any` or `AnyRef`.

docs/content/querying/_index.md
```diff
@@ -90,39 +94,98 @@ method `toSet` is called. The resulting Scala set is converted to a
...
+node type from the previous query, e.g., in our example, we know that
+the node with the ID `34` is a file node. An explicit cast can be
```
GlassAndOneHalf:
```suggestion
the node with the ID `34` is a `File` node. An explicit cast can be
```

PR #169, opened 2019-11-12
Replace Log4J/slf4j-simple with Logback.

GlassAndOneHalf:


joern-cli/src/main/resources/logback.xml
```diff
@@ -0,0 +1,13 @@
+<configuration>
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <!-- encoders are assigned the type
+         ch.qos.logback.classic.encoder.PatternLayoutEncoder by default -->
+    <encoder>
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+
+  <root level="info">
```
fabsx00: I wonder if we should keep it at WARN

GlassAndOneHalf: Agreed, done.

PR #196, opened 2019-12-13
Add new PDG script.

GlassAndOneHalf:
For the following program:
```c
int main() {  
  int a = 2;
  int b = 3;  if (a > 0) print(a);
  int result = foo(a, b);
  return result;  
}

int foo(int x, int y) {
  return x + y;
}

void print(int x) {
  printf(x);  
}
```

This script will print:
```
res2: AnyRef = (
  List(
    (57L, 54L),
    (26L, 20L),
    (24L, 20L),
    (31L, 10L),
    (31L, 15L),
    (24L, 10L),
    (35L, 28L),
    (28L, 31L),
    (46L, 42L),
    (43L, 45L),
    (46L, 41L),
    (7L, 35L),
    (20L, 10L)
  ),
  List(
    (35L, "return result;"),
    (28L, "result = foo(a, b)"),
    (24L, "print(a)"),
    (20L, "a > 0"),
    (7L, "RET"),
    (31L, "foo(a, b)"),
    (46L, "x + y"),
    (57L, "printf(x)"),
    (43L, "RET"),
    (26L, "a")
  )
)
```

The first list contains the edges between the vertices `in_vertex_id => out_vertex_id`. The second contains all of the vertices in the PDG, along with the code that the vertex contains.

### DIFF
--- a/joern-cli/src/main/resources/scripts/general/pdg-for-funcs-dump.sc
+++ b/joern-cli/src/main/resources/scripts/general/pdg-for-funcs-dump.sc
@@ -70,7 +70,6 @@ import io.circe.syntax._
 import io.circe.{Encoder, Json}
 import org.apache.tinkerpop.gremlin.structure.{Edge, VertexProperty}
 
-import io.shiftleft.codepropertygraph.generated.nodes.MethodParameterIn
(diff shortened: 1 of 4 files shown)


fabsx00:
For this script, as well as the scripts we have for ASTs and CFGs, I wonder if we can allow generation only for a specified function. That would be great because code bases may be huge.

PR #200, opened 2020-01-07
Add description for using fuzzyc2cpg cpgs

derpsteb:
We had problems finding methods in cpgs that were built with fuzzy2cpg. This explains why we had those problems and provides a solution. 
@fabsx00 Do we actually want to put this information in the official docs? This place seems the most logical to search for this info, that's why I put it here.


derpsteb:
It seems like GitHub's UI is broken or sth as the build was finished long ago ([build](https://travis-ci.org/ShiftLeftSecurity/joern/builds/633769256?utm_source=github_status&utm_medium=notification)).
That's why I am merging despite UI warnings.

PR #203, opened 2020-01-16
Joern-parse and loadCpg use overflowDB by default now

fabsx00:
Depends on https://github.com/ShiftLeftSecurity/codepropertygraph/pull/550

Joern-parse now generates an overflowDB database file, as opposed to the protobuf-based CPG format. `loadCpg` has been modified to accept these files by default. The advantage is that nodes and edges are now lazily loaded at traversal time and opening of CPGs should be instant, even for large CPGs.


fabsx00:
Yes, enabling the overflow mechanism is intentional. I'd like to achieve lazy loading and the ability to process graphs that do not fit into ram entirely.

fabsx00:
On a 1.2G mvstore file, I see `loadCpg` spin for about 10 seconds and allocate memory. I am wondering whether it is really lazily loading or not. @mpollmeier could you take a look?

fabsx00:
I've done a head-to-head comparison of this branch with master for VLC, and I see the following:
- Loading time for the CPG reduces from over 50 seconds to about 30 seconds on a 2 core laptop with 16 GB ram
- Time for creating the CPG increases by about 30 seconds.

fabsx00:
A few changes are required to allow for application of enhancements and data flow layer without serializing into proto:
http://github.com/ShiftLeftSecurity/codepropertygraph/pull/550/

fabsx00:
That was it: when not writing out _both_ proto and the mvstore, we get similar performance for `joern-parse` with this PR, and a decrease in initial loading time on `loadCpg`. This PR thus already brings in an improvement, I believe we should further optimize `loadCpg`, however.

fabsx00:
Confirmed that a large part of the loading time is spent in index creation - which we do on load. Might be worth considering storing indexes in overflowdb.

mpollmeier:
https://github.com/ShiftLeftSecurity/overflowdb/pull/62 refactored the index implementation in overflowdb, and shortly @lifecoder will add support to serializing/deserializing them to/from mvstore

fabsx00:
@mpollmeier transitioning from string labels to integer labels brings down the loading time further to 23 seconds, as opposed to 30 seconds, so this was already a good step. The remainder is most likely spent in index creation at the moment.

fabsx00:
Updated to create index on scpg creation. Tested with the linux kernel. Loading from the proto CPG and creating the overflowdb from it is too slow as @lifecoder @mpollmeier suspected. I won't merge this now and wait until we have `cpg2overflowdb`.

fabsx00:
Tested on the 2-core laptop with 16GB RAM: we have reduced loading time from over 50 seconds to 6 seconds. Great work everyone :) @lifecoder @mpollmeier

PR #210, opened 2020-01-24
Add script to detect if malloc is not subsequently freed.

GlassAndOneHalf:


joern-cli/src/main/resources/scripts/README.md
```diff
@@ -5,14 +5,17 @@ This is our collection of ready-to-run scripts for `joern`.
...
 ```
 
-inside a running Joern session.
+inside a running Joern session. You can also use the `runScript` method available on the `cpg`:
+```scala
+cpg.runScript("my-script.sc", Map("param" -> "value"))
+```
 
-To get a list of available scripts you might want to call:
+To get a list of available scripts you can want to call:
```
fabsx00: "can want" => can

joern-cli/src/main/resources/scripts/c/malloc-leak.sc
```diff
@@ -0,0 +1,20 @@
+import io.shiftleft.codepropertygraph.Cpg
+import io.shiftleft.codepropertygraph.generated.nodes.Call
+import io.shiftleft.dataflowengine.language._
+import io.shiftleft.semanticcpg.language._
+
+private def mallocCalls(cpg: Cpg): Steps[Call] = {
```
fabsx00: I can imagine that in the future, we'll want to more generically identify allocation calls, but let's get this in first :)

joern-cli/src/main/resources/scripts/README.md
GlassAndOneHalf: I think this calls for a covfefe

PR #228, opened 2020-03-04
Reduce duplication for predefs

fabsx00:
We were using slightly different predefs for ammonite for the interactive shell vs scripts. To reduce duplication and make the difference immediately visible, I've introduced a `Predef` object hosting shared code and the diffs.

### DIFF
--- a/joern-cli/src/main/scala/io/shiftleft/joern/Predefined.scala
+++ b/joern-cli/src/main/scala/io/shiftleft/joern/Predefined.scala
@@ -1,0 +1,33 @@
+package io.shiftleft.joern
+
+object Predefined {
+
+  /* ammonite tab completion is partly broken for scala > 2.12.8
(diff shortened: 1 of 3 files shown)


joern-cli/src/main/scala/io/shiftleft/joern/Predefined.scala
```diff
@@ -0,0 +1,33 @@
+package io.shiftleft.joern
+
+object Predefined {
+
+  /* ammonite tab completion is partly broken for scala > 2.12.8
+   * applying workaround for package wildcard imports from https://github.com/lihaoyi/Ammonite/issues/1009 */
+  val shared: String = """
+        |import gremlin.scala.{`package` => _, _}
+        |import io.shiftleft.console.{`package` => _, _}
```
hubertp: @fabsx00 Generally if one wants to import the whole package object we would simply write
```
import io.shiftleft.console._
```

Issue #229, opened 2020-03-06
Support installation via script on macOS


l0s:
Attempting to run the `joern-install.sh` script on macOS produces the following error:

    readlink: illegal option -- f
    usage: readlink [-n] [file ...]

System information:

    sh-3.2$ uname -a
    Darwin host 19.3.0 Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST 2020; root:xnu-6153.81.5~1/RELEASE_X86_64 x86_64

AlexDenisov:
Closed via #230

PR #230, opened 2020-03-06
Support installation via script on macOS

l0s:
The `readlink` command on macOS does not support the `-f` / `--canonicalize`
parameter available in the GNU version. This adds a check for the Darwin
kernel and provides an alternative approach to finding the script's
directory.

Addresses: #229

### DIFF
--- a/joern-install.sh
+++ b/joern-install.sh
@@ -4,9 +4,9 @@ set -eu
 if [ "$(uname)" = 'Darwin' ]; then
   # get script location
   # https://unix.stackexchange.com/a/96238
-  if [ -n "$BASH_SOURCE" ]; then
+  if [ "${BASH_SOURCE:-x}" != 'x' ]; then
     this_script=$BASH_SOURCE
-  elif [ -n "$ZSH_VERSION" ]; then
+  elif [ "${ZSH_VERSION:-x}" != 'x' ]; then
     setopt function_argzero
     this_script=$0
   elif eval '[[ -n ${.sh.file} ]]' 2>/dev/null; then
--- a/joern-install.sh
+++ b/joern-install.sh
@@ -1,8 +1,26 @@
 #!/usr/bin/env sh
 set -eu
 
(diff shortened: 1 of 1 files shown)


AlexDenisov:
Hi @l0s, thanks a lot for the patch. I tried it under bash on macOS Catalina and it works.

One thing that I believe needs to be addressed is `zsh` support:
when I run the script under `zsh` it fails because `BASH_SOURCE` is undefined which is an error (see `set -eu` above).

It is up to you now: you can either make it work for zsh or just remove it and we stick to a recommendation "use `sh joern-install.sh` on macOS".

What do you think? :)

l0s:
Thanks for reviewing this @AlexDenisov! I'll investigate a solution.

Here is what I found so far. It actually works for me on ZSH when I run `./joern-install.sh` (ZSH 5.7.1). However, it fails when I run `zsh joern-install.sh`. I will investigate some more.

l0s:
@AlexDenisov, I pushed a fix. I tested with the following scenarios:

    % ./joern-install.sh 
    -n This script will download and install the Joern tools on your machine. Proceed? [Y/n]: 
    ^C
    % sh joern-install.sh
    -n This script will download and install the Joern tools on your machine. Proceed? [Y/n]: 
    ^C
    % bash joern-install.sh 
    This script will download and install the Joern tools on your machine. Proceed? [Y/n]: ^C
    % zsh joern-install.sh 
    This script will download and install the Joern tools on your machine. Proceed? [Y/n]: ^C
    % sh -eu joern-install.sh 
    -n This script will download and install the Joern tools on your machine. Proceed? [Y/n]: 
    ^C

Would you mind testing it again? This is the system with which I tested:

    % zsh --version
    zsh 5.7.1 (x86_64-apple-darwin19.0)
    % uname -a
    Darwin 19.3.0 Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST 2020; root:xnu-6153.81.5~1/RELEASE_X86_64 x86_64

PR #262, opened 2020-05-21
use sbt-native-packager to preserve executable file attributes

mpollmeier:
using zt-zip resulted in losing executable file attributes, while
sbt-native-packager does set those just fine - i didn't even know this existed

### DIFF
--- a/joern-cli/build.sbt
+++ b/joern-cli/build.sbt
@@ -117,22 +117,20 @@ generateScaladocs := {
   out
 }
 
-lazy val packageZip = taskKey[File]("create zip with joern-cli, including schema-extender")
-packageZip := {
-  import org.zeroturnaround.zip.ZipUtil
+/* add schemas from codepropertygraph and schema-extender */
+Universal/packageBin/mappings ++= {
+  import better.files.File
   val joernCliStaged = (Universal/stage).value
-  val schemaExtenderStaged = (Projects.schemaExtender/Universal/stage).value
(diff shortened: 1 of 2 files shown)


joern-cli/build.sbt
```diff
@@ -117,22 +117,20 @@ generateScaladocs := {
...
+/* add schemas from codepropertygraph and schema-extender */
+Universal/packageBin/mappings ++= {
+  import better.files.File
   val joernCliStaged = (Universal/stage).value
-  val schemaExtenderStaged = (Projects.schemaExtender/Universal/stage).value
-  val schemaExtenderInstallLocation = joernCliStaged/"schema-extender"
-
-  IO.copyDirectory(schemaExtenderStaged, schemaExtenderInstallLocation, CopyOptions(overwrite = true, preserveLastModified = true, preserveExecutable = true))
+  val schemaExtenderStaged = (Projects.schemaExtender/Universal/stage).value.toPath
+  val tmpDir = File.newTemporaryDirectory("joern-cli-build").deleteOnExit
```
ursachec, line 125: Unless I'm missing something, the `joern-cli-build` prefix is not used anywhere, so might as well not specify it

mpollmeier: it's just a prefix for the temporary dir created in /tmp. it's not necessary, but good practice to do so: just in case something goes wrong and you want to know where that random file came from, it gives you a hint where to start looking

PR #269, opened 2020-05-25
upgrade deps, fix test

mpollmeier:


mpollmeier:
You raise a valid point and made me rethink a previous decision. 

In an attempt to avoid unnecessary overhead I previously changed the
return type of `method.parameter` from `NodeSteps` to `Iterator`.
Given that it complicates things - the user now needs to call
`.start` explicitly, or we add a hard to understand implicit
conversion. Once we use overflowdb traversal, it won't have any
overhead any more, so, I'll change that back then. 

Closing in favor of https://github.com/ShiftLeftSecurity/codepropertygraph/pull/771

PR #272, opened 2020-06-04
upgrade cpg

mpollmeier:


mpollmeier:
will handle as part of https://github.com/ShiftLeftSecurity/joern/pull/271

PR #278, opened 2020-06-08
scalafmt - mostly to ensure joern and ocular versions are identical

mpollmeier:


mpollmeier:
looks like scalafmt rules are different between the projects...

PR #300, opened 2020-07-20
Remove joern server

ursachec:


ursachec:
@fabsx00 Ah, right, missed that, preparing a commit...

Issue #360, opened 2020-10-14
Docker Image does not build


AlexanderProschek:
When I attempt to build a docker image, I get this issue:
```
alex@MacBook-Pro joern $ docker build . --no-cache
[+] Building 50.1s (7/9)                                                                                                                                                                                    
 => [internal] load .dockerignore                                                                                                                                                                      0.0s
 => => transferring context: 2B                                                                                                                                                                        0.0s
 => [internal] load build definition from Dockerfile                                                                                                                                                   0.0s
 => => transferring dockerfile: 37B                                                                                                                                                                    0.0s
 => [internal] load metadata for docker.io/library/alpine:latest                                                                                                                                       0.2s
 => CACHED [1/6] FROM docker.io/library/alpine:latest@sha256:185518070891758909c9f839cf4ca393ee977ac378609f700f60a771a2dfe321                                                                          0.0s
 => [2/6] RUN apk update && apk upgrade && apk add --no-cache openjdk8-jre-base python3 git curl gnupg bash nss ncurses                                                                               17.1s
 => [3/6] RUN ln -sf python3 /usr/bin/python                                                                                                                                                           0.2s
 => ERROR [4/6] RUN curl -sL "https://piccolo.link/sbt-1.3.13.tgz" | gunzip | tar -x -C /usr/local                                                                                                    32.5s 
------                                                                                                                                                                                                      
 > [4/6] RUN curl -sL "https://piccolo.link/sbt-1.3.13.tgz" | gunzip | tar -x -C /usr/local:                                                                                                                
#7 32.48 gunzip: invalid magic                                                                                                                                                                              
#7 32.48 tar: short read                                                                                                                                                                                    
------
failed to solve with frontend dockerfile.v0: failed to build LLB: executor failed running [/bin/sh -c curl -sL "https://piccolo.link/sbt-$SBT_VERSION.tgz" | gunzip | tar -x -C /usr/local]: runc did not terminate sucessfully
```

AlexanderProschek:
The link [https://piccolo.link/sbt-1.3.13.tgz](https://piccolo.link/sbt-1.3.13.tgz) does not seem to work, so I updated it with this download https://github.com/sbt/sbt/releases/download/v$SBT_VERSION/sbt-$SBT_VERSION.tgz

PR #362, opened 2020-10-14
Fix local Dockerfile issue

AlexanderProschek:
This Pull request is in response to #360 

### Issue:
For some reason when the docker build failed on my machine at the step `RUN curl -sL "https://piccolo.link/sbt-$SBT_VERSION.tgz" | gunzip | tar -x -C /usr/local` and I have determined that the issue is downloading the tar file. It seems as though the file is not located at that address anymore.

### Solution:
The official [sbt website](https://www.scala-sbt.org/download.html) offers downloads for each version of sbt straight from GitHub. Swapping the location of sbt from `https://piccolo.link/sbt-$SBT_VERSION.tgz` to `https://github.com/sbt/sbt/releases/download/v$SBT_VERSION/sbt-$SBT_VERSION.tgz` seems to fix the issue

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ RUN ln -sf python3 /usr/bin/python
 ENV SBT_VERSION 1.3.13
 ENV SBT_HOME /usr/local/sbt
 ENV PATH ${PATH}:${SBT_HOME}/bin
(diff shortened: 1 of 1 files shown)


mpollmeier (approved):

mpollmeier:
Looks correct to me and I verified the new url works, so will go ahead and merge. 
cc @itsacoderepo since you initially added the Dockerfile. 

Thank you @AlexanderProschek !

PR #363, opened 2020-10-15
make fuzzycpp download optional

mpollmeier:
This is a bit of a controversy: for local development this is
expected, since the fuzzycpp binary is released to github from the cpg
release build. When using an official cpg release this should not fail
however.

Now, if we fail to download the binary, sbt prints a warning and
allows to retrieve the root exception.


mpollmeier:
if this finds your liking, i'll do the same for ocular

fabsx00 (approved):

PR #370, opened 2020-10-25
Implement `joern-stats` subtool

ursachec:
Useful for printing statistics on Code Property Graph files
when an interactive session is too overkill.

Example usage:
```
$ ./joern-stats example.cpg.bin.zip
METADATA  1
NAMESPACE 3827
TYPE_DECL 29983
METHOD    262613
CALL      5708704
FINDING   275
```

### DIFF
--- a/joern-cli/src/main/scala/io/shiftleft/joern/JoernStats.scala
+++ b/joern-cli/src/main/scala/io/shiftleft/joern/JoernStats.scala
@@ -18,7 +18,7 @@ object JoernStats extends App {
       head("Prints statistics on various nodes found in Code Property Graphs")
       help("help")
 
-      arg[String]("cpg")
+      arg[String]("cpg.bin")
         .text("Path to the Code Property Graph file")
         .action((name, c) => c.copy(cpgFileName = name))
     }
--- a/joern-cli/src/main/scala/io/shiftleft/joern/JoernStats.scala
+++ b/joern-cli/src/main/scala/io/shiftleft/joern/JoernStats.scala
@@ -31,7 +31,6 @@ object JoernStats extends App {
       sys.exit(1)
     }
 
-    println("METADATA  " + cpg.metaData.size)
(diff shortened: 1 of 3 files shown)


joern-cli/src/main/scala/io/shiftleft/joern/JoernStats.scala
```diff
@@ -0,0 +1,42 @@
...
+
+case class StatsConfig(cpgFileName: String = "cpg.bin")
+
+object JoernStats extends App {
+  private def parseConfig: Option[StatsConfig] = {
+    new scopt.OptionParser[StatsConfig]("joern-stats") {
+      head("Prints statistics on various nodes found in Code Property Graphs")
+      help("help")
+
+      arg[String]("cpg")
```
fabsx00, line 21: Let's make `cpg.bin` the default so that it is consistent with `joern-export` and `joern-flow`.

joern-cli/src/main/scala/io/shiftleft/joern/JoernStats.scala
```diff
@@ -0,0 +1,42 @@
...
+      println("ERROR: could not load Code Property Graph file at path `" + config.cpgFileName + "`. Exiting.")
+      sys.exit(1)
+    }
+
+    println("METADATA  " + cpg.metaData.size)
+    println("NAMESPACE " + cpg.namespace.size)
+    println("TYPE_DECL " + cpg.typeDecl.size)
+    println("METHOD    " + cpg.method.size)
+    println("CALL      " + cpg.call.size)
+    println("FINDING   " + cpg.graph.nodes(NodeTypes.FINDING).size)
```
fabsx00, line 39: Since findings aren't part of the base specification and we're not providing any open-source tools for generating them, I'm wondering whether we should leave them out for now. `MetaData` should always be one, so, not sure about that one either.

fabsx00 (approved):

PR #377, opened 2020-11-02
Promt about installing Joern tools

m1cm1c:
The install script includes an option where the default answer once is stated as yes, once as no. This PR removes the "yes" part, leaving the "no" part which is what the code actually does.


ursachec (approved):
Good catch @m1cm1c, thank you for the contribution!

PR #495, opened 2021-04-02
Update build.sbt

itsacoderepo:


itsacoderepo:
Merging because tests pass ..

PR #579, opened 2021-08-23
Add js2cpg to Joern

johannescoetzee:
# Notes
These changes add `js2cpg.sh` to the Joern repo, allowing users to both generate the cpg from javascript code by calling `./js2cpg <SRC_DIR>` from the root directory, as well as in Joern through `importCode.javascript("src_dir")

# Testing
`sbt clean test`, as well as manual tests creating the cpg both by calling `js2cpg.sh` and through Joern. I didn't test whether this specifically works when installing joern via `joern-install.sh`, but I did verify that `c2cpg` is successfully installed by this, and the approach followed here mimics that one almost exactly.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,6 +9,7 @@ ThisBuild / scalaVersion := "2.13.5"
 ThisBuild /Test /fork := true
 val cpgVersion = "1.3.287"
 val ghidra2cpgVersion = "0.0.24"
+val js2cpgVersion = "0.2.2"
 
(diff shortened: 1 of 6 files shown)


fabsx00 (approved):

mpollmeier:
Thank you! 
Going forward, we'll package all frontends as binaries to avoid jvm dependencies by the frontends (which will lead to clashing transitive deps). Let's merge this as is, but before announcing the inclusion of js2cpg I'd like to bring in that change - will send a PR.

PR #581, opened 2021-08-24
change packaging: download frontends as binaries

mpollmeier:
* also: upgrade cpg

### reasons to invoke frontends via shell command rather than jvm dependency
* smaller classpath
  * sbt/idea/... work faster
  * joern/ocular start faster
* no hidden memory leaks (as observed with fuzzyc2cpg)
* less complexity in classpath
* no issues with jre versions
* only a matter of time until we have version clashes between frontends

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,7 @@ jobs:
           ./schema-extender/test.sh
           cd -
       - run: sbt ciReleaseTagNextVersion createDistribution
-      - run: sha512sum ./joern-cli.zip > ./joern-cli.zip.sha512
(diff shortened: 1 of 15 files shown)


ursachec (approved):

johannescoetzee:
It looks like with the current revision, `sbt stage` is no longer sufficient to get a working joern build from a clean repo. Would it be possible to add fetch and extract steps as part of the stage job?

mpollmeier:
yes, good point, I'll take a look

mpollmeier:
@johannescoetzee done, stage and distribution now work
will test `joern-install.sh` again after this was merged/released

johannescoetzee (approved):
Looks like everything is working as expected!

PR #586, opened 2021-08-31
Default to latest release when updating querydb

johannescoetzee:
# Notes
Default to using the latest query-database release when running `joern-scan --updatedb`. This means we wouldn't have to update this version manually. 

__Alternative approach:__ We could also have the default version not be the latest one, but in that case we should move the default version into build.sbt so that it's harder to miss.

# Testing
```
❯ ./joern-scan --updatedb
Downloading default query bundle from: https://github.com/joernio/query-database/releases/latest/download/querydb.zip
Wrote: 34671576 bytes to /tmp/joern-scan16544845908678208747/querydb.zip
Removing current version of query database
Adding updated version of query database
```


johannescoetzee:
Not going with the initial approach, since this could lead to upstream changes unexpectedly breaking Joern. Getting SBT to expose settings to the underlying code is also probably more effort than it's worth (if it's possible), so I'm just closing this instead.

PR #587, opened 2021-09-01
Add a separate C front-end version to decouple front-end and codepropertygraph staging

johannescoetzee:
# Notes
This is a convenience follow-up to https://github.com/joernio/joern/pull/583 and decouples the version used to fetch `c2cpg` and `fuzzyc2cpg` from the `codepropertygraph` version. This lets you test local `codepropertygraph` changes without having to use local `c2cpg` and `fuzzyc2cpg` distributions, provided no changes to the front-ends need to be tested as well (a common case).

# Testing
`sbt clean stage test`, followed by 

```
❯ ./c2cpg.sh
Error: Missing argument input-dirs
Try --help for more information.
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,6 +7,7 @@ ThisBuild / scalaVersion := "2.13.5"
 ThisBuild /Test /fork := true
 val cpgVersion = "1.3.305"
 val c2cpgVersion = "1.3.305"
+val fuzzyc2cpgVersion = "1.3.305"
 val ghidra2cpgVersion = "0.0.25"
 val js2cpgVersion = "0.2.3"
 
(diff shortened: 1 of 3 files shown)


johannescoetzee:
A thought I had about naming is having separate `c2cpg` and `fuzzyc2cpg` versions. It's more obvious what the versions represent in that case, but comes at the cost of more duplication.

build.sbt
```diff
@@ -5,7 +5,8 @@ organization := "io.shiftleft"
 ThisBuild / scalaVersion := "2.13.5"
 // don't upgrade to 2.13.6 until https://github.com/com-lihaoyi/Ammonite/issues/1182 is resolved
 ThisBuild /Test /fork := true
-val cpgVersion = "1.3.304"
+val cpgVersion = "1.3.305"
+val cFrontendVersion = "1.3.305"
```
ursachec, line 9: make this `c2cpgVersion` instead for consistency reasons?

ursachec (approved):
besides the name of the const, lgtm

build.sbt
johannescoetzee: That's a bit of a lie (since it applies to fuzzyc as well), but especially since we're getting rid of fuzzyc soon, sounds fine

ursachec: ahh, I see. how about two versions then?

ursachec (approved):

PR #600, opened 2021-09-08
Add javasrc2cpg support and clean up repo

johannescoetzee:
# Notes
* Remove front-end scripts from repo root. Specific front-ends can now be invoked through joern-parse
* Add `javasrc2cpg` support to Joern. __NOTE: The front-end itself seems broken at the moment, but having it available in Joern would help with testing__
# Testing
`sbt clean stage` followed by manual tests to verify that the front-ends are still working as expected.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,6 +8,7 @@ ThisBuild /Test /fork := true
 val cpgVersion = "1.3.316"
 val ghidra2cpgVersion = "0.0.27"
 val js2cpgVersion = "0.2.3"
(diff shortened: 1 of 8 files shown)


mpollmeier (approved):
i didn't mind those linked executables in the repo root, but I can live without them as well :)

johannescoetzee:
> i didn't mind those linked executables in the repo root, but I can live without them as well :)

The idea of removing them actually came from a conversation with @fabsx00. It's slightly messy to keep them there, and unnecessary now that `joern-parse` supports them all.

PR #623, opened 2021-10-04
Add a list-languages option to joern-scan and joern-parse

johannescoetzee:
# Notes
Adds a `list-languages` option to list available languages with `joern-scan` and `joern-parse`, for example
```
❯ ./joern-scan --list-languages
Available languages (case insensitive):
- golang
- fuzzy_test_lang
- csharp
- java
- php
- c
- kotlin
- ghidra
- javascript
- python
- llvm
- newc
- javasrc
```

### DIFF
--- a/joern-cli/src/main/scala/io/shiftleft/joern/JoernParse.scala
+++ b/joern-cli/src/main/scala/io/shiftleft/joern/JoernParse.scala
@@ -1,7 +1,11 @@
 package io.shiftleft.joern
 
+import better.files.File
+import io.shiftleft.codepropertygraph.generated.Languages
 import io.shiftleft.console.{FrontendConfig, InstallConfig}
 import io.shiftleft.console.cpgcreation.{cpgGeneratorForLanguage, guessLanguage}
+
+import scala.jdk.CollectionConverters._
 
 object JoernParse extends App {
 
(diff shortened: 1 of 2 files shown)


joern-cli/src/main/scala/io/shiftleft/joern/JoernParse.scala
```diff
@@ -85,12 +111,13 @@ object JoernParse extends App {
...
                           enhanceOnly: Boolean = false,
-                          language: String = "c")
+                          language: String = "",
+                          listLanguages: Boolean = false)
 
   def parseConfig(parserArgs: List[String]): Either[String, ParserConfig] =
     new scopt.OptionParser[ParserConfig]("joern-parse") {
       arg[String]("input")
-        .required()
+        .optional()
```
ursachec, line 120: why should this be optional?

johannescoetzee: It should be possible to run `joern-parse --list-languages` without providing an input path. If the `--list-languages` option isn't specified, there's a check later to make sure that a valid input path has been provided.

joern-cli/src/main/scala/io/shiftleft/joern/JoernParse.scala
```diff
@@ -39,6 +51,21 @@ object JoernParse extends App {
     }
   }
 
+  def checkInputPath(config: ParserConfig): Either[String, Unit] = {
+    if (config.inputPath == "" || !File(config.inputPath).exists) {
+      Left("Invalid input path provided")
```
ursachec, line 56:
```suggestion
      Left("Input path does not exist at `" + config.inputPath  + "`, exiting.")
```

joern-cli/src/main/scala/io/shiftleft/joern/JoernParse.scala
ursachec: ah, ok, makes sense :+1:

ursachec (approved):

PR #681, opened 2021-11-04
integrate jimple2cpg

mpollmeier:
* also upgrades jimple's cpg dependency and adapts a few things
* add jimple to ImportCode options


fabsx00 (approved):

fabsx00:
Since we're copying over code and the history no longer shows who contributed what, I think we should list authors of modules in a prominent place to give them credit for their work. What do you guys think?

mpollmeier:
good idea, probably best on the main readme?

DavidBakerEffendi (commented):
Back in the land of internet from my road trip today but it looks good to me 👍

PR #682, opened 2021-11-04
Bump CPG version and fix c2cpg catch block orders

johannescoetzee:


ursachec (approved):
maybe add a test case as well?

fabsx00 (approved):

johannescoetzee:
> maybe add a test case as well?

There are already tests for this in [CfgCreationPassTests](https://github.com/joernio/joern/blob/master/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/CfgCreationPassTests.scala#L441-L474). This PR is is necessary since https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1503 changed how the CFG is generated again, thus breaking these tests again.

ursachec:
ahhh, I see, thanks for the explanation!

PR #686, opened 2021-11-04
rework/improve heuristics for language/frontend detection

mpollmeier:
* if input is a directory, scan it recursively, rather than just the top level
* do a group count of all files in given directory, rather than just searching for one language followed by the next etc. rationale: most projects have a mix of languages
* cleanup, docs, refactor
* add javasrc as potential language option


fabsx00 (approved):

mpollmeier:
tested with ocular, merging now

Issue #689, opened 2021-11-05
[javasrc2cpg] Inconsistent number of paths found in first `try` dataflow test


johannescoetzee:
The first testcase in the dataflow test suite for `try` statements usually finds 3 paths, but on rare occasions 2. It's unclear why this is happening, and I haven't been able to reproduce this issue through the `joern` console. My suspicion is that it's some weird concurrency bug caused by the interaction between the cpg generator and scalatest. If something similar pops up, this should be investigated more thoroughly.

fabsx00:
Fixed by https://github.com/joernio/joern/pull/695

PR #690, opened 2021-11-05
Require at least one path in inconsistent try test

johannescoetzee:
This testcase gives inconsistent results. It mostly passes, but sometimes fails with only 2 paths found. I was unable to reproduce the issue in the `joern` console, so believe it is an issue specific to the test. As such, I've updated the test to only require that some path is found. In the meantime, I've opened https://github.com/joernio/joern/issues/689 to keep track of this in case the same issue appears elsewhere.


johannescoetzee:
Closed in favour of https://github.com/joernio/joern/pull/693

PR #697, opened 2021-11-08
Include `lineNumberEnd` in method full names of calls to macros

fabsx00:
Depends on https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1516

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/MacroHandler.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/MacroHandler.scala
@@ -166,7 +166,7 @@ trait MacroHandler {
       val fileName = fileLocation.getFileName
       fileName + ":" + lineNo + ":" + lineNoEnd + ":" + name + ":" + argAsts.size
     } else {
-      "<empty>:-1:" + name + ":" + argAsts.size
+      "<empty>:-1:-1:" + name + ":" + argAsts.size
     }
   }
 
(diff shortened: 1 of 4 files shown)


ursachec (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/MacroHandler.scala
```diff
@@ -161,6 +151,25 @@ trait MacroHandler {
...
+    val name = macroDef.getName.toString
+    val fileLocation = macroDef.getFileLocation
+
+    if (fileLocation != null) {
+      val lineNo = fileLocation.getStartingLineNumber
+      val lineNoEnd = lineEnd(macroDef).getOrElse("-1")
+      val fileName = fileLocation.getFileName
+      fileName + ":" + lineNo + ":" + lineNoEnd + ":" + name + ":" + argAsts.size
+    } else {
+      "<empty>:-1:" + name + ":" + argAsts.size
```
johannescoetzee, line 169: Very minor nit, but in the failure case we're reporting one line number field, while in the success case we're reporting 2. If we (or anyone) wants to extract this information programmatically, it would be easier if the fields are uniform.

```suggestion
      "<empty>:-1:-1:" + name + ":" + argAsts.size
```

fabsx00: true, good catch, will fix

PR #698, opened 2021-11-08
Make the new C/C++ frontend the default frontend

fabsx00:
After plenty of testing and improving of the macro handling, with this PR, we make the new Eclipse-CDT-based C/C++ the default frontend for C/C++. This enables us to process modern C++ and handle the preprocessor without requiring the user to jump through configuration hoops. As is the true for the old frontend, the new frontend does NOT require a working build environment: if default header files are installed on the system, they will be taken into account. If they can be heuristically identified in the code base, they are picked up. If not, the analyzer continues with what it has and does not simply abort. This makes the tool ideal for analyzing code that is only partially available, code that comes from a decompiler, old code for which configuring a build environment is time consuming, and finally, settings where learning methods are applied to large amounts of code and configuring a working build environment for each project is not a viable option.

### DIFF
--- a/console/src/main/scala/io/joern/console/BridgeBase.scala
+++ b/console/src/main/scala/io/joern/console/BridgeBase.scala
@@ -199,8 +199,19 @@ trait BridgeBase {
     val language = config.language.getOrElse(
       io.joern.console.cpgcreation
         .guessLanguage(src)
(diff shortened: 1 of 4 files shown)


ursachec (approved):
:tada:

johannescoetzee (approved):

PR #710, opened 2021-11-10
Add `RefStoreToLoad` MIPS pass

ursachec:
- might incorrecty add too many edges in cases in which load/store instructions to the same address but with different values overlap in the specified CFG distance
- added `loadWord / storeWord` because 1) store instructions have a semantic of `1->2` while assignment has `2->1` and because it should make it easier to handle more complex DF tracking cases through memory stores


joern-cli/frontends/ghidra2cpg/src/main/scala/io/joern/ghidra2cpg/Ghidra2Cpg.scala
```diff
@@ -189,6 +190,9 @@ class Ghidra2Cpg(
         }
     }
 
+    if (proc == "MIPS") {
```
itsacoderepo, line 193: There is a match case for "MIPS". Why do you rewrite it and add an additional if statement?

itsacoderepo (changes requested):

joern-cli/frontends/ghidra2cpg/src/main/scala/io/joern/ghidra2cpg/Ghidra2Cpg.scala
ursachec: upsi, yes, that doesn't make sense, updated

ursachec:
the change is not very good, closing

PR #757, opened 2021-11-17
Add synchronized statements to javasrc2cpg

johannescoetzee:
Notes
* Add support for `synchronized` statements as named blocks. Conceptually it makes more sense to represent these as control structures, but since `synchronized` blocks specifically are a Java concept, this approach is simpler. 
* Some code cleanup.


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -929,6 +929,28 @@ class AstCreator(filename: String, global: Global) {
     AstWithCtx(switchAst, ctx)
   }
 
+  private def astForSynchronizedStatement(stmt: SynchronizedStmt,
+                                          scopeContext: ScopeContext,
+                                          order: Int): AstWithCtx = {
+    val parentNode =
+      NewBlock()
```
ursachec, line 936: wouldn't it be better for this to be a modifier? https://cpg.joern.io/#node-ref-modifier

ursachec (approved):

Issue #758, opened 2021-11-17
joern-scan: Global structs in tests not parsed correctly


raymontag:
I wanted to create a test case for a joern-scan query that contains a global struct definition which is not parsed. Something like

```
static const struct test test = {
  .test = some_func,
};
```

It follows the query file, the test file, and at the end a joern session to show that it works correctly in the interactive shell. The joern query prints the locals list to show that it the struct's local node is not created

```
package io.joern.scanners.c

import io.joern.scanners._
import io.shiftleft.semanticcpg.language._
import io.joern.console._
import io.shiftleft.console._
import io.shiftleft.macros.QueryMacros._
import io.joern.dataflowengineoss.language._
import io.joern.dataflowengineoss.queryengine.EngineContext
import io.joern.dataflowengineoss.semanticsloader.Semantics

object Test extends QueryBundle {

  @q
  def test(): Query =
    Query.make(
      name = "test",
      author = "me",
      title = "test",
      description = """
        |test
        |""".stripMargin,
      score = 2,
      withStrRep({ cpg =>
        def locals = cpg.local
        println(locals.l)
        locals
      }),
      tags = List(),
      codeExamples = CodeExamples( // Examples are modified excerpts from the Linux kernel
        List("""
          |	static const struct test test = {
          |		.test		= some_func,
          |	};       
          """.stripMargin),
        List("""
          """.stripMargin)
      )
    )
}
```

```
package io.joern.scanners.c

import io.joern.suites.CQueryTestSuite
import io.shiftleft.codepropertygraph.generated.nodes
import io.shiftleft.semanticcpg.language._
import overflowdb.traversal._
import io.shiftleft.semanticcpg.language._
import io.joern.console.scan._
import io.shiftleft.console.Query

class TestTests extends CQueryTestSuite {

  override def queryBundle = Test

  "test" in {
    val query = queryBundle.test()
      query(cpg)
      .flatMap(_.evidence)
      .cast[nodes.Local]
      .name
      .toSet shouldBe Set("test")
  }
}
```

```
joern> val code = """
                 |static const struct test test = {
                 |.test= some_func,
                 |}                
                 """.stripMargin 
code: String = """
static const struct test test = {
.test= some_func,
}                
          """

joern> importCode.c.fromString(code) 
[...]
joern> cpg.local.l 
res43: List[Local] = List(
  Local(
    id -> 3074457345618258674L,
    closureBindingId -> None,
    code -> "test",
    columnNumber -> None,
[...]

PR #767, opened 2021-11-19
use newc frontend instead of oldc in CQueryTestSuite

raymontag:
Fix for #758


fabsx00 (approved):

raymontag:
I noticed, that for the failing tests a `.whereNot(_.name("<global>"))` in the query does fix it but I guess it would be better to modify the test cases?

max-leuthaeuser:
> I noticed, that for the failing tests a `.whereNot(_.name("<global>"))` in the query does fix it but I guess it would be better to modify the test cases?

I guess a `.nameNot("<global>")` is ok. Lets wait for @fabsx00 answer.

fabsx00:
Yes, that should be ok, please go ahead.

raymontag:
I added a fix in the PR

max-leuthaeuser:
> I added a fix in the PR

This needs some `scalafmt`.

Issue #772, opened 2021-11-22
oldc vs newc Parsing-Difference for struct types


raymontag:
I noticed that the oldc and newc frontend parse the type of identifiers differently if the type is a struct. In particular, the oldc frontend writes `struct ...` while the newc frontend leaves the `struct` keyword from the type. That makes it difficult, e.g., to query if a struct was freed.

```
joern> val code = """
                 |
                 |struct test {
                 |int a;
                 |};
                 |
                 |void free_struct() {
                 | struct test *ptr;
                 | ptr = kzalloc(sizeof(struct test), GFP_KERNEL);
                 | free(ptr);
                 |}
                 | 
                 |void kfree_struct() {
                 | struct test *ptr;
                 | ptr = kzalloc(sizeof(struct test), GFP_KERNEL);
                 | kfree(ptr);
                 |}          
                 |""".stripMargin 


joern> importCode.oldc.fromString(code) 
...

joern> cpg.call("free").argument(1).evalType.l 
res15: List[String] = List("struct test *")

joern> importCode.c.fromString(code) 
...

joern> cpg.call("free").argument(1).evalType.l 
res17: List[String] = List("test*")
```

Probably related to #585

Issue #774, opened 2021-11-22
oldc vs newc Parsing-Difference for fixed size arrays


raymontag:
I noticed that in some cases the type of a fixed size array is not correctly parsed and the type of the identifier/local is set to ANY instead in the newc frontend.

```
joern> val code = """
                 |void bad1(size_t a) {
                 | uint8_t src[1], dst[1];
                 | memcpy(dst, src, a);
                 |}
                 |""".stripMargin 

joern> importCode.c.fromString(code) 
...

joern> cpg.call("memcpy").argument(1).evalType.l 
res36: List[String] = List("ANY")

joern> importCode.oldc.fromString(code) 
...

joern> cpg.call("memcpy").argument(1).evalType.l 
res38: List[String] = List("uint8_t [ 1 ]")

```

Probably related to #585

max-leuthaeuser:
Fixed with #775

PR #775, opened 2021-11-22
Fixed some differences for types between old c and new c frontend

max-leuthaeuser:
Fixes: https://github.com/joernio/joern/issues/772 and https://github.com/joernio/joern/issues/774


max-leuthaeuser:
This really is debatable.
The type in the linked issue actually should be `test*` and not `struct test*`. It would also make linking harder. But from a user perspective it's easier to find structs though as the issue creator argues. Otherwise, one would need to write an additional query to check if `test` is a struct.

fabsx00:
Agree with @max-leuthaeuser here... I'm wondering, can we maybe use the CODE field instead to hold `struct foo`?

max-leuthaeuser:
> Agree with @max-leuthaeuser here... I'm wondering, can we maybe use the CODE field instead to hold `struct foo`?

Yes, done.

fabsx00 (approved):

PR #828, opened 2021-12-16
Make javasrc method call AST structure more consistent with c2cpg

johannescoetzee:
* NB!! In the ossdataflowengine, definitions of identifiers will now kill any field access calls accessing an object matching that identifier. For example, the statement `x = new Box()` will kill previous definitions of `x.value()`, `x.value.toString()` etc.
* Add explicit `this` to method `code` fields where required to match `java2cpg`.
* For an explicit method call like `obj.someMethod()` or `this.someMethod()`, represent the left-hand side as a field access operation instead of just an identifier (to match c2cpg).
* Fix some testcases that were broken by this change.


fabsx00 (approved):
Noted, and nice catch!

PR #833, opened 2021-12-20
Preparations for Scala 3 compatibility.

ml86:
The Macros sub project does not dependent on Console anymore. Instead Console depends
on Macros. In addition the classes which need to be treated separately for Scala 2 and
Scala 3 are moved into the Macros sub project. To name them: Query.scala and
QueryDatabase.scala. We need to separate those since the Console project is not yet
ready for cross build because of its Ammonite dependency.
Next step will be enabling the cross build for the Macros sub project and adding the
Scala 3 code adjustements.
Maybe we should rename Macros sub project, what do you think?


fabsx00 (commented):
Yes, since one would not suspect `Query` and `QueryDatabase` to be in `macros`, it probably makes sense to rename it. And the reason for this move and the changes in dependency relations is purely because Ammonite doesn't build with scala3 yet?

fabsx00 (approved):

PR #834, opened 2021-12-20
Make "macros" sub project cross build.

ml86:
This PR contains two parts to enable cross build:
1. Side by side with the Scala 2 macro "withStrRep" we now have the
   Scala 3 macro version of it.
2. The "q" annotation was a Scala 2 annotation and now got changed
   to a Java runtime annotation. That way we can easily support
   both Scala versions without the need for two different code paths.
   This comes at the price of not being binary compatible with old versions
   of QueryDB, but we are source compatible. As discussed on Slack we deem
   this OK.


macros/src/main/scala-3/io/joern/macros/QueryMacros.scala
```diff
@@ -0,0 +1,22 @@
...
+object QueryMacros {
+
+  inline def withStrRep(inline traversal: Cpg => Traversal[_ <: StoredNode]): TraversalWithStrRep =
+    ${withStrRepImpl('{traversal})}
+
+  private def withStrRepImpl(travExpr: Expr[Cpg => Traversal[_ <: StoredNode]])
+                            (using quotes: Quotes): Expr[TraversalWithStrRep] = {
+    import quotes.reflect._
+    val pos = travExpr.asTerm.pos
+    val code = Position.apply(pos.sourceFile, pos.start, pos.end).sourceCode.getOrElse("N/A")
```
mpollmeier, line 19:
```suggestion
    val code = Position(pos.sourceFile, pos.start, pos.end).sourceCode.getOrElse("N/A")
```

macros/src/main/scala/io/joern/console/QueryDatabase.scala
```diff
@@ -1,26 +1,16 @@
 package io.joern.console
 
-import io.joern.console.Query
 import org.reflections8.Reflections
 import org.reflections8.util.{ClasspathHelper, ConfigurationBuilder}
-import org.slf4j.{Logger, LoggerFactory}
 
-import scala.annotation.{StaticAnnotation, unused}
```
mpollmeier, line 8: `unused` was used in this class, there's a compile error..

mpollmeier (approved):

ml86:
I incorporated your suggestions and also removed the Scala 2 reflection workaround code which should not be necessary anymore since we stop using Scala 2 reflection.

mpollmeier (approved):

PR #838, opened 2021-12-21
Add single file input support for javasrc2cpg

johannescoetzee:
JavaParser expects the input to be a source code directory (as opposed to a single file). This is somewhat annoying for testing and tinkering with the front-end, so this PR solves the problem by copying the input file to a temporary directory if a single-file input is given. 

The `fileName` used for CPG nodes will still be the original input file.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
@@ -1,5 +1,6 @@
 package io.joern.javasrc2cpg
 
+import better.files.File
 import io.joern.javasrc2cpg.passes.AstCreationPass
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.codepropertygraph.generated.Languages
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
```diff
@@ -47,4 +50,21 @@ class JavaSrc2Cpg {
...
+    * This is inconvenient for small-scale testing, so if a single source file is created,
+    * copy it to a temp directory.
+    */
+  private def getSourcesFromDir(sourceCodePath: String): (String, List[String]) = {
+    val sourceFile = File(sourceCodePath)
+    if (sourceFile.isDirectory) {
+      val sourceFileNames = SourceFiles.determine(Set(sourceCodePath), sourceFileExtensions)
+      (sourceCodePath, sourceFileNames)
+    } else {
+      val dir = Files.createTempDirectory("javasrc")
```
max-leuthaeuser, line 64: This temporary directory should be deleted / cleaned-up afterwards.

johannescoetzee: Ah, good catch!

max-leuthaeuser (approved):

PR #842, opened 2021-12-22
Removed code smells and cleaned-up

max-leuthaeuser:
Applied formatting, linting, and removal of unused imports and brackets.
Added types to public members.


max-leuthaeuser:
Won't take this, closed for now.
This needs to be done differently.

PR #845, opened 2021-12-23
Change joern-parse output option to -o or --output for consistency

johannescoetzee:
Most (all?) front-ends use the `-o` or `--output` options to specify an output file. It makes sense for `joern-parse` to use the same and this is what one would expect after running the front-ends directly.

This change would break any scripts that rely on the output option being `--out`, but given `joern-parse` was recently rewritten almost entirely, the impact of this seems limited.


max-leuthaeuser:
You could also deprecate `--out` with `.hidden()` instead of removing it completely. This way, old scripts would continue working.

fabsx00 (approved):

PR #846, opened 2021-12-29
Update to SBT 1.6.1

max-leuthaeuser:


mpollmeier:
while we're on it, can you update these to 1.6.1 which was released yesterday? :)

mpollmeier (approved):

PR #865, opened 2022-01-06
Introduce `kotlin2cpg`

ursachec:


joern-cli/frontends/kotlin2cpg/build.sbt
```diff
@@ -0,0 +1,30 @@
+name := "kotlin2cpg"
+
+scalaVersion := "2.13.7"
```
mpollmeier, line 3: you can drop a few things here, which are already defined for the entire build in /build.sbt:

scalaVersion
fork
resolvers
scalacOptions
onChangedBuildSource

mpollmeier (approved):

mpollmeier:
please address the comment before merging though

joern-cli/frontends/kotlin2cpg/build.sbt
ursachec: ah, good catch, thanks, will address

PR #868, opened 2022-01-07
Enable Scala 3 cross build for javasrc2cpg.

ml86:


johannescoetzee:
I'm holding back on approving since non-style-related checks are failing.

johannescoetzee (approved):

PR #881, opened 2022-01-12
Johannes/javasrc fixes

johannescoetzee:
# Notes
* Fix javasrc method calls and constructors for the closed source dataflow engine.
* General fixes for method call names and arguments
* Remove javasrc `ObjectInstantiationTests` since the relevant tests are now included in `ConstructorInvocationTests`
* Add reaching def edges from the last node in a block used as an argument to the call node (to support blocks as call arguments). 
* Fix for flows filtered out by `reachableByFlows` (by @fabsx00) if the source node is a call argument (and therefore invisible).
* Update c2cpg and fuzzyc2cpg dataflow expectations due to the fix above (some previously hidden flows now appear).


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/ReachingDefPass.scala
```diff
@@ -125,6 +127,35 @@ class ReachingDefPass(cpg: Cpg, maxNumberOfDefinitions: Int = 4000) extends Para
...
+              if (edgesToAdd.nonEmpty) {
+                addEdge(block, call)
+              }
+
+            case Some(node: Call) =>
+              addEdge(node, call, nodeToEdgeLabel(node))
+              addEdge(block, call)
+          }
+        }
+      }
```
johannescoetzee, line 158: This turned into a monolithic PR of stuff found while fixing javasrc2cpg constructors. This code snippet is the most important one that may impact other frontends.

max-leuthaeuser: Do we have to handle other types of nodes that may appear as last AST child in a block?
Otherwise, it may crash with a match error.

johannescoetzee: Ah, good catch! There are cases such as `return` and the various control structures that would break this.

fabsx00 (approved):
Looks good to me, go ahead.

PR #886, opened 2022-01-13
Fix Kotlin AST props around static calls

ursachec:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/passes/AstCreator.scala
```diff
@@ -1578,8 +1581,8 @@ class AstCreator(
       expr.getSelectorExpression() match {
         case selectorExpression: KtCallExpression =>
           withOrder(selectorExpression.getValueArguments()) { case (arg, order) =>
-            val selectorOrder = selectorOrderCount + order + 1
-            val selectorArgIndex = selectorOrder - 1
+            val selectorOrder = if (isStaticCall) order else selectorOrderCount + order + 1
+            val selectorArgIndex = if (isStaticCall) order else selectorOrder - 1
```
johannescoetzee, line 1585: Here the `order` and `argIndex` are the same for static calls, but different for dynamic dispatch calls. Should this be the case?

ursachec: yes (consider the case of a call with an implicit `this` - AST order is different from ARG_IDX), but the dynamic dispatch case is not yet handled 100% correctly, doing that in a separate PR.

johannescoetzee (approved):

PR #918, opened 2022-01-17
Parse Kotlin array access calls inside QEs

ursachec:


johannescoetzee (approved):
Looks good (if checks pass)

max-leuthaeuser (approved):

PR #950, opened 2022-01-21
Move javasrc type logic into separate class

johannescoetzee:
typeFullName logic was scattered all over the AstCreator which made it very difficult to figure out where type names came from and why things failed when they did. This PR pulls all the typeFullName logic into a single class to simplify this situation.

This change also moves away from using JavaParser's `getQualifiedName` for types, as this did not give consistent results compared to `java2cpg`. Specifically, for nested classes we had
```
// java2cpg
package.OuterClass$NestedClass

// javasrc2cpg
package.OuterClass.NestedClass
```
whereas now we would use the `$` notation for nested classes (although parsing those is currently broken and will be fixed in a separate PR).


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInfoProvider.scala
```diff
@@ -0,0 +1,308 @@
...
+class TypeInfoProvider(global: Global) {
+
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
+  /** Add `typeName` to a global map and return it. The
+    * map is later passed to a pass that creates TYPE
+    * nodes for each key in the map.
+    */
+  private def registerType(typeName: String): String = {
+    global.usedTypes.put(typeName, true)
```
ursachec, line 53:
```suggestion
    global.usedTypes.putIfAbsent(typeName, true)
```

ursachec (approved):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
```diff
@@ -36,9 +36,11 @@ class AstCreationPass(codeDir: String, filenames: List[String], cpg: Cpg, keyPoo
     val parser = new JavaParser(parserConfig)
     val parseResult = parser.parse(new java.io.File(filename))
 
+    val typeInfoProvider = TypeInfoProvider(global)
```
ursachec, line 39: soft suggestion, I recently renamed `kotlin2cpg`s _TypeInfoProvider_ to _NameGenerator_ because I think the name fits better https://github.com/joernio/joern/commit/b30c221fd0e8fbf1ba21637684e22df5ce67453e

johannescoetzee: The PR where you renamed it was actually what inspired this PR. I was undecided on which name to choose for this one, but figured _TypeInfoProvider_ fit better in this situation (since they are slightly different and this one doesn't exactly _generate_ names per se). I'm not convinced that _TypeInfoProvider_ is the best name in this case either, with something like _TypeResolver_ being closer to accurate, but it doesn't seem overly confusing, at least.

PR #961, opened 2022-01-25
Jimple2Cpg: Array/Multi-array Support + Single File Load Bug Fix

DavidBakerEffendi:
- Fixed access path issue where array index accesses were reported to be invalid ASTs. This was just a change in AST children's `order` from `(0, 1)` to `(1, 2)`
- Updated frontend to leverage `Call(<operator>.arrayInitializer)` instead of `Unknown(new)` vertices
- Support for multi-array creation added
- Array tests derived from JavaSrc2Cpg included
- Fixed bug where if a single file was specified then all files in the directory were loaded

### DIFF
--- a/joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/Jimple2Cpg.scala
+++ b/joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/Jimple2Cpg.scala
@@ -60,17 +60,17 @@ class Jimple2Cpg {
     try {
       // Determine if the given path is a file or directory and sanitize accordingly
       val rawSourceCodeFile = new JFile(rawSourceCodePath)
-      val normSourceCodePath = rawSourceCodeFile.toPath.toAbsolutePath.normalize.toString
-      val sourceCodePath = if (rawSourceCodeFile.isDirectory) {
-        normSourceCodePath
+      val sourceTarget = rawSourceCodeFile.toPath.toAbsolutePath.normalize.toString
+      val sourceCodeDir = if (rawSourceCodeFile.isDirectory) {
+        sourceTarget
       } else {
         Paths
-          .get(new JFile(normSourceCodePath).getParentFile.getAbsolutePath)
+          .get(new JFile(sourceTarget).getParentFile.getAbsolutePath)
           .normalize
           .toString
       }
 
-      configureSoot(sourceCodePath)
+      configureSoot(sourceCodeDir)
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/Jimple2Cpg.scala
```diff
@@ -116,6 +113,26 @@ class Jimple2Cpg {
...
+      sourceCodePath: String,
+      sourceFileExtensions: Set[String],
+      archives: List[String]
+  ) = {
+    (archives
+      .map(new ZipFile(_))
+      .flatMap(x => {
+        unzipArchive(x, sourceCodePath) match {
+          case Failure(e) =>
+            logger.error(s"Error extracting files from archive at ${x.getName}", e); null
```
mpollmeier, line 126: `null` is evil, please don't use it, to avoid NullPointers. 
In Scala, we have Try, which you already use in `unzipArchive`. You can decide if you want to return a `Try` here or just deal with the happy path, in which case you should throw an exception in this line.

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/Jimple2Cpg.scala
```diff
@@ -116,6 +113,26 @@ class Jimple2Cpg {
...
   }
 
+  private def loadSourceFiles(
+      sourceCodePath: String,
+      sourceFileExtensions: Set[String],
+      archives: List[String]
+  ) = {
+    (archives
+      .map(new ZipFile(_))
+      .flatMap(x => {
```
mpollmeier, line 123: This method is quite hard to read, partly because it does many things, partly because it's quite nested. A few ideas:
* extract into smaller methods
* introduce named helper variables
* rewrite map/flatMap into for comprehension
* less parens: `.flatMap(x => {...})` -> `.flatMap { x => ... }`

mpollmeier (approved):

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/Jimple2Cpg.scala
DavidBakerEffendi: Yeah I realise this method has become a bit of a mess as edge cases have been coded for on top of this method over time. I'll take some time to fix this up

mpollmeier (approved):

PR #990, opened 2022-02-03
Stub TypeDecls for Types from Header Files

max-leuthaeuser:
For C/C++ types from header files without any accessible implementation
we might end up with CPG Types without a corresponding CPG TypeDecl.
As these types may be referenced via `TypeDecl.aliasTypeFullName` e.g., for C++ type alias or typedefs we previously crashed e.g., in `TypeDeclTraversal.canonicalType`.

This PR stubbs them.

Fixes: https://github.com/ShiftLeftSecurity/codescience/issues/5784


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/HeaderContentPass.scala
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  private def typeNeedsTypeDeclStub(t: Type): Boolean =
+    cpg.graph.nodes(NodeTypes.TYPE_DECL).has(Properties.FULL_NAME, t.typeDeclFullName).isEmpty
+
+  private def createMissingTypeDecls(dstGraph: DiffGraph.Builder): Unit = {
+    Traversal(cpg.graph.nodes(NodeTypes.TYPE))
+      .foreach {
+        case t: Type if typeNeedsTypeDeclStub(t) =>
+          val newTypeDecl = NewTypeDecl()
```
mpollmeier, line 91:
```suggestion
    Traversal(cpg.graph.nodes(NodeTypes.TYPE))
      .filter(typeNeedsTypeDeclStub)
      .foreach { t =>
        val newTypeDecl = NewTypeDecl()
```
matter of personal style preference, so feel free to ignore, but IMO the above is more readable and concise. you can leave out the `case _` as well.

mpollmeier (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/HeaderContentPass.scala
max-leuthaeuser: Hm... we need to know that `t` is a `Type` to have access to `name` and `typeDeclFullName`.
Hence, the explicit `case`.

mpollmeier: ah i see, yeah that makes sense, ignore my comment :)

PR #993, opened 2022-02-04
Fixed handling of static initializers and calls

max-leuthaeuser:
Calls for statically initialized variables/members (in enums) or static calls in general (in elaborated types) are now attached to a fake initialization method rather than their defining TypeDecl.


mpollmeier:
re https://discord.com/channels/832209896089976854/832229141355954176/939054357578674206

mpollmeier (approved):
tested locally, fixes the issue from the above linked discord

PR #1018, opened 2022-02-15
Fixed c2cpg system header file discovery

max-leuthaeuser:
Before this PR we always included all system header files when creating a CDT scanner.
Without checking whether it is a C or a CPP file.

This PR checks the file to parse beforehand.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/parser/CdtParser.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/parser/CdtParser.scala
@@ -50,11 +50,10 @@ class CdtParser(parseConfig: ParserConfig, headerFileFinder: HeaderFileFinder)
   }
 
   private def createScannerInfo(file: Path): ScannerInfo = {
-    if (FileDefaults.isCPPFile(file.toString)) {
-      new ScannerInfo(definedSymbols, (includePaths ++ parseConfig.systemIncludePathsCPP).map(_.toString).toArray)
-    } else {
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/parser/CdtParser.scala
```diff
@@ -49,12 +49,21 @@ class CdtParser(parseConfig: ParserConfig, headerFileFinder: HeaderFileFinder)
     }
   }
 
+  private def createScannerInfo(file: Path): ScannerInfo = {
+    if (FileDefaults.isCPPFile(file.toString)) {
+      new ScannerInfo(definedSymbols, (includePaths ++ parseConfig.systemIncludePathsCPP).map(_.toString).toArray)
+    } else {
+      new ScannerInfo(definedSymbols, (includePaths ++ parseConfig.systemIncludePathsC).map(_.toString).toArray)
+    }
+  }
```
mpollmeier, line 58: something like this to remove duplication?
```suggestion
    val additionalIncludes = 
      if (FileDefaults.isCPPFile(file.toString)) parseConfig.systemIncludePathsCPP
      else parseConfig.systemIncludePathsC
      
    new ScannerInfo(definedSymbols, (includePaths ++ additionalIncludes).map(_.toString).toArray)
  }
```

mpollmeier (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/parser/CdtParser.scala
max-leuthaeuser: Done. Thanks!

PR #1019, opened 2022-02-15
Fix some javasrc crashes

johannescoetzee:


ursachec (approved):
looks good, would be even nicer to have a small test case for the scenario (useful for example as a template for other situations in which this type of error might occur)

johannescoetzee:
> looks good, would be even nicer to have a small test case for the scenario (useful for example as a template for other situations in which this type of error might occur)

Noted. I still need to figure out exactly what is triggering this, so will update with a testcase once I'm back.

PR #1023, opened 2022-02-16
Johannes/javasrc diffgraph

johannescoetzee:


johannescoetzee:
This still needs to be done, but it'll probably be easier to just to it in a new branch by now.

PR #1032, opened 2022-02-20
Fix #974

### LINKED ISSUE #974: [newc] Missing data flow
Given the following code it seems that it is not possible to find a data flow starting from `foo`:

```
#include <stdio.h>
#include <string.h>

typedef struct {
  char *buf1;
} FooStruct;

void doFoo(FooStruct *str) {
  char onebuf[100];
  memset(onebuf,'1',sizeof onebuf);
  str->buf1 = onebuf;
}

int main(void) {
  FooStruct foo;
  doFoo(&foo);
  return 0;
}
```

Joern queries:
```
importCode("/tmp/test")
run.ossdataflow
def source = cpg.local.name("foo").referencingIdentifiers
def sink = cpg.call.code("doFoo.*").argument
sink.reachableByFlows(source).p
```

image

fabsx00:


joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/querying/DataFlowTests.scala
```diff
@@ -1741,3 +1741,27 @@ class DataFlowTest58 extends DataFlowCodeToCpgSuite {
...
+      | FooStruct foo;
+      | doFoo(&foo);
+      | return 0;
+      |}
+      |""".stripMargin
+
+  "should find flow from local to 'doFoo'" in {
+    def source = cpg.local.name("foo").referencingIdentifiers
+    def sink   = cpg.call.code("doFoo.*").argument
+    sink.reachableByFlows(source).p.foreach(println)
```
max-leuthaeuser, line 1764: Proper test the result here?

itsacoderepo (approved):

fabsx00:
Thanks, yeah, that was still debug code rather than the actual test.

PR #1044, opened 2022-02-23
Several improvements for c2cpg

max-leuthaeuser:
* added support for C++ Array Designators
 * added support for dangling declarations from unresolved macros (happens e.g., in/from system header files)
 * replaced CDT file reading in our custom file content provider with the one from IOUtils dealing with non-UTF8 stuff correctly
 * simplified the CdtParser interface

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForTypesCreator.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForTypesCreator.scala
@@ -248,7 +248,10 @@ trait AstForTypesCreator {
               case d =>
                 astForDeclarator(declaration, d, order)
             }
-          case _ if nodeSignature(declaration) == ";"  => Seq.empty
+          case _ if nodeSignature(declaration) == ";" =>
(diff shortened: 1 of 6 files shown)


ursachec (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -248,7 +248,10 @@ trait AstForTypesCreator {
               case d =>
                 astForDeclarator(declaration, d, order)
             }
-          case _ if nodeSignature(declaration) == ";"  => Seq.empty
+          case _ if nodeSignature(declaration) == ";" =>
+            Seq.empty // dangling decls from unresolved macros; we ignore them
```
ursachec, line 252: i'd add a test for these

max-leuthaeuser: I was not yet able to reproduce this in a clean unit test case.
But it did happen e.g., when scanning VLC (wild combination of missing defines and system header files).
We want some actual tests for that - I agree. But I need to investigate this further and do a separate PR.

PR #1062, opened 2022-02-28
Exit early if no `.kt` files present in input dir

ursachec:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Main.scala
```diff
@@ -61,6 +61,12 @@ object Main extends App {
           println("The input path provided is not a directory `" + sourceDir + "`. Exiting.")
           System.exit(1)
         }
+
+        val filesWithKtExtension = File(sourceDir).listRecursively.filter(f => f.extension == Some(".kt"))
```
max-leuthaeuser, line 65: You could re-use `io.joern.x2cpg.SourceFiles.determine` for this job like we do in other frontends already.

ursachec: Ah, yes, thanks, updated

max-leuthaeuser (approved):

PR #1073, opened 2022-03-03
Migrate cpg passes to new style, using odb BatchedUpdate

bbrehm:
This relies on https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1597

### DIFF
--- a/joern-cli/frontends/x2cpg/build.sbt
+++ b/joern-cli/frontends/x2cpg/build.sbt
@@ -30,7 +30,9 @@ scalacOptions ++= Seq() ++ (
         "-language:implicitConversions", // Allow definition of implicit functions called views
         "-unchecked",                    // Enable additional warnings where generated code depends on assumptions.
         "-Xcheckinit",                   // Wrap field accessors to throw an exception on uninitialized access.
-        "-Xfatal-warnings",              // Fail the compilation if there are any warnings.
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/x2cpg/build.sbt
```diff
@@ -30,7 +30,9 @@ scalacOptions ++= Seq() ++ (
         "-language:implicitConversions", // Allow definition of implicit functions called views
         "-unchecked",                    // Enable additional warnings where generated code depends on assumptions.
         "-Xcheckinit",                   // Wrap field accessors to throw an exception on uninitialized access.
-        "-Xfatal-warnings",              // Fail the compilation if there are any warnings.
+        //"-Xfatal-warnings",            //-Werror is incompatible with the concept of @deprecate.
```
max-leuthaeuser, line 33: Can we put a `@scala.annotation.nowarn` where required instead?

fabsx00 (approved):

joern-cli/frontends/x2cpg/build.sbt
bbrehm: Can you explain in more detail?

The deprecation warnings are intended and we want them. But they are no error and should not be treated as such.

max-leuthaeuser: > The deprecation warnings are intended and we want them.

This was the missing piece, sorry. I thought suppressing them for now would be ok.

PR #1101, opened 2022-03-10
Exit 1 if resulting Kotlin CPG has no METHOD nodes

ursachec:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Main.scala
```diff
@@ -145,7 +146,12 @@ object Main extends App {
 
         val typeInfoProvider = new DefaultTypeInfoProvider(environment)
         val cpg = new Kt2Cpg().createCpg(filesWithMeta, fileContentsAtPath, typeInfoProvider, Some(config.outputPath))
+        val hasAtLeastOneMethodNode = cpg.method.take(1).size > 0
         cpg.close()
+        if (!hasAtLeastOneMethodNode) {
+          println("Error: resulting CPG does not contain any METHOD nodes.")
+          System.exit(1)
```
mpollmeier, line 153: what if i scan an empty artifact, wouldn't that result in 0 method nodes and still be correct?

max-leuthaeuser:
Kotlin allows global variables outside of methods. Where are they stored? We had the same issue with c2cpg and js2cpg where we fixed that by adding a fake global method. With a similar situation here we would always have a method at the end, even if the input code does not contain any user defined method.

ursachec:
@max-leuthaeuser thanks, that's a good note. Global variables are not stored anywhere at the moment, that is still a to-do but not with very high priority since i don't see them as critical as other things (like gathering as much type information as possible for a specific project) for finding vulns. i'd merge this as is and change the `nonEmpty` to `> 1` when globals are in

max-leuthaeuser (approved):

PR #1103, opened 2022-03-11
Added commonly used data structures to x2cpg

DavidBakerEffendi:
Following off of @max-leuthaeuser's structure on `js2cpg` where these are separated into a package, I'm pulling some of these commonly used data structures into `x2cpg` to avoid duplication on this end.

Note `ExternalCommand` is added since `solidity2cpg` will be calling an `npm` package too.

Should this PR also reformat other frontends to adopt these or shall each frontend's maintainer do it in their own?

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/Global.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/Global.scala
@@ -1,0 +1,9 @@
+package io.joern.x2cpg.datastructures
+
+import java.util.concurrent.ConcurrentHashMap
+
+class Global {
+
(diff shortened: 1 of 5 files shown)


fabsx00:
This means we just have copies of these utilities in x2cpg now, right? Maybe best to remove them from `jssrc2cpg` then and have `jssrc2cpg` call these utilities in `x2cpg`. I guess we'd have to merge `jssrc2cpg` to master first for that though.

DavidBakerEffendi:
> 

Alternatively, we could merge this into master and have @max-leuthaeuser rebase `js2cpg`.

max-leuthaeuser:
> > I guess we'd have to merge jssrc2cpg to master first for that though.
> 
> Alternatively, we could merge this into master and have @max-leuthaeuser rebase `js2cpg`.

Yes, will do that.

fabsx00:
Feel free to merge then.

PR #1105, opened 2022-03-11
jimple2cpg: use `X2CpgFrontend`

fabsx00:
Re: https://github.com/joernio/joern/pull/1102


joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/Jimple2Cpg.scala
```diff
@@ -74,7 +69,6 @@ class Jimple2Cpg {
       }
 
       configureSoot()
-      val cpg             = newEmptyCpg(outputPath)
       val metaDataKeyPool = new IntervalKeyPool(1, 100)
```
DavidBakerEffendi, line 72: Do we need these key pools?  It's only Plume that needs these. Can remove if it's within the scope of this PR.

DavidBakerEffendi (approved):
Happy with the update

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/Jimple2Cpg.scala
fabsx00: Done

max-leuthaeuser (approved):

PR #1113, opened 2022-03-15
Default to type ANY instead of <empty> if javasrc type info is missing

johannescoetzee:
Using a default `<empty>` type instead of `ANY` if a type could not be resolved resulted in missing `TYPE_EVAL` edges causing cpg2sp crashes (specifically tested with HelloShiftLeft). Defaulting to the `ANY` type instead fixes this issue, while also being a better representation for the missing types in question.


mpollmeier (approved):
💯 just tested: this fixes the `NoSuchElementException` in cpg2sp

PR #1116, opened 2022-03-17
Fix javasrc super types

johannescoetzee:
This PR fixes some inconsistencies between static/dynamic dispatch logic and when `this` nodes are created for implicit `this` arguments. It also fixes the type of the `this` node for the `<init>` call used to represent `super`.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInfoProvider.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInfoProvider.scala
@@ -326,16 +326,11 @@ class TypeInfoProvider(global: Global) {
 
   def scopeType(scopeContext: ScopeContext, isSuper: Boolean = false): String = {
     scopeContext.typeDecl match {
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInfoProvider.scala
```diff
@@ -323,6 +323,22 @@ class TypeInfoProvider(global: Global) {
...
+        if (isSuper) {
+          val parentType =
+            typ.inheritsFromTypeFullName.headOption
+              .getOrElse("ANY")
+          registerType(parentType)
+        } else {
+          registerType(typ.fullName)
+        }
+
+      case None => "ANY"
```
mpollmeier, line 339:
```suggestion
      case Some(typ) if isSuper =>
        val parentType = typ.inheritsFromTypeFullName.headOption.getOrElse("ANY")
        registerType(parentType)
      case Some(typ) =>
        registerType(typ.fullName)
      case None => "ANY"
```

matter of personal style I guess, so feel free to disregard

mpollmeier (approved):

PR #1131, opened 2022-03-21
Add "inference jar" functionality to javasrc2cpg

johannescoetzee:
This PR adds an option to include "inference jars" when creating CPGs with javasrc2cpg. Currently all information about external dependencies is missing, so with these changes at least the type information about the dependencies is retrieved. Note that the method and type decl stubs that are created with the usual passes are still "empty", so the benefits when attempting to find data flows may be limited.


johannescoetzee:
There is still a discussion to be had about how to deal with the larger issue of missing types and finding this information, but I'm merging this PR since the changes included are entirely optional, seem to work as expected and are already potentially useful

PR #1160, opened 2022-03-23
Improve c2cpg performance.

ml86:
- Moved ast creation out of global lock in order to avoid threads waiting for the
  lock the majority of the time.
- Improve column calculation algorithm.

    The old version managed to be slow enough to be the number one hot spot
    during profiling of c2cpg. The new version makes this fast so that it does
    not show up during performance measuring anymore.

    Column number now start at one instead of zero like the line numbers and
    the new implemention obsoleted a bug in the old column end calculation
    which led to complete bogus numbers.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
@@ -1,6 +1,5 @@
 package io.joern.c2cpg.astcreation
 
-import io.joern.c2cpg.utils.IOUtils
 import io.shiftleft.codepropertygraph.generated.nodes.{NewBlock, NewNode}
 import io.shiftleft.codepropertygraph.generated.{DispatchTypes, Operators}
 import io.shiftleft.codepropertygraph.generated.nodes.NewMethod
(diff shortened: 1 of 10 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -58,9 +59,25 @@ trait AstCreatorHelper {
     }
   }
 
-  private def fileLines(node: IASTNode): Seq[Int] = {
+  private def fileOffsetTable(node: IASTNode): Array[Int] = {
     val f = fileName(node)
-    global.file2LinesCache.computeIfAbsent(f, _ => IOUtils.readLineLengthsInFile(Paths.get(f)))
+    global.file2OffsetTable.computeIfAbsent(f, _ => genFileOffsetTable(Paths.get(f)))
```
max-leuthaeuser, line 64: Then we could remove `IOUtils.readLineLengthsInFile` as its not needed any longer.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -58,9 +59,25 @@ trait AstCreatorHelper {
...
+    asString = asString.replaceAll("\r\n", "\n")
+    asString = asString.replaceAll("\r", "\n")
+    val asCharArray = asString.toCharArray
+    val offsets     = mutable.ArrayBuffer.empty[Int]
+
+    for (i <- Range(0, asCharArray.length)) {
+      if (asCharArray(i) == '\n') {
+        offsets.append(i + 1)
+      }
+    }
```
max-leuthaeuser, line 79: Is this truly platform-independent? The IOUtils file reading from cpg does quite some stuff to handle all kinds of problems with charset encoding and unmappable characters.

max-leuthaeuser (approved):
Just some nits. Thanks a lot! Should bring in some drastic performance improvements.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
ml86: I changed the file reading to use IOUtils.

PR #1179, opened 2022-03-25
Fix props and structure for kt componentN fns

ursachec:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/passes/AstCreator.scala
```diff
@@ -658,19 +665,21 @@ class AstCreator(fileWithMeta: KtFileWithMeta, xTypeInfoProvider: TypeInfoProvid
             NewIdentifier()
               .typeFullName(typeDecl.fullName)
               .code(Constants.this_)
-              .argumentIndex(0)
+              .name(Constants.this_)
+              .argumentIndex(1)
```
johannescoetzee, line 669: This seems odd to me. Why should this not be 0?

johannescoetzee: From an offline discussion and double checking at least javasrc2cpg, this isn't actually odd at all. It looks like this is expected for field access expressions.

With that being said, a note somewhere in documentation would be helpful.

johannescoetzee (approved):

PR #1182, opened 2022-03-25
Remove unnecessary prefix from test helper class

ursachec:
the substring _kotlin2cpg_ is already part of the FQN, no need to have it twice


johannescoetzee (approved):

johannescoetzee:
This one is a matter of style. This approach is cleaner than the old one, but does come at the cost of consistency with other frontends' code.

ursachec:
yes, good point, we should change the other ones as well soonish

PR #1212, opened 2022-04-01
Use tmp dir as a gradle home

note:
Right now `kt2cpg` fails if you don't have `~/.gradle`. You can reproduce it by `mv ~/.gradle ~/.gradle2`


ursachec:
Discussed offline: we'll keep the user's default gradle home because it might contain logic necessary for resolving a project's depenencies; setting it explicitly could break it

ursachec:
see https://github.com/joernio/joern/pull/1218 for a fix of the currently broken functionality

PR #1215, opened 2022-04-02
dataflowengineoss: Option to Disable Cache Use + Some Simple Documentation

DavidBakerEffendi:
- For experimental purposes to evaluate the effectiveness of the cache, an option to disable re-use of already calculated results. Right now we see the overhead of collecting results to give back to the engine context being quite slow despite the improve cache hit ratio. By measuring the speed of a fully disabled cache we can approximate how useful the current caching strategy is.
- Added some basic documentation to make the engine friendlier

PS: Running experiments locally show that disabling the cache basically doubles memory consumption and increases analysis time by loads!

Example of results from giving `initialTable` to `EngineContext` to re-use results on an m5d.large EC2 instance:

## Cache Hit Ratios
dflow_cache
## `reachableByDetailed` performance (+ collecting results overhead)
taint

### DIFF
--- a/dataflowengineoss/README.md
+++ b/dataflowengineoss/README.md
@@ -1,0 +1,49 @@
+# Data Flow Engine
+
+A taint-tracking system based on whole-program data-dependence 
+representation. External library calls can be defined by 
+semantic models (see `src/test/resources/default.semantics`)
+
+Basic usage:
+
+```scala
+// If using Joern shell, imports and engine context will be pre-configured and available already
+import io.shiftleft.semanticcpg.language._
+import io.joern.dataflowengineoss.language.toExtendedCfgNode
+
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi:
Disabling the cache under `ReachableByCallable::results:354` was too hardcore i.e.: 
```scala
val cachedResult = if (context.disableCache) None else table.createFromTable(parent, path)
```
Since the cache is really baked into how the problem is solved. Have instead decided to not share cache between local searches and see how that affects the speed.

DavidBakerEffendi:
Here are the results on my laptop. "Do Not Share Cache With Local Search" is this PR with `disableCache = true`.
taint
dflow_cache

fabsx00 (approved):

PR #1231, opened 2022-04-05
Automatically set `order` property

fabsx00:
* If `order` is not set explicitly, set it automatically according to the child's position among its siblings.
* Compatibility pass: for method parameters, if `index` is not set, copy `order` to `index`
* In traversals, to select the n'th parameter, use the `index` field as opposed to the `order` field
* For all frontends except `pysrc2cpg`, do not manually set `order` on parameters
* Quite a bit of work on removing setting of the `order` field from `c2cpg` and `javasrc2cpg`

Do not merge yet. Still need to look into how this affects `codescience`.


ursachec (approved):

fabsx00:
Closing to bring in step by step.

PR #1238, opened 2022-04-07
jimple2cpg: Fixed fields by using correct canonical name and code property

DavidBakerEffendi:
- Added field access tests
- Corrected `canonicalName` property
- Corrected LHS of the `code` property


johannescoetzee (approved):
The changes look fine, so approving assuming that tests will pass.

PR #1249, opened 2022-04-09
WIP: Pointer Assignment Graph (PAG) Pass

DavidBakerEffendi:
- The first step in making Joern handle pointers correctly is adding "may point to" information for each identifier
- The pass only works intraprocedurally and any subsequent pass (e.g. call graph, reaching def) that wants to use pointer information must resolve inter-procedural pointer info itself
- This pass first builds an "assignment graph" which identifies where identifiers were first declared (relies on `<operator>.alloc` and `<operator>.arrayInitializer`. This is then used to finds potential roots of where identifiers may have been aliased.

Misc:
- Under [TypeTests.scala](https://github.com/joernio/joern/pull/1249#diff-d097c85d6f5e20fcdaa16a8a7f43ae2c37ca2dc3b92d6f830eb1a8937b8dafd3)  I've fixed a wild `println` and fixed a typo.

TODO:
- [x] This PR is dependent on other PRs coming in from `codepropertygraph` and `joern` 
- [x] Create a dot generator for these edges
- [x] Fix bug where PAG only points to one alloc site

Design decisions to consider:

In order to get the pointer assignment graph, an assignment graph is built. *Should this be it's own pass?* 

This simply points to where a variable was created/declared. This chain is then followed until an allocation point and this is how the pointer assignment graph is created. Could be useful for variables created as receivers of methods when resolving interprocedural points-to info.

Right now method parameters are not considered as allocation sites but perhaps they should be considered as assignment sites for the assignment graph. Example: the following code will lack points-to information since there is no allocation site present - rather the allocation site must be outside of `faz`. If we consider parameter `a` an allocation site then we will see a result of `r` and `p` pointing to `a`.

```java
 static void faz(Object a) {
   Object r = a;
   Object p = r;
 }
 ```

### CODE REVIEW

**Comment by DavidBakerEffendi** [comment]
Example points to graph for:
```java
static void foo() {
   var p = new A(); // alloc 1
   var q = p;
   var r = new B(); // alloc 2
   p.f = r;
   var t = bar(q);
   t.m();
 }
 ```

graphviz

File: joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/typerelations/PointerAssignmentPass.scala [hunk 1]
```diff
@@ -0,0 +1,143 @@
...
+  override def generateParts(): Array[Method] = cpg.method.toArray
+
+  override def runOnPart(builder: DiffGraphBuilder, method: Method): Unit = {
+    val allocSites = method.ast.isCall.name(Operators.alloc, Operators.arrayInitializer).l
+    if (allocSites.nonEmpty) {
+      val assignmentGraph = buildAssignmentGraph(method)
+      val pointsToGraph   = buildPointsToGraph(assignmentGraph)
+      // All references to nodes in the method body should now have their
+      // "may point to" edge added to an allocation site if it is in the map
+      method.ast
```
bbrehm, line 26: cc @ml86 -- I vaguely remember that some frontends do the terrible thing that they emit inner methods below the AST of the outer method. Is this still correct?

If so, then we need to go via the contains-edge here (the corresponding pass takes that possibility into account)

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/typerelations/PointerAssignmentPass.scala
```diff
@@ -0,0 +1,143 @@
...
+
+/** Creates an intraprocedural pointer assignment graph. This provides may point to information for each identifier or
+  * field identifier.
+  */
+class PointerAssignmentPass(cpg: Cpg) extends ForkJoinParallelCpgPass[Method](cpg) {
+
+  override def generateParts(): Array[Method] = cpg.method.toArray
+
+  override def runOnPart(builder: DiffGraphBuilder, method: Method): Unit = {
+    val allocSites = method.ast.isCall.name(Operators.alloc, Operators.arrayInitializer).l
```
bbrehm, line 20: Same potential issue as below wrt inner methods.

You only need to check whether allocSites is nonempty, not find all of them, right? You can use headOption for that.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/typerelations/PointerAssignmentPass.scala
```diff
@@ -0,0 +1,143 @@
...
+          }
+        }
+    }
+  }
+
+  /** Builds a graph where identifiers on the left of assignment statements are mapped to nodes on the right hand side.
+    */
+  private def buildAssignmentGraph(m: Method): Map[VarInCtx, Set[VarInCtx]] = {
+    val assignmentGraph               = mutable.HashMap.empty[VarInCtx, Set[VarInCtx]]
+    var worklist: ListBuffer[CfgNode] = mutable.ListBuffer.from(m.ast.isExpression.assignment.l)
```
bbrehm, line 46: Consider something like `val worklist = mutable.ArrayDeque.from(...)`.

bbrehm: `assignment` is a very API, basically because its implementation is abusing the database (it violates the assumption that every NodeDb has a unique NodeRef refering to it). We could not find consensus on a solution or removal of the API, but I would not be surprised if this triggered bugs when swapping. I would prefer if you did this directly here (i.e. duplicate the logic that checks whether the call is `Operators.assignment`)

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/typerelations/PointerAssignmentPass.scala
```diff
@@ -0,0 +1,143 @@
...
+
+    while (worklist.nonEmpty) {
+      val assignment = worklist.head
+      worklist = worklist.tail
+      // Extract LHS and RHS of each expression
```
bbrehm, line 51: If you want to keep using the problematic asignment API, then consider `.source` and `.target`. Since you already do the extraction manually, it's very cheap to stop using the assignment API.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/typerelations/PointerAssignmentPass.scala
```diff
@@ -0,0 +1,143 @@
...
+      val allocLhs: Set[VarInCtx] = assignment.astChildren
+        .orderLte(1)
+        .collect {
+          case x: Identifier => VarInCtx(x)
+          case x: Call if x.methodFullName == Operators.fieldAccess =>
+            x.astChildren.collect { case x: FieldIdentifier => VarInCtx(x) }.head
```
bbrehm, line 57: This is super problematic. An assignment
```
foo.fieldname = y;
```
will now seemingly overwrite the `fieldname` field of all objects that have a field named `fieldname` and forget about `foo`.

Also, `alloclhs` is an Option, no need to make a Set out of it.

bbrehm (commented):
Some nits. 

Can we have a call and you explain to me what you're trying to do? Or can you link me a description of the alg you're implementing here?

Apologies that the review is so late, was out sick last week.

DavidBakerEffendi:
> Some nits.
> 
> Can we have a call and you explain to me what you're trying to do? Or can you link me a description of the alg you're implementing here?
> 
> Apologies that the review is so late, was out sick last week.

Glad to see you're back and feeling well.

Yeah a call would be good - given some thinking I realise there were some things I was creating that was duplicating work. I suppose I should also clarify that this is a intra-procedural version of an exhaustive pointer analysis algorithm (which is slow so most real-world implementations have an on-the-fly version). So let's first arrange for a call before I get to making review changes.

Using the PAG algorithm from [here](http://www.sable.mcgill.ca/publications/techreports/sable-tr-2002-9.ps) and there is a nice video explanation [here](https://www.youtube.com/watch?v=DtYZEKdMcYI).

The overall purpose of this is to address issues I've been finding in Joern benchmarks for false positives being raised due to object/alias insensitivity. I'm also attempting to formalise a collection of static analysis benchmarks on [this branch](https://github.com/joernio/joern/tree/dave/joern-benchmarks).

DavidBakerEffendi:
@bbrehm and I discussed this and the notion of our dynamic call graph linker as a pass.

For something worthwhile one may want to look at something inter procedural for better precision but this would be expensive as a pass. A better position may be to look at this and dynamic call resolution as being on-the-fly.

The next steps will be developing a set of benchmarks (e.g. [this branch](https://github.com/joernio/joern/tree/dave/joern-benchmarks)) to accurately identify the flaws in Joern and make better decisions on what to do next.

bbrehm:
Per call: 

Between which nodes are we supposed to create edges? The algorithm operates on the level of `VarInCtx`, but there are potentially many graph nodes that correspond to the same `VarInCtx`. The answer is of course: We need to create edges where the user later needs to find edges (well, duh).

So I would suggest we keep this in a branch until we have an actual use-case for the computed information. Then we can look at this prospective pass and see how we need the info encoded.

In other words: There is no canonical way of mapping the concept of "PAG" to CPG, and therefore we should in this instance start from the problems we want to solve, instead of the shiny solutions we can compute.

Otherwise nice work to approach the thorny issue of pointer analysis in CPG. We definitely need something here, just not yet 100% clear what we need (also: I think alias analysis is a better frame that points-to-set, for the reason of locality. ).

To clarify above link, this here is the "field-based" pointer analysis.

PR #1269, opened 2022-04-14
javasrc: Clean up typeInfoProvider a bit

johannescoetzee:
This moves a lot of the ugly JavaParser type resolution logic into a separate class and merges some type methods where possible. Some of the scope type logic has also been moved into the `TypeInfoProvider` in an attempt to contain that as much as possible.

There's still work to do for separating everything cleanly, but this should give a decent indication of the general idea.

This is PR  1269 so, for no reason in particular, I  checked to see if anything interesting happened in the year 1269. From Wikipedia, there were a few mildly interesting candidates, but one stood out in particular:
* [Pélerin de Maricourt](https://en.wikipedia.org/wiki/Peter_of_Maricourt) first describes [magnetic poles](https://en.wikipedia.org/wiki/Magnet), and remarks on the nonexistence of [isolated magnetic poles](https://en.wikipedia.org/wiki/Magnetic_monopole).

This has nothing to do with joern, java, or code in general, but hopefully whoever reads this will find it interesting as well.


ursachec (approved):
nice one!

PR #1270, opened 2022-04-15
jimple2cpg: Annotation Support

DavidBakerEffendi:
Borrowed `javasrc2cpg` tests for annotations and have implemented these in `jimple` with subtle differences:

- Included missing string implementation for annotation nodes in `DotSerializer`
- All `code` properties for annotations include call parenthesis e.g. `@NormalAnnotation()`
- `code` properties for string literals are surrounded by double quotations (`"`)

### DIFF
--- a/joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/passes/AstCreator.scala
@@ -110,19 +110,23 @@ class AstCreator(filename: String, cls: SootClass, global: Global) extends AstCr
       .withChildren(methodAsts)
   }
 
-  private def astForField(v: SootField, order: Int): Ast = {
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/passes/AstCreator.scala
```diff
@@ -993,4 +1116,73 @@ object AstCreator {
       f(x, i + 1)
     }.toSeq
   }
+
+  private val PRIMITIVES: Map[Char, String] = Map[Char, String](
```
mpollmeier, line 1120:
```suggestion
  private val Primitives: Map[Char, String] = Map[Char, String](
```
scala devs don't tend to SHOUT :)
up to you though, this may have some meaning, and is really just a style guide, not a mandate

(copied from https://github.com/joernio/joern/pull/1294/)

mpollmeier (approved):

PR #1277, opened 2022-04-19
Downgrade ANTLR to 4.7

ml86:
Using 4.10 causes problems with upstream since 4.10 requires java 11.

### DIFF
--- a/project/Versions.scala
+++ b/project/Versions.scala
@@ -2,7 +2,9 @@
 object Versions {
   val cpg           = parseVersion("cpgVersion")
   val js2cpg        = parseVersion("js2cpgVersion")
-  val antlr         = "4.10"
+  // Dont upgrade antlr to 4.10 or above since those versions require java 11 or higher which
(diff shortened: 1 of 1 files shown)


project/Versions.scala
```diff
@@ -2,7 +2,9 @@
 object Versions {
   val cpg           = parseVersion("cpgVersion")
   val js2cpg        = parseVersion("js2cpgVersion")
-  val antlr         = "4.10"
+  // Dont upgrade antlr to 4.10 or above since those versions require java 11 or higher which
```
max-leuthaeuser, line 5: Then we perhaps want to add it to: https://github.com/joernio/joern/blob/master/.scala-steward.conf?

max-leuthaeuser (approved):

project/Versions.scala
ml86: Good point. Separate PR https://github.com/joernio/joern/pull/1281

PR #1284, opened 2022-04-21
Ghidra2cpg: 'argumentIndex' initial

itsacoderepo:


itsacoderepo:
Opening a new PR because it is more convenient

PR #1294, opened 2022-04-22
jimple2cpg: Correctly Handling Class Constants

DavidBakerEffendi:
- Created a test suite for reflection-related operations
- Implemented the `Literal` for class constants and nulls
- Brought in `parseAsmType` from the annotations branch for jimple2cpg

### DIFF
--- a/joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/passes/AstCreator.scala
@@ -847,47 +847,29 @@ class AstCreator(filename: String, cls: SootClass, global: Global) extends AstCr
 
   private def astForConstantExpr(constant: Constant, order: Int): Ast = {
     constant match {
-      case _: ClassConstant => Ast()
-      case _: NullConstant  => Ast()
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/passes/AstCreator.scala
```diff
@@ -993,4 +975,73 @@ object AstCreator {
       f(x, i + 1)
     }.toSeq
   }
+
+  private val PRIMITIVES: Map[Char, String] = Map[Char, String](
```
mpollmeier, line 979:
```suggestion
  private val Primitives: Map[Char, String] = Map[Char, String](
```
scala devs don't tend to SHOUT :)
up to you though, this may have some meaning, and is really just a style guide, not a mandate

mpollmeier (approved):

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/passes/AstCreator.scala
DavidBakerEffendi: Lol, got it. Will make suggested change.

PR #1299, opened 2022-04-25
Interprocedural Dot File Generator: Call Graphs

DavidBakerEffendi:
- jimple2cpg method nodes now use `astParent` properties
- Extended dot serialised `Graph` case classes to include subgraph clusters
- Added entry point for dot serialisers to now be called from the `cpg` itself
- Type clusters are obtained via `TypeDecl` via AST in info but defaults to `astParentFullName` otherwise

Example call graph on Java
```java
package test;
 class Foo {
   int add(int x, int y) {
     return x + y;
   }

   int main(int argc, char argv) {
     return add(argc, 3);
   }

   int bar(int argc) {
     foo(argc);
   }
 }

class MyObject {
    public static String staticCall(String s) {
        return s;
    }

    public String myMethod(String s) {
        return s;
    }
}

public class Bar {
    MyObject obj = new MyObject();

    public static void staticMethod() {}

    public String foo(MyObject myObj) {
        return myObj.myMethod("Hello, world!");
    }

    public void bar() {
        foo(obj);
    }

    public void baz() {
        this.foo(obj);
    }

    public void qux() {
        staticMethod();
    }

    public void quux() {
      bar();
    }
}
```
graphviz-2

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/dotgenerator/DotCpg14Generator.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/dotgenerator/DotCpg14Generator.scala
@@ -15,7 +15,7 @@ object DotCpg14Generator {
     val cfg = new CfgGenerator().generate(method)
     val ddg = new DdgGenerator().generate(method)
     val cdg = new CdgGenerator().generate(method)
-    DotSerializer.dotGraph(method, ast ++ cfg ++ ddg ++ cdg, withEdgeTypes = true)
+    DotSerializer.dotGraph(Some(method), ast ++ cfg ++ ddg ++ cdg, withEdgeTypes = true)
   }
 
 }
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/dotgenerator/DotCpg14Generator.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/dotgenerator/DotCpg14Generator.scala
@@ -15,7 +15,7 @@ object DotCpg14Generator {
     val cfg = new CfgGenerator().generate(method)
     val ddg = new DdgGenerator().generate(method)
     val cdg = new CdgGenerator().generate(method)
-    DotSerializer.dotGraph(Some(method), ast ++ cfg ++ ddg ++ cdg, withEdgeTypes = true)
+    DotSerializer.dotGraph(Option(method), ast ++ cfg ++ ddg ++ cdg, withEdgeTypes = true)
   }
 
 }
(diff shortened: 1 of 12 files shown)


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/dotgenerator/DotCpg14Generator.scala
```diff
@@ -15,7 +15,7 @@ object DotCpg14Generator {
     val cfg = new CfgGenerator().generate(method)
     val ddg = new DdgGenerator().generate(method)
     val cdg = new CdgGenerator().generate(method)
-    DotSerializer.dotGraph(method, ast ++ cfg ++ ddg ++ cdg, withEdgeTypes = true)
+    DotSerializer.dotGraph(Some(method), ast ++ cfg ++ ddg ++ cdg, withEdgeTypes = true)
```
mpollmeier, line 18:
```suggestion
    DotSerializer.dotGraph(Option(method), ast ++ cfg ++ ddg ++ cdg, withEdgeTypes = true)
```

Always prefer `Option(xyz)` over `Some(xyz)` because it performs a null-check, therefor you'll never end up with a `Some(null)`. There's more occurrences in this PR, can you please change them as well?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/dotgenerator/CallGraphGenerator.scala
```diff
@@ -0,0 +1,36 @@
...
+    val vertices = cpg.all.collect { case m: Method => m }.l
+    val edges = vertices.flatMap { srcMethod =>
+      storeInSubgraph(srcMethod, subgraph)
+      srcMethod.call.flatMap { child =>
+        child.callOut.map { tgt =>
+          storeInSubgraph(tgt, subgraph)
+          Edge(srcMethod, tgt, label = child.dispatchType.stripSuffix("_DISPATCH"))
+        }
+      }
+    }.distinct
```
mpollmeier, line 23:
```suggestion
    val edges = for {
      srcMethod <- vertices
      _ = storeInSubgraph(srcMethod, subgraph)
      child <- srcMethod.call
      tgt <- child.callOut
    } yield {
      storeInSubgraph(tgt, subgraph)
      Edge(srcMethod, tgt, label = child.dispatchType.stripSuffix("_DISPATCH"))
    }.distinct
```

The for comprehension saves the reader from parsing the flatMap/map etc... a matter of style, up to you really.
Note that I didn't ask the compiler for feedback on the above yet...

mpollmeier (commented):
Some nits from the style perspective. @fabsx00 thoughts from the technical perspective will be more relevant.

fabsx00 (approved):

PR #1339, opened 2022-05-03
Render primitive types in `kotlin2cpg`

ursachec:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/TypeRenderer.scala
```diff
@@ -55,7 +57,20 @@ object TypeRenderer {
...
+            val fqName                   = mappedType.asSingleFqName()
+            val nonUnwrappedRender       = stripped(renderer.renderFqName(fqName.toUnsafe))
+            val isWrapperOfPrimitiveType = JvmPrimitiveType.isWrapperClassName(fqName)
+
+            if (unwrapPrimitives && isWrapperOfPrimitiveType) {
+              JvmPrimitiveType
+                .values()
+                .toList
+                .filter(_.getWrapperFqName.toString == fqName.toString)
+                .map(_.getJavaKeywordName)
```
max-leuthaeuser, line 69: Could be also done with collect. Just a matter of style.

max-leuthaeuser (approved):

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/TypeRenderer.scala
ursachec: keeping it as it is for now, will tackle on a cleanup round of the whole frontend

PR #1350, opened 2022-05-04
Implement overriden binding signatures.

ml86:
For methods which override a method from a super class or interface, we
also need to add bindings with the erased signature of the overriden
methods.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
@@ -1,6 +1,6 @@
 package io.joern.javasrc2cpg.passes
 
-import com.github.javaparser.ast.`type`.TypeParameter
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -357,6 +377,73 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
     }
   }
 
+  private def substituteTypeVariable(resolvedType: ResolvedType, typeParamValues: ResolvedTypeParametersMap): String = {
+    substituteTypeVariableInternal(resolvedType, typeParamValues)
+  }
```
johannescoetzee, line 382: Is this wrapper method necessary? Given that both `substituteTypeVariable` and `substituteTypeVariableInternal` are private, they seem to be equivalent.

johannescoetzee (approved):

PR #1356, opened 2022-05-05
cleanup/fix default.semantics

mpollmeier:
* add mappings for `<operators>.xyz` while we have those typos
* symlink duplicated default semantics, delete unused


dataflowengineoss/src/test/resources/default.semantics
```diff
@@ -1,39 +0,0 @@
-# "<operator>.sizeOf" reduces the FPs
```
max-leuthaeuser, line 1: This one is used in many tests. I guess we need a symlink here too.

max-leuthaeuser (approved):

PR #1363, opened 2022-05-06
WIP: switch to `callAst` method in x2cpg in javasrc

johannescoetzee:
WIP because some dataflow tests are failing with the new structure where there is no longer an argument edge to the receiver of the call. I'm leaving this aside for now to avoid possible conflicts with work @fabsx00 is doing on the dataflow engine.


johannescoetzee:
I'll close this for now and just include the changes in a larger javasrc cleanup PR later

PR #1375, opened 2022-05-09
Replace TypeInfoProvider and JP2JavaSrcTypeAdapter by new implementation.

ml86:
This new implementation is TypeInfoCalculator which aims to be reduced
in size and more condense.


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/TypeInfoCalculator.scala
```diff
@@ -0,0 +1,185 @@
...
+  def name(typ: ResolvedType): String = {
+    nameOrFullName(typ, false)
+  }
+
+  def fullName(typ: ResolvedType): String = {
+    registerType(nameOrFullName(typ, true))
+  }
+
+  private def nameOrFullName(typ: ResolvedType, fullyQualified: Boolean): String = {
+    typ match {
```
johannescoetzee, line 41: I'm not sure that this is where the tests failed, but a default case should be added here.

johannescoetzee (approved):

Issue #1009, opened 2022-02-10
Dataflow inconsistency related to reassignment and subcalls


jryerb:
For the following source code:
```
oid reassignThenFree(char * ptr)
{
    ptr = malloc(0x80);
    free(ptr);
    return;
}

void reassign(char * ptr)
{
    ptr = malloc(0x80);
    return;
}

int case0()
{
    char * data = malloc(0x100);
    free(data);
    free(data);

    return 0;
}

int case1()
{
    char * data = malloc(0x100);
    free(data);
    data = malloc(0x80);
    free(data);

    return 0;
}

int case2()
{
    char * data = malloc(0x100);
    free(data);
    reassignThenFree(data);

    return 0;
}

int case3()
{
    char * data = malloc(0x100);
    free(data);
    reassign(data);
    free(data);

    return 0;
}
```
The following queries and results are:
```
joern> importCode("./test") 
...
joern> run.ossdataflow 
...
joern> def sink = cpg.call("free").argument(1) 
defined function sink

joern> def source = cpg.call("free").argument(1) 
defined function source

joern> sink.reachableByFlows(source).filter(path => path.elements.size > 1).p 
res166: List[String] = List(
  """________________________________________________________________________________________________________________
| tracked        | lineNumber| method| file                                                                     |
|===============================================================================================================|
| free(data)     | 45        | case3 | /Users/jryerby/Documents/Query dev/RES-489/simple/called_func_behavior.c |
| reassign(data) | 46        | case3 | /Users/jryerby/Documents/Query dev/RES-489/simple/called_func_behavior.c |
| free(data)     | 47        | case3 | /Users/jryerby/Documents/Query dev/RES-489/simple/called_func_behavior.c |
""",
  """____________________________________________________________________________________________________________
| tracked    | lineNumber| method| file                                                                     |
|===========================================================================================================|
| free(data) | 17        | case0 | /Users/jryerby/Documents/Query dev/RES-489/simple/called_func_behavior.c |
| free(data) | 18        | case0 | /Users/jryerby/Documents/Query dev/RES-489/simple/called_func_behavior.c |
""",
  """_________________________________________________________________________________________________________________________________________
| tracked                      | lineNumber| method           | file                                                                     |
|========================================================================================================================================|
| free(data)                   | 36        | case2            | /Users/jryerby/Documents/Query dev/RES-489/simple/called_func_behavior.c |
[...]

PR #1376, opened 2022-05-10
Cleanup and fixes for the OSS data flow engine

fabsx00:
* Refactoring of code for DDG creation
* METHOD_PARAMETER_OUT nodes are now correctly placed in the DDG
* Test cases added for the `ReachingDefFlowGraph` (the CFG from the perspective of the DDG generator)
* Cleanup and fixes for the query engine


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/ResultTable.scala
```diff
@@ -30,7 +32,7 @@ class ResultTable(
       res.map { r =>
         val pathToFirstNode = r.path.slice(0, r.path.map(_.node).indexOf(first.node))
         val completePath    = pathToFirstNode ++ (first +: remainder)
-        r.copy(path = completePath)
+        r.copy(path = Vector(completePath.head) ++ completePath.tail)
```
DavidBakerEffendi, line 35: Is this to make a deep copy of the path? If so there is also `completePath.map(_.copy())`

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/TaskCreator.scala
```diff
@@ -0,0 +1,125 @@
...
+            val callSiteStack = result.callSiteStack.clone()
+            callSiteStack.push(call)
+            ReachableByTask(returnStatement, sources, new ResultTable, newPath, callDepth + 1, callSiteStack)
+          }
+        }
+      }
+    }
+
+    val forArgs = outArgsAndCalls.flatMap { case (result, args, path, callDepth) =>
+      args.toList.flatMap { arg =>
```
DavidBakerEffendi, line 106: If you want to get rid of unchecked `.asInstanceOf` casts
```suggestion
      args.toList.flatMap { case (arg: Expression) =>
```

DavidBakerEffendi (approved):
The review was brain hurting juice but hopefully my myriad of suggested changes helps to make it more robust!

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala
```diff
@@ -0,0 +1,332 @@
...
+        }
+      }
+
+      // Handle block arguments
+      call.argument.isBlock.foreach { block =>
+        block.astChildren.lastOption match {
+          case None => // Do nothing
+          case Some(node: Identifier) =>
+            val edgesToAdd = in(node).toList.flatMap { inDef =>
+              numberToNode(inDef) match {
```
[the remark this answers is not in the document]
johannescoetzee: @DavidBakerEffendi this will crash with a match error if a non-identifier node is returned by the lookup. 
```
numberToNode.get(inDef).collect { case identifier: Identifier => identifier }
```
should cover all the cases.

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/ResultTable.scala
fabsx00: Good catch. This is more of a refactoring artifact. Should be possible to just say `completePath` here.

fabsx00: Ok, no, this needs to be a copy.

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala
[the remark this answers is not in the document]
fabsx00: It would still need to be `case _ => None`, I think.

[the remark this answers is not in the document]
johannescoetzee: I wondered about that too, but turns out it doesn't (at least if I understood the use case correctly):
```
ocular> val opt: Option[Object] = Some("Hello") 
opt: Option[Object] = Some(value = "Hello")

ocular> opt.collect { case x: Integer => x } 
res51: Option[Integer] = None

ocular> opt.collect { case x: String => x } 
res52: Option[String] = Some(value = "Hello")
```

fabsx00:
This should fix https://github.com/joernio/joern/issues/1009

PR #1378, opened 2022-05-10
Fixed DF semantics file location

max-leuthaeuser:
`relativise` and `Semantics.fromList(new Parser().parseFile(semanticsFilename))` can't handle symlinks (Windows only).
We ended up with empty semantics there.


mpollmeier (approved):
out of curiosity: did that result in failed tests on windows? if not, can we add something that tells us next time around?

max-leuthaeuser:
> out of curiosity: did that result in failed tests on windows? if not, can we add something that tells us next time around?

Unfortunately not with the existing tests. But with the new jssrc2cpg frontend there will be plenty soon.

PR #1383, opened 2022-05-11
Add `1` to line numbers of Kotlin CPGs

ursachec:
it seems that the `package` directive is not taken into consideration

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/psi/Extractor.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/psi/Extractor.scala
@@ -6,7 +6,7 @@ object Extractor {
   def line(element: PsiElement): Int = {
     try {
       element.getContainingFile.getViewProvider.getDocument
-        .getLineNumber(element.getTextOffset)
+        .getLineNumber(element.getTextOffset) + 1
     } catch {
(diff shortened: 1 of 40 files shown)


johannescoetzee:
With the changes as they are here, would the line numbers in files without the package directive be wrong or is there some sort of implicit package directive added? If they would be wrong, is it possible to check whether the file contains a package directive?

ursachec:
@johannescoetzee 

```
$ cat /tmp/simple/h.kt 
fun main() {
  println("Helloo, world")
}
```

```
ocular> cpg.method.fullName(".*main.*").l 
res8: List[Method] = List(
  Method(
    id -> 5L,
    astParentFullName -> "<empty>",
    astParentType -> "<empty>",
    binarySignature -> None,
    code -> "main",
    columnNumber -> Some(value = 4),
    columnNumberEnd -> Some(value = 0),
    depthFirstOrder -> 1,
    filename -> "h.kt",
    fullName -> "main:void()",
    hasMapping -> None,
    hash -> None,
    internalFlags -> 4,
    isExternal -> false,
    lineNumber -> Some(value = 1),     <---- LINE_NUMBER after this change
    lineNumberEnd -> Some(value = 3),
    name -> "main",
    order -> 1,
    signature -> "void()",
    varargParameter -> -1
  )
)
```

johannescoetzee (approved):

Issue #1385, opened 2022-05-11
Java Parser failing with ERROR on a source file


tuxology:
While using the new Javasrc2cpg frontend on tarpit-java, the Java Parser is failing with following error. 

```
joern> importCode.java("/home/suchakra/Projects/tarpit-java/src", "tarpit-src") 
Creating project `tarpit-src` for code at `/home/suchakra/Projects/tarpit-java/src`
2022-05-11 10:50:43.750 ERROR AstCreator: Parsing file /home/suchakra/Projects/tarpit-java/src/main/java/io/shiftleft/tarpit/Insider.java failed with scala.MatchError: WildcardUsage{type=null, boundedType=null} (of class com.github.javaparser.resolution.types.ResolvedWildcard)
moving cpg.bin.zip to cpg.bin because it is already a database file
Creating working copy of CPG to be safe
```

As a consequence only 3 files are included int eh resulting CPG and much of the graph is missing important nodes. 

#### Joern Version 
v1.1.796

#### JVM 
```
openjdk version "1.8.0_312"
OpenJDK Runtime Environment (build 1.8.0_312-8u312-b07-0ubuntu1~20.04-b07)
OpenJDK 64-Bit Server VM (build 25.312-b07, mixed mode)
```

mpollmeier:
I can confirm the fix works. Once it's merged we still need to release it to prod, I can do that, but here's what it'll look like:

![grafik](https://user-images.githubusercontent.com/506752/168015182-4de100d4-6a9f-401f-b155-414c40f04a33.png)

(I was just testing javasrc anyway)

johannescoetzee:
The code that was merged is slightly different from the version @mpollmeier used when testing the fix, although the merged code still fixed the crash (at least locally). Reopening this until there's confirmation that the code in master fixes this issue.

mpollmeier:
depends on https://github.com/ShiftLeftSecurity/javasrc2cpg-internal/pull/22

mpollmeier:
tested this again, just to be safe - all good!
![image](https://user-images.githubusercontent.com/506752/168846726-c6a2e71b-7d98-49fd-8a24-ef60080be0cb.png)

PR #1386, opened 2022-05-11
javasrc: Fix crash when getting wildcard type name

johannescoetzee:
Fixes https://github.com/joernio/joern/issues/1385.


mpollmeier:
Looking good from a high level view...
https://github.com/joernio/joern/issues/1385#issuecomment-1124627443
https://github.com/ShiftLeftSecurity/product/issues/10361#issuecomment-1124647089

ml86 (approved):

johannescoetzee:
After a discussion with @ml86, I removed the catch-all case that was present when @mpollmeier did the test. With the current version of the PR, I was able to successfully create a CPG for the tarpits-java source mentioned in the ticket, so it shouldn't affect the findings we're getting.

Issue #1221, opened 2022-04-04
Bash file in Windows - CreateProcess error=193, %1 is not a valid Win32 application


kiranravindran90:
While I tried to create a project, I am getting this below error. I believe its probable because of Windows trying to execute Bash files. But not sure what would be a work around.

Error message :
_____________________________________________________
**java.io.IOException: Cannot run program ".......\joern-master\joern-cli\target\universal\stage\c2cpg.sh": CreateProcess error=193, %1 is not a valid Win32 application**
  java.lang.ProcessBuilder.start(ProcessBuilder.java:1128)
  java.lang.ProcessBuilder.start(ProcessBuilder.java:1071)
  scala.sys.process.ProcessBuilderImpl$Simple.run(ProcessBuilderImpl.scala:85)
  scala.sys.process.ProcessBuilderImpl$AbstractBuilder.run(ProcessBuilderImpl.scala:117)
  scala.sys.process.ProcessBuilderImpl$AbstractBuilder.run(ProcessBuilderImpl.scala:116)
  io.joern.console.cpgcreation.CpgGenerator.runShellCommand(CpgGenerator.scala:27)
  io.joern.console.cpgcreation.CCpgGenerator.generate(CCpgGenerator.scala:21)
  io.joern.console.cpgcreation.CpgGeneratorFactory.runGenerator(CpgGeneratorFactory.scala:63)
  io.joern.console.cpgcreation.ImportCode.$anonfun$apply$5(ImportCode.scala:132)
  scala.Option.flatMap(Option.scala:283)
  io.joern.console.cpgcreation.ImportCode.io$joern$console$cpgcreation$ImportCode$$apply(ImportCode.scala:130)
  io.joern.console.cpgcreation.ImportCode.apply(ImportCode.scala:38)
  ammonite.$sess.cmd1$.<clinit>(cmd1.sc:1)
java.io.IOException: CreateProcess error=193, %1 is not a valid Win32 application
  java.lang.ProcessImpl.create(Native Method)
  java.lang.ProcessImpl.<init>(ProcessImpl.java:487)
  java.lang.ProcessImpl.start(ProcessImpl.java:154)
  java.lang.ProcessBuilder.start(ProcessBuilder.java:1107)
  java.lang.ProcessBuilder.start(ProcessBuilder.java:1071)
  scala.sys.process.ProcessBuilderImpl$Simple.run(ProcessBuilderImpl.scala:85)
[...]

PR #1398, opened 2022-05-16
Added Windows executables

max-leuthaeuser:
Fixes: https://github.com/joernio/joern/issues/1221

_Note_: we have to add `LauncherJarPlugin` to the frontends. Otherwise we run into "line too long" issues on Windows.

### DIFF
--- a/console/src/main/scala/io/joern/console/cpgcreation/CCpgGenerator.scala
+++ b/console/src/main/scala/io/joern/console/cpgcreation/CCpgGenerator.scala
@@ -8,7 +8,7 @@ import java.nio.file.Path
   * preprocessing.
   */
 case class CCpgGenerator(config: FrontendConfig, rootPath: Path) extends CpgGenerator {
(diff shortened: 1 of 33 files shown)


mpollmeier (approved):

mpollmeier:
not entirely sure if this would help here as well, but [elsewhere](https://github.com/ShiftLeftSecurity/ocular/blob/04a7266/build.sbt#L59) we configured `scriptClasspath := Seq("*")` to avoid long lines for windows

max-leuthaeuser:
> not entirely sure if this would help here as well, but [elsewhere](https://github.com/ShiftLeftSecurity/ocular/blob/04a7266/build.sbt#L59) we configured `scriptClasspath := Seq("*")` to avoid long lines for windows

Yes, seen that. Both solutions are valid.
But `scriptClasspath := Seq("*")` does not keep the order of classpath elements which is currently not a problem as it seems but might change with new frontends coming in in the future.

PR #1400, opened 2022-05-16
upgrade cpg/odb, integrate neo4jcsv with joern-export

mpollmeier:


fabsx00 (approved):
As discussed, it would probably be better to have an option to dump the entire graph as opposed to AST/CFG/... and then specify the output format as dot/neo4j, but let's do that at a later point in time.

PR #1404, opened 2022-05-17
Initial commit of jssrc2cpg

max-leuthaeuser:


fabsx00 (approved):
As discussed, let's provide the user with a command for installing astgen if it is not installed.

max-leuthaeuser:
> As discussed, let's provide the user with a command for installing astgen if it is not installed.

Will do 👍

PR #1406, opened 2022-05-17
Extract `classes.jar` from aar dependency files

ursachec:
A large number of Android dependencies are distributed as `aar`
files which are just a specially laid-out zip file:
https://developer.android.com/studio/projects/android-library

The embedded Kotlin compiler does not accept `aar` files directly, so 
their `classes.jar` need to be extracted first, which is what this 
PR accomplishes.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/GradleDependencies.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/GradleDependencies.scala
@@ -27,6 +27,13 @@ case class GradleProjectInfo(gradleVersion: String, tasks: Seq[String], hasAndro
   }
 }
 
+object Constants {
+  val aarFileExtension            = "aar"
+  val gradleAndroidPropertyPrefix = "android."
+  val gradlePropertiesTaskName    = "properties"
+  val jarInsideAarFileName        = "classes.jar"
+}
+
 case class GradleDepsInitScript(contents: String, taskName: String, destinationDir: Path)
 
 object GradleDependencies {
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/GradleDependencies.scala
```diff
@@ -200,6 +199,30 @@ object GradleDependencies {
...
   }
 
+  private def extractClassesJarFromAar(aar: File): Option[Path] = {
+    val newPath              = aar.path.toString.replaceFirst("aar$", "jar")
+    val jarInsideAarFileName = "classes.jar"
+    val aarUnzipDirSuffix    = ".unzipped"
+    val outDir               = File(aar.path.toString + aarUnzipDirSuffix)
+    aar.unzipTo(outDir, _.getName == jarInsideAarFileName)
+    val outFile = File(newPath)
+    def classesJarEntries =
```
max-leuthaeuser, line 209: Could be a val doesn't it?

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/GradleDependencies.scala
```diff
@@ -227,7 +250,19 @@ object GradleDependencies {
                 Try(makeConnection(projectDir.toFile)) match {
                   case Success(connection) =>
                     Using.resource(connection) { c =>
-                      runGradleTask(c, initScript, initScriptFile.pathAsString)
+                      runGradleTask(c, initScript, initScriptFile.pathAsString) match {
+                        case Some(deps) =>
+                          val aarFileExtension = "aar"
```
max-leuthaeuser, line 255: Maybe put all those special strings  
like aar or classes.jar in constants somewhere?

ursachec: good idea, done

PR #1414, opened 2022-05-20
Cleanup signature and methodFullName calculation for...

ml86:
method and construction <init> calls.
This also fixes a bug int signature and thus the methodFullName of
construction <init> calls which have been incorporating the argument
types instead of the parameter types. The argument types are now only
used if the parameter types cannot be resolved.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
@@ -343,19 +343,41 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
     }
   }
 
+  private def constructorSignature(
+    constructor: ResolvedConstructorDeclaration,
+    typeParamValues: ResolvedTypeParametersMap
+  ): String = {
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -2598,32 +2616,37 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
-      case (Some(targetType), Some(returnType)) =>
-        val signature      = s"$returnType(${argumentTypes.mkString(",")})"
-        val methodFullName = s"$targetType.${call.getNameAsString}:$signature"
-        (signature, methodFullName)
+    val methodFullName =
+      receiverTypeOption match {
+        case Some(receiverType) =>
+          s"$receiverType.${call.getNameAsString}:$signature"
+        case None =>
+          s"<unresolvedReceiverType.${call.getNameAsString}:$signature"
```
johannescoetzee, line 2642: Is there a missing `>` here?

johannescoetzee (approved):
Looks good, apart from the possibly missing `>` in the `unresolvedReceiverType` comment above

PR #1423, opened 2022-05-23
CfgNode extension "passesNot" to filter out flows through sanitisers

DavidBakerEffendi:
- While we have `reachableBy`, as it stands there is no way to kill paths that flow through sanitisers
- Here is a prototype `passes` and `passesNot` that probably needs fine tuning but gets the job done with `postDominatedBy`

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CfgNodeMethods.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CfgNodeMethods.scala
@@ -94,6 +94,26 @@ class CfgNodeMethods(val node: CfgNode) extends AnyVal with NodeExtension {
     }
   }
 
+  /** Using the post dominator tree, will determine if this node passes through the included set of nodes and filter it
+    * in.
+    * @param included
+    *   the nodes this node must pass through.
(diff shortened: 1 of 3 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CfgNodeMethods.scala
```diff
@@ -94,6 +94,26 @@ class CfgNodeMethods(val node: CfgNode) extends AnyVal with NodeExtension {
...
+
+  /** Using the post dominator tree, will determine if this node passes through the excluded set of nodes and filter it
+    * out.
+    * @param excluded
+    *   the nodes this node must not pass through.
+    * @return
+    *   the traversal of this node if it does not pass through the excluded set.
+    */
+  def passesNot(excluded: Set[CfgNode]): Traversal[CfgNode] =
+    node.filterNot(_.postDominatedBy.toSet.exists(excluded.contains))
```
mpollmeier, line 115:
```suggestion
    node.filterNot(_.postDominatedBy.exists(excluded.contains))
```
no need to convert the Traversal to a Set afaics (?)

mpollmeier (approved):

PR #1424, opened 2022-05-23
Improved Java False Positives with a println Semantic

DavidBakerEffendi:
- Added default semantic for `PrintWriter.println`
- This has drastically improved the results for Securibench

### DIFF
--- a/benchmarks/src/test/scala/io/joern/benchmarks/testfixtures/BenchmarkFixture.scala
+++ b/benchmarks/src/test/scala/io/joern/benchmarks/testfixtures/BenchmarkFixture.scala
@@ -30,7 +30,7 @@ class BenchmarkFixture(
     with Matchers
     with BeforeAndAfterAll {
 
-  val semanticsFile: String = ProjectRoot.relativise("dataflowengineoss/src/test/resources/default.semantics")
+  val semanticsFile: String = ProjectRoot.relativise("benchmarks/src/test/resources/default.semantics")
   lazy val defaultSemantics: Semantics           = Semantics.fromList(new Parser().parseFile(semanticsFile))
   implicit val resolver: ICallResolver           = NoResolve
   implicit lazy val engineContext: EngineContext = EngineContext(defaultSemantics, EngineConfig(maxCallDepth = 4))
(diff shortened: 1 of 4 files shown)


benchmarks/src/test/scala/io/joern/benchmarks/testfixtures/BenchmarkFixture.scala
```diff
@@ -30,7 +30,7 @@ class BenchmarkFixture(
     with Matchers
     with BeforeAndAfterAll {
 
-  val semanticsFile: String = ProjectRoot.relativise("dataflowengineoss/src/test/resources/default.semantics")
+  val semanticsFile: String = ProjectRoot.relativise("benchmarks/src/test/resources/default.semantics")
```
mpollmeier, line 33: Did you forget to add that file? It doesn't seem to exist...

mpollmeier (commented):
Apart from my comment, and this is probably unrelated to this specific PR: I tried to run the benchmark on master, and that also failed, albeit with a different error: 
```
[info] - should report secure *** FAILED ***
[info]   [False positive] Source was found to taint the sink (BenchmarkFixture.scala:77)
```

It's probably due to missing prereqs - any idea what's wrong?

benchmarks/src/test/scala/io/joern/benchmarks/testfixtures/BenchmarkFixture.scala
DavidBakerEffendi: Ah yes thanks my bad

DavidBakerEffendi:
> Apart from my comment, and this is probably unrelated to this specific PR: I tried to run the benchmark on master, and that also failed, albeit with a different error:
> 
> ```
> [info] - should report secure *** FAILED ***
> [info]   [False positive] Source was found to taint the sink (BenchmarkFixture.scala:77)
> ```
> 
> It's probably due to missing prereqs - any idea what's wrong?

Currently it uses ScalaTest to drive the benchmark - a failed test will mean either we got a false positive or a false negative. `BenchmarkFixture.scala` overrides `withFixture` to determine if this should tally on either false positive, false negative, true positive, or true negative and then write this to either console or the specified file. `joern-benchmarks` makes sure to route the output to `benchmarks/README.md`.

PR #1432, opened 2022-05-25
Ignore tests requiring network access

ursachec:


ml86 (approved):

mpollmeier:
wouldn't it be better to do a quick check if we have network access, and run these tests in the 99% case where we have it? and if we don't, then just print a warning?

to be clear: i'm happy to add that logic, just checking with you first

ursachec:
@mpollmeier maybe, not sure, i've just received complaints about one of the test cases taking too long, so disabled it for now until a concrete decision is made on how to handle that test suite. the cleanest solution seems to be including the artifacts in the test resources, but not yet sure how that'll be implemented

PR #1437, opened 2022-05-25
Re-implement javasrc lambdas and add more comprehensive tests

johannescoetzee:
This is effectively a full re-write of lambda support for javasrc2cpg. It adds a `TYPE_DECL` corresponding to the lambda method (and inheriting from the implemented functional interface) and bindings to the implemented method, all of which were missing with the previous implementation.

It relies heavily on the `ExpectedType` information to create correct signatures for the lambda method node and bindings, so in cases where we don't know the expected type, these are left as `java.lang.Object`, even if JavaParser could resolve the lambda type. In cases where neither the expected type nor the lambda type could be resolved, the results will be effectively useless. In this case we don't know which method the lambda implements, so the relevant bindings aren't created at all.

A specific case where JavaParser could resolve the lambda but where we don't know the expected type occurs in the example:
```
import java.util.List;
import java.util.stream.Collectors;

public class Foo {
  public List<String> method(List<String> list) {
    return list.stream().map(string -> string).collect(Collectors.toList());
  }
}
```
In this case, we know `string -> string` implements `Function<T,R>` since it could be resolved, but don't know that the type parameters map to `String` in this context.

# Update on 20.06.22 revision

__PR for codescience dataflow tests:__ https://github.com/ShiftLeftSecurity/codescience/pull/6200

This PR is now no longer a WIP. but ended up adding a fair amount of stuff. These can be roughly categorized as:

## Lambda re-implementation
Lambda support was effectively rewritten entirely. Dataflow through lambdas using the OSS dataflow engine doesn't work, probably due to the open source call linker not being able to link the lambda call to the `$lambda$X` method created for the implementation. Closed source dataflow is now on par with `java2cpg`, at least as far as covered by the closed source dataflow tests. A PR to fix expected flows for those tests will follow once this has been merged. 

Since java lambdas (conceptually) create a class implementing an interface method, a fair amount of type information is needed for lambdas to work. If that type information is not available (e.g. due to using a dependency that couldn't be fetched), this implementation will give unusable results. It may be possible to use some heuristics to figure out some of this type information, but in my opinion the effort is not worth it.

## ForEach re-implemenation
The only other big change included in this PR. To fix dataflow through foreach loops, javasrc now represents a foreach loop more similarly to java bytecode. Specifically, for foreach loops over native arrays:
```
// ex1
for (int x : xs) { ... }
// where xs has type int[] is represented as
for (int idx = 0; idx < xs.length; idx++) {
  int x = xs[idx];
}

// ex2
for (int x : new int[] {1, 2 }) { ... }
// is represented as
int[] tmpIterable = new int[] {1, 2}
for (int idx = 0; idx < tmpIterable.length; idx++) {
  int x = tmpIterable[idx];
  ...
}
```

whereas for foreach loops over collections:
```
for (Integer x : xs) { ... }
// where xs has type List<Integer> is represented as
Iterator tmpIterator = xs.iterator
while (tmpIterator.hasNext()) {
  Integer x = tmpIterator.next();
  ...
}
```

## Fixed varargs handling
Before, varargs were treated as regular arguments. This led to crashes in the worst case and potentially missing dataflows otherwise. With this PR, varargs are now treated as arrays, similar to compiled Java. Variadic parameters are given an array type, while variadic arguments are treated as `arrayInitializationExpressions`. For example:
```
public static void foo(String... inputs) {}
public static void bar() {
  foo("Hello", "world");
}

// is now represented as
public static void foo(String[] inputs) {}
public static void bar() {
  foo(new String[] {"Hello", "world"});
}
```

## Missing modifiers and other misc additions
This adds some missing method and typeDecl modifiers and improves the code fields. Most notably, this no longer uses a simple `toString` for the code for method calls, since lambdas in the scope of the call could make this code field incredibly long.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
@@ -1495,6 +1495,7 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
   ): NewIdentifier = {
     NewIdentifier()
       .name(local.name)
+      .code(local.name)
       .typeFullName(local.typeFullName)
       .lineNumber(lineNumber)
       .order(order)
@@ -1551,15 +1552,15 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
         .name(iteratorLocalName)
         .code(iteratorLocalName)
         .typeFullName(TypeConstants.Iterator)
-        .order(1)
+        .order(order)
         .lineNumber(lineNo)
 
     // Create local assign from iterator
     val iteratorAssignNode =
       assignmentNode()
         .typeFullName(TypeConstants.Iterator)
-        .order(2)
-        .argumentIndex(2)
+        .order(order + 1)
+        .argumentIndex(order + 1)
         .lineNumber(lineNo)
     val iteratorAssignIdentifier = identifierFromNamedVarType(iteratorLocalNode, lineNo, 1)
     val iteratorMethodName = "iterator"
@@ -1601,18 +1602,18 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
         .typeFullName(TypeConstants.Boolean)
         .dispatchType(DispatchTypes.DYNAMIC_DISPATCH)
         .code(hasNextCallFullName)
-        .order(3)
-        .argumentIndex(3)
+        .order(1)
+        .argumentIndex(1)
         .lineNumber(lineNo)
-    val iteratorHasNextCallReceiver = identifierFromNamedVarType(iteratorLocalNode, lineNo, 0)
+    val iteratorHasNextCallReceiver = identifierFromNamedVarType(iteratorLocalNode, lineNo, 1).argumentIndex(0)
     val iteratorHasNextCallAst =
       callAst(iteratorHasNextCallNode, Seq(Ast(iteratorHasNextCallReceiver)))
         .withRefEdge(iteratorHasNextCallReceiver, iteratorLocalNode)
         .withReceiverEdge(iteratorHasNextCallNode, iteratorHasNextCallReceiver)
 
     // Empty block update
-    val emptyUpdateNode = NewBlock().order(4).argumentIndex(4).lineNumber(lineNo)
-    val emptyUpdateAst  = Ast(emptyUpdateNode)
+//    val emptyUpdateNode = NewBlock().order(4).argumentIndex(4).lineNumber(lineNo)
+//    val emptyUpdateAst  = Ast(emptyUpdateNode)
 
     // Body
     // According to the Java language specification, there should be exactly one variable defined here.
@@ -1643,20 +1644,20 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
     val varLocalAssignIdentifier = identifierFromNamedVarType(variableLocal, lineNo, 1)
 
     val iterNextCallName = "next"
-    val iterNextCallSignature = composeMethodLikeSignature(forVariableType, parameterTypes = Nil)
+    val iterNextCallSignature = composeMethodLikeSignature(TypeConstants.Object, parameterTypes = Nil)
     val iterNextCallFullName = composeMethodFullName(TypeConstants.Iterator, iterNextCallName, iterNextCallSignature)
     val iterNextCallNode =
       NewCall()
         .name(iterNextCallName)
         .methodFullName(iterNextCallFullName)
         .signature(iterNextCallSignature)
-        .typeFullName(forVariableType)
+        .typeFullName(TypeConstants.Object)
         .dispatchType(DispatchTypes.DYNAMIC_DISPATCH)
         .code(iterNextCallFullName)
         .order(2)
         .argumentIndex(2)
         .lineNumber(lineNo)
-    val iterNextCallReceiver = identifierFromNamedVarType(iteratorLocalNode, lineNo, 0)
+    val iterNextCallReceiver = identifierFromNamedVarType(iteratorLocalNode, lineNo, 1).argumentIndex(0)
     val iterNextCallAst =
       callAst(iterNextCallNode, Seq(Ast(iterNextCallReceiver)))
         .withRefEdge(iterNextCallReceiver, iteratorLocalNode)
@@ -1670,10 +1671,10 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
     val bodyPrefixAsts = Seq(Ast(variableLocal), varLocalAssignAst)
     val bodyAst = stmt.getBody match {
       case block: BlockStmt =>
-        astForBlockStatement(block, 5, prefixAsts = bodyPrefixAsts)
+        astForBlockStatement(block, 2, prefixAsts = bodyPrefixAsts)
 
       case bodyStmt =>
-        val bodyBlockNode = NewBlock().order(5).argumentIndex(5).lineNumber(lineNo)
+        val bodyBlockNode = NewBlock().order(2).argumentIndex(2).lineNumber(lineNo)
         val bodyStmtAsts = astsForStatement(bodyStmt, bodyPrefixAsts.size + 1)
         Ast(bodyBlockNode)
         .withChildren(bodyPrefixAsts)
@@ -1682,50 +1683,55 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
 
     val forNode =
       NewControlStructure()
-        .controlStructureType(ControlStructureTypes.FOR)
-        .order(order)
-        .argumentIndex(order)
+        .controlStructureType(ControlStructureTypes.WHILE)
+        .order(order + 2)
+        .argumentIndex(order + 3)
         .code(ControlStructureTypes.FOR)
         .lineNumber(lineNo)
         .columnNumber(column(stmt))
 
+//    val forAst = Ast(forNode)
+//      .withChild(Ast(iteratorLocalNode))
+//      .withChild(iteratorAssignAst)
+//      .withChild(iteratorHasNextCallAst)
+//      .withChild(emptyUpdateAst)
+//      .withChild(bodyAst)
+
     val forAst = Ast(forNode)
-      .withChild(Ast(iteratorLocalNode))
-      .withChild(iteratorAssignAst)
       .withChild(iteratorHasNextCallAst)
-      .withChild(emptyUpdateAst)
       .withChild(bodyAst)
+      .withConditionEdge(forNode, iteratorHasNextCallNode)
 
-    val javaUtilIteratorTypeDecl = NewTypeDecl().name("Iterator").fullName(TypeConstants.Iterator)
-    val constructedIteratorTypeDecl = NewTypeDecl().fullName(iteratorLocalNode.typeFullName).name(TypeNodePass.fullToShortName(iteratorLocalNode.typeFullName))
+//    val javaUtilIteratorTypeDecl = NewTypeDecl().name("Iterator").fullName(TypeConstants.Iterator)
+//    val constructedIteratorTypeDecl = NewTypeDecl().fullName(iteratorLocalNode.typeFullName).name(TypeNodePass.fullToShortName(iteratorLocalNode.typeFullName))
+//
+//    val javaUtilIteratorBindings = List(
+//      BindingTableEntry(
+//        hasNextCallName,
+//        hasNextCallSignature,
+//        hasNextCallFullName
+//      ),
+//      BindingTableEntry(
+//        iterNextCallName,
+//        iterNextCallSignature,
+//        iterNextCallFullName
+//      )
+//    )
+//
+//    val javaUtilIteratorBindingTable = getFlatBindingTable(javaUtilIteratorTypeDecl, javaUtilIteratorBindings)
+//    createBindingNodes(javaUtilIteratorTypeDecl, javaUtilIteratorBindingTable)
+//
+//    val constructedIteratorBindings = List(
+//      BindingTableEntry(
+//        iteratorMethodName,
+//        iteratorMethodSignature,
+//        iteratorMethodFullName
+//      )
+//    )
+//    val constructedIteratorBindingTable = getFlatBindingTable(constructedIteratorTypeDecl, constructedIteratorBindings)
+//    createBindingNodes(constructedIteratorTypeDecl, constructedIteratorBindingTable)
 
-    val javaUtilIteratorBindings = List(
-      BindingTableEntry(
-        hasNextCallName,
-        hasNextCallSignature,
-        hasNextCallFullName
-      ),
-      BindingTableEntry(
-        iterNextCallName,
-        iterNextCallSignature,
-        iterNextCallFullName
-      )
-    )
-
-    val javaUtilIteratorBindingTable = getFlatBindingTable(javaUtilIteratorTypeDecl, javaUtilIteratorBindings)
-    createBindingNodes(javaUtilIteratorTypeDecl, javaUtilIteratorBindingTable)
-
-    val constructedIteratorBindings = List(
-      BindingTableEntry(
-        iteratorMethodName,
-        iteratorMethodSignature,
-        iteratorMethodFullName
-      )
-    )
-    val constructedIteratorBindingTable = getFlatBindingTable(constructedIteratorTypeDecl, constructedIteratorBindings)
-    createBindingNodes(constructedIteratorTypeDecl, constructedIteratorBindingTable)
-
-    Seq(forAst)
+    Seq(Ast(iteratorLocalNode), iteratorAssignAst, forAst)
   }
 
   def astForForEach(stmt: ForEachStmt, order: Int): Seq[Ast] = {
@@ -3083,7 +3089,7 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
     val lambdaMethodNode = createLambdaMethodNode(expr, parametersWithoutThis, returnType)
     addLambdaMethodBindingToDiffGraph(lambdaMethodNode)
 
-    val methodReturnNode = methodReturnNode(returnType, None,  line(expr), column(expr))
+    val returnNode = methodReturnNode(returnType, None,  line(expr), column(expr))
     val virtualModifier = Some(NewModifier().modifierType(ModifierTypes.VIRTUAL))
     val staticModifier  = Option.when(thisParam.isEmpty)(NewModifier().modifierType(ModifierTypes.STATIC))
     val privateModifier = Some(NewModifier().modifierType(ModifierTypes.PRIVATE))
@@ -3106,7 +3112,7 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
         .withChildren(modifiers)
         .withChildren(parameters)
         .withChild(lambdaMethodBody)
-        .withChild(Ast(methodReturnNode))
+        .withChild(Ast(returnNode))
 
     val lambdaMethodAst = identifiersMatchingParams.foldLeft(lambdaMethodAstWithoutRefs)((ast, identifier) =>
       ast.withRefEdge(identifier, lambdaParameterNamesToNodes(identifier.name))
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
@@ -6,6 +6,7 @@ import com.github.javaparser.ast.body.{
   AnnotationDeclaration,
   BodyDeclaration,
   CallableDeclaration,
+  ClassOrInterfaceDeclaration,
   ConstructorDeclaration,
   EnumConstantDeclaration,
   FieldDeclaration,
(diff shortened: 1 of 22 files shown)


joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
```diff
@@ -1,130 +1,667 @@
...
+        |    public static void sink(String s) {}
+        |
+        |    public Consumer<String> test() {
+        |        return input -> sink(input + s);
+        |    }
+        |}
+        |""".stripMargin)
+
+    // TODO: Add 0th parameter logic
+    "create a 0th `this` parameter" ignore {
```
ursachec, line 283: why the ignore?

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
```diff
@@ -1,130 +1,667 @@
...
+        |    public void sink(String s) {}
+        |
+        |    public Consumer<String> test() {
+        |        return input -> sink(input + s);
+        |    }
+        |}
+        |""".stripMargin)
+
+    // TODO: Add 0th parameter logic
+    "create a 0th `this` parameter" ignore {
```
ursachec, line 318: again, why the ignore?

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
```diff
@@ -1,130 +1,667 @@
...
+          interfaceBinding.methodFullName shouldBe "Foo.baz:java.lang.String(java.lang.Object,java.lang.Object)"
+          interfaceBinding.signature shouldBe "java.lang.String(java.lang.Object,java.lang.Object)"
+          binding.inE.collectAll[Binds].map(_.outNode().asInstanceOf[TypeDecl]).l match {
+            case List(typeDecl) => typeDecl.fullName shouldBe "Foo"
+            case result         => fail(s"Expected Binds edge from typeDecl but got $result")
+          }
+
+          binding.methodFullName shouldBe "TestClass.lambda$0:java.lang.String(java.lang.Integer,java.lang.Float)"
+          binding.signature shouldBe "java.lang.String(java.lang.Integer,java.lang.Float)"
+          binding.inE.collectAll[Binds].map(_.outNode().asInstanceOf[TypeDecl]).l match {
```
ursachec, line 563: maybe it's possible to not do the hard cast there

ursachec:
all in all looks good, as discussed offline, i'd include more DF tests before getting this in. no need to rush i'd say, but these are only my two cents on the matter ; )

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
johannescoetzee: You're right! I'm not sure why I chose to do it this way, but it's much nicer to match the specific type instead.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
johannescoetzee: Because if I didn't ignore it, then it would fail ;) Fixing the issue isn't a bad idea though

johannescoetzee:
Commit [92a424f](https://github.com/joernio/joern/pull/1437/commits/92a424fa351caba70dd94c17849b26f71850e119) which will break tests revealed yet another issue with resolving types. In the example 
```
public class TestClass {

    public static String foo(Integer x, Float y, String z) { return z; }

    public static Foo<Integer, Float> test(String captured) {
        return (input, moreInput) -> foo(input, moreInput, captured);
    }
}
```
(where `Foo` is defined in scope), JavaParser resolves the tyeps of `input` and `moreInput` as `Object`, so can't resolve the call to `foo`. This means that any call in a lambda that takes one of the lambda parameters as argument will most likely fail to resolve. We need to double check fallback types to fix this, but in the meantime this PR will be red.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
johannescoetzee: See above

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
johannescoetzee: Also done

johannescoetzee:
> all in all looks good, as discussed offline, i'd include more DF tests before getting this in. no need to rush i'd say, but these are only my two cents on the matter ; )

I added one OSS dataflow tests for lambdas which is currently `pendingUntilFixed`. Fixing this would probably require significant changes to open source call linking. With that being said, these changes do fix the full suite of closed source dataflow tests (PR to update those will follow once this is merged)

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -932,25 +987,44 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+  }
 
-      case _ => ""
+  private def modifierAstsForMethod(methodDeclaration: MethodDeclaration): Seq[Ast] = {
+    val isInterfaceMethod         = scopeStack.getEnclosingTypeDecl.exists(_.code.contains("interface "))
+    val isAbstractMethod          = methodDeclaration.isAbstract || (isInterfaceMethod && !methodDeclaration.isDefault)
+    val abstractModifier          = Option.when(isAbstractMethod)(NewModifier().modifierType(ModifierTypes.ABSTRACT))
+    val staticVirtualModifierType = if (methodDeclaration.isStatic) ModifierTypes.STATIC else ModifierTypes.VIRTUAL
+    val staticVirtualModifier     = Some(NewModifier().modifierType(staticVirtualModifierType))
+    val accessModifierType = if (methodDeclaration.isPublic) {
```
ursachec, line 1003: please remove brackets in if-else's to make things more readable

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -1292,22 +1360,477 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+        .methodFullName(Operators.lessThan)
+        .code(s"$idxName < $iterableName.length")
+        .typeFullName(TypeConstants.Boolean)
+        .dispatchType(DispatchTypes.STATIC_DISPATCH)
+        .order(3)
+        .argumentIndex(3)
+        .lineNumber(lineNo)
+    val comparisonIdxIdentifier = identifierFromNamedVarType(idxLocal, lineNo, 1)
+    val comparisonFieldAccess =
+      NewCall()
```
ursachec, line 1477: if you make a fn like this one https://github.com/joernio/joern/blob/master/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/Nodes.scala#L210-L226 you'll make the code a lot more readable. highly recommended

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -2323,183 +2926,353 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+            expectedTypeParamTypes.getOrElse(returnType.asTypeParameter, returnType)
 
-    Ast(methodRef)
+          case returnType => returnType
+        }
+      }
+      .orElse {
+        maybeResolvedLambdaType
+      }
+      .map {
```
ursachec, line 3028: also here, good pattern for inlining

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/Scope.scala
```diff
@@ -1,46 +1,125 @@
...
+  }
+
+  def getCapturedVariables: List[NewNode with HasTypeFullName with HasNameMutable] = {
+    stack
+      .foldLeft(List.empty[NewNode]) { (acc, stackEntry) =>
+        acc ++ stackEntry.variables.values.map(_.node)
+      }
+      .collect {
+        case param: NewMethodParameterIn => param
+        case local: NewLocal             => local
```
ursachec, line 54: if i'm not mistaken, a _case_ for `NewMember` is missing here

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -691,14 +741,34 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
     Ast(entryNode).withChildren(args)
   }
 
+  private def modifiersForFieldDeclaration(decl: FieldDeclaration): Seq[Ast] = {
+    val staticModifier = if (decl.isStatic) {
```
ursachec, line 745: rm brackets

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -1292,22 +1360,477 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+      NewLocal()
+        .name(iterableName)
+        .code(iterableName)
+        .typeFullName(iterableType)
+        .order(order)
+        .lineNumber(lineNo)
+    val iterableLocalAst = Ast(iterableLocalNode)
+
+    val iterableAssignNode =
+      NewCall()
```
ursachec, line 1396: extract to `operatorCallNode` fn

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -1292,22 +1360,477 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+        .lineNumber(lineNo)
+        .order(1)
+    scopeStack.addToScope(idxName, idxLocal)
+    idxLocal
+  }
+
+  private def nativeForEachIdxInitializerAst(lineNo: Option[Integer], idxLocal: NewLocal): Ast = {
+    val idxName = idxLocal.name
+    val idxInitializerCallNode =
+      NewCall()
```
ursachec, line 1435: extract to `operatorCallNode` fn

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -1292,22 +1360,477 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+    iterableLocal: NamedVariableNodeType,
+    idxLocal: NewLocal
+  ): Ast = {
+    val idxName      = idxLocal.name
+    val iterableName = iterableLocal.name
+
+    val compareNode =
+      NewCall()
+        .name(Operators.lessThan)
+        .methodFullName(Operators.lessThan)
```
ursachec, line 1468: extract to `operatorCallNode` fn

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -1292,22 +1360,477 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+
+    callAst(compareNode, compareArgs)
+      .withRefEdge(comparisonIdxIdentifier, idxLocal)
+      .withRefEdge(fieldAccessIdentifier, iterableLocal)
+  }
+
+  private def nativeForEachIncrementAst(lineNo: Option[Integer], idxLocal: NewLocal): Ast = {
+    val incrementNode =
+      NewCall()
+        .name(Operators.postIncrement)
```
ursachec, line 1506: extract to `operatorCallNode` fn

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -2323,183 +2926,353 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+    NewClosureBinding()
+      .closureBindingId(closureBindingId)
+      .closureOriginalName(originalName)
+      .evaluationStrategy(EvaluationStrategies.BY_SHARING)
+  }
+
+  private def closureBindingsForCapturedNodes(captured: List[NamedVariableNodeType]): List[ClosureBindingEntry] = {
+    captured
+      .map { capturedNode =>
+        val closureBindingId   = randomUUID().toString
```
ursachec, line 3044: add a TODO that this should not be random. we want a CPG generated from the same source to be identical on a new run

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -2508,15 +3281,29 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
-    maybeResolvedCall.toOption.flatMap { methodDecl =>
-      Try(typeInfoCalc.fullName(methodDecl.getParam(idx).getType)).toOption
+  private def getExpectedParamType(
+    maybeResolvedCall: Try[ResolvedMethodLikeDeclaration],
+    idx: Int
+  ): Option[ExpectedType] = {
+    maybeResolvedCall.toOption.map { methodDecl =>
+      val paramCount = methodDecl.getNumberOfParams
+
+      val resolvedType = if (idx < paramCount) {
```
ursachec, line 3297: rm brackets

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -2572,21 +3359,68 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+    tryResolvedDecl match {
+      case Success(_) if hasVariadicParameter =>
+        val expectedVariadicTypeFullName =
+          getExpectedParamType(tryResolvedDecl, paramCount - 1)
+            .map(_.fullName)
+            .getOrElse(TypeConstants.UnresolvedType)
+        val (regularArgs, varargs) = argsAsts.splitAt(paramCount - 1)
+        val arrayInitializer =
+          NewCall()
+            .name(Operators.arrayInitializer)
```
ursachec, line 3390: `operatorCallNode` fn

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
```diff
@@ -1,130 +1,755 @@
...
+          }
+          lambdaMethod.methodReturn.typeFullName shouldBe "java.lang.String"
+
+        case result => fail(s"Expected single lambda method but got $result")
+      }
+    }
+
+    "create the correct binding for the lambda method" in {
+      cpg.all.collectAll[Binding].filter(_.name == "lambda$0").l match {
+        case List(lambdaBinding) =>
```
ml86, line 98: There should not be a binding with the name of the implementing method. Just the bindings of the implemented methods name is enough. It does not really hurt but it is noise so please remove those bindings.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/Scope.scala
johannescoetzee: I originally had one there, but don't think there should be. The justification is that, for lambdas using object members, a `this` reference is added as the 0th lambda param (consistent with the java compiler), so the member would be accessed as `this.member`.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
```diff
@@ -1,130 +1,755 @@
...
     }
   }
 
-  "CPG for `identity-like` supplier" - {
-    lazy val cpg = JavaSrc2CpgTestContext.buildCpg("""
+  // This is an example of a case where all the type info we need to figure out that the lambda implements
+  // `Function<String, String>`, but from JavaParser and the current expectedType propagation we only get
+  // `java.util.Function<Object, Object>`. Ideally we'd want all occurrences of `java.lang.Object` to be
+  // `java.lang.String` instead.
+  "a lambda implementing a mapper in stream" should {
```
ml86, line 672: Is this another bug in JavaParser?

ml86 (changes requested):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -2323,183 +2926,353 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
-      (Nil, Nil)
+    if (paramTypesList.size != lambdaParameters.size) {
+      logger.error(s"Found different number lambda params and param types for $expr. Some parameters will be missing.")
     }
 
-    val closureBindings = closureBindingsForLambdas(identifiersNotMatchingParams)
-    val refEdgePairs    = buildRefEdgePairs(identifiersMatchingParams, namesToMethodParams)
+    val parameterNodes = lambdaParameters
+      .lazyZip(paramTypesList)
+      .lazyZip(LazyList.from(1))
```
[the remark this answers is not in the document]
johannescoetzee: To answer this and the above comment: I did it this way, since the lazyZips give a result of `List[(NewParameter, String, Int)]` instead of `List[((NewParameter, String), Int)]`, the former being a bit cleaner in my opinion. I don't really have a strong opinion either way, however, so will update.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -2323,183 +2926,353 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
-    }
+  private def genericParamTypeMapForLambda(
+    expectedType: Option[ExpectedType]
+  ): Map[ResolvedTypeParameterDeclaration, ResolvedType] = {
+    expectedType
+      .flatMap(_.resolvedType)
+      // This should always be true for correct code
+      .collect { case r: ResolvedReferenceType => r }
+      .map { resolvedType =>
+        resolvedType.getTypeParametersMap.asScala.map { tp =>
```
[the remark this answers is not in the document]
johannescoetzee: This isn't really something I gave explicit thought. I did it this way for the convenience of working with Scala maps, but you're right that consistency is better here. Updated for next revision.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
```diff
@@ -1,130 +1,755 @@
...
+
+    "create 2 method nodes for the respective lambdas" in {
+      cpg.method.name(".*lambda.*").l match {
+        case List(lambda0, lambda1) =>
+          lambda0.name shouldBe "lambda$0"
+
+          lambda1.name shouldBe "lambda$1"
+
+        case result => fail(s"Expected 2 lambda methods but got $result")
+      }
```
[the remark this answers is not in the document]
johannescoetzee: Noted and fixed.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
```diff
@@ -1,130 +1,755 @@
...
+    }
+
+    "create the correct binding for the lambda method" in {
+      cpg.all.collectAll[Binding].filter(_.name == "lambda$0").l match {
+        case List(lambdaBinding) =>
+          lambdaBinding.methodFullName shouldBe "Foo.lambda$0:java.lang.String(java.lang.String)"
+          lambdaBinding.signature shouldBe "java.lang.String(java.lang.String)"
+
+          lambdaBinding.inE.collectAll[Binds].map(_.outNode()).l match {
```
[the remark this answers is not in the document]
johannescoetzee: That's much cleaner, thanks!

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
johannescoetzee: I haven't looked too deeply into this, but as far as I can tell it's a limitation of the JavaParser type solver rather than a bug.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -2323,183 +2926,353 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
-      // bodyAst.ctx.identifiers.values.toSeq.partition(identifier => namesToMethodParams.contains(identifier.name))
-      (Nil, Nil)
+    if (paramTypesList.size != lambdaParameters.size) {
+      logger.error(s"Found different number lambda params and param types for $expr. Some parameters will be missing.")
     }
 
-    val closureBindings = closureBindingsForLambdas(identifiersNotMatchingParams)
-    val refEdgePairs    = buildRefEdgePairs(identifiersMatchingParams, namesToMethodParams)
+    val parameterNodes = lambdaParameters
+      .lazyZip(paramTypesList)
```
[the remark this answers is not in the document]
johannescoetzee: See below.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -2323,183 +2926,353 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+      .map { boundMethod =>
+        boundMethod.getReturnType match {
+          case returnType: ResolvedTypeVariable =>
+            expectedTypeParamTypes.getOrElse(returnType.asTypeParameter, returnType)
 
-    Ast(methodRef)
+          case returnType => returnType
+        }
+      }
+      .orElse {
```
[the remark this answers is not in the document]
johannescoetzee: Hmm, I don't think inlining is appropriate in this specific case, but the fact that this was brought up does mean that some refactoring is required. I've broken this chain into separate assignments that should hopefully be more readable.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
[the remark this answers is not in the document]
johannescoetzee: Noted and done. I'm using `s"$lambdaMethodName:${capturedNode.name}"` now, which should be a unique combination.

ml86 (approved):

ursachec (approved):

PR #1445, opened 2022-05-27
Make `--optimize-dependencies` the default for `js2cpg`

fabsx00:
`js2cpg` recently introduced the `--optimize-dependencies` flag. When used, `js2cpg` will be more selective about downloading dependencies, which significantly reduces CPG construction time. While the flag is not enabled by default for `js2cpg` runs, I am making it the default for `joern`. It is expected that it will be the default in a few months, at which point we do not need to specify the flag manually anymore.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,8 +2,8 @@ name                     := "joern"
 ThisBuild / organization := "io.joern"
 ThisBuild / scalaVersion := "2.13.8"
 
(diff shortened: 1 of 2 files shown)


mpollmeier (approved):

max-leuthaeuser:
Please also for the js2cpg.bat for Windows 🙂

PR #1451, opened 2022-05-30
run kotlin tests in parallel: execution time 62s -> 25s

mpollmeier:
note: the `kotlin2cpgdependencies*` temp files are now being created
in the system temp dir rather than the project dir. it wouldn't really
matter, but because we're not forking the jvm for test execution, they
are now only being deleted when the user exits sbt...


mpollmeier:
Unfortunately the tests don't clean up after themselves and so we end up with OutOfMemory errors in github actions. Shouldn't be a big issue though, I'll TAL later. 
In general though, that's the risk with not forking the JVM...

ursachec:
:raised_hands: @mpollmeier

fabsx00 (approved):

PR #1467, opened 2022-06-02
alternative fix for CfgCreator complexity class woes

bbrehm:
This is an alternative to https://github.com/joernio/joern/pull/1455

This does change behavior: We generated a lot of incorrect CFGs. Generally speaking, I would very much appreciate a source-code comment if we test for known incorrect behavior. Such tests are still valuable for documenting known shortcomings, but a comment (like e.g. below wrt empty loops) makes life much easier when testing.

@fabsx00 -- are you happy with readability?


fabsx00:
This fixes the incorrect CFGs, which were related to nesting of continues/breaks/case-labels: https://github.com/joernio/joern/pull/1471

PR #1485, opened 2022-06-09
Fix NPE blindly

ursachec:
test tbd


ursachec:
found while running `kotlin2cpg` on https://github.com/tachiyomiorg/tachiyomi

johannescoetzee (approved):
It sucks that this is necessary, but

PR #1487, opened 2022-06-09
Refactor single class in `kotlin2cpg`

ursachec:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/DefaultTypeInfoProvider.scala
```diff
@@ -105,54 +100,45 @@ class DefaultTypeInfoProvider(environment: KotlinCoreEnvironment) extends TypeIn
...
   }
 
   def isRefToCompanionObject(expr: KtNameReferenceExpression): Boolean = {
     val mapForEntity = bindingsForEntity(bindingContext, expr)
-    mapForEntity.getKeys
-      .contains(BindingContext.SHORT_REFERENCE_TO_COMPANION_OBJECT.getKey)
+    Option(mapForEntity)
+      .map(_.getKeys)
+      .map(_.contains(BindingContext.SHORT_REFERENCE_TO_COMPANION_OBJECT.getKey))
+      .getOrElse(false)
```
johannescoetzee, line 141: See comment above

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/DefaultTypeInfoProvider.scala
```diff
@@ -326,56 +296,40 @@ class DefaultTypeInfoProvider(environment: KotlinCoreEnvironment) extends TypeIn
...
-        }
-      case None => None
-    }
+    resolvedCallDescriptor(expr)
+      .collect {
+        case _: ClassConstructorDescriptorImpl     => Some(true)
+        case _: TypeAliasConstructorDescriptorImpl => Some(true)
+        case _                                     => Some(false)
+      }
+      .getOrElse(None)
```
johannescoetzee, line 309: This can be cleaned up further by having `isConstructorCall` return a `Boolean`, since it always seems to be used as
```
isConstructorCall(expr).getOrElse(false)
```

The shortest implementation I can think of is
```
resolvedCallDescriptor(expr).exists {
  case _: ClassConstructorDescriptorImpl     => Some(true)
  case _: TypeAliasConstructorDescriptorImpl => Some(true)
  case _                                     => Some(false)
}
```
which will return false if `resolvedCallDescriptor(expr)` is `None` or contains anything except the 2 matched types.

max-leuthaeuser (approved):

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/DefaultTypeInfoProvider.scala
```diff
@@ -504,36 +430,34 @@ class DefaultTypeInfoProvider(environment: KotlinCoreEnvironment) extends TypeIn
...
-            val extType = originalDesc.getExtensionReceiverParameter.getType
-            val extName = originalDesc.getName
-            val rendered =
-              if (renderedFqNameForDesc.startsWith(TypeConstants.kotlinApplyPrefix)) {
-                TypeConstants.javaLangObject
-              } else {
-                TypeRenderer.render(extType, false, false)
-              }
-            s"$rendered.$extName"
+            Some((originalDesc.getName, originalDesc.getExtensionReceiverParameter))
```
johannescoetzee, line 436: The change to this block in general makes it SOOOOO much cleaner than before, but the new version can be rewritten as 
```
Option(originalDesc.getExtensionReceiverParameter).map { extensionReceiverParameter =>
  (originalDesc.getName, extensionReceiverParameter)
}
```
which is, in my opinion, nicer than the explicit null check (I really hate null ;))

johannescoetzee: This comment should actually include the 
```
 } else {
  None
  }
  ```
  below

File: joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/DefaultTypeInfoProvider.scala [hunk 4]
```diff
@@ -706,28 +608,24 @@ class DefaultTypeInfoProvider(environment: KotlinCoreEnvironment) extends TypeIn
...
       .collect {
-        case _: LazyJavaClassDescriptor =>
-          true
-        case _: LazyClassDescriptor =>
-          true
+        case _: LazyJavaClassDescriptor => true
+        case _: LazyClassDescriptor     => true
       }
       .getOrElse(false)
   }
```
johannescoetzee, line 631: This could be replaced by an `exists` call with a catch-all third case (see my comment about `isConstructorCall`). It's less of a thing here since this already returns a `Boolean`, however.

johannescoetzee (approved):
These comments are all very subjective, so feel free to pick which ones you want to incorporate :)

ursachec:
thxies @johannescoetzee! all good points

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/DefaultTypeInfoProvider.scala
ursachec: yess

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/DefaultTypeInfoProvider.scala
ursachec: :+1:

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/DefaultTypeInfoProvider.scala
ursachec: i'll leave this one as it is for now

PR #1499, opened 2022-06-15
kotlin2cpg: Bootstrap DF test suite

ursachec:


johannescoetzee (approved):

johannescoetzee:
These changes are fine as is, but for consistency between frontends, I'd suggest extending the new `Code2CpgFixture` introduced by @ml86 recently. You can see an example of how I defined my new test fixture for javasrc in https://github.com/joernio/joern/blob/johannes/javasrc-lambdas/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/testfixtures/JavaSrcCodeToCpgFixture.scala

ursachec:
ur right, that is nicer, thx

PR #1500, opened 2022-06-15
jssrc2cpg: performance fix for file reading

max-leuthaeuser:
Massively improves performance for larger JS/TS projects (e.g., https://github.com/appsmithorg/appsmith 40 sec -> 10 sec)
by using scala.io.Source.fromFile instead of our own IOUtils from x2cpg.
Have to figure out why.


max-leuthaeuser:
That's the plan 🙂

mpollmeier:
my guess would be that it's this regex replace that we invoke for every line: https://github.com/ShiftLeftSecurity/codepropertygraph/blob/master/codepropertygraph/src/main/scala/io/shiftleft/utils/IOUtils.scala#L47

max-leuthaeuser:
Yes, profiling indicates that.
I guess having that regex precompiled would help already.

max-leuthaeuser:
Yes, this was the problem. Will do a separate PR for it.

PR #1521, opened 2022-06-23
Allow only a single input path for CPG generation

fabsx00:
This is a simplification that I've wanted to bring in for a while and that has now become more important as we're fixing the FILENAME field to contain relative paths: CPGs should only be created from a single input file or input directory. If multiple projects are to be included in one CPG, the user can move them into a parent directory, but that shouldn't be of concern to us.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
@@ -7,7 +7,7 @@ object SourceFiles {
   /** For a given input path, determine all source files by inspecting filename extensions.
     */
   def determine(inputPath: String, sourceFileExtensions: Set[String]): List[String] = {
-    determine(inputPath, sourceFileExtensions)
+    determine(Set(inputPath), sourceFileExtensions)
   }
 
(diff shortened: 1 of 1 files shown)


mpollmeier (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
```diff
@@ -4,6 +4,12 @@ import better.files._
 
 object SourceFiles {
 
+  /** For a given input path, determine all source files by inspecting filename extensions.
+    */
+  def determine(inputPath: String, sourceFileExtensions: Set[String]): List[String] = {
+    determine(inputPath, sourceFileExtensions)
```
max-leuthaeuser, line 10: `determine(Set(inputPath),...`

max-leuthaeuser (changes requested):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
max-leuthaeuser, line 10:
```suggestion
    determine(Set(inputPath), sourceFileExtensions)
```

PR #1528, opened 2022-06-27
Set Metadata.root field for all frontends

fabsx00:
For all frontends included in the joern repo, we are now setting the Metadata.root field to contain the absolute path of the root directory of the imported code. The next step will be to store relative paths in `FILENAME` fields and adapt `.dump` to make use of both the `Metadata.root` field and the relative filename to determine which file to display.


max-leuthaeuser (approved):

max-leuthaeuser:
jssrc2cpg is missing.

max-leuthaeuser (changes requested):
jssrc2cpg

max-leuthaeuser (approved):

PR #1538, opened 2022-06-30
Johannes/javasrc remove order

johannescoetzee:
* Remove most explicit order setting from javasrc2cpg. The only remaining order setting is along with the `index` and `argumentIndex` setting for parameters and arguments respectively.
* Make the `closureBindingId` field actually unique. This was only unique for `closureBindingIds` contained in a single file before and not across files.


joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/CallTests.scala
```diff
@@ -218,34 +218,34 @@ class CallTests extends JavaSrcCodeToCpgFixture {
...
 
   "should allow traversing from call to arguments" in {
     cpg.call("add").argument.size shouldBe 3
-    val List(arg0) = cpg.call("add").argument(0).l
+    val List(arg0) = cpg.call("add").receiver.l
     arg0.isInstanceOf[nodes.Identifier] shouldBe true
     arg0.asInstanceOf[nodes.Identifier].name shouldBe "this"
     arg0.code shouldBe "this"
-    arg0.order shouldBe 0
+//    arg0.order shouldBe 0
```
ml86, line 234: Why is this commented out?

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/CallTests.scala
```diff
@@ -334,6 +334,8 @@ class CallTests extends JavaSrcCodeToCpgFixture {
 
     argument.name shouldBe Operators.fieldAccess
     argument.typeFullName shouldBe "test.MyObject"
+    argument.order shouldBe 1
```
ml86, line 337: For easier readability please also check for `argument.code shouldBe "obj"`.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
```diff
@@ -288,17 +288,18 @@ class LambdaTests extends JavaSrcCode2CpgFixture {
...
     // TODO: Add 0th parameter logic
     "create a 0th `this` parameter" in {
+      val lambdaChildren = cpg.method.name(".*lambda.*").astChildren.l
       cpg.method.name(".*lambda.*").parameter.l match {
         case List(thisParam, inputParam) =>
           thisParam.name shouldBe "this"
           thisParam.code shouldBe "this"
           thisParam.typeFullName shouldBe "Foo"
-          thisParam.order shouldBe 0
+//          thisParam.order shouldBe 0
```
ml86: Why is this commented out?

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -2293,89 +2116,92 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+      a.root.collect { case x: ExpressionNew =>
+        x.argumentIndex = i
+        x.order = i
+      }
+    }
+  }
+
+  // TODO: This is an almost exact copy of the method in x2cpg. Fix this.
+  // The reason this exists here is that some dataflow tests are failing without the arg edge
+  // to the receiver.
```
ml86, line 2155: It depends on the language whether the receiver is also an argument to the called method or not. So the x2cpg version of this method should support both. What is its name? I could not find a `astWithRecv...` named method in x2cpg.

ml86 (changes requested):

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/CallTests.scala
johannescoetzee: Good question. I commented out a bunch of order tests while testing argument indices and have since fixed most of those, but it seems this and `thisParam` below slipped through. Will fix.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/CallTests.scala
johannescoetzee: Will do!

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
johannescoetzee: See above

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
johannescoetzee: That's a good point. The method in `x2cpg` is actually called `callAst`. I originally called this something similar, but later changed it to avoid confusing the two without also updating the comment.

johannescoetzee: To address your actual point here, I'll update the `callAst` method in x2cpg as well and switch back to that.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
```diff
@@ -112,8 +113,21 @@ abstract class AstCreatorBase(filename: String) {
...
+  ): Ast = {
+    astWithRecvAndArgs(callNode, arguments, receiver, withRecvArgEdge)
+  }
 
+  def astWithRecvAndArgs(
+    rootNode: NewNode,
+    arguments: Seq[Ast] = List(),
+    receiver: Option[Ast] = None,
+    withRecvArgEdge: Boolean = false
+  ): Ast = {
```
ml86, line 130: We dont need the same functionality under 2 names. Please just move the body of this function in `callAst`.

ml86 (changes requested):

PR #1545, opened 2022-07-05
javasrc2cpg cleanup 1

johannescoetzee:
This is the first in a series of javasrc2cpg cleanup PRs. The main changes here are:
* Moving the NodeBuilders object to x2cpg
* Adding (and using) the identifierNode method.
* Moving some javasrc magic strings into a constants object.
There are also a few misc improvements that I noticed as I was working on the rest.


ursachec (approved):
lgtm

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/NodeBuilders.scala
```diff
@@ -0,0 +1,30 @@
...
+    column: Option[Integer] = None,
+    dynamicTypeHintFullName: Seq[String] = Seq.empty
+  ): NewIdentifier = NewIdentifier()
+    .name(name)
+    .code(name)
+    .typeFullName(typeFullName)
+    .lineNumber(line)
+    .columnNumber(column)
+    .dynamicTypeHintFullName(dynamicTypeHintFullName)
+}
```
ml86, line 30: I would have expected those methods to be also in `AstCreatorBase`. Is there a reasons why you did not do that?

PR #1547, opened 2022-07-06
javasrc cleanup #2

johannescoetzee:
Another small cleanup PR, mostly for modifiers.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
@@ -684,14 +684,15 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
     val staticModifier =
       Option.when(decl.isStatic)(modifierNode(ModifierTypes.STATIC))
 
-    val accessModifierType = if (decl.isPublic)
-      Some(ModifierTypes.PUBLIC)
-    else if (decl.isPrivate)
-      Some(ModifierTypes.PRIVATE)
-    else if (decl.isProtected)
-      Some(ModifierTypes.PROTECTED)
-    else
-      None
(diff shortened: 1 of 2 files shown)


ursachec (approved):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -310,7 +309,7 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
     val parameterTypes = calcParameterTypes(method, typeParamValues)
 
     val returnType =
-      Try(method.getReturnType).toOption
+      Try(method.getReturnType)
         .map(returnType => typeInfoCalc.fullName(returnType, typeParamValues))
```
ml86, line 313: Unless you have good reason to catch exceptions from the `map` please do not remove `toOption`.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -429,20 +428,17 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
   }
 
   private def identifierForResolvedTypeParameter(typeParameter: ResolvedTypeParameterDeclaration): NewIdentifier = {
     val name = typeParameter.getName
-    val typeFullName = Try(typeParameter.getUpperBound) match {
-      case Success(upperBound) => typeInfoCalc.fullName(upperBound)
-      case Failure(_)          => TypeConstants.Object
-    }
+    val typeFullName = Try(typeParameter.getUpperBound)
+      .map(typeInfoCalc.fullName)
```
ml86, line 440: Same as above. You are catching exception from the `map` now and silently ignore them. Is that intended?

PR #1469, opened 2022-06-03
Make jssrc2cpg default

pr0me:
Regarding issue #1468.

This commit makes the necessary changes to register `jssrc2cpg` as the default frontend for javascript projects.

PR #1563, opened 2022-07-08
Make jssrc2cpg default (updated version)

max-leuthaeuser:
Replaces: https://github.com/joernio/joern/pull/1469


fabsx00 (commented):
Have we checked here what happens when `astgen` isn't installed? Would be good to tell the user to install it using `npm`, possibly with the correct command line.

PR #1565, opened 2022-07-08
Fix name for  identifier nodes

johannescoetzee:
Fixes https://github.com/ShiftLeftSecurity/codescience/issues/6289


mpollmeier (approved):
verified: it works, thank you!

PR #1566, opened 2022-07-08
joern-scan: add config for query dump destination

ursachec:


mpollmeier (approved):
👍🏻 
if you want to take it one step further: define a static `DefaultDumpDestination` e.g. in `object JoernScan` and refer to it in both the helptext and the parameter

joern-cli/src/main/scala/io/joern/joerncli/JoernScan.scala
```diff
@@ -59,7 +60,11 @@ object JoernScan extends App with BridgeBase {
 
     opt[Unit]("dump")
       .action((_, c) => c.copy(dump = true))
-      .text("Dump available queries to file")
+      .text("Dump available queries to a temporary file")
```
mpollmeier, line 63:
```suggestion
      .text("Dump available queries to a file")
```
I'd keep it as before - the file isn't automatically removed, at least not by us...

mpollmeier (approved):

joern-cli/src/main/scala/io/joern/joerncli/JoernScan.scala
ursachec: :+1: good point, thx

ursachec:
> 👍🏻 if you want to take it one step further: define a static `DefaultDumpDestination` e.g. in `object JoernScan` and refer to it in both the helptext and the parameter

yess, much nicer, danke

PR #1574, opened 2022-07-11
javasrc: New test fixture migration #1

johannescoetzee:
This PR moves the first couple of test classes to the new `JavaSrcCode2CpgFixture`. In some cases I left the tests unchanged, using a single code block, since the tests themselves are very simple and, in my opinion, the effort of restructuring the tests outweighs the benefits of doing so. For more complex test suites (especially once I get to the dataflow tests) this will not be the case.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
@@ -141,7 +141,6 @@ import io.shiftleft.codepropertygraph.generated.nodes.{
   NewMethodReturn,
   NewModifier,
   NewNamespaceBlock,
-  NewNode,
   NewReturn,
   NewTypeDecl,
   NewTypeRef,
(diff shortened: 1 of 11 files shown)


max-leuthaeuser:
Also note, that adding lazy vals does increase compilation- and run-time slightly due to some overhead.
This is described in detail [here](https://docs.scala-lang.org/sips/improved-lazy-val-initialization.html#version-6---no-synchronization-on-this-and-concurrent-initialization-of-fields) (v6 is the implementation of lazy in Scala that is currently used).
Also see: [performance comparison]( https://d-d.me/tnc/30/lazy-sip-perf/report/#config=%7B%22filterConfig%22%3A%7B%22curves%22%3A%5B%2216%22%2C%2217%22%2C%2218%22%2C%2219%22%2C%2221%22%2C%2222%22%2C%2223%22%5D%2C%22order%22%3A%5B%22param-size%22%2C%22date%22%5D%2C%22filters%22%3A%5B%5B%22100000%22%2C%22300000%22%2C%22500000%22%2C%221000000%22%2C%223000000%22%2C%225000000%22%5D%2C%5B%221477397877000%22%5D%5D%7D%2C%22chartConfig%22%3A%7B%22type%22%3A0%2C%22showCI%22%3Afalse%7D%7D).

In our tests here it will not impact the performance in any way (just some ms when compiling perhaps) but as Markus pointed out it also does not bring anything to the table.

PR #1575, opened 2022-07-11
kotlin2cpg: test dataflows through try and if-else

ursachec:


joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/dataflow/SimpleDataFlowTests.scala
```diff
@@ -103,4 +103,125 @@ class SimpleDataFlowTests extends KotlinCode2CpgFixture(withOssDataflow = true)
...
+    val cpg = code("""
+        |package mypkg
+        |
+        |import kotlin.random.Random
+        |
+        |fun doSomething(p: Int): Int {
+        |  val someVal = 41
+        |
+        |  val afterIfNoBlocks = if (p % 2 == 0) someVal else 42
+        |  println(afterIfNoBlocks)
```
johannescoetzee, line 117: For easier debugging with a debugger, I'd recommend breaking all the tests into smaller chunks. For example, code I highlighted for this comment could stand alone as a separate cpg, independent of the others. That way, if something does break, one wouldn't have to step through ast creation for the entire method (which includes code that's irrelevant for this particular test) and could rather just step through relevant code.

The same comment also applies to the other tests in this PR, but I only mention it here as an example

johannescoetzee (approved):

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/dataflow/SimpleDataFlowTests.scala
ursachec: alright, good point

PR #1585, opened 2022-07-13
querydb: Bootstrap Kotlin Query bundle

ursachec:


querydb/src/test/scala/io/joern/scanners/kotlin/NetworkProtocolsTests.scala
```diff
@@ -0,0 +1,15 @@
...
+import io.joern.suites.KotlinQueryTestSuite
+import io.shiftleft.codepropertygraph.generated.nodes.Call
+
+class NetworkProtocolsTests extends KotlinQueryTestSuite {
+
+  override def queryBundle = NetworkProtocols
+
+  "should find calls relevant to insecure network protocol usage" in {
+    val query = queryBundle.usageOfInsecureProtocol()
+    query(cpg).flatMap(_.evidence).collect { case c: Call => c }.size shouldBe 1
```
johannescoetzee, line 13: Does this line only check the code for the positive example? If not, then I think this test would still pass if the query does the exact opposite of what you want it to (in this case flagging the `https` URL as insecure).

ursachec: No, it checks for both (the `cpg` that the query is run on contains the concatenated positive _and_ negative examples
````
class KotlinQueryTestSuite extends Code2CpgFixture(new Kotlin2CpgFrontend()) {
//...

  def concatQueryCodeExamples: String =
    allQueries
      .map { q =>
        q.codeExamples.positive
          .mkString("\n")
          .concat("\n")
          .concat(
            q.codeExamples.negative
              .mkString("\n")
          )
      }
      .mkString("\n")

  protected val cpg: TestCpg = code(concatQueryCodeExamples)
}
```

ursachec: But good that you point it out, it's not a very intuitive way of testing for negative examples, I'll see if I can think of something different

johannescoetzee (approved):

PR #1586, opened 2022-07-13
Johannes/javasrc abstract fix

johannescoetzee:
I wrote the explanation in code since this result was very surprising to me.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
@@ -2699,8 +2699,13 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
       logger.warn(s"Could not resolve the interface implemented by the lambda $expr. Type info may be missing.")
     }
 
-    // By definition, a functional interface will declare exactly one abstract method, so `find` is fine.
-    val maybeBoundMethod = maybeImplementedInterface.flatMap(_.getDeclaredMethods.asScala.find(_.isAbstract))
+    // While, in theory, there should only be a single abstract method defined by a functional interface, it turns
+    // out this is sort of not true for at least the java.util.Comparator interface, which re-declares the equals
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -2699,8 +2699,23 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
+    // example.
+    val maybeBoundMethod = maybeImplementedInterface.flatMap { interface =>
+      val abstractMethodsWithMatchingParamCount = interface.getDeclaredMethods.asScala
+        .filter(_.isAbstract)
+        .filter(_.getNumberOfParams == expr.getParameters.size())
+      if (abstractMethodsWithMatchingParamCount.size > 1) {
+        logger.warn(
+          s"Found multiple abstract methods matching param count for implementing lambda. Ignoring type information for stability"
+        )
+        None
```
ml86, line 2714: This is not required. According to https://docs.oracle.com/javase/8/docs/api/java/lang/FunctionalInterface.html it is allowed to have multiple abstract methods as long as only one of them is not already defined in `java.lang.Object`. So you can filter those away and we should be fine and not need this untyped fallback.

ml86 (changes requested):

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/TypeTests.scala
```diff
@@ -129,6 +129,28 @@ class NewTypeTests extends JavaSrcCode2CpgFixture {
         case result => fail(s"Expected identifier and arrayInitializer arguments but got $result")
       }
     }
+
+    // This test relies on instability in JavaParser results, so running it multiple times is a way of increasing
+    // confidence in the correctness of the code, albeit at the cost of slower tests.
```
ml86, line 134: That is not correct anymore. So running the test once is enough. I would also expect a different lambda signature in the test below.

ml86 (approved):

PR #1589, opened 2022-07-14
jssrc2cpg: CFG pass test refactorings

max-leuthaeuser:
Split the CfgCreationPassTest into smaller units just like I did with the AST pass tests.


ursachec (approved):
:+1:  :+1:  looks much nicer to me

PR #1594, opened 2022-07-14
jssrc ast root might also be an identifier - add warning for anything else

mpollmeier:


max-leuthaeuser:
I am still not sure if that cuts it.
Typescript allows to export almost anything: https://www.typescriptlang.org/docs/handbook/modules.html
As the name of the exported entity is used in the actual lowering later on (see the tests) I still think it should allow for anything that has a name here. That's why is was using HasName in first place.

max-leuthaeuser:
A better approach including tests: https://github.com/joernio/joern/pull/1596

PR #1602, opened 2022-07-15
StoredNode.tagList: no need to instantiate many `NewTag`

mpollmeier:
instead, just use the existing nodes, nothing special needed here...

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/StoredNodeMethods.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/StoredNodeMethods.scala
@@ -15,16 +15,7 @@ class StoredNodeMethods(val node: StoredNode) extends AnyVal with NodeExtension 
   def file: Traversal[File] =
     Traversal.fromSingle(node).file
 
-  def tagList: Traversal[NewTag] = {
-    node._taggedByOut.asScala
-      .map { case tagNode: HasName with HasValue =>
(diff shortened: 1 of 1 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/StoredNodeMethods.scala
```diff
@@ -15,16 +15,7 @@ class StoredNodeMethods(val node: StoredNode) extends AnyVal with NodeExtension
...
-      .map { case tagNode: HasName with HasValue =>
-        (tagNode.name, Option(tagNode.value))
-      }
-      .distinct
-      .collect { case (name, Some(value)) =>
-        NewTag()
-          .name(name)
-          .value(value)
-      }
+  def tagList: Traversal[Tag] = {
```
ml86, line 18: This does not deduplicate anymore. If i remember correctly that was sometime requried.

mpollmeier: indeedly - here e.g. in codescience TaggingTests:
```
[info] TaggingTests:
[info] Types tagging
[info] - should find only the method1 of class1
[info] - should find only the method1 of class2 which extends from class1 *** FAILED ***
[info]   ArraySeq("y", "y", "z") was not equal to List("y", "z") (TaggingTests.scala:25)
```

ok, will leave as is for now and focus on the replacement of both

PR #1617, opened 2022-07-19
jssrc2cpg: bring in native astgen invocation

max-leuthaeuser:
I was finally able to build a native custom version of astgen (based on @joernio/astgen 2.0.4) using 'pkg' (available [here](https://github.com/max-leuthaeuser/astgen)).
This enables us to run it entirely without node (as a stripped down node comes with the binary), and it is ultimately much faster.


max-leuthaeuser:
Todo: update the docs with description on how to build the astgen binaries yourself.

PR #1619, opened 2022-07-20
kotlin2cpg: fix ast for fns without body-blocks

ursachec:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/KtPsiToAst.scala
```diff
@@ -314,9 +314,23 @@ trait KtPsiToAst {
...
+      case Some(ast) => ast
+      case None =>
+        val bodyBlock = blockNode("", "")
+        ktFn.getBodyExpression match {
+          case expr if expr != null =>
+            blockAst(bodyBlock, astsForExpression(expr, None).toList)
+          case _ =>
+            blockAst(bodyBlock, List())
+        }
     }
```
max-leuthaeuser, line 334: This could be one match. E.g., you could wrap `ktFn.getBodyBlockExpression` like `Option(ktFn.getBodyBlockExpression)` which will automatically result in `None` if its `null`.

ursachec: yess, much nicer, done

max-leuthaeuser (approved):

PR #1621, opened 2022-07-20
kotlin2cpg: test interprocedural dataflows

ursachec:


joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/dataflow/InterproceduralTests.scala
```diff
@@ -0,0 +1,82 @@
...
+        |""".stripMargin,
+        "file2.kt"
+      )
+
+    "find flows through the function" in {
+      val source = cpg.literal.codeExact("41414141")
+      val sink   = cpg.method.name("println").callIn.argument
+      val flows  = sink.reachableByFlows(source)
+      flows.map(flowToResultPairs).toSet shouldBe
+        Set(List(("val aVal = 41414141", Some(4)), ("f2(aVal)", Some(5)), ("f2(p)", Some(3)), ("println(p)", Some(4))))
```
johannescoetzee, line 36: Although this does fit on one line, I think it would be more readable to write this across multiple lines as in the test below.

johannescoetzee:
A general comment about these tests is that they only test dataflow one way, i.e. trough values passed as arguments in function calls. A test for tracking dataflow through a return value would be useful (assuming this doesn't exist already and was not planned for a separate future PR).

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/dataflow/InterproceduralTests.scala
ursachec: alright, maybe, but _scalafmt_ currently moves everything in one line and i'd rather make changes there after we've decided on them...

johannescoetzee: Ah yes, scalafmt exists. In this case, please ignore my comment :)

ursachec: alrighty ;)

Issue #1606, opened 2022-07-16
scala.MatchError while importCode


vishesh27k:
importCode throwing error for repository - https://github.com/keycloak/keycloak
Cpg got generated but couldn't find some methods.

<img width="1435" alt="image" src="https://user-images.githubusercontent.com/45128646/179346284-9c9173a9-3874-4615-86c4-7703c7447197.png">
<img width="666" alt="image" src="https://user-images.githubusercontent.com/45128646/179346298-9b672312-a739-466f-965c-9220e9cdeb36.png">

mpollmeier:
I tested with keycloak, this is fixed
@vishesh27k thanks for repoorting
@johannescoetzee thanks for fixing

PR #1622, opened 2022-07-20
javasrc: Add error handling for some javaparser UnsolvedSymbolExceptions and StackOverflowErrors

johannescoetzee:
These changes were motivated by https://github.com/joernio/joern/issues/1606 and include a couple of error-handling changes for JavaParser crashes when processing the keycloak repo. Most of my testing was done without dependency fetching or information, since this represents the worst case for available information, but some of the crashes occurred even with dependency info. More investigation into why this happened is required.

There are a couple of changes in this PR, but the most notable changes are:
* Added a `tryWithSafeStackOverflow` method to catch `StackOverflowErrors` caused by an unterminated recursion bug in the JavaParser symbol solver. 
* Added `Try` around some expressions that threw `UnsolvedSymbolExceptions`

The biggest issues with this PR are:
* Added a `tryWithSafeStackOverflow` method to catch `StackOverflowErrors` caused by an unterminated recursion bug in the JavaParser symbol solver. <- this seems like a bad fix, but I can't think of anything better to do. There is a "TODO" comment in the javaparser source code that would break the loop when implemented, but for now I think just catching the StackOverflow and treating this as a failed resolution is the best approach. 
* There are no unit tests for any of these crashes. I looked into the first couple of crashes, but was unable to reproduce them in smaller examples. In many cases the crashes didn't occur when running the frontend on just the single affected source file, so they probably have to do with some weird and specific interaction between missing type info and different files. My testing was therefore done by hand, generating a CPG for the entire keycloak repo and seeing what breaks.


ml86 (approved):
Please spend some more time trying to fix the StackOverflowException. There is a reason this is not caught by `Try`. This is not something you want to have during any normal operation.

PR #1632, opened 2022-07-22
querydb: add new kotlin query `nopTrustManagerUsed`

ursachec:


querydb/src/main/scala/io/joern/scanners/kotlin/NetworkCommunication.scala
```diff
@@ -0,0 +1,96 @@
...
+        val nopTrustManagerFullNames =
+          cpg.typeDecl
+            .isExternal(false)
+            .filter(_.inheritsFromTypeFullName.contains("javax.net.ssl.X509TrustManager"))
+            .filter { node =>
+              node.method.name("checkClientTrusted").block.expressionDown.isEmpty ||
+              node.method.name("checkServerTrusted").block.expressionDown.isEmpty
+            }
+            .fullName
+            .l
```
maltek, line 36:
```suggestion
            .toSeq
```

querydb/src/main/scala/io/joern/scanners/kotlin/NetworkCommunication.scala
```diff
@@ -0,0 +1,96 @@
...
+      title = "NOP trust manager used to initialize SSL context",
+      description = "Traffic sent using this context can be intercepted by attackers on the same network",
+      score = 5,
+      withStrRep({ cpg =>
+        val nopTrustManagerFullNames =
+          cpg.typeDecl
+            .isExternal(false)
+            .filter(_.inheritsFromTypeFullName.contains("javax.net.ssl.X509TrustManager"))
+            .filter { node =>
+              node.method.name("checkClientTrusted").block.expressionDown.isEmpty ||
```
maltek, line 32:
```suggestion
              node.method.nameExact("checkClientTrusted").block.expressionDown.isEmpty ||
```

querydb/src/main/scala/io/joern/scanners/kotlin/NetworkCommunication.scala
```diff
@@ -0,0 +1,96 @@
...
+              node.method.name("checkServerTrusted").block.expressionDown.isEmpty
```
maltek, line 33:
```suggestion
              node.method.nameExact("checkServerTrusted").block.expressionDown.isEmpty
```

maltek (approved):
this is kind of interesting because we have a Java query for the same thing. If we get more of those we should try to see how much code can be shared between them.

ursachec:
@maltek true, i remember seeing it and then forgot about it. re/ sharing code, it seems like the logical step forward (this started to become more and more clear while writing android queries, most of which require the same manifest/build.gradle parsing logic)

PR #1633, opened 2022-07-25
Add script for publishing to `SDKMAN!`

claudiuvursache:
see: https://sdkman.io/vendors

```
$ sdk list
// output trimmed...

Joern (1.1.997)                                                https://joern.io/

Joern is a platform for analyzing source code, bytecode, and binary executables.
It generates code property graphs (CPGs), a graph representation of code for
cross-language code analysis. Code property graphs are stored in a custom graph
database. This allows code to be mined using search queries formulated in a
Scala-based domain-specific query language. Joern is developed with the goal of
providing a useful tool for vulnerability discovery and research in static
program analysis.

                                                             $ sdk install joern

```

### DIFF
--- a/sdkman_release_new_candidate_version.sh
+++ b/sdkman_release_new_candidate_version.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+# This script abstracts away API calls to sdkman's HTTP API for publishing new candidate versions.
+#
+# Example usage:
+# ./sdkman_release_new_candidate_version.sh 1.2.3
+
+set -o errexit
+set -o pipefail
+
+# Script starts here
+
+command -v curl >/dev/null 2>&1 || { echo "'curl' is required but not installed. Exiting." >&2; exit 1; }
+
+if [ "$#" -ne 1 ]; then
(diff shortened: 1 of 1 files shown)


sdkman_release_new_candidate_version.sh
```diff
@@ -0,0 +1,38 @@
...
+
+if [[ -z "${SDKMAN_CONSUMER_KEY}" ]]; then
+  echo "ENV var 'SDKMAN_CONSUMER_KEY' required but not set. Exiting."
+  exit 1
+fi
+
+if [[ -z "${SDKMAN_CONSUMER_TOKEN}" ]]; then
+  echo "ENV var 'SDKMAN_CONSUMER_TOKEN' required but not set. Exiting."
+  exit 1
+fi
```
johannescoetzee, line 28: Where would we get these values if we wanted to use the script?

johannescoetzee (approved):
Looks fine, but some documentation regarding the `SDKMAN_CONSUMER_{KEY,TOKEN}` would be nice.

sdkman_release_new_candidate_version.sh
claudiuvursache: those are secrets i've received at SDKMAN! vendor registration time. I've shared them with one or two Joern folk

PR #1634, opened 2022-07-25
Misc javasrc fixes

johannescoetzee:
# Changes:
* flatten the AST for expressions like `new int[] {1, 2, 3}`, which were previously represented as an `<operator>.alloc` call with an `<operator>.arraiInitializer` call as a child. 
* add resource initialisers to the AST directly before the `try-with-resources` block in which they are defined.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
@@ -1066,24 +1066,30 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
     astForBlockStatement(catchClause.getBody)
   }
 
-  def astForTry(stmt: TryStmt): Ast = {
+  def astsForTry(stmt: TryStmt): Seq[Ast] = {
     val tryNode = NewControlStructure()
       .controlStructureType(ControlStructureTypes.TRY)
       .code("try")
       .lineNumber(line(stmt))
       .columnNumber(column(stmt))
 
+    val resources = stmt.getResources.asScala.flatMap(astsForExpression(_, expectedType = None)).toList
     val tryAst    = astForBlockStatement(stmt.getTryBlock, codeStr = "try")
     val catchAsts = stmt.getCatchClauses.asScala.map(astForCatchClause)
-    val catchBlock = Ast(NewBlock().code("catch"))
-      .withChildren(catchAsts)
+    val catchBlock = Option
+      .when(catchAsts.nonEmpty) {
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/ControlStructureTests.scala
```diff
@@ -59,7 +138,7 @@ class NewControlStructureTests extends JavaSrcCode2CpgFixture {
       assignment.order shouldBe 2
       assignment.typeFullName shouldBe "java.lang.String[]"
 
-      val (iterIdentifier, arrayAlloc) = assignment.argument.l match {
+      val (iterIdentifier, arrayInitializer) = assignment.argument.l match {
         case List(iterIdentifier: Identifier, arrayAlloc: Call) => (iterIdentifier, arrayAlloc)
```
maltek, line 142:
```suggestion
        case List(iterIdentifier: Identifier, arrayInitializer: Call) => (iterIdentifier, arrayInitializer)
```

nit: also rename the variable inside the match

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -1785,24 +1791,25 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
...
-    val levelAsts = expr.getLevels.asScala.flatMap { lvl =>
-      lvl.getDimension.toScala match {
-        case Some(dimension) => astsForExpression(dimension, Some(ExpectedType.Int))
-
-        case None => Seq.empty
-      }
-    }
+    maybeInitializerAst match {
+      case Some(initializerAst) =>
+        initializerAst.root.collect { case call: NewCall => call.code(expr.toString) }
```
maltek, line 1798: does this line create a collection and then throw it away for side-effects, or is it a left-over from debugging?

ml86 (approved):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
maltek: oh, it sets the code of the call to `expr.toString` doesn't it? Sometimes you just have to ask a question to find the answer...

johannescoetzee: That is indeed what it does! The fact that you had to ask means I should probably clean it up a bit :)

PR #1638, opened 2022-07-26
Add ConfigFile nodes to javasrc2cpg ASTs

johannescoetzee:
The list of config file filters is based on that used by `java2cpg` in [ConfigFile.java](https://github.com/ShiftLeftSecurity/java2cpg/blob/master/java2cpg/src/main/java/io/shiftleft/java2cpg/configfiles/ConfigFile.java).

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/ConfigFileCreationPass.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/ConfigFileCreationPass.scala
@@ -10,8 +10,7 @@ import org.slf4j.LoggerFactory
 import java.nio.file.Paths
 import scala.util.{Failure, Success, Try}
 
-class ConfigFileCreationPass(projectDir: String, cpg: Cpg)
-  extends ConcurrentWriterCpgPass[File](cpg) {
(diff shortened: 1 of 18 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/ConfigFileCreationPass.scala
```diff
@@ -0,0 +1,95 @@
...
+      case Success(file) if file.isDirectory =>
+        file.listRecursively
+          .filter(isConfigFile)
+          .toArray
+
+      case Success(file) if isConfigFile(file) =>
+        Array(file)
+
+      case _ => Array.empty
+    }
```
max-leuthaeuser, line 28: We have https://github.com/joernio/joern/blob/master/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala#L16 for this.

johannescoetzee: That method would work for some of these, but `isConfigFile` is more complicated than a simple extension check. It might not be necessary to do the more complex checks, but I did it this way for the sake of consistency with java2cpg.

ml86 (approved):

PR #1639, opened 2022-07-26
kotlin2cpg: change signature of lambda invocations

claudiuvursache:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/DefaultTypeInfoProvider.scala
```diff
@@ -443,12 +443,31 @@ class DefaultTypeInfoProvider(environment: KotlinCoreEnvironment) extends TypeIn
...
+        (s"$renderedFqName:$fullNameSignature", signature)
       case None => defaultValue
     }
   }
 
+  def lambdaInvocationSignature(expr: KtLambdaExpression): String = {
+    val paramsString =
+      if (expr.getValueParameters.isEmpty) TypeConstants.javaLangObject
+      else
+        TypeConstants.javaLangObject + s"${("," + TypeConstants.javaLangObject).repeat(expr.getValueParameters.size())}"
```
johannescoetzee, line 466: Mixing string concatenation and interpolation looks a bit strange here. In my opinion, something like
`List.fill(expr.getValueParameters.size)(TypeConstants.javaLangObject).mkString(",") is cleaner, although that's very subjective.

johannescoetzee (approved):
The code looks fine in general, but I'm approving with the caveat that I'm not sure whether the new signatures are correct.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/DefaultTypeInfoProvider.scala
claudiuvursache: :+1: changed

PR #1641, opened 2022-07-27
jssrc2cpg: removed test resource files

max-leuthaeuser:
Test files and their content that is required during jssrc2cpg tests is now generated on-the-fly.
This is way easier to maintain as the single point of truth is now directly visible in the test code.
Hence, we do not have to rely on external files any longer. Also, the handling of resource paths / URIs during testing within Intellij or a SBT shell is quite error prone.


claudiuvursache (approved):
++, looks much nicer to me as well

PR #1645, opened 2022-07-28
querydb: add new kotlin query bundle

claudiuvursache:


querydb/src/main/scala/io/joern/scanners/kotlin/PathTraversals.scala
```diff
@@ -0,0 +1,124 @@
...
+        val zipEntryTypeFullNames =
+          List("org.apache.commons.compress.archivers.ArchiveEntry", "java.util.zip.ZipEntry")
+        def zipEntryParams =
+          cpg.method.parameter.typeFullNameExact(zipEntryTypeFullNames: _*)
+        def fileOutStreamInit =
+          cpg.method.fullNameExact("java.io.FileOutputStream.<init>:void(java.io.File)").callIn
+        def pathStartsWithCalls =
+          cpg.method.fullNameExact("java.nio.file.Path.startsWith:boolean(java.nio.file.Path)").callIn
+
+        def uncheckedZipEntryParameters = zipEntryParams.filterNot { param =>
```
maltek, line 33: Nit: Personally, I try to avoid double-negation like here `filterNot(...nonEmpty)`

querydb/src/main/scala/io/joern/scanners/kotlin/PathTraversals.scala
```diff
@@ -0,0 +1,124 @@
...
+          pathStartsWithCalls.argument.reachableByFlows(param).nonEmpty
```
maltek, line 34:
```suggestion
          pathStartsWithCalls.argument(0).reachableByFlows(param).nonEmpty
```

(Not tested - but I think this check should include the argument index.)

maltek (approved):
this pass should just work as-is for Java as well as Kotlin - if we had a way to share them

querydb/src/main/scala/io/joern/scanners/kotlin/PathTraversals.scala
claudiuvursache: yess, good catch, thx

querydb/src/main/scala/io/joern/scanners/kotlin/PathTraversals.scala
claudiuvursache: yepp, looks better without

PR #1647, opened 2022-07-29
Add option to run delombok on java source code before generating AST

johannescoetzee:
This PR adds a bare-bones implementation for running lombok before AST generation which is enabled with the `--run-delombok` flag. The `JAVASRC2CPG_JAVA_HOME` environment variable can be set to manually specify which Java version should be used to run delombok (since this does affect the results).


max-leuthaeuser:
I think you also need to copy the `lombok.jar` like we do the astgen binaries for jssrc2cpg (https://github.com/joernio/joern/blob/master/joern-cli/frontends/jssrc2cpg/build.sbt#L77) or otherwise is wont be available in the packaged final distribution.

maltek:
It looks like Lombok is on Maven. Is there any problem with adding it as a regular dependency that made you commit the .jar as a binary blob?

max-leuthaeuser:
Oh good question @maltek!
If its actually available there and can be used as is we should just add it as dep and use its api.

johannescoetzee:
There are 2 reasons I didn't use lombok as a regular Maven dependency. The most important is that delombok is only available as a task in a package private class, so actually executing the delombok task would require some hacky reflection stuff and even then I'm not sure if it'll work (since it uses further reflection and all of this just sounds like a headache waiting to happen). The second is that, by running the jar separately, we can decouple the java versions used to run joern and lombok.

maltek:
Could we run `s"/some/other/java -cp ${System.getProperty("java.class.path")} lombok.launch.Main ...` after adding lombok as a dependency? (I don't really know enough about our deployment or the JVM compatibility story.)

I'll stop bickering if the answer is no.

johannescoetzee:
That's a good thought. I'll give it a shot and update the PR one way or the other

ml86 (changes requested):
Maybe github does not show me that you used git-lfs to add `lombok.jar` but if it is correct in that regard remove it and use git-lfs. In general: Do not add binary to our git repos without git-lfs. After you did that please make sure that the commits containing the binary are not referenced anymore so that they will not be delivered in future git-clone to keep the download size small.

I would also favor a solution which fetches lombok like a the other dependencies. To invoke it with a specific jvm version you will likely have to have a look at the `sbt-native-packager` and as well as our distro packaging in order to find our where the dependency jars end up.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/Main.scala
```diff
@@ -32,7 +34,16 @@ private object Frontend {
         .action((path, c) => c.copy(inferenceJarPaths = c.inferenceJarPaths + path)),
       opt[Unit]("fetch-dependencies")
         .text("attempt to fetch dependencies jars for extra type information")
-        .action((_, c) => c.copy(fetchDependencies = true))
+        .action((_, c) => c.copy(fetchDependencies = true)),
+      opt[String]("feature-set-version")
```
ml86, line 38: This seems to be unrelated to `lombok` as it is only used as input to JavaParser. Please put that into a different PR.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/Delombok.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import java.nio.file.Paths
+import scala.util.{Failure, Success, Try}
+
+object Delombok {
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
+  private val javaPath: String = {
+    sys.env
+      .get("JAVASRC2CPG_JAVA_HOME")
```
ml86, line 15: Three things about this env variable:
1. It is not really descriptive
2. I suggest, that however we will name this variable, it will also be used to define the java version for the JavaParser type resolution. After all it just makes sense to have that the same as the lombok target version.
3. I suggest we make the env variable a parameter and call it "--analysis-java-home".

johannescoetzee:
> Maybe github does not show me that you used git-lfs to add `lombok.jar` but if it is correct in that regard remove it and use git-lfs. In general: Do not add binary to our git repos without git-lfs. After you did that please make sure that the commits containing the binary are not referenced anymore so that they will not be delivered in future git-clone to keep the download size small.
> 
> I would also favor a solution which fetches lombok like a the other dependencies. To invoke it with a specific jvm version you will likely have to have a look at the `sbt-native-packager` and as well as our distro packaging in order to find our where the dependency jars end up.

I did not use git-lfs, so thanks for the heads up! I'll look the regular dependency fetching we do if the direct-through-maven approach doesn't work.

mpollmeier:
General note: please avoid using git-lfs. Github imposed a 1GB rule on bandwidth (!), i.e. each PR build and anyone downloading the repo (git fetch) will contribute towards that and very quickly we've rendered the repo unusable for everyone. 

Example from joern: we added 150M files to git lfs on friday, and by sunday no one was able to fetch the repo any longer, until Fabs inserted some coins...

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/Delombok.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  def run(projectDir: String, analysisJavaHome: Option[String]): String = {
+    Try(File.newTemporaryDirectory(prefix = "delombok").deleteOnExit()) match {
+      case Success(tempDir) =>
+        ExternalCommand.run(delombokToTempDirCommand(tempDir, analysisJavaHome), cwd = projectDir) match {
+          case Success(_) =>
+            tempDir.path.toAbsolutePath.toString
+
+          case Failure(t) =>
+            logger.warn(s"Executing delombok failed with $t")
```
mpollmeier, line 51:
```suggestion
            logger.warn(s"Executing delombok failed", t)
```
otherwise you'll miss out on the stack trace

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/Delombok.scala
```diff
@@ -0,0 +1,61 @@
...
+            logger.warn("Creating AST with original source instead. Some methods and type information will be missing.")
+            projectDir
+        }
+
+      case Failure(e) =>
+        logger.warn(s"Failed to create temporary directory for delomboked source. Methods and types may be missing: $e")
```
mpollmeier, line 57:
```suggestion
        logger.warn(s"Failed to create temporary directory for delomboked source. Methods and types may be missing", e)
```

maltek (approved):

ml86 (approved):

PR #1649, opened 2022-07-29
jssrc2cpg: use git-lfs for astgen binaries

max-leuthaeuser:


mpollmeier (approved):
Unfortunately these files are already in the git history and will forever be part of the repo. Unless we're a bit nasty and change the history...

max-leuthaeuser:
I tried that using the Git lfs migration tool. Works fine but requires a force push to master which is protected.

PR #1655, opened 2022-08-02
upgrade deps (RIP property mixins)

mpollmeier:


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/accesspath/TrackedBase.scala
```diff
@@ -14,12 +14,22 @@ case class TrackedLiteral(literal: Literal) extends TrackedBase {
...
   }
 }
 
+case class TrackedMethod(method: MethodRef) extends TrackedMethodOrTypeRef {
+  override def code = method.code
+}
+case class TrackedTypeRef(typeRef: TypeRef) extends TrackedMethodOrTypeRef {
+  override def code = typeRef.code
+}
+
```
ml86, line 32: This would require downstream changes in the data flow tracker. It is not much of a deal but i would like to suggest using alternative constructors instead. So that we would have:
```
case class TrackedMethodOrTypeRef(methodOrTypeRef: StoredNode, code: String) {
  def this(methodRef: MethodRef) ...
 def this(typeRef: TypeRef)...
}

**Review by ml86 — APPROVED** [review]

File: semanticcpg/src/main/scala/io/shiftleft/semanticcpg/accesspath/TrackedBase.scala (hunk 1 above)
Comment (mpollmeier) [reply semanticcpg/src/main/scala/io/shiftleft/semanticcpg/accesspath/TrackedBase.scala:32]: mmh, afaics all affected places downstream are actually beneficial... e.g. 
```
.collect { case TrackedMethodOrTypeRef(methodRef: MethodRef) => methodRef }
->
.collect { case TrackedMethod(methodRef) => methodRef }
```

another example
```
        .collectAll[TrackedMethodOrTypeRef]
        .foreach {
          case TrackedMethodOrTypeRef(ref: MethodRef) =>
            ...
          case _ =>

->
        .collectAll[TrackedMethod].foreach { ref => 
           ...
        }
```

ml86: Hmm yeah, there are cases where the new representation is better and in others the old. So i would rather stick to what it is.

mpollmeier: resolved via zoom - we'll take the new representation

PR #1656, opened 2022-08-03
kotlin2cpg: decrease code duplication

claudiuvursache:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/KtPsiToAst.scala
```diff
@@ -944,10 +932,7 @@ trait KtPsiToAst {
       callNode(expr.getText, methodName, fullName, signature, retType, dispatchType, line(expr), column(expr)),
       argIdx
     )
-    Ast(node)
-      .withChild(receiverAst)
-      .withChildren(argAsts)
-      .withArgEdges(node, argAsts.map(_.root.get))
+    callAst(node, List(receiverAst) ++ argAsts)
```
johannescoetzee, line 935: Should the receiver be part of the arg list here (instead of being provided as a separate argument?

johannescoetzee (approved):

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/KtPsiToAst.scala
johannescoetzee: Approved if this is intentional, but please change it if not

PR #1657, opened 2022-08-03
Fix some javasrc type bugs

johannescoetzee:
This PR fixes a couple of bugs where types in method full names for unresolvable calls were `<unresolved...>` even though there was enough type information available to use the actual types.

It also replaces the `<unresolvedReceiver>` and `<unresolvedSignature>` types with `<unresolvedType>` for the sake of uniformity.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/TypeInfoCalculator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/TypeInfoCalculator.scala
@@ -242,7 +242,7 @@ object TypeInfoCalculator {
     val Class: String          = "java.lang.Class"
     val Iterator: String       = "java.util.Iterator"
     val Void: String           = "void"
-    val UnresolvedType: String = "<unresolvedType>"
+    val UnresolvedType: String = "codepropertygraph.Unresolved"
   }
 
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/TypeInfoCalculator.scala
```diff
@@ -232,26 +230,21 @@ object TypeInfoCalculator {
...
+    val Long: String           = "long"
+    val Float: String          = "float"
+    val Double: String         = "double"
+    val Char: String           = "char"
+    val Boolean: String        = "boolean"
+    val Object: String         = "java.lang.Object"
+    val Class: String          = "java.lang.Class"
+    val Iterator: String       = "java.util.Iterator"
+    val Void: String           = "void"
+    val UnresolvedType: String = "<unresolvedType>"
```
maltek, line 245: I just remembered Kotlin has a name for unresolved types - can we use the same one here please?

https://github.com/joernio/joern/blob/c1a0cf499f4a9d5f670db746db8f903f5ece8c05/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/Constants.scala#L5

maltek (approved):

PR #1663, opened 2022-08-06
Make gradle dependency mandatory

fabsx00:
Using `javasrc` in joern plugins fails when using gradle dependency resolution because the gradle tooling jars aren't present. I noticed that the dependencies were marked as optional. Now they are mandatory.


claudiuvursache (approved):

mpollmeier:
note that this means that we need to add the following to each downstream build that depends on javasrc:
```
ThisBuild / resolvers += "Gradle Releases" at "https://repo.gradle.org/gradle/libs-releases/"
```

PR #1668, opened 2022-08-09
kotlin2cpg: tweak ASTs around classes

claudiuvursache:
- various METHOD nodes of class methods were missing the VIRTUAL
  modifier
- METHOD_RETURN nodes of ctors had two incorrect props set


ml86 (approved):

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/KtPsiToAst.scala
```diff
@@ -249,8 +254,8 @@ trait KtPsiToAst {
 
     val typeFullName = typeInfoProvider.typeFullName(ktClass.getPrimaryConstructor, TypeConstants.any)
```
maltek, line 255: seems we could delete this line now

claudiuvursache: ugh, thx, missed it, i'll bring the removal in in a next pr

PR #1670, opened 2022-08-09
refactor dsl: tighten return types, use generated steps rather than graph-generic dsl, ...

mpollmeier:
Mostly refactoring, but one noteworthy change: in `TagTraversal` I dropped the `.sort` because it's a bad default: it only works if the entire traversal is in memory

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/TagTraversal.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/TagTraversal.scala
@@ -1,79 +1,39 @@
 package io.shiftleft.semanticcpg.language
 
+import io.shiftleft.codepropertygraph.generated.EdgeTypes
 import io.shiftleft.codepropertygraph.generated.nodes._
-import io.shiftleft.codepropertygraph.generated.{EdgeTypes, NodeTypes}
 import overflowdb.traversal._
 
 class TagTraversal(val traversal: Traversal[Tag]) extends AnyVal {
 
   def member: Traversal[Member] =
-    traversal
-      .in(EdgeTypes.TAGGED_BY)
-      .hasLabel(NodeTypes.MEMBER)
-      .sortBy(_.id)
-      .cast[Member]
+    traversal.in(EdgeTypes.TAGGED_BY).collectAll[Member]
 
   def method: Traversal[Method] =
-    traversal
-      .in(EdgeTypes.TAGGED_BY)
-      .hasLabel(NodeTypes.METHOD)
-      .sortBy(_.id)
-      .cast[Method]
+    traversal.in(EdgeTypes.TAGGED_BY).collectAll[Method]
 
   def methodReturn: Traversal[MethodReturn] =
-    traversal
-      .in(EdgeTypes.TAGGED_BY)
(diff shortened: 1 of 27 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/TagTraversal.scala
```diff
@@ -1,79 +1,39 @@
...
 
 class TagTraversal(val traversal: Traversal[Tag]) extends AnyVal {
 
   def member: Traversal[Member] =
-    traversal
-      .in(EdgeTypes.TAGGED_BY)
-      .hasLabel(NodeTypes.MEMBER)
-      .sortBy(_.id)
-      .cast[Member]
+    traversal.in(EdgeTypes.TAGGED_BY).collectAll[Member]
```
ml86, line 10: Not sorting anymore changes the order which may have influence on findings.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/MethodMethods.scala
```diff
@@ -60,17 +51,16 @@ class MethodMethods(val method: Method) extends AnyVal with NodeExtension with H
 
   /** The type declaration associated with this method, e.g., the class it is defined in.
     */
-  def definingTypeDecl: Traversal[TypeDecl] =
-    Traversal.fromSingle(method).definingTypeDecl
+  def definingTypeDecl: Option[TypeDecl] =
+    Traversal.fromSingle(method).definingTypeDecl.headOption
```
ml86, line 55: The steps on the traversals should be defined in terms of the steps on a single element and not the other way around. I know this was already not the case before your change but since you are already touching it.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/TagTraversal.scala
mpollmeier: I was going to add the sort at the usage side...

context (copied from the pr description):  I dropped the .sort because it's a bad default: it only works if the entire traversal is in memory.

ml86: Ok but please make sure not to miss any usage sites.

ml86 (approved):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/MethodMethods.scala
mpollmeier: I was considering that, but some steps (like this one) are defined with a repeat/until traversal, and it's probably best to keep it that way, since those are quite expressive/readable...

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/TagTraversal.scala
mpollmeier: I changed my mind :) It's used in too many places in codescience...

ml86 (approved):

Issue #1685, opened 2022-08-16
jssrc: wrong arguments for calls with field access receivers


maltek:
e.g. the call at 
https://github.com/OWASP/NodeGoat/blob/7c293e721bd1e95be6f82475d295b9b10e3b584e/app/assets/js/chart/chart-data-morris.js#L10 has no arg 0, some weird identifier as arg 1 and the actual first argument as arg 2. It has a field access as receiver, but without argument edge.

```
ocular> cpg.call.code("Morris.Area\\(.*").argument.l 
res15: List[Expression] = List(             
  Identifier(                 
    id -> 1847L,
    argumentIndex -> 1,                      
    argumentName -> None,            
    code -> "Morris",     
    columnNumber -> Some(value = 8),      
    depthFirstOrder -> 68,
    dynamicTypeHintFullName -> ArraySeq(),
    internalFlags -> 4,
    lineNumber -> Some(value = 10),
    name -> "Morris",
    order -> 1,
    typeFullName -> "ANY"
  ),     
  Block(
    id -> 1162L,
    argumentIndex -> 2,
    argumentName -> None,
    code -> """{
            // ID of the element in which to draw the chart.
            element: "morris-chart-area",
            // Chart data records -- each entry in this array corresponds to a point on the chart.
            data: [{
....
ocular> cpg.call.code("Morris.Area\\(.*").receiver.l 
res16: List[Expression] = List(
  Call(
    id -> 1160L,
    argumentIndex -> 0,
    argumentName -> None,
    code -> "Morris.Area",
    columnNumber -> Some(value = 8),
    depthFirstOrder -> 67,
    dispatchType -> "STATIC_DISPATCH",
    dynamicTypeHintFullName -> ArraySeq(),
    internalFlags -> 4,
    lineNumber -> Some(value = 10),
    methodFullName -> "<operator>.fieldAccess",
    methodInstFullName -> None,
    name -> "<operator>.fieldAccess",
    order -> 0,
    resolved -> None,
    signature -> "",
    typeFullName -> "ANY"
  )
)
```

max-leuthaeuser:
It looks weird. But the behavior should be identical to js2cpg. Could that be verified with a simpler minimal reproducing example in both frontends?

maltek:
In joern, just a file with `a.b(x);` shows the same behavior for both frontends. (But there's no `importCode.jssrc.fromString` or `importCode.javascript.fromString` so this is way harder to test than it should be :smile: )

[...]

PR #1687, opened 2022-08-17
jssrc2cpg: fix call receiver and base handling

max-leuthaeuser:
This is for: https://github.com/joernio/joern/issues/1685

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/SimpleAstCreationPassTest.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/SimpleAstCreationPassTest.scala
@@ -480,11 +480,11 @@ class SimpleAstCreationPassTest extends AbstractPassTest {
 
       // barBaseTree constructs is tested for in another test.
 
-      val List(thisArg) = barCall.astChildren.isIdentifier.argumentIndex(1).l
+      val List(thisArg) = barCall.astChildren.isIdentifier.argumentIndex(0).l
       thisArg.name shouldBe "_tmp_0"
       thisArg.order shouldBe 1
 
-      val List(zArg) = barCall.astChildren.isIdentifier.argumentIndex(2).l
+      val List(zArg) = barCall.astChildren.isIdentifier.argumentIndex(1).l
       zArg.name shouldBe "z"
       zArg.order shouldBe 2
     }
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/SimpleAstCreationPassTest.scala
```diff
@@ -515,6 +513,44 @@ class SimpleAstCreationPassTest extends AbstractPassTest {
...
+
+      val List(receiver) = call.receiver.isCall.l
+      receiver.code shouldBe "a.b"
+      receiver.methodFullName shouldBe Operators.fieldAccess
+
+      val List(receiverAsArg) = call.argument(0).cast[Call].l
+      receiver shouldBe receiverAsArg
+
+      val List(base) = receiver.astChildren.isIdentifier.argumentIndex(1).l
+      base.name shouldBe "a"
```
maltek, line 537: I don't think that's we want yet.

We want

- either the `a` identifier as arg 0 - apparently that's how it's always been with js2cpg
- or the `a.b` field access as arg 0 - that matches the original code more precisely, but I don't know if there was a reason why we've gone with the identifier previously
- but neither can be arg 1. That must be reserved for the "x" identifier that's actually in position 1 in the original code.

maltek (changes requested):

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/SimpleAstCreationPassTest.scala
max-leuthaeuser: Look closely. Thats the base of the receiver (a field access call here).

maltek: oh, I see :facepalm:

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/SimpleAstCreationPassTest.scala
```diff
@@ -515,6 +513,44 @@ class SimpleAstCreationPassTest extends AbstractPassTest {
...
+      val List(accessElement) = receiver.astChildren.isFieldIdentifier.argumentIndex(2).l
+      accessElement.canonicalName shouldBe "b"
+
+      val List(recvArg, baseArg, arg) = call.argument.l.sortBy(_.order)
+      recvArg shouldBe receiver
+      recvArg.argumentIndex shouldBe 0
+      recvArg.order shouldBe 0
+      baseArg.code shouldBe "a"
+      baseArg.argumentIndex shouldBe 0
+      baseArg.order shouldBe 1
```
maltek, line 548: When tracking flows into parameter 0 of the callee of such a call, I think we'd get duplicate dataflows here - one through `recvArg` and another through `baseArg`.

My vote would be to just get rid of the `baseArg` - but I don't know the reason why we generated the `baseArg` in js2cpg to begin with.

maltek (approved):

PR #1689, opened 2022-08-17
jssrc2cpg: fix declaration handling

max-leuthaeuser:
We had dangling sub ASTs for multiple declarators in the same declaration.
This fixes crashes with the closed-source DF engine.

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
@@ -279,8 +279,16 @@ trait AstForDeclarationsCreator { this: AstCreator =>
     } else {
       MethodScope
     }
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
```diff
@@ -279,8 +279,16 @@ trait AstForDeclarationsCreator { this: AstCreator =>
     } else {
       MethodScope
     }
-    declaration.json("declarations").arr.foldLeft(Ast()) { (ast, d) =>
-      ast.merge(astForVariableDeclarator(d, scopeType))
+    val declAsts = declaration.json("declarations").arr.toList.map(astForVariableDeclarator(_, scopeType))
+    declAsts match {
+      case ::(head, tail) =>
```
maltek, line 284: why not write this infix?
```suggestion
      case head :: tail =>
```

maltek (approved):

Issue #1688, opened 2022-08-17
jssrc: add some flag to identify constructor calls


maltek:
a constructor call like `new String(1)` gets turned into a block like:

```
_tmp_0 = .alloc
String(this=_tmp_0, 1)
_tmp_0
```

In the CPG we've lost the information that `String` is used as a constructor.

Therefore, we have no automatic way to deduce the type of `_tmp_0` even when we figure out the callee of `String(1)`. And I have to write a policy for every single constructor stating the type of its parameter 0.

I wish for some flag on the `String(1)` call or its `_tmp_0` argument that allows our dynamic typing engine to figure that out by itself.

PR #1690, opened 2022-08-18
jssrc2cpg: use "<operator>.new" as NAME for constructor calls

max-leuthaeuser:
This is for: https://github.com/joernio/joern/issues/1688

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForExpressionsCreator.scala
@@ -143,7 +143,7 @@ trait AstForExpressionsCreator { this: AstCreator =>
 
     val receiverNode = astForNodeWithFunctionReference(callee)
     Ast.storeInDiffGraph(receiverNode, diffGraph)
-    
+
     // TODO: place "<operator>.new" into the schema
     val callNode = handleCallNodeArgs(newExpr, receiverNode.nodes.head, tmpAllocNode2, callName = "<operator>.new")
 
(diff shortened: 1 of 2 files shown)


maltek (commented):
I fear this doesn't solve my problem quite yet. Our dynamic type engine uses carrot paths - and it won't do anything useful with the receiver code in the dynamic type field.

Of course, jssrc has no concept of carrot paths and no way to set them to correctly! That's why I asked for some flag - if the type engine has a way of telling whether a call is being used as a constructor, then it should be straightforward to teach it to do the right thing.

max-leuthaeuser:
Well, if we do not want to extend the schema for `CALL` we could simply set a special `NAME` (e.g., `<operator>.new`) for these constructor calls.

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -110,11 +123,25 @@ trait AstForExpressionsCreator { this: AstCreator =>
...
+    receiverNode.nodes.head match {
+      case i: NewIdentifier =>
+        i.typeFullName = typeHint
+        i.dynamicTypeHintFullName = ArraySeq(typeHint)
+      case _ =>
+    }
+
     val tmpAllocName      = generateUnusedVariableName(usedVariableNames, "_tmp")
-    val localTmpAllocNode = createLocalNode(tmpAllocName, Defines.ANY.label)
+    val localTmpAllocNode = createLocalNode(tmpAllocName, typeHint).dynamicTypeHintFullName(ArraySeq(typeHint))
```
maltek, line 140: I think we can drop this part. With the `<operator>.new` in place there's no longer a reason do this in the frontend. It would be easy to do the exact same thing in a pass - and possible to do arbitrarily more complex things.

maltek (approved):
yes, I think that should work :+1:

PR #1700, opened 2022-08-23
[ml4joern] initial version of `joern-vectors`

fabsx00:
This is the initial version of the command line tool `joern-embed`, which allows embedding code into a vector space. For now, it implements a very simplistic bag of words embedding over AST nodes. It employs feature hashing, making it possible to generate features one by one and streaming them to disk. For now, it only outputs json to stdout.

### DIFF
--- a/joern-cli/src/main/scala/io/joern/joerncli/CpgBasedTool.scala
+++ b/joern-cli/src/main/scala/io/joern/joerncli/CpgBasedTool.scala
@@ -52,4 +52,16 @@ object CpgBasedTool {
     }
   }
 
+  def exitIfInvalid(outDir: String, cpgFileName: String): Unit = {
+    if (File(outDir).exists)
+      exitWithError(s"Output directory `$outDir` already exists.")
+    if (File(cpgFileName).notExists)
+      exitWithError(s"CPG at $cpgFileName does not exist.")
+  }
+
+  def exitWithError(msg: String): Unit = {
+    System.err.println(s"error: $msg")
+    System.exit(1)
+  }
+
 }
--- a/joern-cli/src/main/scala/io/joern/joerncli/JoernEmbed.scala
+++ b/joern-cli/src/main/scala/io/joern/joerncli/JoernEmbed.scala
@@ -1,0 +1,24 @@
+package io.joern.joerncli
+
+object JoernEmbed extends App {
+
+  case class Config(cpgFileName: String = "cpg.bin", outDir: String = "out")
+
+  private def parseConfig: Option[Config] =
+    new scopt.OptionParser[Config]("joern-export") {
+      head("Dump intermediate graph representations (or entire graph) of code in a given export format")
+      help("help")
+      arg[String]("cpg")
+        .text("input CPG file name - defaults to `cpg.bin`")
+        .optional()
+        .action((x, c) => c.copy(cpgFileName = x))
+      opt[String]('o', "out")
+        .text("output directory - will be created and must not yet exist")
+        .action((x, c) => c.copy(outDir = x))
+    }.parse(args, Config())
+
+  parseConfig.foreach { config =>
+    println(config)
+  }
+
+}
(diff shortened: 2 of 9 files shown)


joern-cli/src/main/scala/io/joern/joerncli/CpgBasedTool.scala
```diff
@@ -52,4 +52,16 @@ object CpgBasedTool {
     }
   }
 
+  def exitIfInvalid(outDir: String, cpgFileName: String): Unit = {
+    if (File(outDir).exists)
+      exitWithError(s"Output directory `$outDir` already exists.")
```
mpollmeier, line 57: in most other situations we delete the existing output...

mpollmeier: (personally I prefer if tools just override so I can invoke them multiple times without having to worry about deleting files etc... - ymmv)

joern-cli/src/main/scala/io/joern/joerncli/JoernEmbed.scala
```diff
@@ -0,0 +1,144 @@
...
+
+  case class Embedding(data: (() => Traversal[(T, SparseVectorWithExplicitFeature)])) {
+    lazy val dimToStructure: Map[Int, String] = {
+      val m = mutable.HashMap[Int, String]()
+      data().foreach { case (_, vector) =>
+        vector.foreach { case (hash, (_, structureAsString)) =>
+          m.put(hash, structureAsString)
+        }
+      }
+      m.toMap
```
mpollmeier, line 44:
```suggestion
      val tuples = for {
        (_, vector) <- data()
        (hash, (_, structureAsString)) <- vector
      } yield hash -> structureAsString
      
      tuples.toMap
```

joern-cli/src/main/scala/io/joern/joerncli/JoernEmbed.scala
```diff
@@ -0,0 +1,144 @@
...
+  /** Extract a sequence of (object, vector) pairs from a cpg.
+    */
+  def embed(cpg: Cpg): Embedding = {
+    Embedding({ () =>
+      extractObjects(cpg)
+        .map { obj => (obj, enumerateSubStructures(obj)) }
+        .map { case (obj, substructures) =>
+          obj -> vectorize(substructures)
+        }
+    })
```
mpollmeier, line 66:
```suggestion
    Embedding(() =>
      extractObjects(cpg).map { obj =>
          val substructures = enumerateSubStructures(obj)
          obj -> vectorize(substructures)
        }
    )
```

joern-cli/src/main/scala/io/joern/joerncli/JoernEmbed.scala
```diff
@@ -0,0 +1,144 @@
...
+      opt[Unit]("features")
+        .text("Provide map from dimensions to features")
+        .action((_, c) => c.copy(dimToFeature = true))
+    }.parse(args, Config())
+
+  parseConfig.foreach { config =>
+    exitIfInvalid(config.outDir, config.cpgFileName)
+    Using.resource(CpgBasedTool.loadFromOdb(config.cpgFileName)) { cpg =>
+      val embedding = new BagOfAPISymbolsForMethods().embed(cpg)
+      println("{")
```
mpollmeier, line 123: my favorite json lib :)

mpollmeier (approved):

mpollmeier:
joern-embed sounds a little generic... how about joern-vector-embed, or joern2vector?

joern-cli/src/main/scala/io/joern/joerncli/JoernEmbed.scala
fabsx00: In this particular use-case, the tuples are `(dimension, structure)` pairs, and we know that for a given dimension, it is always associated with the same structure and vice versa. The reason I used this mutable map here is that I want to ensure that we don't keep all of these duplicates in memory. Otherwise, I would have preferred your version.

fabsx00:
`joern-vectors`?

joern-cli/src/main/scala/io/joern/joerncli/JoernEmbed.scala
mpollmeier: 👍🏻

mpollmeier:
`joern-vectors` sounds good and is less generic 👍🏻

itsacoderepo (approved):

PR #1711, opened 2022-08-26
Add UnresolvedNamespace and UnresolvedSignature defines.

ml86:
They are supposed to be used by statically typed frontend with
package/namespace structure to indicate unresolved static type
information.
For more details see code comment.

@fabs I did not find an already existing suitable place for those
defines. Please suggest some alternative if you have a better idea
on naming/location.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/Defines.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/Defines.scala
@@ -1,0 +1,19 @@
+package io.joern.x2cpg
+
+object Defines {
+  // The following two defines should be used for type and method full names to
+  // indicate unresolved static type information. Using them enables
+  // the closed source backend to apply policies in a less strict fashion.
+  // The most notable case is the METHOD_FULL_NAME property of a CALL node.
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/Defines.scala
```diff
@@ -0,0 +1,19 @@
+package io.joern.x2cpg
+
+object Defines {
+  // The following two defined should be used for type and method full names to
```
fabsx00, line 4: "The following two defines"

fabsx00:
I believe we already have other defines somewhere, but it may be in `semanticcpg`. Don't know from the top of my head. If this can wait until Monday, I'll try to find out.

johannescoetzee:
This is more of a question about how these will be used than it is a comment on the changes themselves: With javasrc2cpg, we currently take a reasonably good guess at what argument types should be, but can sometimes (often?) end up with a subclass of the true parameter type. With these new `<undefidned*>` types, would it be better to not try to guess at all?

ml86:
@johannescoetzee That depends on the quality and details of your guess.

johannescoetzee (approved):

PR #1716, opened 2022-08-30
Run delombok by default when a lombok dependency is detected

johannescoetzee:
With these changes, `javasrc2cpg` will automatically run delombok and scan the delomboked code when a lombok dependency is found with `--fetch-dependencies`. This isn't a particularly clever default and can't be switched off as is (another `--no-delombok` option could be added, but at that point we'd pretty much need to add a decision tree picture to the documentation to make sense of delombok logic). I'm not convinced this is something we actually want to do, but is a way of getting better WebGoat results by default (at least in sptests which currently fetch dependencies by default).

Update: This also fixes filenames in the CPG when delombok is used, so the filenames will now match the original source file and not the temporary delomboked files.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
@@ -8,8 +8,10 @@ import io.shiftleft.codepropertygraph.generated.Languages
 import io.joern.x2cpg.passes.frontend.{MetaDataPass, TypeNodePass}
 import io.joern.x2cpg.{SourceFiles, X2CpgFrontend}
 import io.joern.x2cpg.X2Cpg.withNewEmptyCpg
+import io.joern.x2cpg.utils.dependency.DependencyResolver
 import org.slf4j.LoggerFactory
 
+import java.nio.file.Paths
 import scala.jdk.CollectionConverters.EnumerationHasAsScala
 import scala.util.Try
 
@@ -29,14 +31,17 @@ class JavaSrc2Cpg extends X2CpgFrontend[Config] {
   def createCpg(config: Config): Try[Cpg] = {
     withNewEmptyCpg(config.outputPath, config: Config) { (cpg, config) =>
       new MetaDataPass(cpg, language, config.inputPath).createAndApply()
-      val sourcesInfo = getSourcesFromDir(config)
+      val dependencies        = getDependencyList(config)
+      val hasLombokDependency = dependencies.exists(_.contains("lombok"))
+      val sourcesInfo         = getSourcesFromDir(config, hasLombokDependency)
       if (sourcesInfo.sourceFiles.isEmpty) {
         logger.error(s"no source files found at path ${config.inputPath}")
       } else {
         logger.info(s"found ${sourcesInfo.sourceFiles.size} source files")
       }
 
-      val astCreator = new AstCreationPass(sourcesInfo.typeSolverSourceDir, sourcesInfo.sourceFiles, config, cpg)
+      val astCreator =
+        new AstCreationPass(sourcesInfo.typeSolverSourceDir, sourcesInfo.sourceFiles, config, cpg, dependencies)
       astCreator.createAndApply()
       new ConfigFileCreationPass(config.inputPath, cpg).createAndApply()
       new TypeNodePass(astCreator.global.usedTypes.keys().asScala.toList, cpg)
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
```diff
@@ -29,25 +33,66 @@ class JavaSrc2Cpg extends X2CpgFrontend[Config] {
...
+  }
+
   /** JavaParser requires that the input path is a directory and not a single source file. This is inconvenient for
     * small-scale testing, so if a single source file is created, copy it to a temp directory.
     */
-  private def getSourcesFromDir(config: Config): SourceDirectoryInfo = {
+  private def getSourcesFromDir(config: Config, hasLombokDependency: Boolean): SourceDirectoryInfo = {
+    val delombokMode = getDelombokMode(config)
+    if (hasLombokDependency) {
+      logger.info(s"Analysing delomboked code as lombok dependency was found.")
```
maltek, line 88: probably should move that into the match below, otherwise it might seem we're ignoring the command-line choice

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
```diff
@@ -59,25 +104,35 @@ class JavaSrc2Cpg extends X2CpgFrontend[Config] {
...
-      if (config.delombokFullAnalysis)
+      if (runDelombok && (delombokMode != DelombokMode.TypesOnly))
         delombokSourcesDir.get
       else
         originalSourcesDir
 
     val typeSourcesPath =
-      if (config.delombokFullAnalysis || config.delombokTypesOnly)
+      if (runDelombok)
         delombokSourcesDir.get
```
maltek, line 119: we could collapse this whole if into `delombokSourcesDir.getOrElse(originalSourcesDir)`

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/Main.scala
```diff
@@ -40,12 +39,13 @@ private object Frontend {
...
         .action((path, c) => c.copy(delombokJavaHome = Some(path))),
-      opt[Unit]("delombok-full-analysis")
-        .text("run delombok on source before scanning for more accurate methods and type results")
-        .action((_, c) => c.copy(delombokFullAnalysis = true)),
-      opt[Unit]("run-delombok-types-only")
-        .text("run delombok on source but use output only for type information")
-        .action((_, c) => c.copy(delombokTypesOnly = true))
+      opt[String]("delombok-mode")
+        .text("""Specifies how delombok should be executed. Options are
+				 | no-delombok => to not run delombok under any circumstances.
```
maltek, line 44:
```suggestion
				| no-delombok => to not run delombok under any circumstances.
```

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
```diff
@@ -29,25 +33,66 @@ class JavaSrc2Cpg extends X2CpgFrontend[Config] {
...
+      val sourcesInfo         = getSourcesFromDir(config, hasLombokDependency)
       if (sourcesInfo.sourceFiles.isEmpty) {
         logger.error(s"no source files found at path ${config.inputPath}")
       } else {
         logger.info(s"found ${sourcesInfo.sourceFiles.size} source files")
       }
 
-      val astCreator = new AstCreationPass(sourcesInfo.typeSolverSourceDir, sourcesInfo.sourceFiles, config, cpg)
+      val astCreator =
+        new AstCreationPass(sourcesInfo, config, cpg, dependencies)
```
maltek, line 46:
```suggestion
      val astCreator = new AstCreationPass(sourcesInfo, config, cpg, dependencies)
```

maltek (approved):
just some small nits

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
johannescoetzee: Good catch!

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
johannescoetzee: This is true (and avoids the `get` call, which is better)

PR #1718, opened 2022-08-31
Solidity2cpg Cleanup #1

johannescoetzee:
The first in a series of cleanup PRs. Any TODOs can be ignored since I just added those as notes to myself for later, more ambitious changes.

Note: Checks on these PRs will fail until I've updated the setup scripts to install surya, so ignore any red crosses for now. I'll verify that unit tests pass locally before opening any of these PRs.

### DIFF
--- a/joern-cli/frontends/solidity2cpg/src/main/scala/io/joern/solidity2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/solidity2cpg/src/main/scala/io/joern/solidity2cpg/passes/AstCreator.scala
@@ -3,11 +3,12 @@ import io.joern.x2cpg.Ast.storeInDiffGraph
 import io.joern.solidity2cpg.domain.SuryaObject._
 import io.joern.x2cpg.{Ast, AstCreatorBase}
 import io.joern.x2cpg.datastructures.Global
-import io.joern.x2cpg.utils.NodeBuilders.modifierNode
+import io.joern.x2cpg.utils.NodeBuilders.{identifierNode, modifierNode, operatorCallNode}
+import io.shiftleft.codepropertygraph.generated.nodes.Identifier.{PropertyDefaults => IdentifierDefaults}
+import io.shiftleft.codepropertygraph.generated.nodes.MethodParameterIn.{PropertyDefaults => ParameterDefaults}
 import io.shiftleft.codepropertygraph.generated.{
   ControlStructureTypes,
   DispatchTypes,
-  EdgeTypes,
   EvaluationStrategies,
   ModifierTypes,
   NodeTypes,
@@ -15,39 +16,26 @@ import io.shiftleft.codepropertygraph.generated.{
   PropertyNames
 }
 import io.shiftleft.codepropertygraph.generated.nodes.{
-  NewAnnotation,
-  NewAnnotationLiteral,
-  NewAnnotationParameter,
-  NewAnnotationParameterAssign,
-  NewArrayInitializer,
-  NewBinding,
   NewBlock,
   NewCall,
-  NewClosureBinding,
   NewControlStructure,
   NewFieldIdentifier,
-  NewFile,
   NewIdentifier,
-  NewJumpTarget,
   NewLiteral,
   NewLocal,
   NewMember,
   NewMethod,
   NewMethodParameterIn,
-  NewMethodRef,
   NewMethodReturn,
   NewModifier,
   NewNamespaceBlock,
-  NewNode,
   NewReturn,
   NewTypeDecl,
-  NewTypeRef,
-  NewUnknown
+  NewTypeRef
 }
 import org.slf4j.LoggerFactory
 import overflowdb.BatchedUpdate.DiffGraphBuilder
 
-import java.util
 import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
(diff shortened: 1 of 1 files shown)


MatthewEthanTam:
> The first in a series of cleanup PRs. Any TODOs can be ignored since I just added those as notes to myself for later, more ambitious changes.
> 
> Note: Checks on these PRs will fail until I've updated the setup scripts to install surya, so ignore any red crosses for now. I'll verify that unit tests pass locally before opening any of these PRs.

@johannescoetzee Okay sounds good and to install Surya use `npm install -g @matthewethantam/surya`, this is the version used and has a bit more addition to Surya

joern-cli/frontends/solidity2cpg/src/main/scala/io/joern/solidity2cpg/passes/AstCreator.scala
```diff
@@ -341,52 +327,39 @@ class AstCreator(filename: String, sourceUnit: SourceUnit, global: Global) exten
...
-    varDecl.visibility match {
-      case x: String => visibility = " " + x
-      case _         => visibility = ""
-    }
+    var storage = ""
 
-    varDecl.storageLocation match {
-      case x: String => storage = " " + x
-      case _         => storage = ""
+    // TODO: I assume the original match exists because of a possible null value. Fix that during parsing.
```
DavidBakerEffendi, line 340: A note on this, if you have a deeper look into the JSON parser where these case classes are being constructed you could potentially re-define how some of these missing values are handled there already.

joern-cli/frontends/solidity2cpg/src/main/scala/io/joern/solidity2cpg/passes/AstCreator.scala
```diff
@@ -648,80 +589,93 @@ class AstCreator(filename: String, sourceUnit: SourceUnit, global: Global) exten
...
+    val memberAsts =
       stateVariableDeclaration.variables.collect { case x: VariableDeclaration => astForMemberVarDecl(x, order) }
 
-    Ast().withChildren(fieldType)
+    // TODO Return a sequence of Asts here.
+    Ast().withChildren(memberAsts)
   }
 
   private def getMappingKeyAndValue(mapping: Mapping): String = {
+    // TODO Are these matches comprehensive?
```
DavidBakerEffendi, line 600: We could consider constructing a new trait from what we expect to come from `Mapping` examples and have a better bound of expected types.

DavidBakerEffendi: This is again something to add to the case class construction side

joern-cli/frontends/solidity2cpg/src/main/scala/io/joern/solidity2cpg/passes/AstCreator.scala
```diff
@@ -910,29 +859,30 @@ class AstCreator(filename: String, sourceUnit: SourceUnit, global: Global) exten
     }
     else
       operation.operator match {
+        // TODO Can `NOT` really be postfix?
```
DavidBakerEffendi, line 862: From [TutorialsPoint](https://www.tutorialspoint.com/solidity/solidity_operators.htm) it doesn't seem like it lol. @MatthewEthanTam can you confirm this? e.g. is `x!` a valid expression versus prefix `!x`.

DavidBakerEffendi (commented):
Some general feedback but otherwise cleanup is looking good. One thing to note is that I don't believe we really got method call full names resolved? I think @MatthewEthanTam eventually found a way to dump Surya's call resolution or add it somehow into the parsing but he can elaborate if he got this working.

DavidBakerEffendi:
I'll leave the Accepted Review to @MatthewEthanTam

MatthewEthanTam:
> Some general feedback but otherwise cleanup is looking good. One thing to note is that I don't believe we really got method call full names resolved? I think @MatthewEthanTam eventually found a way to dump Surya's call resolution or add it somehow into the parsing but he can elaborate if he got this working.

Yes, so I added method call full names as an argument in the Surya and we then parse it from there so those should be working !

MatthewEthanTam (approved):
Looks good!

joern-cli/frontends/solidity2cpg/src/main/scala/io/joern/solidity2cpg/passes/AstCreator.scala
```diff
@@ -910,29 +859,30 @@ class AstCreator(filename: String, sourceUnit: SourceUnit, global: Global) exten
...
+    val callNode = operatorCallNode(operatorName, code = code, line = lineNumber, column = columnNumber)
       .order(order)
       .argumentIndex(order)
-      .columnNumber(operation.columnNumber.get)
-      .lineNumber(operation.lineNumber.get)
 
     Ast(callNode).withChild(subExpression)
   }
   private def astForTypeNameExpression(expression: TypeNameExpression, order: Int): Ast = {
+    // TODO If this really should be empty, remove the method.
```
MatthewEthanTam, line 885: I don't think it should be empty, I didn't get to creating it because in the contracts views it wasn't used yet :)

joern-cli/frontends/solidity2cpg/src/main/scala/io/joern/solidity2cpg/passes/AstCreator.scala
```diff
@@ -1179,6 +1075,8 @@ class AstCreator(filename: String, sourceUnit: SourceUnit, global: Global) exten
   }
 
   private def astForMemberAccess(memberAccess: MemberAccess, order: Int): Ast = {
+    // TODO None of the open source frontends currently use <operator>.memberAccess, so this
+    //  should probably just be a fieldAccess instead.
```
MatthewEthanTam, line 1079: Yes, I agree. Should change the method name

Issue #1717, opened 2022-08-30
C declaration type parsing error


jryerb:
The code field for locals does not correctly show all type information when declared in a comma separated list.
Example code:
```
int main()
{
    char* buf1, buf2;
    int count = 20;
    int a = strncpy(buf1, buf2, count);

    return 0;
}
```
Output:
```
joern> importCode(".../dec_test.c") 
res28: Cpg = Cpg (Graph [26 nodes])
joern> cpg.local.code.l 
res35: List[String] = List("char* buf1", "char buf2", "int count", "int a")
```
As can be seen above, the code field for `buf1` includes the `char *` but the for `buf2` only `char`. Since pointer information is not included as part of the local/identifier type by design, this error prevents correct identification of pointers.

PR #1719, opened 2022-09-01
c2cpg: fixed locals code for multiple decls at one line

max-leuthaeuser:
Fixes: https://github.com/joernio/joern/issues/1717


maltek:
I think this is wrong, and the old behavior was correct. (That's why the usual recommendation is to put the `*` on the variable name, not the type name. Personally, I just don't declare multiple pointers/arrays in the same line at all.)

```c
#include <stdio.h>

void main() {
    char *x, y;
    printf("%d\n", sizeof(x));
    printf("%d\n", sizeof(y));
}
```

```
$ gcc /tmp/a.c
$ ./a.out 
8
1
```

max-leuthaeuser:
Oh, you are right!

PR #1721, opened 2022-09-02
Ignore php setup and test for now

johannescoetzee:
This is causing integration tests to fail as php isn't available on the test runner. Ignoring these is fine at the moment since the frontend is very far from usable anyways.


maltek:
not needed after all

PR #1722, opened 2022-09-02
joern-vectors: fixed json output

fabsx00:


joern-cli/src/main/scala/io/joern/joerncli/JoernVectors.scala
```diff
@@ -121,14 +121,14 @@ object JoernVectors extends App {
...
+      println("\"objects\":")
       traversalToJson(embedding.objects)
       if (config.dimToFeature) {
-        println("dimToFeature: ")
+        println(",\"dimToFeature\": ")
         println(Serialization.write(embedding.dimToStructure))
         println(",")
       }
-      println("vectors:")
+      println(",\"vectors\":")
```
max-leuthaeuser, line 131: With config.dimToFeature we would end up with two consecutive ','.

fabsx00: thanks!

max-leuthaeuser (approved):

PR #1726, opened 2022-09-05
Fix `astgen` invocation

fabsx00:
When `jssrc2cpg` is used as a library in an extension, then the current code for resolving the location of `astgen` fails because it is looking for a directory called `jssrc2cpg` in the staging directory. When searching the path backwards for `lib`, it works in the normal setting and when used in an extension.


fabsx00:
... causes problems with the unit testing setup. Looking.

max-leuthaeuser (approved):

PR #1734, opened 2022-09-08
Add a hacky call methodFullName resolution pass

johannescoetzee:
There are situations where we can currently infer the typeFullName for method return and parameter types by using import information, but since JavaParser cannot resolve these methods, this type information is missing when these methods are called from classes in other files. This PR adds a very rough implementation to find the implementation for internal methods that are called from different files. It doesn't handle overloaded methods or inheritance well, but should give better results than not having it.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInferencePass.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInferencePass.scala
@@ -1,0 +1,47 @@
+package io.joern.javasrc2cpg.passes
+
+import io.joern.javasrc2cpg.util.TypeInfoCalculator.TypeConstants
+import io.shiftleft.codepropertygraph.Cpg
+import io.shiftleft.codepropertygraph.generated.nodes.Call.PropertyNames
+import io.shiftleft.codepropertygraph.generated.nodes.Call
+import io.shiftleft.passes.ConcurrentWriterCpgPass
+import io.shiftleft.semanticcpg.language._
+import org.slf4j.LoggerFactory
+
+class TypeInferencePass(cpg: Cpg) extends ConcurrentWriterCpgPass[Call](cpg) {
+
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
+  override def generateParts(): Array[Call] = {
+    cpg.call.methodFullName(s".*${TypeConstants.UnresolvedType}.*").toArray
+  }
+
+  override def runOnPart(diffGraph: DiffGraphBuilder, call: Call): Unit = {
+    val candidateMethods = cpg.method.internal.nameExact(call.name).l
+
+    candidateMethods match {
+      case Nil           => // Call is to an unresolved external method, so we can't get any more information here.
+      case method :: Nil =>
+        // This is probably the correct method if the number of arguments is correct.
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInferencePass.scala
```diff
@@ -0,0 +1,47 @@
...
+class TypeInferencePass(cpg: Cpg) extends ConcurrentWriterCpgPass[Call](cpg) {
+
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
+  override def generateParts(): Array[Call] = {
+    cpg.call.methodFullName(s".*${TypeConstants.UnresolvedType}.*").toArray
+  }
+
+  override def runOnPart(diffGraph: DiffGraphBuilder, call: Call): Unit = {
+    val candidateMethods = cpg.method.internal.nameExact(call.name).l
```
maltek, line 20: Please use `.nextOption` and `.hasNext` instead of creating a potentially large linked list. This will probably entail moving the parameter count check up here and losing one of the log messages, but that seems OK to me.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInferencePass.scala
```diff
@@ -0,0 +1,47 @@
...
+
+    candidateMethods match {
+      case Nil           => // Call is to an unresolved external method, so we can't get any more information here.
+      case method :: Nil =>
+        // This is probably the correct method if the number of arguments is correct.
+        if (method.parameter.size == call.argument.size) {
+          diffGraph.setNodeProperty(call, PropertyNames.MethodFullName, method.fullName)
+          diffGraph.setNodeProperty(call, PropertyNames.Signature, method.signature)
```
maltek, line 28: What are the values of `dispatchType` and `typeFullName` for these calls? Do we want to also override them?

maltek (approved):
I can't review the design - I don't know the full story how this fits in with e.g. the policy fullname guessing (or how that works).

So just some minor suggestions/questions from me for the implementation.

johannescoetzee:
> I can't review the design - I don't know the full story how this fits in with e.g. the policy fullname guessing (or how that works).
> 
> So just some minor suggestions/questions from me for the implementation.

The changes for the policy fullname guessing still need to be done, but since these changes will usually give the correct method signature, I think the names here are still preferable to the new unresolved format in cases where we can find them.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInferencePass.scala
johannescoetzee: Noted. I'll refactor this to avoid the explicit list conversion.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInferencePass.scala
johannescoetzee: Good question! `typeFullName` should also be overridden, but I'm not sure about `dispatchType`. I'll look into that.

PR #1753, opened 2022-09-15
javasrc2cpg: rewrite unresolved methodFullNames and signatures in TypeInferencePass

johannescoetzee:


johannescoetzee:
Closing this PR since I don't get any WebGoat findings while testing locally. Clearly something here is broken.

PR #1774, opened 2022-09-21
Moved the JS specific call linker to x2cpg

max-leuthaeuser:
Can now be re-used in js2cpg and CS (will do separate PRs).

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/JavascriptCallLinker.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/JavascriptCallLinker.scala
@@ -1,17 +1,22 @@
-package io.joern.jssrc2cpg.passes
+package io.joern.x2cpg.passes.frontend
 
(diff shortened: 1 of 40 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/JavascriptCallLinker.scala
```diff
@@ -1,17 +1,22 @@
...
+import overflowdb.traversal.jIteratortoTraversal
+import overflowdb.traversal.toNodeTraversal
 
 import scala.collection.mutable
 
 /** The Javascript specific call linker links static call sites (by full name) and call sites to methods in the same
   * file (by name).
   */
-class CallLinkerPass(cpg: Cpg) extends SimpleCpgPass(cpg) {
+class JavascriptCallLinker(cpg: Cpg) extends SimpleCpgPass(cpg) {
```
maltek, line 19: the copy in CS has changed lines 46-47 to `.flatMap(_._methodRefViaRefIn)` to avoid the `.in()` call

maltek (approved):

PR #1777, opened 2022-09-22
Add PHP if and while statements

johannescoetzee:


joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
```diff
@@ -0,0 +1,436 @@
...
+  "if statements" should {
+    "work without a body, an elseif or an else" in {
+      val cpg = code("<?php\nif ($a) {}")
+
+      val ifAst = cpg.controlStructure.l match {
+        case List(ast) =>
+          ast.controlStructureType shouldBe ControlStructureTypes.IF
+          ast
+
+        case result => fail(s"Expected single control structure but found $result")
```
max-leuthaeuser, line 20: You could also use Scalatests `Inside` trait and `inside` function to save you from writing the fail case manually.

max-leuthaeuser (approved):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
johannescoetzee: Ah, that's very useful! Thanks!!

PR #1780, opened 2022-09-26
jssrc2cpg: lowered import declarations

max-leuthaeuser:
Import declarations are now additionally lowered to good-old ES6 style require assignments. With this the CS require preprocessor does not need modifications.

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
@@ -326,16 +326,22 @@ trait AstForDeclarationsCreator { this: AstCreator =>
     }
     diffGraph.addNode(createDependencyNode(name, referenceName, IMPORT_KEYWORD))
     createImportNodeAndAttachToAst(impDecl, referenceName, name)
-    astForRequireCallFromImport(name, referenceName, impDecl)
+    astForRequireCallFromImport(name, None, referenceName, impDecl)
   }
 
-  private def astForRequireCallFromImport(name: String, from: String, nodeInfo: BabelNodeInfo): Ast = {
-    val id        = createIdentifierNode(name, nodeInfo)
-    val localNode = createLocalNode(id.code, Defines.ANY.label)
-    scope.addVariable(id.code, localNode, BlockScope)
+  private def astForRequireCallFromImport(
+    name: String,
+    alias: Option[String],
+    from: String,
+    nodeInfo: BabelNodeInfo
+  ): Ast = {
+    val destName  = alias.getOrElse(name)
+    val destNode  = createIdentifierNode(destName, nodeInfo)
+    val localNode = createLocalNode(destName, Defines.ANY.label)
+    scope.addVariable(destName, localNode, BlockScope)
     diffGraph.addEdge(localAstParentStack.head, localNode, EdgeTypes.AST)
 
-    val destAst           = Ast(id)
+    val destAst           = Ast(destNode)
     val sourceCallArgNode = createLiteralNode(s"\"$from\"", None, nodeInfo.lineNumber, nodeInfo.columnNumber)
     val sourceCall = createCallNode(
       s"$REQUIRE_KEYWORD(${sourceCallArgNode.code})",
@@ -345,15 +351,24 @@ trait AstForDeclarationsCreator { this: AstCreator =>
       nodeInfo.columnNumber
     )
     val sourceAst = createCallAst(sourceCall, List(Ast(sourceCallArgNode)))
+    Ast.storeInDiffGraph(sourceAst, diffGraph)
+
+    val fieldAccessCall = createFieldAccessCallAst(
+      sourceAst.nodes.head,
+      createFieldIdentifierNode(name, nodeInfo.lineNumber, nodeInfo.columnNumber),
+      nodeInfo.lineNumber,
+      nodeInfo.columnNumber
+    )
+    Ast.storeInDiffGraph(fieldAccessCall, diffGraph)
+
     val assigmentCallAst =
       createAssignmentCallAst(
         destAst.nodes.head,
-        sourceAst.nodes.head,
-        s"var ${codeOf(destAst.nodes.head)} = ${codeOf(sourceAst.nodes.head)}",
+        fieldAccessCall.nodes.head,
+        s"var ${codeOf(destAst.nodes.head)} = ${codeOf(fieldAccessCall.nodes.head)}",
         nodeInfo.lineNumber,
         nodeInfo.columnNumber
       )
-    Ast.storeInDiffGraph(sourceAst, diffGraph)
     assigmentCallAst
   }
 
@@ -364,7 +379,7 @@ trait AstForDeclarationsCreator { this: AstCreator =>
     if (specifiers.isEmpty) {
       diffGraph.addNode(createDependencyNode(source, source, IMPORT_KEYWORD))
       createImportNodeAndAttachToAst(impDecl, source, source)
-      astForRequireCallFromImport(source, source, impDecl)
+      astForRequireCallFromImport(source, None, source, impDecl)
     } else {
       val specs = impDecl.json("specifiers").arr.toList
       val depNodes = specs.map { importSpecifier =>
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
```diff
@@ -352,7 +326,35 @@ trait AstForDeclarationsCreator { this: AstCreator =>
     }
     diffGraph.addNode(createDependencyNode(name, referenceName, IMPORT_KEYWORD))
     createImportNodeAndAttachToAst(impDecl, referenceName, name)
-    Ast()
+    astForRequireCallFromImport(name, referenceName, impDecl)
+  }
+
+  private def astForRequireCallFromImport(name: String, from: String, nodeInfo: BabelNodeInfo): Ast = {
```
maltek, line 332: this doesn't seem to handle [named imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#description
) yet.

Also, do we have to do hoist this to the start of `:program`? I remember @ml86 saying that imports work at any level, but I'm not sure to what extent our dynamic type engine cares about statement order.

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
```diff
@@ -146,7 +146,7 @@ trait AstForDeclarationsCreator { this: AstCreator =>
     val specifiers = declaration
       .json("specifiers")
       .arr
-      .toSeq
+      .toList
```
maltek, line 149: why are you converting things to linked lists? I tend to avoid them because of concerns about pointer chasing. But I never really looked into the topic in the scala ecosystem, so maybe I'm missing some reasons why they aren't so bad after all?

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
max-leuthaeuser: Named imports are all handled, see the test cases. 
The resulting AST is added to the exact same position where the import decl itself would appear.

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
max-leuthaeuser: Because certain helper functions in x2cpg require a List. I did not want to change that API in this PR.
But this is a good target for future refactorings I guess.

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
maltek: I don't see any field access calls here, so the handling can't be correct.

max-leuthaeuser: https://github.com/joernio/joern/blob/4f15589e72e982642aef6fc59412131a37178ada/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/DependencyAstCreationPassTest.scala#L159

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/DependencyAstCreationPassTest.scala
```diff
@@ -168,6 +167,35 @@ class DependencyAstCreationPassTest extends AbstractPassTest {
...
+        "alias2",
+        "defaultMember1",
+        "alias3",
+        "defaultMember2",
+        "module-name"
+      )
+      cpg.assignment.code.l shouldBe List(
+        "var name = require(\"module-name\")",
+        "var otherName = require(\"module-name\")",
+        "var member1 = require(\"module-name\")",
```
maltek, line 187: we can't figure out from such a call that `member1` refers to `^module-name^.member1`. It would have to be

```javascript
var member1 = require(\"module-name\").member1
```

ml86:
I do not think that is a route we want to pursue because this adds `require` calls to the CPG which could have been in the code as well, making it hard to judged where the `require` call originates from. It deviates more than necessary from the original code which we try to avoid. So joern/ocular user who wants to write a query finding files which import a library with `require` and `import` now has to resort to something like looking on a not present line number.

I think adjusting the backend passes to handle `require` and `import` nodes is the better way to go here.

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/DependencyAstCreationPassTest.scala
max-leuthaeuser: I see, let me fix that.

max-leuthaeuser:
@ml86 Yes, I guess that would be the cleanest solution.
But in the current state that would require a total rewrite of the CS RequirePreprocessor and dependents as we set `DYNAMIC_TYPE_HINT_FULL_NAME`s on the require calls there (these will later be used as parts of carrot paths). 
Without this PR here we do not have any calls for the import decl at all. Where do we then set the `DYNAMIC_TYPE_HINT_FULL_NAME`?

(And we do generate a ton of nodes for synthetic / de-sugared code already - e.g., for static init block, struct like objects / dicts, certain loop types)

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/DependencyAstCreationPassTest.scala
maltek: now we have too many field accesses. We only want them for what MDN calls `exportN` at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#syntax

maltek:
can't we just put it on the `local` nodes for these variables? (But I have no strong opinions either way.)

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/DependencyAstCreationPassTest.scala
max-leuthaeuser: Yes, done. I read over that section of the docs too quickly.

ml86:
@max-leuthaeuser You could just put a DYNAMIC_TYPE_HINT_FULL_NAME on every IDENTIFIER referencing the LOCAL representing the import. Since the imported module names are always static in `import` statements, the frontend can generate a property type name analog to RequirePreprocessor.

max-leuthaeuser:
I am not sure if that would work with named imports.

Example:
```
// what jssrc2cpg sees
import { member } from "module-name";

member.call();
```
This transpiles to:
```
// what js2cpg sees
var moduleName = require("module-name");

moduleName.member.call();
```

Setting the `DYNAMIC_TYPE_HINT_FULL_NAME` to `module-name` on `member` doesnt look right. The actual field access `moduleName.member` goes missing. We cant deal with carrot paths in the frontend either.

ml86:
There are two things we could do for named imports:
1. On the named import identifiers we could put DYNAMIC_TYPE_HINT_FULL_NAME for the member type. You would need to make sure to use the same type name as in https://github.com/ShiftLeftSecurity/codescience/blob/5e44e1389d128c5a21431f14f191515fd6c29312/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/passes/languagespecific/javascript/ExportTypeDeclCreator.scala#L273.
2. Create a variable for the complete import object e.g. named `_moduleNameObject` and than have a `var member = _moduleNameObject.member`.

Both have the advantage of not lowering to `require` calls and thus avoid confusion.

maltek (approved):

PR #1784, opened 2022-09-27
jssrc2cpg: set type hints for locals and identifiers based on imports

max-leuthaeuser:
Replaces: https://github.com/joernio/joern/pull/1780

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/build.sbt
+++ b/joern-cli/frontends/jssrc2cpg/build.sbt
@@ -6,6 +6,7 @@ dependsOn(Projects.dataflowengineoss % Test, Projects.x2cpg % "compile->compile;
 
 libraryDependencies ++= Seq(
   "io.shiftleft"              %% "codepropertygraph" % Versions.cpg,
+  "commons-io"                 % "commons-io"        % "2.11.0",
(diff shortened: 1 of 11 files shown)


maltek (changes requested):
Either we set dynamic type hints to proper carrot paths (`^module-name^.member`) or we find a way to avoid it (i.e. generate some intermediate field accesses). But let's not set dynamic type hints to some incompatible values.

max-leuthaeuser:
Isnt that what Markus mentioned (https://github.com/ShiftLeftSecurity/codescience/blob/5e44e1389d128c5a21431f14f191515fd6c29312/dataflowengine/src/main/scala/io/shiftleft/dataflowengine/passes/languagespecific/javascript/ExportTypeDeclCreator.scala#L273)?

maltek:
Ugh, I was only thinking about imports of external modules. That code is creating types for exported objects in the CPG. The two cases use incompatible naming schemes (avoiding collisions). For external libraries, we need carrot-paths (`^module-name^.member`). For internal it would currently need to be something like `./cpg/file/name.member` (but that one we're free to change as long as we change `ExportTypeDeclCreator` in lock-step).

max-leuthaeuser:
Closed in favor of: https://github.com/joernio/joern/pull/1780

PR #1788, opened 2022-09-27
Add PHP switch statements and fix variable code

johannescoetzee:
Switch statements felt like a good candidate for a common x2cpg method, but at least from comparing php2cpg and javasrc2cpg, it looks like actually extracting shared code will be difficult to do without adding just as much preprocessing code back in.


max-leuthaeuser (approved):
Maybe add a CFG test for case/switch?

johannescoetzee:
> Maybe add a CFG test for case/switch later on?

Good point. I'll add CFG tests for all of the control structures in a follow-up PR.

PR #1795, opened 2022-09-29
Apply data flow semantics at generation time

fabsx00:
Previously, we did not apply data flow semantics at DDG generation time, but only at query-time. While that's enough for reachableBy queries and is the most flexible solution, it means that raw exports of the graph will show a lot of `REACHING_DEF` edges that shouldn't exist. Since a lot of people seem to be interested in exporting the raw graph, I want to make sure they don't mistakenly believe that we don't honor data flow semantics. Therefore, semantics are now also applied at generation time.

### DIFF
--- a/benchmarks/src/test/scala/io/joern/benchmarks/testfixtures/BenchmarkFixture.scala
+++ b/benchmarks/src/test/scala/io/joern/benchmarks/testfixtures/BenchmarkFixture.scala
@@ -152,7 +152,9 @@ class BenchmarkCpgContext {
     applyDefaultOverlays(cpg.get)
     val context = new LayerCreatorContext(cpg.get)
     val options = new OssDataFlowOptions()
-    new OssDataFlow(options).run(context)
+    val semanticsFile: String            = ProjectRoot.relativise("benchmarks/src/test/resources/default.semantics")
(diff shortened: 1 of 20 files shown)


itsacoderepo (approved):

DavidBakerEffendi:
This is a great idea, so essentially the `REACHING_DEF` edges would be less over-approximated from the start? This will be great for change detection as we can use these edges to determine code changes with much more precision and improve the speed of incremental graph updates.

I wonder if this shouldn't perhaps be configurable for users who want to load large programs in and only run data-flow queries on subgraphs of the potentially large CPG? Then have some sensible default one way or the other.

fabsx00:
My guess would be that the performance hit is negligible, but I haven't measured it. The only real downside I see to this approach is that it's not possible to change a data-flow semantic to be more permissive after the fact. Given our overtainting approach, this shouldn't be too relevant though: usually, we want to reduce FPs by making semantics more restrictive.

DavidBakerEffendi:
Right, so since we're still generating `REACHING_DEF` edges intraprocedurally the restriction comes from the model semantics and won't have much of a performance hit as it's query equivalent. Got it 👍

DavidBakerEffendi (approved):

PR #1797, opened 2022-09-30
Rework semantics handling

fabsx00:
Idea: while we want to allow loading semantics from a file, in most cases, we just want to use the default semantic, so, only being able to load from file makes the code a lot more complex. Instead, we now provide the default semantics directly in code but also allow serializing it out. Finally, a custom semantic can be loaded on the shell or in scripts by setting the `semantics` variable.

* We now load semantics from the code, which is less brittle than loading from file
* Custom semantics can be specified on the shell and in joern scripts by setting `semantics`
* `report` was broken, a newline was missing.


itsacoderepo (approved):

console/src/main/scala/io/joern/console/Reporting.scala
```diff
@@ -8,7 +8,7 @@ trait Reporting {
   def reportOutStream: OutputStream = System.err
 
   def report(string: String): Unit = {
-    reportOutStream.write(string.getBytes("UTF-8"))
+    reportOutStream.write((string + "\n").getBytes("UTF-8"))
```
mpollmeier, line 11: `System.lineSeparator` ftw

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/semanticsloader/Parser.scala
```diff
@@ -29,6 +29,15 @@ class Semantics private (methodToSemantic: mutable.Map[String, FlowSemantic]) {
 
   def forMethod(fullName: String): Option[FlowSemantic] = methodToSemantic.get(fullName)
 
+  def serialize: String = {
+    elements
+      .sortBy(_.methodFullName)
+      .map { elem =>
+        s"\"${elem.methodFullName}\" " + elem.mappings.map { case (x, y) => s"$x -> $y" }.mkString(" ")
+      }
+      .mkString("\n")
```
mpollmeier, line 38: System.lineSeparator

PR #1801, opened 2022-10-04
Remove `semanticcpg/testing` deprecated class use

claudiuvursache:
doing this in smaller chunks than the previous attempt https://github.com/joernio/joern/pull/1778/files

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NewTagNodePairTraversal.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NewTagNodePairTraversal.scala
@@ -3,9 +3,11 @@ package io.shiftleft.semanticcpg.language
 import io.shiftleft.codepropertygraph.generated.EdgeTypes
 import io.shiftleft.codepropertygraph.generated.nodes.{NewNode, NewTagNodePair, StoredNode}
 import io.shiftleft.passes.DiffGraph
+import overflowdb.BatchedUpdate.DiffGraphBuilder
 import overflowdb.traversal._
 
 class NewTagNodePairTraversal(traversal: Traversal[NewTagNodePair]) extends HasStoreMethod {
+  val original = traversal
 
   override def store()(implicit diffGraph: DiffGraph.Builder): Unit = {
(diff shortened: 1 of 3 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NewTagNodePairTraversal.scala
```diff
@@ -3,9 +3,11 @@ package io.shiftleft.semanticcpg.language
 import io.shiftleft.codepropertygraph.generated.EdgeTypes
 import io.shiftleft.codepropertygraph.generated.nodes.{NewNode, NewTagNodePair, StoredNode}
 import io.shiftleft.passes.DiffGraph
+import overflowdb.BatchedUpdate.DiffGraphBuilder
 import overflowdb.traversal._
 
 class NewTagNodePairTraversal(traversal: Traversal[NewTagNodePair]) extends HasStoreMethod {
+  val original = traversal
```
max-leuthaeuser, line 10: Is this used somewhere?

max-leuthaeuser (approved):

mpollmeier (approved):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NewTagNodePairTraversal.scala
claudiuvursache: yesss, here https://github.com/joernio/joern/pull/1801/files#diff-3991921d186ab499e2c9c975dea554375dc7ec17624aed0cd22094e0cc7ab968R24

fabsx00: If the idea here is to be able to reaccess the traversal after it has been used, then you need a clone here.

claudiuvursache: that val was only temporary, forgot to add a comment about it. removing it soon with https://github.com/joernio/joern/pull/1814/files

PR #1802, opened 2022-10-05
Introduce `HasStoreAndPersistMethods` trait

claudiuvursache:
helper trait to ease the switch to the new DiffGraph API

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NewNodeSteps.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/NewNodeSteps.scala
@@ -2,18 +2,30 @@ package io.shiftleft.semanticcpg.language
 
 import io.shiftleft.codepropertygraph.generated.nodes.NewNode
 import io.shiftleft.passes.DiffGraph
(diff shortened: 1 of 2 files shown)


bbrehm:
I'm against this way of changing it. The coding pattern of using an implicit builder and then hitting ".store()" on a node is imo terrible. Furthermore, `tagNodePair` is an abomination. This is not a node in the graph; somebody added it to the schema to make the "DSL" more pretty.

It is a nice side-effect of the migration that we can get rid of that cruft. It is a further nice side-effect that we get to fix lots and lots of bugs in old code because someone looks at it with new eyes.

claudiuvursache:
@bbrehm i'm with you in that it's not ideal, and it may well be we can find better alternatives, at the moment my goal is to get rid of the deprecated DiffGraph API usage and for that, the change is very well suited.

bbrehm:
We removed almost all other deprecated DiffGraph API usage without porting cruft. The `.store()` function will get deprecated next, and it makes imo no sense to do the work twice. When we change the API on a fundamental level, then we should also do the necessary refactoring of coding patterns.

claudiuvursache:
discussed offline, we'll go with the change as it is for now

PR #1805, opened 2022-10-06
jssrc2cpg: remove astgen binaries with sbt clean

max-leuthaeuser:


fabsx00:
I guess that means that an `sbt clean compile` would lead to redownloading of `astgen`?

max-leuthaeuser:
Exactly.

fabsx00 (approved):

PR #1816, opened 2022-10-09
Fix data flow of js object notation

Liyw979:
https://github.com/joernio/joern/blob/518992dd540498b51f7acd1e66e57bb80c66be8a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala#L108

```
if call.name != Operators.fieldAccess
```
This line is blocking the data flow from inside object notation to outside. Here is the test case.
```
    val cpg: Cpg = code(
      """
        |const a = { b : 47 } ;
        |fn(a);
        |""".stripMargin)

    val sink = cpg.call.nameExact("fn")
    val src = cpg.literal("47")
    sink.reachableBy(src).nonEmpty shouldBe true
```

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala
@@ -1,9 +1,9 @@
 package io.joern.jssrc2cpg.dataflow
 
-import io.joern.jssrc2cpg.testfixtures.DataFlowCodeToCpgSuite
 import io.joern.dataflowengineoss.language._
-import io.shiftleft.codepropertygraph.generated.EdgeTypes
+import io.joern.jssrc2cpg.testfixtures.DataFlowCodeToCpgSuite
 import io.shiftleft.codepropertygraph.Cpg
+import io.shiftleft.codepropertygraph.generated.EdgeTypes
 import io.shiftleft.semanticcpg.language._
 import overflowdb.traversal._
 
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala
```diff
@@ -586,4 +586,15 @@ class DataflowTest extends DataFlowCodeToCpgSuite {
     println(sink.reachableBy(src).l)
```
fabsx00, line 586: While reviewing this, I noticed that this test here is also broken: could you possibly replace this with `sink.reachableBy(src).size shouldBe 1` as part of your PR? Many thanks!

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala
```diff
@@ -586,4 +586,15 @@ class DataflowTest extends DataFlowCodeToCpgSuite {
...
 
+  "Flow from inside object notation to call argument" in {
+    val cpg: Cpg = code("""
+        |const a = { b : 47 } ;
+        |fn(a);
+        |""".stripMargin)
+
+    val sink = cpg.call.nameExact("fn")
+    val src  = cpg.literal("47")
+    sink.reachableBy(src).nonEmpty shouldBe true
```
fabsx00, line 597: Please make this `sink.reachableBy(src).size shouldBe 1` so that we can detect if we ever get more than one flow here as well.

fabsx00 (commented):
Thank you for catching this! As you can see below, it turns out that one of the tests that was supposed to catch this was also broken: we were only printing out the flows but not actually checking them. This then probably broke in a refactor.

fabsx00:
Actually, let me merge this. I'll cleanup the tests in a follow-up PR.

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala
Liyw979: > While reviewing this, I noticed that this test here is also broken: could you possibly replace this with `sink.reachableBy(src).size shouldBe 1` as part of your PR? Many thanks!

Will try!

fabsx00 (approved):

Liyw979:
@fabsx00 Should we split ```Block``` into 2 kinds of block: ```StatementBlock``` for method body and ```ExpressionBlock``` for lowering like js object nation？And deal with them separately？

PR #1817, opened 2022-10-10
pysrc2cpg: added NAMEs for CALLs

Liyw979:
Howerver, FULL_NAME is not handled. It's still the same with NAME.

https://github.com/joernio/joern/blob/518992dd540498b51f7acd1e66e57bb80c66be8a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/NodeBuilder.scala#L15-L18


fabsx00 (changes requested):
Thanks for this: filling out the NAME field is/was definitely on the list for Python. If I understand correctly, you are introducing a couple of different placeholders in situations where we do not yet fill out the name correctly. I saw `"todo"`, `"<todo>"`, `"unresolved"` and `"<empty>"`. Let's just use the empty string `""` in all of these cases, I'd suggest. Otherwise, people may start matching on these placeholders.

Liyw979:
By the way

- In python and js
- - the receiver of ```x.y()``` call is ```x.y```.
- - the receiver of ```x()``` call is ```x```.

- But in java
- - the receiver of ```x.y()``` call is ```x```.
- - the receiver of static method call ```x()``` is empty.

According to [cpg.joern.io](https://cpg.joern.io/#edge-ref-receiver), shoule we change the implementation in python/js？

fabsx00:
Yep, and that leads to the core of the problem here: in JS/Python, we're modeling the call `x.y()` as the field access `x.y`, followed by a call to it's apply method `(*)`, while in Java, we know that `y` is a method for sure, and so, this is a call of `y` on the receiver `x`. I agree with you and think it should be done as the Java frontend does, however, I'd say we do that in a separate PR.

fabsx00 (approved):

fabsx00:
Short update on this one: I'm planning to merge this by end of week. Right now, it requires some changes downstream that are implemented as we speak.

Liyw979:
@fabsx00 
> in JS/Python, we're modeling the call `x.y()` as the field access `x.y`, followed by a call to it's apply method `(*)`

I think so. We just need to change the receiver from ```x.y``` to ```x``` in this case.

fabsx00:
Update: scheduled to be merged early next week. Also, we're looking to remove the receiver-concept altogether because it's not well defined and not used in the OSS code.

Liyw979:
```OSS``` is the abbreviation of ? 🤔

mpollmeier:
open source software - the thing we're doing here :)

PR #1822, opened 2022-10-11
Adjust code in dataflowengineoss

Liyw979:


Liyw979:
One question: What is the difference between Identifier and NewIdentifier?

mpollmeier:
Identifier is already stored in the underlying graph, `NewIdentifier` is a new node, typically in a DiffGraph, which will be stored in the graph at a later point (when the DiffGraph get's applied to a graph).

PR #1825, opened 2022-10-11
refactor DdgGenerator.scala

Liyw979:
**Slightly modified one test case of kotlin dataflow**
I am not sure the refator is 100% correct, needs further discussion.


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala
```diff
@@ -31,96 +29,77 @@ class DdgGenerator(semantics: Semantics) {
...
-          .filterNot(x => numberToNode.get(x).contains(use))
-          .flatMap(numberToNode.get)
-          .foreach { inElemNode =>
-            addEdge(inElemNode, ret, nodeToEdgeLabel(inElemNode))
-          }
-        if (inElements.isEmpty) {
-          addEdge(method, ret)
-        }
-      }
+      val uses = ret.astChildren.collect { case x: Expression => x }.toList
```
mpollmeier, line 100:
```suggestion
      val uses = ret.astChildren.collectAll[Expression].toList
```

mpollmeier (commented):
Thank you! From a coding style perspective this is a step forward IMO. 
@fabsx00 can you check if the test outcome change is ok?

Liyw979:
According to the comment
https://github.com/joernio/joern/blob/43d332f6db7563b1f5dd3824cf032c841913f264/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala#L45
A node can only be linked by the ```method``` node if this node has no other reachingdef edge
However, the old ddg generator does not do the same
image

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala
Liyw979: Looks like a lot of tests are broken. I will try fix them

Liyw979:
@fabsx00 In the first commit, I used 
```
    def incomingDefsForNode(node: StoredNode): Set[StoredNode] = {
      in(node).toSet
```
which doesn't follows the original logic
In the last commit, It's changed to use the origin logic
```
    def incomingDefsForNode(current: StoredNode, use: StoredNode): Set[StoredNode] = {
      in(current).toSet
```

Liyw979:
This pr is still needs improvement. leave this pr open.

Liyw979:
Closing this since bugs I found in DdgGenerator have being fixed.

PR #1824, opened 2022-10-11
Fixed tons of Linting and Compiler Warnings

max-leuthaeuser:
Did not touch any deprecation warning (DiffGraph etc.)

From sbt clean +compile (some Scala 3 Compiler warnings) + IntelliJ Linter

PR #1827, opened 2022-10-11
Add PHP consts

johannescoetzee:


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
```diff
@@ -208,8 +221,11 @@ class AstCreator(filename: String, phpAst: PhpFile, global: Global) extends AstC
       case printExpr: PhpPrint       => astForPrintExpr(printExpr)
       case ternaryOp: PhpTernaryOp   => astForTernaryOp(ternaryOp)
       case throwExpr: PhpThrowExpr   => astForThrow(throwExpr)
-      case null =>
-        logger.warn(s"expr was null")
```
johannescoetzee, line 212: This was changed in https://github.com/joernio/joern/pull/1824, but the way it was written before is more useful for development purposes. Removing the unhandled case to get a matchError instead would also work, but I prefer to keep this explicit for now.

max-leuthaeuser: It can't be anything other than null as you are using a sealed class hierarchy there. This is why the compiler is complaining there.

johannescoetzee: This is (and should be) true for code in a PR, but is not necessarily the case during development. Would leaving the `unhandled` case as it is (that is, not specifically matching on null) be an issue later on?

max-leuthaeuser: I do not quite get how this won't be the case during development.
As long as the signature is `def astForExpr(expr: PhpExpr): Ast` you can't feed anything other than an instance of a subclass of `PhpExpr` in there where the compiler will ensure that the cases are covered (sealed hierarchy). 
So do you change the signature during your local development? 
Sorry for holding you up here, just want to understand this as I think these compiler warning are very useful here (to miss a pattern match case).

max-leuthaeuser (approved):

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
johannescoetzee: Ah, I think we're just following different approaches to the same issue here. I don't know how much of a difference it makes, but I'm also still running Scala 2. I typically don't look at warnings very much since there have historically been quite a lot of them (although that seems much better now), so the catch-all case was just to crash while outputting a more informative warning. Without the catchall case, I'd get a MatchError with similar-enough information though, so there isn't really that much to gain either way.

PR #1837, opened 2022-10-13
Fix pending dataflow test case in C

Liyw979:
With ```EngineContext(config = EngineConfig(maxCallDepth = 4))``` 😂

### DIFF
--- a/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/dataflow/DataFlowTests.scala
+++ b/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/dataflow/DataFlowTests.scala
@@ -2,11 +2,9 @@ package io.joern.c2cpg.dataflow
 
 import io.joern.c2cpg.testfixtures.DataFlowCodeToCpgSuite
 import io.joern.dataflowengineoss.language._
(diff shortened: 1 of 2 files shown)


Liyw979:
By the way, can you provide way to set ```maxCallDepth``` in the ```joern console``` since the default ```maxCallDepth``` is ```2```

image

fabsx00:
`maxCallDepth` is a `var`, so, on the shell, you can issue the following:
```
context.config.maxCallDepth = 4
```

fabsx00:
Makes me wonder whether we should increase the default value of `maxCallDepth` as well. Otherwise, tests pass but the flows aren't found in practice.

Liyw979:
Sounds good. It takes me one month to find out there is such a configuration of dataflow algrithm.

fabsx00:
This one changes the default value to 4, so, on this PR, you can omit the `=4`:
https://github.com/joernio/joern/pull/1841

Liyw979:
@fabsx00 How about leaving this PR unchanged and create a new PR the clean up ```EngineConfig(maxCallDepth = 4)``` in all test fixtures.
image

fabsx00:
Fine with me as well.

PR #1839, opened 2022-10-13
Remove `undo`

claudiuvursache:
_undo_ was introduced at some point to allow for edits while using Joern as a platform. Because it's not been used very extensively, and in order to help with removing deprecated logic, we decided to remove it for now.


fabsx00 (approved):
... there is a lot more that we can remove now, e.g., storing of undo information in proto files.

claudiuvursache:
> ... there is a lot more that we can remove now, e.g., storing of undo information in proto files.

gtk, will see if i can get that out as well since we're going down this path

PR #1846, opened 2022-10-13
Remove path.size > 1 for this test case

Liyw979:
```
  "DataFlowTest70" should {
    val cpg = code("""
        | int source() {
        |   return 42;
        | }
        | void main() {
        |   sink(source());
        | }
        | """.stripMargin)

    "Test Interprocedural" should {
      "have a flow from argument(which itself is a call) to return" in {
        val source = cpg.literal("42")
        val sink   = cpg.call("sink").argument
        sink.reachableByFlows(source).size shouldBe 1
      }
    }
  }
```

### DIFF
--- a/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/dataflow/DataFlowTests.scala
+++ b/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/dataflow/DataFlowTests.scala
@@ -1337,7 +1337,15 @@ class DataFlowTests extends DataFlowCodeToCpgSuite {
       val source = cpg.call("source")
       val sink   = cpg.call("sink")
       val flows  = sink.reachableByFlows(source)
-      flows.map(flowToResultPairs).toSetMutable shouldBe Set(List(("source()", Some(6)), ("sink(source())", Some(6))))
+      flows.map(flowToResultPairs).toSetMutable shouldBe Set(
+        List(
+          ("source()", Some(6)),
+          ("sink(int arg)", Some(2)),
+          ("return arg;", Some(2)),
+          ("int", Some(2)),
+          ("sink(source())", Some(6))
+        )
+      )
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/dataflow/DataFlowTests.scala
```diff
@@ -1337,7 +1337,15 @@ class DataFlowTests extends DataFlowCodeToCpgSuite {
       val source = cpg.call("source")
       val sink   = cpg.call("sink")
       val flows  = sink.reachableByFlows(source)
-      flows.map(flowToResultPairs).toSetMutable shouldBe Set(List(("source()", Some(6)), ("sink(source())", Some(6))))
+      flows.map(flowToResultPairs).toSetMutable shouldBe Set(
+        List(
```
fabsx00, line 1341: This flow looks a bit strange. I think we gotta take a closer look here before merging this one. Looks like we're now descending into the sink, coming up from it, and then traversing to the argument `source()`.

fabsx00: Actually, it's correct.

fabsx00 (approved):

Liyw979:
@fabsx00 I am wondering should we do the same to 
https://github.com/joernio/joern/blob/aa7e153ab314649cb7f245c3525fc067655c1827/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/TaskSolver.scala?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L135

PR #1852, opened 2022-10-15
add dataflow test template for python

Liyw979:


ml86 (changes requested):
Since the data flow tests for the pysrc frontend are a new group of tests for which so far there has not yet been a test fixture, please use: https://github.com/joernio/joern/blob/6514e297e363ed64e07debec38ea3b933e4d0352/joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/testfixtures/Code2CpgFixture.scala#L11.
This is the new base fixture we want to use across all frontends. This fixture has the advantage of requiring less code to setup the test code. For an example on how to derive from it, please have a look here: https://github.com/joernio/joern/blob/dea4c5218df811b1da7afa4ba02bdc7b58deb1d3/joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/testfixtures/KotlinCodeToCpgFixture.scala#L27

fabsx00:
Yes, it was developed in somewhat of a disconnect with the other frontends. Refactoring it to follow the format of the other frontends is definitely something we should target. In fact, I believe there are more utility methods that one could create in `x2cpg` that would allow deduplicating code across various frontends.

ml86 (approved):

PR #1854, opened 2022-10-16
Improve table

Liyw979:
**A. introduce a new column in the table**
Before: It's confusing to have 2 identical line here
image
After
image

**B. handle line break**
Before 
image
After
image

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/Path.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/Path.scala
@@ -35,7 +35,7 @@ object Path {
             val paramsPretty = method.parameter.toList.sortBy(_.index).map(_.code).mkString(", ")
             s"$methodName($paramsPretty)"
           case _ => cfgNode.statement.repr
-        }).replace("\n", "")
+        }).replace("\n", " ")
 
         Array(nodeType, trackedSymbol, lineNumber, methodName, fileName)
       }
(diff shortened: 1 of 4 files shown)


fabsx00:
@max-leuthaeuser I believe you mentioned there is a standard library we can use to shorten `code` strings as well, given that they can become very long when working with `jssrc2cpg`. Can you review this one and see if we can introduce that as part of this PR?

max-leuthaeuser:
Yes, we should use https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#normalizeSpace-java.lang.String-
to handle this.
And https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#abbreviate-java.lang.String-int- for shortening.

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/Path.scala
```diff
@@ -22,21 +22,22 @@ object Path {
...
         val fileName   = method.file.name.headOption.getOrElse("N/A")
 
-        val trackedSymbol = cfgNode match {
+        val trackedSymbol = (cfgNode match {
           case _: MethodParameterIn =>
             val paramsPretty = method.parameter.toList.sortBy(_.index).map(_.code).mkString(", ")
             s"$methodName($paramsPretty)"
           case _ => cfgNode.statement.repr
-        }
+        }).replace("\n", " ")
```
max-leuthaeuser, line 38: See my comment. \n is too simple and will not handle platform dependent line breaks or tabs etc.

max-leuthaeuser (changes requested):

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/Path.scala
Liyw979: Thanks for pointing out this!

max-leuthaeuser (approved):

PR #1855, opened 2022-10-16
Allow visiting same node multiple times

Liyw979:
Test case failed before this PR
 ```
   "can go in a method multiple times" in {
      val cpg = code("""
          |class Foo{
          |    String name;
          |    String getName() {
          |        return name;
          |    }
          |}
          |class Main{
          |    public void bar(Foo foo) {
          |        check(foo.getName());
          |        sink(foo.getName());
          |    }
          |}
          |""".stripMargin)

      def source = cpg.method("bar").parameter.name("foo")
      def sink   = cpg.method("sink").parameter.index(1)
      sink.reachableBy(source).size shouldBe 1
```
cause:
```
check(foo.getName());
sink(foo.getName());
```
with
https://github.com/joernio/joern/blob/210e867fa1921fd036d6388d56eb4b8751f97fc7/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L189

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name                     := "joern"
 ThisBuild / organization := "io.joern"
 ThisBuild / scalaVersion := "2.13.8"
 
-val cpgVersion = "1.3.568"
+val cpgVersion = "1.3.570"
 
 lazy val joerncli          = Projects.joerncli
 lazy val querydb           = Projects.querydb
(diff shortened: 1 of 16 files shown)


fabsx00 (changes requested):
This, to me, looks like it changes the runtime complexity of the algorithm, something we need to be careful with. I'll take a more detailed look into this problem when I have the time.

PR #1857, opened 2022-10-17
Fix bug in DynamicCallLinker

Liyw979:
https://github.com/joernio/joern/blob/210e867fa1921fd036d6388d56eb4b8751f97fc7/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/DynamicCallLinker.scala#L85-L86

**shoule be .fullNameExact(...)**

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/dataflow/FunctionCallTests.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/dataflow/FunctionCallTests.scala
@@ -1,11 +1,51 @@
 package io.joern.javasrc2cpg.querying.dataflow
 
-import io.joern.javasrc2cpg.testfixtures.{JavaDataflowFixture, JavaSrcCode2CpgFixture}
 import io.joern.dataflowengineoss.language._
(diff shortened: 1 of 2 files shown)


fabsx00:
I believe, right now, we are doing this on purpose. While it may lead to some invalid call edges, in languages such as Python and Javascript, we rarely have full names at the moment - something that is being worked on. Leaving this one open though in case we do want to go this route.

Liyw979:
> we rarely have full names at the moment

I don't quite understand🤔 , ```fullName``` of ```typ``` is correct for Js and Python at the moment. 
Why do we use ```fullName``` to match ```name``` here? 
In the new test case, this will use ```a.Foo``` to match ```Foo``` and will not find the correct ```typ```

Liyw979:
If this is ```cpg.typ.nameExact(typDeclName)``` or ```cpg.typ.fullNameExact(typDeclFullName)```, I can get the idea.
But I don't undestand using ```name``` to match ```fullname``` in https://github.com/joernio/joern/blob/210e867fa1921fd036d6388d56eb4b8751f97fc7/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/DynamicCallLinker.scala#L85-L86

Liyw979:
Closing this since this pr is included in #2143

PR #1858, opened 2022-10-17
jssrc2cpg: add `IMPORT` nodes for `require` statements

fabsx00:
Previously, we only added `IMPORT` nodes for `import` statements while `require` calls were only accessible via `cpg.call("require")`. With this PR, it doesn't matter whether modules are imported via `import` or `require`, in both cases, we create `IMPORT` nodes.

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/MixedAstCreationPassTest.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/MixedAstCreationPassTest.scala
@@ -1264,4 +1264,17 @@ class MixedAstCreationPassTest extends AbstractPassTest {
     }
   }
 
+  "AST generation for import/require" should {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/MixedAstCreationPassTest.scala
```diff
@@ -1264,4 +1264,20 @@ class MixedAstCreationPassTest extends AbstractPassTest {
     }
   }
 
+  "AST generation for import/require" should {
```
max-leuthaeuser, line 1267: There is a test suite somewhere already testing many import / dependency related stuff. Could you put this there? E.g., https://github.com/joernio/joern/blob/master/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/DependencyAstCreationPassTest.scala

max-leuthaeuser (approved):

fabsx00:
Yes, it's called `DependenciesPassTest`. I didn't put it there though because we're not testing that pass but rather the AST creation.

PR #1863, opened 2022-10-18
Add PHP method calls and misc small statements

johannescoetzee:
This also adds namespaces and some methodFullNames (with unresolved signatures)

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/Domain.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/Domain.scala
@@ -206,11 +206,21 @@ object Domain {
     attributes: PhpAttributes
   ) extends PhpStmt
 
+  final case class PhpNamespaceStmt(name: Option[PhpNameExpr], stmts: List[PhpStmt], attributes: PhpAttributes)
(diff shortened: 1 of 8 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/Domain.scala
```diff
@@ -576,7 +598,12 @@ object Domain {
 
   private def readVariable(json: Value): PhpVariable = {
     // TODO Figure out when the variable has an expr name
```
max-leuthaeuser, line 600: This TODO is done now?

max-leuthaeuser (approved):

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/Domain.scala
johannescoetzee: It is! Good catch

fabsx00:
There seems to be something that changed about the macos-11 runner related to PHP.

johannescoetzee:
It looks like it was just a network blip. A maven package fetch failed, but all's good on a rerun.

PR #1871, opened 2022-10-20
Stop the taskSolver from going in a method visited before

Liyw979:
Another approach for #1855 

Test case failed before this PR
 ```
   "do not go in a method twice" in {
      val cpg = code("""
          |class Foo{
          |    String name;
          |    String getName() {
          |        return name;
          |    }
          |}
          |class Main{
          |    public void bar(Foo foo) {
          |        check(foo.getName());
          |        sink(foo.getName());
          |    }
          |}
          |""".stripMargin)

      def source = cpg.method("bar").parameter.name("foo")
      def sink   = cpg.method("sink").parameter.index(1)
      sink.reachableBy(source).size shouldBe 1
```
Cause: two ```call``` to the same ```method```, leading the algorithm to go in ```getName``` twice:
```
check(foo.getName());
sink(foo.getName());
```
When we reach ```sink(foo.getName())``` we go in ```getName``` once, and when we reach ```check(foo.getName());``` we don't need to go in ```getName``` again, beacaue we have searched ```getName``` and if we do go in
https://github.com/joernio/joern/blob/210e867fa1921fd036d6388d56eb4b8751f97fc7/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L189
is triggered, and the task will fail.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/TaskSolver.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/TaskSolver.scala
@@ -22,6 +22,8 @@ class TaskSolver(task: ReachableByTask, context: EngineContext) extends Callable
 
   import Engine._
 
+  implicit val implicitEngineContext: EngineContext = context
+
   /** Entry point of callable. First checks if the maximum call depth has been exceeded, in which case an empty result
     * list is returned. Otherwise, the task is solved and its results are returned.
     */
@@ -126,7 +128,7 @@ class TaskSolver(task: ReachableByTask, context: EngineContext) extends Callable
       // this isn't the start node => return partial result and stop traversing
       case call: Call
           if isCallToInternalMethodWithoutSemantic(call)
-            && !isArgOrRetOfMethodWeCameFrom(call, path) =>
+            && !isMethodWeCameFrom(call, path) =>
         createPartialResultForOutputArgOrRet()
 
       // Case 4: we have reached an argument to an internal method without semantic (output argument) and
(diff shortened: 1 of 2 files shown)


fabsx00:
You're right in that we should expand into `getName` again. What should happen is that we already have a summary for `getName` that we use instead of traversing into `getName`. Then the check on 189 should never be triggered in the first place. The fix discovers that we've taken a wrong turn, but I'd like to see if we can avoid doing so in the first place. Investigating...

fabsx00:
I'm still debugging this, but it's on the top of the list now.

fabsx00:
I have a better understand of the problem now than before, and in summary, disallowing a function to be explored twice is actually incorrect. We want to disallow entering a function twice via the same node. That's not happening here though: the first time, we enter `getName` via its `METHOD_RETURN` node. The second time, we enter it via the `METHOD_OUTPUT_PARAMETER` of `this`. In both cases, we eventually reach the `METHOD_INPUT_PARAMETER` of `this`, which is where the tracker exits, and that's the bug.

The problem with this PR is that we're essentially saying that one cannot visit a method twice, and that's not true. Say, for example, we first trace it going backwards from its first argument, and then, in another call, we end up at its second output argument, we would still have to traverse again, or risk false positives.

Introducing a "number of times a function can be traversed" is also not what we want. Instead, that check on 189 is incorrect, and we need to ensure that we are making correct use of the cache so that we don't enter a function twice via the same node.

I'll look into it tomorrow, I hope.

Liyw979:
> The problem with this PR is that we're essentially saying that one cannot visit a method twice, and that's not true.

Yes, I realise this too later this PR.

Liyw979:
> We want to disallow entering a function twice via the same node.

Maybe we should create another PR to do this, because this is not the casue of current PR.

We can introduce a ```Map[Method, Set[CfgNode]]``` to  ```ReachableByTask``` and ```ReachableByResult``` to write down  the entry nodes used.
The entry node in a method can only be one of ```MethodReturn```, ```Return```, ```ParameterOut``` and ```Call```
Then update and check ```map``` in ```taskCreator``` where we have access to ```the same node```

fabsx00:
Got a pretty simple fix - still needs cleanup though. The key idea is that the path should not just be a path of nodes, but it should be a `(path, callStack)` pair, and we only want to stop traversing when we reach the same `(path, callStack)` pair:

https://github.com/joernio/joern/pull/1878

I will let you know when this is ready for review.

PR #1878, opened 2022-10-21
Fix for data flow issue when visiting function twice

fabsx00:
We used to check whether we visit a node in the graph twice, in which case we terminated. That's unfortunately not correct if a node is visited twice but was reached via a different call chain. Instead, we now only terminate if we reach a node a second time via the same call chain.


max-leuthaeuser (approved):

Liyw979:
🎉

Liyw979:
🎉

DavidBakerEffendi (approved):

PR #1879, opened 2022-10-21
Frontends only need to depend on dataflowengineoss for tests.

ml86:
Also fixed up dependencies for querydb since it so far only got
dataflowengineoss transitively which is now not the case anymore.


fabsx00 (changes requested):
I'm using it an extension I'm working on.

ml86:
Cant you pull in `dataflowengineoss` in that extension? Like it is at the moment, `dataflowengineoss` is always on the classpath for every frontend user which sees overkill.

max-leuthaeuser:
Closing as superseded: all of this landed via #5862 (commit 96ce1bb71). Frontends now depend on dataflowengineoss only in test scope, querydb has the explicit dependency, and createCpgWithAllOverlays was removed from JsSrc2Cpg/SwiftSrc2Cpg.

PR #1889, opened 2022-10-24
Fix bug in ddg generator

Liyw979:
casue: 
https://github.com/joernio/joern/blob/0807206a83582935d9473180814b5befefa6e816/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala#L88
```in(node)``` brings in unsed reaching definition, need to filter out unused node.


fabsx00 (approved):
Good catch! Thank you.

fabsx00:
I wonder why we didn't at least catch this at query time. I'll investigate.

PR #1890, opened 2022-10-24
A call linker for python

Liyw979:
This PR introduce a call linker for python.
The linker links ```calls``` and ```methods``` only by their **name**(not full name), because it's hard to decide thier full name in a dynamic language.
Though this will result in wrong edges, it will make interprocedural data flow analysis work for the moment.


fabsx00 (approved):
Ok, sure, let's do this for now. It's better than not having CALL edges at all.

Liyw979:
Let me resolve the conflicts

fabsx00:
Cool, thanks.

Liyw979:
Actually, I think this linker can also be used for ```js``` and ```php``` for the moment.
By the way, please review #1847

Issue #1893, opened 2022-10-24
joern-export: allow exporting separate methods


fabsx00:
Specifically (or at least for) the graphson output, it would be great if we could export graphs per method as we do with the dot exporter. It's clear that we end up discarding parts of the CPG that aren't inside methods, but that's fine. It would also be important here that we can match method names and graphson file in the output.

mpollmeier:
initial version done via https://github.com/joernio/joern/pull/1902

mpollmeier:
"initial" because we may modify it based on customer feedback. anyway, closing for now

PR #1902, opened 2022-10-26
joern-export: SplitByMethod

mpollmeier:
implements https://github.com/joernio/joern/issues/1893
depends on https://github.com/ShiftLeftSecurity/overflowdb/pull/331 and https://github.com/ShiftLeftSecurity/overflowdb/pull/332 - won't compile until upgrade train has completed

Also includes some refactorings - most notably: Representation|Format: changed to uppercase 
(as per Scala convention and support in match clauses), but match on name ignore case when parsing config

### DIFF
--- a/joern-cli/src/main/scala/io/joern/joerncli/JoernExport.scala
+++ b/joern-cli/src/main/scala/io/joern/joerncli/JoernExport.scala
@@ -33,7 +33,7 @@ object JoernExport extends App {
   /** Choose from either a subset of the graph, or the entire graph (all).
     */
   object Representation extends Enumeration {
-    val Ast, Cfg, Ddg, Cdg, Pdg, Cpg14, All, SplitByMethod = Value
+    val Ast, Cfg, Ddg, Cdg, Pdg, Cpg14, Cpg, All = Value
 
     lazy val byNameLowercase: Map[String, Value] =
       values.map { value =>
@@ -128,7 +128,7 @@ object JoernExport extends App {
     val ExportResult(nodeCount, edgeCount, _, additionalInfo) = repr match {
       case Representation.All =>
         exporter.runExport(cpg.graph, outDir)
-      case Representation.SplitByMethod =>
+      case Representation.Cpg =>
         splitByMethod(cpg).iterator.map { subGraph =>
           val name = subGraph.name
           val sanitizedFileName =
--- a/joern-cli/src/main/scala/io/joern/joerncli/JoernExport.scala
+++ b/joern-cli/src/main/scala/io/joern/joerncli/JoernExport.scala
@@ -5,17 +5,17 @@ import better.files.File
 import io.joern.dataflowengineoss.DefaultSemantics
 import io.joern.dataflowengineoss.layers.dataflows._
 import io.joern.dataflowengineoss.semanticsloader.Semantics
-import io.joern.joerncli.CpgBasedTool.{exitIfInvalid, exitWithError}
+import io.joern.joerncli.CpgBasedTool.exitIfInvalid
 import io.joern.x2cpg.layers._
+import io.shiftleft.codepropertygraph.Cpg
+import io.shiftleft.semanticcpg.language.{toAstNodeMethods, toNodeTypeStarters}
 import io.shiftleft.semanticcpg.layers._
-import overflowdb.Graph
 import overflowdb.formats.ExportResult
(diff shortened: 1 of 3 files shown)


joern-cli/src/main/scala/io/joern/joerncli/JoernExport.scala
```diff
@@ -5,35 +5,55 @@ import better.files.File
...
-    format: Format.Value = Format.dot
+    repr: Representation.Value = Representation.Cpg14,
+    format: Format.Value = Format.Dot
   )
 
   /** Choose from either a subset of the graph, or the entire graph (all).
     */
   object Representation extends Enumeration {
-    val ast, cfg, ddg, cdg, pdg, cpg14, all = Value
+    val Ast, Cfg, Ddg, Cdg, Pdg, Cpg14, All, SplitByMethod = Value
```
fabsx00, line 36: I guess it's a bit strange that ASTs, CFGs, Ddgs, ... are also split by method, and then there's a `SplitByMethod`, but I also can't think of a different way of doing this right now. I guess as long as the docs explain this, we're fine.

fabsx00 (approved):

joern-cli/src/main/scala/io/joern/joerncli/JoernExport.scala
mpollmeier: discussed via slack: we'll rename `SplitByMethod` to `Cpg`
...  there's no optimal solution 🤷🏻

joern-cli/src/main/scala/io/joern/joerncli/JoernExport.scala
```diff
@@ -156,13 +159,46 @@ object JoernExport extends App {
...
+    *   utility map to ensure we don't override output files for identical method names
+    */
+  private def sanitizedFileName(
+    methodName: String,
+    methodFilename: String,
+    fileExtension: String,
+    windowsFilenameDeduplicationHelper: mutable.Set[String]
+  ): String = {
+    val sanitizedMethodName = methodName.replaceAll("[^a-zA-Z0-9-_\\.]", "_")
+    val isWindows           = System.getProperty("os.name").toLowerCase.contains("win")
```
max-leuthaeuser, line 178: There is `scala.util.Properties.isWin` for this.

mpollmeier: nice, thanks!

mpollmeier:
@max-leuthaeuser I had to lower the 'expected export file count' for windows: https://github.com/joernio/joern/pull/1902/files#diff-8857d40c75e3a90bdfd9c46e94a01e8662591f4d6edc7f3c58bce4077fb58523R19-R25

For some reason the cpg produced on windows is much smaller. I didn't look into the details and for this PR it doesn't really matter - I'm just wondering if that's a known thing.

joern-cli/src/test/scala/io/joern/joerncli/JoernExportTests.scala
```diff
@@ -15,7 +15,14 @@ class JoernExportTests extends AnyWordSpec with Matchers with AbstractJoernCliTe
       val tempDir = os.temp.dir(prefix = "joern-export-test")
       JoernExport.exportCpg(cpg, Representation.Cpg, JoernExport.Format.Graphson, tempDir.toNIO)
       val exportedFiles = os.walk(tempDir).filter(_.toIO.isFile)
-      exportedFiles.size should be >= 100
+
+      // c frontend creates a different cpg on windows - not sure about the details, but purely for the export it doesn't
```
max-leuthaeuser, line 19: The windows GH action runner does not provide a gcc installation. Thus, scanning system header files results in 0 files. You may fix this by running the c2cpg with `--no-include-auto-discovery`. Maybe we should make this the default as it pops up every now and then (leads to different CPGs on different environments all the time).

max-leuthaeuser: @mpollmeier There is https://github.com/joernio/joern/pull/1909 for this.

mpollmeier: 👍🏻 I see
re https://github.com/joernio/joern/pull/1909 i'll drop the distinction in this test then

PR #1928, opened 2022-11-02
QueryDB: Java & Android Crypto Vulnerabilities

DavidBakerEffendi:
- Removed simple constant detection query (I don't feel like it fits the theme of a "vulnerability query DB")
- Using the [following paper](https://files.sri.inf.ethz.ch/website/papers/diffcode-pldi2018.pdf) for reference:
  - Added a rule to check for MD5 or SHA-1 hash usage
  - Added a check for low iterations on PBEKey constructor
  - Added a check for PRNG usage on Android minSdk v16-18


DavidBakerEffendi:
The paper has a few more rules that I will implement in the coming days, but just want to avoid one large PR.

claudiuvursache (approved):

querydb/src/test/scala/io/joern/suites/JavaQueryTestSuite.scala
```diff
@@ -0,0 +1,48 @@
...
+          .concat("\n")
+          .concat(
+            q.codeExamples.negative
+              .mkString("\n")
+          )
+      }
+      .mkString("\n")
+
+  def findMatchingCalls(cpg: Cpg, q: Query): List[String] = {
+    q(cpg).flatMap(_.evidence).collect { case c: Call => c.code }
```
mpollmeier, line 36:
```suggestion
    q(cpg).flatMap(_.evidence).collectAll[Call].code.l
```

not sure if it's really better, just putting it out there to consider. 
same goes for the two queries below

mpollmeier (approved):

querydb/src/test/scala/io/joern/suites/JavaQueryTestSuite.scala
DavidBakerEffendi: Screenshot 2022-11-02 at 14 07 46
Unfortunately, we lose access to the property traversal steps. So I'll keep it as is for now.

PR #1930, opened 2022-11-02
Publish test artifacts for all projects.

ml86:
Some of the code we have under tests is or will be reused in other
projects. Mainly our test fixtures will be reused but creating
individual projects which just test utilities for each already existing
project seem overkill.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -90,3 +90,5 @@ publish / skip := true // don't publish the root project
 lazy val root = project
   .in(file("."))
   .aggregate(aggregatedProjects: _*)
+
+ThisBuild / Test / packageBin / publishArtifact := true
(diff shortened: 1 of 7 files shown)


Liyw979:
Thanks for this

mpollmeier (approved):
as discussed on slack

just checked: the kotlin2cpg-tests jar is 4M, since it contains some jars... if the release job slows down significantly by this change, we should externalise those jars and potentially reconcider this change, and instead whitelist project by project

ml86:
I missed that. For now the 4MB will not hurt us much. But i will take a look and try to get rid of those jars.

mpollmeier:
looks like release runtimes haven't changed
https://github.com/joernio/joern/actions/workflows/release.yml

PR #1932, opened 2022-11-02
Fix ast of JS arrow functions

Liyw979:
For `const foo = () => 42`, there should be a ```Return``` node for the data flow algrithm.
image
For `const foo = () => {return 42;}`, there should be only one ```Block``` node.
image


Liyw979:
After: 
`const foo = () => 42`
image
`const foo = () => {return 42;}`
image

max-leuthaeuser (approved):

PR #1934, opened 2022-11-02
Upgrade CPG

fabsx00:
Bring in modified CPG where no more proto serialization is performed. Drastically reduces memory usage and increases runtime performance.


ml86 (approved):

mpollmeier:
importing keycloak required 20g before this PR and now works with 6g 🎉

mpollmeier:
Actually, even works with 4g

Issue #1895, opened 2022-10-25
Dispatch Type?


Liyw979:
```
  "Flow into method defined as lambda and assigned to constant" in {
    val cpg: Cpg = code("""
        | const foo = (x, y) => {
        |   sink(x);
        | };
        |
        | foo(1, 2);
        |""".stripMargin)

    val sink = cpg.call("sink").l
    val src  = cpg.literal.code("1").l
    sink.size shouldBe 1
    src.size shouldBe 1
    sink.reachableBy(src).size shouldBe 1
  }
```
<img width="550" alt="image" src="https://user-images.githubusercontent.com/16919236/197719094-a6cff3a7-cc2c-4423-b310-148110b0ddfa.png">

I think the ```dispatchType``` of ```sink``` and ```foo``` should be ```Static```

max-leuthaeuser:
As we do not know what `sink` actually is we set it correctly I guess (also required for call linking at a later stage). For `foo` we could set it here (`const` and directly in the same scope) - but this would  be no general solution.

Liyw979:
I get it. Thank you!

Liyw979:
But in Js, We must do ```this.sink``` or ```bar.sink``` to do a dynamic call. Am I right ?

max-leuthaeuser:
Well, the semantics for `this` are rather complicated (https://www.w3schools.com/js/js_this.asp).
We did not want to rebuild all these rules here so the linking avoids looking at `this`.

See:
 - https://github.com/joernio/joern/blob/master/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/StaticCallLinker.scala
 - https://github.com/joernio/joern/blob/master/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/DynamicCallLinker.scala
 - https://github.com/joernio/joern/blob/master/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/JavascriptCallLinker.scala

Liyw979:
Ah, I mean that the ```sink(x)``` is not ```this.sink(x)``` or ```others.sink(x)``` so we can say  ```sink(x)``` is a static call.

max-leuthaeuser:
Hm yes, I guess we could do this. I'll test that in a separate PR.

Liyw979:
Thanks

max-leuthaeuser:
[...]

PR #1937, opened 2022-11-03
jssrc2cpg: set dispatch type for calls

max-leuthaeuser:
This is for: https://github.com/joernio/joern/issues/1895

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/RequirePass.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/RequirePass.scala
@@ -1,7 +1,6 @@
 package io.joern.jssrc2cpg.passes
 
 import io.shiftleft.codepropertygraph.Cpg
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/RequirePass.scala
```diff
@@ -89,11 +88,8 @@ class RequirePass(cpg: Cpg) extends SimpleCpgPass(cpg) {
   override def run(diffGraph: BatchedUpdate.DiffGraphBuilder): Unit = {
     cpg.call("require").where(_.inAssignment.target).map(Require.apply(_)).foreach { require =>
       require.methodFullName.foreach { fullName =>
-        cpg.method.fullNameExact(fullName).foreach { method =>
-          require.nodesToPatch.foreach { node =>
-            diffGraph.setNodeProperty(node, "METHOD_FULL_NAME", fullName)
-            diffGraph.addEdge(node, method, EdgeTypes.CALL)
```
max-leuthaeuser, line 95: The JavascriptCallLinker will do this for us now.

PR #1938, opened 2022-11-03
Use common fixture class in CFG tests of different frontend.

ml86:
While introducing this common class i switched the fixture style to the
new Code2CpgFixture.


max-leuthaeuser:
Then we also can remove stuff from https://github.com/joernio/joern/blob/master/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/AbstractPassTest.scala ?

max-leuthaeuser (approved):

PR #1940, opened 2022-11-03
Change METHOD_RETURN CODE property to RET in AstCreatorBase

johannescoetzee:
From a request by @ml86


max-leuthaeuser (approved):

max-leuthaeuser:
Might break some tests in codescience.

ml86:
Sorry for the misunderstanding. Changing the METHOD_RETURN CODE property for other frontends that php2cpg was not my intention. Espacially for frontends which actually have something meaningful like the return type to put as CODE property. Just for languages where there is actually nothing meaningful to put as CODE property, i would rather use `RET` instead of `ANY`.

johannescoetzee:
Ah, in that case I'll close this.

PR #1941, opened 2022-11-03
pysrc2cpg: Call `methodFullName` for Locally Defined or Imported Static Calls

DavidBakerEffendi:
- Modified import creation to add an optional third argument being the alias if one is present
- Split the Python call linker into a static and dynamic linker
  - The static linker runs first, gathering information about what static methods are in scope and modifying caller properties respectively
  - The dynamic linker then runs, linking calls with methods with the same name
- The static call linker generates full names to consider both named modules defined under Python files e.g. `foo/bar.py` and modules with `__init__.py` files, e.g., `foo/bar/__init__.py`. This is then checked against what exists in the CPG, and preference is given to named Python file definitions if both exist.
- `TODO` notes have been made to address some assumptions made in the process


ml86:
Changing the dispatch type of calls would break the closed source data flow tracker. Declaring calls as statically dispatched even if they really are not also deprives other users from a correct code representation. 

This PR is very similar to an issue raised for the jssrc2cpg frontend (https://github.com/joernio/joern/issues/1895). Did i miss some arrangements here? I would advocate for associating this kind of calls to methods in a python specific call linker.

ml86:
@DavidBakerEffendi Have there already been changes to the dispatch type which are already on master?
@DavidBakerEffendi Whether the method full name resolution looks good depends on the level of correctness you want to achieve. I assume now that you just want to be correct 99% of the time and under that assumption the PR is good. But it wont work if the code does things usually considered bad practice like 
```
def func():
  print("foo")

def bar():
  print("bar")
func = bar
func()
```
With this example in mind i want to reiterate my point:
Your 99% effort is good and reasonable but i would ask to do it in a manner which does not change the code representation in way which prevents others to implement a more refined call resolution which also catches the above example. 

The way to do it would be in a separate Python specific call linker (which i think we already have). And i know this would be the more cumbersome version because all the scoping context needs to be recalculated but thats seems to be the cleanest version to me.

DavidBakerEffendi:
@ml86 Right I get you, the soundness is important. So the direction would be not to rely on static properties and request callees on-the-fly or from a post-processing step. I'll see if I can reproduce these test assertions by following and making changes in the call linker pass.

fabsx00:
Does this only work well for Python? If it works well for multiple languages, then we should rather have a component that we use in all of those. And if there currently isn't a better or more correct way of implementing this, then I see no reason not to do this by default rather than having less full names.

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitorHelpers.scala
```diff
@@ -67,8 +68,13 @@ trait PythonAstVisitorHelpers { this: PythonAstVisitor =>
           createIdentifierNode("import", Load, lineAndCol),
           "import",
           lineAndCol,
-          nodeBuilder.stringLiteralNode(from, lineAndCol) ::
-            nodeBuilder.stringLiteralNode(alias.name, lineAndCol) :: Nil,
+          Seq(
+            nodeBuilder.stringLiteralNode(from, lineAndCol),
+            nodeBuilder.stringLiteralNode(alias.name, lineAndCol)
+          ) ++ (alias.asName match {
+            case Some(aliasName) => Seq(nodeBuilder.stringLiteralNode(aliasName, lineAndCol))
```
DavidBakerEffendi, line 75: Here we add an optional third argument, which aims to be the alias name. Should we make this a special named argument?

ml86: What do you mean by "special named"?

DavidBakerEffendi:
Added the capability to discern between named Python classes and `__init__` instances, e.g., `foo.py` and `foo/__init__.py`

fabsx00 (approved):

ml86:
I would very much argue to not use the term "static call" for the implementation of PythonStaticCallLinker in this PR.

It implies that it is responsible of linking static calls and thereby the group of calls it handles will be thought of as "static calls". This is dangerously close to the already existing concept of a statically dispatched calls which for abbreviation reasons so far have already been called "static call". This will lead to confusion so i would argue for a different name of this linker.

DavidBakerEffendi:
@ml86 From reading the other issue (#1895) it is mentioned that all calls in JS are dynamically dispatched. Is this the same in Python, except for those with the `@staticmethod` decorator? From Googling there seems to be no clear answer, but likely there exists some Python textbook somewhere to explain. In this PR, it is mostly assumed that if the call is off of a module, it is treated as a static dispatch, is this false?

ml86:
@DavidBakerEffendi So the dispatch type property of call nodes was introduced with the follow semantics:
STATIC_DISPATCH: The call method is know prior program execution time.
DYNAMIC_DISPATCH: The call method is determined at runtime by the call receiver.

Applied to Python this means all the `<operator>...` calls can be represented as STATIC_DISPATCH but all other calls are DYNAMIC_DISPATCH because they are all just calls via function/method pointers either stored in local variables or in some member/field variable. The `@staticmethod` does not make a difference, it just means that not `self` object is passed into the method.

Like we discussed, it is fine that you now write a pass which applies some heuristics to rewrite DYNAMIC_DISPATCH to STATIC_DISPATCH but i would like to reserve the term "static call" to the original scope. So i would suggest renaming `PythonStaticCallLinker` to `PythonDynamicToStaticCallRewriter`. I just fear that people will start thinking about those calls which can be rewritten by a proper heuristic to a 99% correct call target as "static calls" since that clashes with the original intend.

DavidBakerEffendi:
@ml86 Thanks for the answer! I understand now, will make a follow-up PR to address the correctness of what the pass aims to do.

PR #1946, opened 2022-11-04
Dataflow Engine: Slowdown When Gathering Large Number of Starting Points

DavidBakerEffendi:
A massive slowdown can be observed when performing the query `cpg.method("executeUpdate").methodReturn.reachableByFlows(cpg.literal).l` on [opencrx](https://github.com/opencrx/opencrx). I have not observed termination for this.

It was observed that a single thread gets stuck on `sourceTravsToStartingPoints` and expanding usages. This change restructures this as a `RecursiveTask` which forks into new tasks when obtaining usages. This has reduced runtime of the query above to about 1 min.


fabsx00 (approved):
Thanks! I will also look into optimizing the operation itself. This really shouldn't be a bottleneck, but is obviously is.

PR #1947, opened 2022-11-04
ossdataflowengine: do not solve the same task twice

fabsx00:
Detect when we are trying to solve a task again (with the same context), and if so, don't do it.

Also disables a flaky test.


Liyw979:
Haha, I was planning make a PR to do this.  I meet this problem when trying to find the flow of log4j vulnerability.

PR #1960, opened 2022-11-08
querydb: add new android query

claudiuvursache:


mpollmeier (approved):

querydb/src/main/scala/io/joern/querylib/Android.scala
```diff
@@ -0,0 +1,46 @@
...
+            .filter(_.label == "category")
+            .filter { node =>
+              val categoryName = node.attribute(Constants.androidUri, "name")
+              categoryName match {
+                case Some(n) => n.toString == "android.intent.category.DEFAULT"
+                case None    => false
+              }
+            }
+        }
+        if (validIntentFilters.nonEmpty) {
```
maltek, line 38: could change the last `.filter` into a `.exists` here

maltek (approved):

querydb/src/main/scala/io/joern/querylib/Android.scala
claudiuvursache: yess, much better, thx

PR #1962, opened 2022-11-08
PHP query init

itsacoderepo:
For something like https://github.com/brooksgarrett/dvwa/blob/master/vulnerabilities/exec/source/low.php
```
"""___________________________________________________________________________________________________________
| nodeType   | tracked              | lineNumber| method   | file                                          |
|==========================================================================================================|
| Call       | $target = $_REQUE... | 5         | <global> | /tmp/DVWA/vulnerabilities/exec/source/low.php |
| Identifier | $target = $_REQUE... | 5         | <global> | /tmp/DVWA/vulnerabilities/exec/source/low.php |
| Identifier | "ping -c 4 " . $...  | 14        | <global> | /tmp/DVWA/vulnerabilities/exec/source/low.php |
| Call       | "ping -c 4 " . $...  | 14        | <global> | /tmp/DVWA/vulnerabilities/exec/source/low.php |
""",

### DIFF
--- a/querydb/src/main/scala/io/joern/scanners/QueryTags.scala
+++ b/querydb/src/main/scala/io/joern/scanners/QueryTags.scala
@@ -23,5 +23,6 @@ object QueryTags {
   val insecureNetworkTraffic = "insecure-network-traffic"
   val pathTraversal          = "path-traversal"
   val cryptography           = "cryptography"
+  val remoteCodeExecution    = "remote-code-execution"
 
 }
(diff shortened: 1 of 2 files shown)


johannescoetzee (approved):
~Needs a scalafmt run~ Needed, but otherwise looks good. Just a note about PHP queries in general: at the moment a lot of type information is missing, which doesn't work very well with the dataflow engine. Purely AST based queries will be more useful than dataflow-based ones until this situation is resolved.

itsacoderepo:
> Just a note about PHP queries in general: at the moment a lot of type information is missing, which doesn't work very well with the dataflow engine. 

This is fine for now, i also haven't added any hard tests yet. Maybe you can make use of the queries and create some in the future as soon as the frontend is more stable :)

PR #1963, opened 2022-11-08
Add PHP anonymous functions

johannescoetzee:
This supports both the long-form syntax as well as arrow functions. Both are represented as a regular method belonging to the typeDecl they were defined in.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
@@ -1413,14 +1413,15 @@ class AstCreator(filename: String, phpAst: PhpFile, global: Global) extends AstC
 
     val localsForUses = expr.uses.flatMap { closureUse =>
       val variableAst = astForExpr(closureUse.variable)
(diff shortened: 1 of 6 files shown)


johannescoetzee:
Test failures are CFG related failures that seem to have been introduced by a rebase.

max-leuthaeuser (approved):

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
```diff
@@ -55,7 +55,7 @@ class AstCreator(filename: String, phpAst: PhpFile, global: Global) extends AstC
 
     val globalTypeDecl = NewTypeDecl()
       .name(namespaceBlock.name)
-      .fullName(namespaceBlock.fullName)
+      .fullName(namespaceBlock.name)
```
max-leuthaeuser, line 58: This is intended?

johannescoetzee: It is. The namespaceBlock fullname contains too much information (the filename, for example). Whether this is correct is a different discussion, but for the purposes of this PR the name is actually what we want.

PR #1964, opened 2022-11-09
pysrc2cpg: Refactor Call Linkers

DavidBakerEffendi:
- No longer change calls to STATIC_DISPATCH
- Renamed call linkers to `PythonModuleDefinedCallLinker` and `PythonNaiveCallLinker` to reflect their linking strategies


ml86 (approved):
Thank you. That is much more self descriptive.

PR #1965, opened 2022-11-09
jssrc2cpg: Link anonymous function exports

### LINKED ISSUE #1914: jssrc2cpg: Require statements do not capture external module names
JavaScript frontend does not capture imported module names from external dependencies. Below is a test case one can use with proposed outputs. Right now, both `express` and `app` return `ANY` at this point.

```javascript
  "external modules imported via `require` should be resolved correctly" in {
    val cpg = code(
      """
        |const express = require('express')
        |const app = express()
      """.stripMargin,
      "sample.js"
    )

    implicit val callResolver: NoResolve.type = NoResolve
    cpg.identifier("express").typeFullName.l shouldBe List("express")
    cpg.identifier("app").typeFullName.l shouldBe List("express::constructor")
  }
 ```

### PR DESCRIPTION
**Pull request by DavidBakerEffendi**
There is a disconnect with `module.export` functions where the key is named and there is a lambda on the right (no name present). For example, calls to the following will not be successfully linked:

```javascript
module.exports = {
  sayhi: function() {            // this will be called anonymous
   console.log("Hello World!");
 },
  saybye: function() {           // this will be called anonymous1
   console.log("Good-bye!");
 }
}
```

The above is example can be found in test cases and the following changes are made:
- Set `methodFullName` property on call nodes when they are linked
- Set `typeFullName` of identifier receiving the result of require to the path with `<export>::` prefix
- Using details from the above, will be able to resolve an exported function without an identifier

Resolves #1914

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/RequirePass.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/RequirePass.scala
@@ -1,13 +1,14 @@
 package io.joern.jssrc2cpg.passes
 
 import io.shiftleft.codepropertygraph.Cpg
-import io.shiftleft.codepropertygraph.generated.EdgeTypes
-import io.shiftleft.codepropertygraph.generated.nodes.Call
+import io.shiftleft.codepropertygraph.generated.{EdgeTypes, PropertyNames}
+import io.shiftleft.codepropertygraph.generated.nodes.{Call, Local}
 import io.shiftleft.passes.SimpleCpgPass
 import overflowdb.BatchedUpdate
 import io.shiftleft.semanticcpg.language._
-import overflowdb.traversal.{NodeOps, Traversal}
+import overflowdb.traversal.{NodeOps, Traversal, iterableToTraversal}
 
+import java.io.File
 import java.nio.file.Paths
 
 /** This pass enhances the call graph and call nodes by interpreting assignments of the form `foo = require("module")`.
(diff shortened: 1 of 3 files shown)


max-leuthaeuser (approved):

max-leuthaeuser:
Very nice!

DavidBakerEffendi:
@max-leuthaeuser Thanks for the review, while falling asleep I realized that this is unsound in that it assumes that a variable that receives data from a `require` will not be reassigned. I've gone ahead and adapted the implementation to make use of `DYNAMIC_TYPE_HINT_FULL_NAME` instead and have added a test where a mutable variable is assigned to different requires where functions of the same name are exported, and thus the call could be to either.

max-leuthaeuser:
Do we need a special handling for `const` variables where this cannot happen?

DavidBakerEffendi:
> Do we need a special handling for `const` variables where this cannot happen?

This can be implemented, is there already some kind of modifier or node indicating immutability? 

Though if it is valid JavaScript and the `const` variable is only assigned once, then it will only have 1 call and the `METHOD_FULL_NAME` will be set to that method. Only if there are more than 1 call edges is it set to `<unknownFullName>`.

DavidBakerEffendi:
Good idea, then in the case of a const shall I set the type under TYPE_FULL_NAME and otherwise under DYNAMIC_TYPE_HINT_FULL_NAME? I can make that change before merging.

max-leuthaeuser:
Sounds reasonable.

DavidBakerEffendi:
Okay, the tests have been adjusted to check which type info is set and where depending on mutability of the variable declared by a `require`.

PR #1966, opened 2022-11-09
Implement own CharStreamImpl for JavaCC based Python parser.

ml86:
This became necessary since we used the JavaCC provided default
CharStream implementation in a hackish way which caused exceptions on
some input files.


mpollmeier (approved):
I guess you checked and double checked if anything like this is available in (standard) libraries..?

PR #1975, opened 2022-11-13
Add missing symlinks to joern-install.sh

VainlyStrain:
Add symlinks for the Python, Java, Kotlin, PHP and Jimple CPG construction tools to joern-install.sh, so that these tools can be called without needing to specify the path in a script.

If these symlinks have been omitted on purpose, feel free to ignore this pull request.


mpollmeier:
sorry about the delay, yes I believe these were missing indeed
the PR build is red due to a timeout, will restart
thank you!

mpollmeier (approved):

PR #1983, opened 2022-11-15
querydb: add bundle `android.JavaScriptInterface`

claudiuvursache:


claudiuvursache:
really need to find a good abstraction for building the code examples so that PRs don't get as large...

querydb/src/main/scala/io/joern/querydb/language/android/ManifestXmlTraversal.scala
```diff
@@ -10,6 +10,23 @@ object Constants {
...
+      .filter(_.label == "manifest")
+      .flatMap(_.child)
+      .filter(_.label == "application")
+      .flatMap { applicationNode =>
+        val activityName = applicationNode.attribute(Constants.androidUri, "usesCleartextTraffic")
+        activityName match {
+          case Some(n) if n.toString == "true" => Some(true)
+          case Some(_)                         => Some(false)
+          case None                            => None
+        }
```
maltek, line 27:
```suggestion
        activityName.map(_.toString == "true")
```

querydb/src/main/scala/io/joern/scanners/android/JavaScriptInterface.scala
```diff
@@ -0,0 +1,518 @@
...
+        def exposedJavaScriptInterfaceObjects =
+          if (appUsesCleartextTraffic)
+            webViewsWithInsecureLoadUrlCalls.referencingIdentifiers.inCall
+              .nameExact("addJavascriptInterface")
+              .argument(1)
+          else Traversal.empty
+        val exposedJavaScriptInterfaceObjectNames =
+          exposedJavaScriptInterfaceObjects.collect {
+            case expression: Identifier => expression.typeFullName
+            case expression: Call       => expression.typeFullName
```
maltek, line 65:
```suggestion
            case ident: Identifier => ident.typeFullName
            case call: Call        => call.typeFullName
```

since `Expression` is a node type, I think it's preferable to not use it as a variable name for a different type (even if it's a sub-type).

maltek (approved):
just some small nits

querydb/src/main/scala/io/joern/scanners/android/JavaScriptInterface.scala
claudiuvursache: yepp, nicer, thx

PR #1984, opened 2022-11-15
Use more caching to speed up javasrc2cpg for large projects

johannescoetzee:
The biggest performance gains from this PR are that `solveType` calls resulting in SOEs are now cached. There is still a WIP replacement for the JavaParserTypeSolver (that gets type information from source files) coming that should give much better results, but for now this reduced the AST creation time for elasticsearch to about 11 minutes on my workstation.


johannescoetzee:
Closing in favour of https://github.com/joernio/joern/pull/1986

PR #2007, opened 2022-11-21
kotlin2cpg: use imports to set unresolved types

claudiuvursache:


max-leuthaeuser (approved):

johannescoetzee (approved):

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/DefaultTypeInfoProvider.scala
```diff
@@ -743,6 +754,13 @@ class DefaultTypeInfoProvider(environment: KotlinCoreEnvironment) extends TypeIn
       }
       .getOrElse(defaultValue)
   }
+  def typeFromImports(name: String, file: KtFile): Option[String] = {
+    file.getImportList.getImports.asScala.flatMap { directive =>
+      if (directive.getImportedName != null && directive.getImportedName.toString == name.stripSuffix("?"))
```
johannescoetzee, line 759: Small style nit (and it's debatable whether this is better), but you can do `Option(directive.getImportedName).contains(name.stripSuffix("?"))` here

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/types/DefaultTypeInfoProvider.scala
```diff
@@ -743,6 +754,13 @@ class DefaultTypeInfoProvider(environment: KotlinCoreEnvironment) extends TypeIn
...
+        Some(directive.getImportPath.getPathStr)
+      else None
```
johannescoetzee, line 761: The `if (condition) { Some(thing) } else { None }` pattern can also be rewritten as `Option.when(condition) { thing }`

claudiuvursache: same as above

PR #2014, opened 2022-11-23
upgrade to Scala 3

mpollmeier:
The biggest visible change is the repl, which is a complete reimplementation. 
Ammonite is no longer in our dependency tree, and I factored out the core repl
functionality into a separate build: https://github.com/mpollmeier/scala-repl-pp/

The repl options and usage are identical though, apart from an additional `--dependency` 
option which I'll document on joern-website once this is merged. 

The main downstream builds are ready to get upgraded as well. 

Note: the code is ready to get reviewed, however we won't merge anything until we're 
happy with the feedback from the manual user testing.


joern-cli/src/main/resources/scripts/c/malloc-overflow.sc
```diff
@@ -1,18 +1,21 @@
...
-    }
-    .l
+    }.code
+
+  val expected = Set(
+    "malloc(sizeof(int) * 42)",
+    "malloc(sizeof(int) * 3)",
+    "malloc(sizeof(int) + 55)"
+  )
+  assertContains("calls", calls, expected)
```
ml86, line 20: Did this turn into some kind of test? Before we did not assert anything. Did the use of this script change. Some question applies to the other `.sc` files.

benchmarks/build.sbt
```diff
@@ -16,12 +14,8 @@ Compile / doc / sources ~= (_ filter (_ => false))
...
 // we want to consume this from a java8 build
 compile / javacOptions ++= Seq("--release", "8")
-scalacOptions ++= Seq() ++ (
-  CrossVersion.partialVersion(scalaVersion.value) match {
-    case Some((3, _)) => Seq("-Xtarget:8")
-    case _            => Seq("-target:jvm-1.8")
-  }
-)
+
+scalacOptions += "-Xtarget:8"
```
ml86, line 18: Cant we set this for all subprojects in the main `build.sbt`?

joern-cli/src/main/resources/scripts/c/malloc-overflow.sc
ml86: I just noticed that those expectation came from `RunTestScripts.scala`. Why did you change the setup? Having the tests in the actual implementation seems strange.

macros/src/main/scala-3/io/joern/macros/QueryMacros.scala
```diff
@@ -10,13 +10,14 @@ import scala.quoted.{Expr, Quotes}
...
+    travExpr: Expr[Cpg => Traversal[_ <: StoredNode]]
+  )(using quotes: Quotes): Expr[TraversalWithStrRep] = {
     import quotes.reflect._
-    val pos = travExpr.asTerm.pos
+    val pos  = travExpr.asTerm.pos
     val code = Position(pos.sourceFile, pos.start, pos.end).sourceCode.getOrElse("N/A")
-    '{TraversalWithStrRep(${travExpr}, ${Expr(code)})}
+    '{ TraversalWithStrRep(${ travExpr }, ${ Expr(code) }) }
   }
 }
```
ml86, line 23: There is a scala 2 version of the `QueryMacros` implementation which i would have expected to be deleted. The file location can than also be adjusted (no more `scala-3` in path).

benchmarks/build.sbt
mpollmeier: good idea, done

macros/src/main/scala-3/io/joern/macros/QueryMacros.scala
mpollmeier: good point, done

PR #2015, opened 2022-11-23
Create CFG for match expressions

johannescoetzee:


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/cfgcreation/CfgCreator.scala
```diff
@@ -536,6 +538,48 @@ class CfgCreator(entryNode: Method, diffGraph: DiffGraphBuilder) {
...
+            val nodeCfg = cfgFor(node)
+            Cfg.empty :: (currCfg ++ nodeCfg) :: prevCfgs
+        }
+      }
+      .reverse
+  }
+
+  /** CFG creation for match expressions of the form `match { case condition: expr ... }`
+    */
+  protected def cfgForMatchExpression(node: ControlStructure): Cfg = {
```
max-leuthaeuser, line 564: Isn't that mostly like cfgForSwitchStatement?

max-leuthaeuser (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/cfgcreation/CfgCreator.scala
johannescoetzee: They are indeed very similar. I've refactored cfgForMatch/Switch to reduce code duplication.

PR #2018, opened 2022-11-24
WIP add (failing) test to verify that two calls are part of the DDG

mpollmeier:
I'm not the best person to judge wether those calls really _should_
be part of the DDG, but from our discussion on
[discord](https://discord.com/channels/832209896089976854/832209896089976857/1044687371800678420)
it sounds like the answer is 'yes'.
Anywhere, since I was asked about this I just added a (failing) test
to hopefully make it easier to grok the context and take it from
there.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/io/dotgenerator/DotDdgGeneratorTests.scala
+++ b/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/io/dotgenerator/DotDdgGeneratorTests.scala
@@ -3,7 +3,8 @@ package io.joern.c2cpg.io.dotgenerator
 import io.joern.c2cpg.testfixtures.DataFlowCodeToCpgSuite
 import io.joern.dataflowengineoss.dotgenerator.DdgGenerator
 import io.joern.dataflowengineoss.language._
-import io.shiftleft.codepropertygraph.generated.nodes.Call
+import io.shiftleft.codepropertygraph.Cpg
+import io.shiftleft.codepropertygraph.generated.nodes.{AstNode, Call, CfgNode, Method}
 import io.shiftleft.semanticcpg.language._
 
 class DotDdgGeneratorTests extends DataFlowCodeToCpgSuite {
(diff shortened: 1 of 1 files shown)


max-leuthaeuser:
Thanks! I also asked @bbrehm about this.

mpollmeier:
fwiw here's the dot graph:

ddg-minified

[ddg-minified.dot.txt](https://github.com/joernio/joern/files/10086159/ddg-minified.dot.txt)

bbrehm:
I think this is a question of what people mean with ddg?

Cursory googling reveals two conflicting definitions that appear to be in common use.

That is, consider the program
```
void foo(int* x){
print(*x);
*x = 0;
}
```
Following e.g. https://home.cs.colorado.edu/~kena/classes/5828/s99/lectures/lecture25.pdf slide 19, the first statement (the memory load) and the second (the memory store) would have a data dependency from load to store (basically because the two statements cannot be reordered). Following that definition, you would expect a dependency in your example.

Following instead the definition in https://compilers.cs.uni-saarland.de/teaching/spa/2014/slides/ProgramDependenceGraph.pdf slide 12, there is no data dependence: The two statements cannot affect each other (and the superscalar CPU can execute them in parallel). Following that definition, you would expect no dependency in your program.

So I guess this is a documentation issue?

I don't know, I am a mathematician, not a computer scientist. I don't particularly care about official definitions, we just should be clear in the docs what we compute.

FWIW, I think the concept that has no data dependencies here is much more useful, because it behaves pretty well under SSA and mem2reg like compiler transforms; nobody cares whether you can reorder lines of code, the question is what dependencies the CPU sees and which ones it can kill via register renaming.

bbrehm (approved):
From my point of view, we could merge this. What do you think @max-leuthaeuser @mpollmeier ?

mpollmeier:
Sounds good to me - certainly better and clearer than before, thanks @bbrehm 
As discussed on discord, hopefully @fabsx00 can add some more clarity and direction, once he's back.

max-leuthaeuser:
Looks good, thanks!

PR #2023, opened 2022-11-25
kotlin2cpg: inner class naming

BFergerson:
I wanted to try contributing a fix but got stuck. The following code isn't correct but does solve the errornous inner class naming issue present in kotlin2cpg. Inner classes should follow `Outer$Inner` naming but currently Joern names as `Outer.Inner`.


claudiuvursache:
your contribution is much appreciated @BFergerson! i am in the middle of a different task at the moment, but will pick up where you left off soonish and ping when ready

claudiuvursache:
i have a fix up fix @ https://github.com/joernio/joern/pull/2024 @BFergerson
i've refactored the `TypeRenderer` a bit, and the fullnames of inner classes now follow the expected convention https://github.com/joernio/joern/pull/2024
thanks again for the prompt

PR #2042, opened 2022-11-30
Perform set lookup in parallel for started tasks.

fabsx00:


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
```diff
@@ -48,8 +48,8 @@ class Engine(context: EngineContext) {
   private val completionService =
     new ExecutorCompletionService[TaskSummary](executorService)
 
-  private val started: mutable.Set[TaskFingerprint] = mutable.Set()
-  private var held: List[ReachableByTask]           = List()
+  private val started: mutable.Set[(CfgNode, List[CfgNode])] = mutable.Set()
```
DavidBakerEffendi, line 51: Is this not just the `TaskFingerprint` class now? Why the use of tuple?

fabsx00: True, let me get the task fingerprint back in. It's more descriptive.

DavidBakerEffendi: Once resolved and the merge conflict is sorted, then it's approved for me 👍

PR #2061, opened 2022-12-06
A fix for handling of METHOD_RETURN

fabsx00:
The `METHOD_RETURN` node currently represents two things: the formal return parameter and the exit node. This is problematic because the semantics of using `METHOD_RETURN` nodes as sources/sinks in data flow queries are not clear.  I'm proposing the following semantics and adapting the tests accordingly:

`METHOD_RETURN` represents the exit node, not the formal return parameter. In consequence:

- METHOD_RETURN as a sink means: flows to the exit node
- METHOD_RETURN as a source is not allowed, because in a DDG, all nodes eventually flow to the exit node, and this is unlikely what a user is looking for. For backward compatibility, we allow it anyway and translate METHOD_RETURN sources into corresponding CALL nodes.
- To use all return values of `foo` as a source, we use `cpg.call("foo")` as the CALL node represents the actual output parameter.
- To identify all flows to returns, one uses `RETURN` nodes as sinks

With these rules in place, when the interprocedural data flow engine reaches a CALL node in backward tracking, it is clear that it should traverse to the corresponding RETURN nodes, and not to the METHOD_RETURN node.

*Change in behavior* If there are no RETURN nodes inside the method with a given METHOD_RETURN, then we should not just expand to the METHOD_RETURN node as this would mean expanding to the exit node of the method. I believe this is a bug fix, but it may now result in fewer (invalid) flows.

This PR makes these adaptions.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
@@ -13,7 +13,6 @@ import overflowdb.traversal.{NodeOps, Traversal}
 
 import scala.collection.parallel.CollectionConverters._
 import java.util.concurrent._
-import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 import scala.util.{Failure, Success, Try}
 
@@ -85,7 +84,6 @@ class Engine(context: EngineContext) {
     def handleSummary(taskSummary: TaskSummary): Unit = {
       val newTasks = taskSummary.followupTasks
       submitTasks(newTasks, sources)
-      val task       = taskSummary.task
       val newResults = taskSummary.results
       completedResults ++= newResults
     }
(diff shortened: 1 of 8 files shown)


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
```diff
@@ -120,6 +136,8 @@ class Engine(context: EngineContext) {
 
 object Engine {
 
+  private val logger: Logger = LoggerFactory.getLogger(Engine.getClass)
```
max-leuthaeuser, line 139: This is unused in `object Engine` is'nt it?

DavidBakerEffendi (approved):
Looks good, but in general, this PR addresses an additional change to the cache which should maybe also be reflected in the PR description. Suppose the unused `table` is simply removed, but good to have that documented here.

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
fabsx00: Correct, let me remove it.

ml86:
I think we should not give up the possibility to represent the formal method return value via one note. How about we just split METHOD_RETURN into two nodes, one for the CFG exit and one for the formal method return?

PR #2075, opened 2022-12-08
Fix semantics for assignment and post/pre increment

### LINKED ISSUE #1921: Dataflow semantic of assignments
Hi, I just noticed that data does not propagate over multiple assignments, e.g., `a = b = c`.  The dataflow semantic for assignments should include the return value as output.

fabsx00:
Fix #1921


tuxology (approved):

max-leuthaeuser:
```
[info] - should find that there is no flow from `y = 1` to exit node *** FAILED ***
[info]   1 was not equal to 0 (DataFlowTests.scala:1207)
```

That's the failing test w.r.t. the return node problem I mentioned.

fabsx00:
Yep, just fixed that. It's in fact correct. There's a follow-up PR now that I'm working on that addresses the return value handling across functions.

PR #2080, opened 2022-12-09
[jssrc2cpg] Built-in Types Now Set as External Types

### LINKED ISSUE #2079: [jssrc2cpg] ECMA Types are Detected as Internal Types
Types such as:
* `ANY`
* `__ecma.Number`
* `__ecma.String`

Contain `.isExternal = false` property.

DavidBakerEffendi:
When built-in types are created, they are set as external, to be consistent with other frontends.

Resolves #2079


max-leuthaeuser (approved):

max-leuthaeuser:
`.isExternal(false)` was the behavior in js2cpg.
This PR will change sptests results (once jssrc2cpg-internal is bumped). We need to check if there is anything dependent in codescience.

DavidBakerEffendi:
@max-leuthaeuser So shall we wait on merging this to avoid unintended side-effects?

max-leuthaeuser:
Could not find anything problematic there.
pysrc2cpg also sets ANY to external=false. I wonder whats the initial reason for doing so.
@ml86 Any idea? Any objection to this PR here?

ml86:
Setting those types to external makes sense. So no objects from my side.

PR #2081, opened 2022-12-09
Fix interprocedural tracking via METHOD_RETURN nodes

fabsx00:
The tests illustrate the problem quite nicely: because `METHOD_RETURN` is also the exit node, we ended up overtainting in the past. We revert to this behavior for external methods and method stubs. For internal methods, we traverse to RETURN nodes instead.


DavidBakerEffendi (approved):
Nice! Curious to see what the results are with `joern-benchmark` at this point, since I did realize that whether a sink/source is `.methodReturn` or `.return` there would be different results.

PR #2090, opened 2022-12-09
c2cpg: get rid of absolut paths

max-leuthaeuser:
For: https://shiftleftinc.atlassian.net/browse/SEN-300


itsacoderepo (approved):

max-leuthaeuser:
Works nicely. Some time ago I made .dump and .dumpRaw robust against both relative (with the metadata root field) and absolute paths.

PR #2093, opened 2022-12-11
[CfgNodeTraversal] Match Error on MethodParameter_

DavidBakerEffendi:
When traversing to the method from a `MethodParameterIn` node as a `CfgNode` I get a match error. I have added this default case, which will then refer the traversal to `io.shiftleft.semanticcpg.language.nodemethods.CfgNodeMethods.method` that handles the `Method[Return | Parameter[In | Out]]` case

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/generalizations/CfgNodeTraversal.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/generalizations/CfgNodeTraversal.scala
@@ -1,6 +1,5 @@
 package io.shiftleft.semanticcpg.language.types.expressions.generalizations
 
-import io.shiftleft.Implicits.JavaIteratorDeco
 import io.shiftleft.codepropertygraph.generated.nodes._
 import io.shiftleft.codepropertygraph.generated.{EdgeTypes, NodeTypes}
 import io.shiftleft.semanticcpg.language._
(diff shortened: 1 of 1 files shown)


mpollmeier:
You also dropped the `case methodReturn` - is that because it's the same as the new default?
I'd like to get @ml86's blessing for this.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/generalizations/CfgNodeTraversal.scala
```diff
@@ -22,14 +22,13 @@ class CfgNodeTraversal[A <: CfgNode](val traversal: Traversal[A]) extends AnyVal
...
         method
-      case methodReturn: MethodReturn =>
-        methodReturn.method
       case expression: Expression =>
         expression.method
       case callRepr: CallRepr =>
         callRepr._astIn.onlyChecked.asInstanceOf[Method]
       case jumpTarget: JumpTarget =>
         jumpTarget.method
+      case x => x.method
```
mpollmeier, line 31:
```suggestion
      case cfgNode => cfgNode.method   // refers to `semanticcpg.language.nodemethods.CfgNodeMethods.method`
```

ml86:
The fact that those two methods have slightly different semantics in the first place is problem here. The version of the method for the traversal should just call the version on the node itself to ensure the same semantics: `traversal.map(_.method)`.
`

DavidBakerEffendi:
@ml86 So if I understand correctly, the method targeted in the change should actually just pass off the call as is done in the default case?

DavidBakerEffendi:
@ml86 Change made 👍

ml86 (approved):

PR #2094, opened 2022-12-12
CodeDumper: safer filename calculation

max-leuthaeuser:
For methods generated by the MethodStubCreator (e.g., for operators) we end up with artificial method nodes in the `<global>` namespace and file but without line numbers. The fake filename (`<global>`) would crash the filename calculation here. So I moved that piece of code where it is guarded by lineNumber existence.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
@@ -34,12 +34,6 @@ object CodeDumper {
         logger.info(s"dump not supported for language '$lang'")
         ""
       case (Some(node), Some(lang)) =>
-        val filename = location.filename match {
-          case f if Paths.get(f).isAbsolute =>
-            f
-          case f =>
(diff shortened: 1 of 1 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
```diff
@@ -49,9 +43,12 @@ object CodeDumper {
         method
           .collect {
             case m: Method if m.lineNumber.isDefined && m.lineNumberEnd.isDefined =>
-              val rawCode = if (lang == Languages.GHIDRA) {
-                m.code
-              } else {
+              val rawCode = if (lang == Languages.GHIDRA) { m.code }
```
mpollmeier, line 46: that indentation seems a little off and doesn't help readability - I guess scalafmt will complain as well...?

mpollmeier (approved):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
max-leuthaeuser: Its fine for scalafmt.

PR #2095, opened 2022-12-12
c2cpg: fix relative/absolute path handling

max-leuthaeuser:
It is relative for project files but absolute for system header files.


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -48,8 +48,11 @@ trait AstCreatorHelper { this: AstCreator =>
   }
 
   private def fileOffsetTable(node: IASTNode): Array[Int] = {
-    val f = fileName(node)
-    file2OffsetTable.computeIfAbsent(f, _ => genFileOffsetTable(Paths.get(config.inputPath, f)))
+    val path = fileName(node) match {
```
mpollmeier, line 51:
```suggestion
    // file path is relative for project files but absolute for system header files
    val path = fileName(node) match {
```

mpollmeier (approved):

PR #2096, opened 2022-12-12
Thread optimisations

pandurangpatil:


.gitignore
```diff
@@ -29,3 +29,6 @@ joern-cli/workspace
 **/astgen-linux
 **/php2cpg/bin/vendor/
 **/php2cpg/bin/composer*
+
+
+.DS_Store
```
mpollmeier, line 34: this doesn't hurt but should rather go into your `~/.gitignore`, since it's specific to your personal setup, in this case MacOS.

pandurangpatil: This is removed.

fabsx00 (approved):

fabsx00:
I will merge this first and then apply some cleanups in a follow-up PR.

Issue #2086, opened 2022-12-09
jssrc: invalid desugaring for spread syntax in object literals


maltek:
example of actual behavior from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals

```javascript
const array = [1, 2, 3];
const obj = { ...array }; // { 0: 1, 1: 2, 2: 3 }
```

i.e. it does a shallow copy of `array` into the resulting object. The only real way to model that in the CPG is as a call to an operator (or possibly `Object.assign` which has the same semantics, but I'd prefer something like `<operator>.shallowCopy` or `<operator>.spreadObject`).

------

Currently, we try to do weird things based on the name of the variable that's being spread:

```
joern> importCode.jssrc.fromString("""({...bar})""") 
res80: Cpg = Cpg (Graph [55 nodes])

joern> cpg.call.code.p 
res81: List[String] = List("_tmp_0.bar = bar", "_tmp_0.bar")
```

and even more wrong:

```
joern> importCode.jssrc.fromString("""({...foo.bar})""") 
res78: Cpg = Cpg (Graph [58 nodes])

joern> cpg.call.code.p 
res79: List[String] = List("_tmp_0._tmp_1 = foo.bar", "_tmp_0._tmp_1", "foo.bar")
```

max-leuthaeuser:
This is not an easy one. 
Currently, we desugar ObjectExpressions like this in a general case:
```javascript
const x = {
  a,
  b,
  ...rest
}
// => desugaring:
const x = {
  var tmp;
  tmp.a = a;
  tmp.b = b;
  tmp.rest = rest;
  tmp;
}
```
which is wrong for `...rest` as you described. We simply handle it here as it would be a normal identifier. You suggest to add a specific call `<operator>.spreadObject`?
```javascript
const x = {
  var tmp;
  tmp.a = a;
  tmp.b = b;
  tmp.rest = CALL(name=<operator>.spreadObject) with arg 1: IDENTIFIER(name=rest);
  tmp;
}
```
?

maltek:
Ah, yes I didn't think about how these expressions can be combined this with a regular object literal. This complicates things... Apparently you can have an arbitrary number of spreads and explicit properties and in case of conflicts the last value wins, i.e.

```javascript

const rest1 = {
  "a": "rest1",
  "b": "rest1",
  "c": "rest1",
};

const rest2 = {
  method() {
    return { "b": "rest2" }
[...]

PR #2101, opened 2022-12-13
jssrc2cpg: fix for desugaring of rest and spread elements

max-leuthaeuser:
Fixes: https://github.com/joernio/joern/issues/2086

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForPrimitivesCreator.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForPrimitivesCreator.scala
@@ -4,7 +4,6 @@ import io.joern.jssrc2cpg.parser.BabelNodeInfo
 import io.joern.jssrc2cpg.passes.Defines
 import io.joern.x2cpg.Ast
 import io.shiftleft.codepropertygraph.generated.DispatchTypes
-import io.shiftleft.codepropertygraph.generated.nodes.NewIdentifier
 
 trait AstForPrimitivesCreator { this: AstCreator =>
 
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForPrimitivesCreator.scala
```diff
@@ -35,10 +34,16 @@ trait AstForPrimitivesCreator { this: AstCreator =>
...
+  protected def astForSpreadOrRestElement(spreadElement: BabelNodeInfo, arg1Ast: Option[Ast] = None): Ast = {
+    val ast = astForNodeWithFunctionReference(spreadElement.json("argument"))
+    val callNode = createCallNode(
+      spreadElement.code,
+      "<operator>.starredUnpack",
+      DispatchTypes.STATIC_DISPATCH,
+      spreadElement.lineNumber,
+      spreadElement.columnNumber
+    )
+    callAst(callNode, arg1Ast.toList :+ ast)
```
maltek, line 46: I'm not sure when this would even happen, but if this is called with `None` for `arg1Ast`, will the argument indices still be correct? (i.e. will the object that's being spread still have argument index 2?)

maltek (commented):
`<operator>.starredUnpack` is a bit weird in JavaScript-land because it doesn't use any asterisks for this operator. My preference would be either something generic (e.g. `<operator>.shallowCopy`, ideally also for Python) or something that makes sense for JS (e.g. `<operator>.spread`). But I don't feel like bikeshedding on naming too much, so just take this as a comment and not a veto.

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForPrimitivesCreator.scala
max-leuthaeuser: `.toList` will result in an empty List for `None` so this should work.

max-leuthaeuser:
@ml86 What do you thing about the name?
Should we simply keep `<operator>.starredUnpack` to be consistent even if it does not fit the JS world.
Or should we use a JS specific one like `<operator>.spread`? Or even a completely new one for both Python and JS?

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForPrimitivesCreator.scala
maltek: yes, it won't cause an error. My question is if the argument indices would change in that case. Because that would be a problem when writing policies.

max-leuthaeuser: No, it would not. Its an empty List not a List(Ast()).

ml86:
I think for now it is better to have a name fitting the language. So for JS `...spread`. In case we will treat this in a special way in the future we can change it to something more general.

PR #2107, opened 2022-12-15
c2cpg: cleanups and refactorings

max-leuthaeuser:
Mostly for relative/absolute path handling (left-overs from last PRs on that)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/utils/IOUtils.scala
```diff
@@ -1,11 +1,28 @@
...
 
 object IOUtils {
 
+  def toAbsolutePath(path: String, config: Config): String = path match {
+    // the path may be relative for project files but absolute for system header files
+    case f if Paths.get(f).isAbsolute => Paths.get(f).toString
+    case f                            => Paths.get(config.inputPath, f).toString
+  }
+
+  def toRelativePath(path: String, config: Config): String = {
```
claudiuvursache, line 16: this could live in x2cpg pbb

PR #2112, opened 2022-12-16
Parallel execution of tasks

rahul-privado:
### Description
1. Changes for parallel processing to execute tasks
2. Parallel processing for dedup of completed results

Used future for every set of subtasks. Each future will execute of the default fork-join pool.

### Testing done
1. Tested for reduction in execution time. Alovoa finishes in 2 mins on my 10 core 32 GB RAM machine.
2. Compared flow counts and paths with the HEAD of master without my changes. Verified that they match.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
@@ -13,9 +13,11 @@ import overflowdb.traversal.{NodeOps, Traversal}
 
 import scala.collection.parallel.CollectionConverters._
 import java.util.concurrent._
-import scala.collection.mutable
 import scala.jdk.CollectionConverters._
-import scala.util.{Failure, Success, Try}
+import scala.concurrent.{Future}
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
 
 case class ReachableByTask(
   sink: CfgNode,
@@ -37,14 +39,8 @@ class Engine(context: EngineContext) {
   import Engine._
 
   private val logger: Logger                   = LoggerFactory.getLogger(this.getClass)
-  private var numberOfTasksRunning: Int        = 0
-  private val executorService: ExecutorService = Executors.newWorkStealingPool()
-  private val completionService =
-    new ExecutorCompletionService[TaskSummary](executorService)
 
-  def shutdown(): Unit = {
-    executorService.shutdown()
-  }
+  def shutdown(): Unit = {}
 
   /** Determine flows from sources to sinks by exploring the graph backwards from sinks to sources. Returns the list of
     * results along with a ResultTable, a cache of known paths created during the analysis.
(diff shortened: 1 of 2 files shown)


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/TaskSolver.scala
```diff
@@ -1,13 +1,55 @@
...
+      return
+    }
+
+    val totalTasks     = totalTaskCounter.get()
+    val doneTasks      = doneTaskCounter.get()
+    val futuresStarted = futuresStartedCounter.get()
+    val futuresEnded   = futuresEndedCounter.get()
+    val backlog        = futuresStarted - futuresEnded
+
+    println(
```
pandurangpatil, line 41: We need to convert these prints to trace logs.

rahul-privado: Converted to debug logs

rahul-privado: They print the progress and give visibility on what is happening. That is why I thought they should be debug logs. Please let me know if you still think they should be trace

fabsx00:
Thanks! I'll bring this in today/tomorrow.

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
```diff
@@ -73,45 +68,61 @@ class Engine(context: EngineContext) {
...
+    val startTimeSec: Long = System.currentTimeMillis / 1000
     submitTasks(tasks.toVector, sources)
-    runUntilAllTasksAreSolved()
-    deduplicate(completedResults.toVector)
-  }
 
-  private def submitTasks(tasks: Vector[ReachableByTask], sources: Set[CfgNode]) = {
-    numberOfTasksRunning += tasks.size
-    tasks.foreach(t => completionService.submit(new TaskSolver(t, context, sources)))
+    do {
```
fabsx00, line 102: Do I understand correctly that we're waiting for 1 second in a loop here and polling for results? Is there maybe a more elegant way where we block until all calculations are done?

rahul-privado: We are polling the counters every second to indicate whether all jobs are done. 

A more elegant way could be to use semaphores, condition variables or barriers - depending on which ones are available in Scala. Signal task completion from futures and wake up the main thread for exit after all tasks are done. This will involve some complex logic to achieve sync but I thought that wasn't necessary since this rudimentary way also does the job.

fabsx00: Ok. That means though that every `reachableBy` invocation takes at least a second, even if it is a very simple query. I can imagine we'd notice this during interactive use, so, we may not be able to get around building something a bit more evolved. Let me think about it.

rahul-privado: I have reduced the poll time to 100 milliseconds

rahul-privado: @fabsx00 

I find that the number of tasks and flows is 70-80% higher on this branch ( with your changes ) compared to `master`

rahul-privado: @fabsx00 

I have merged latest master into this branch. The result table is now the critical section and is under lock. I see that it has a higher locking time causing threads to get stuck on it. Is there a way to reduce the locking time?

I am referring to

`Engine.scala -> addResultsToMainTable() -> synchronized(mainResultTable.add(parentTask, Vector(r.copy(taskStack = newTaskStack, path = newPath))))`

rahul-privado:
> ### Description
> 1. Changes for parallel processing to execute tasks
> 2. Parallel processing for dedup of completed results
> 
> Used future for every set of subtasks. Each future will execute of the default fork-join pool.
> 
> ### Testing done
> 1. Tested for reduction in execution time. Alovoa finishes in 2 mins on my 10 core 32 GB RAM machine.
> 2. Compared flow counts and paths with the HEAD of master without my changes. Verified that they match.



> Thanks! I'll bring this in today/tomorrow.

@fabsx00

I have merged latest master into this branch. The result table is now the critical section and is under lock. I see that it has a higher locking time causing threads to get stuck on it. Is there a way to reduce the locking time?

I am referring to

Engine.scala -> addResultsToMainTable() -> synchronized(mainResultTable.add(parentTask, Vector(r.copy(taskStack = newTaskStack, path = newPath))))

PR #2128, opened 2022-12-22
Improved caching

fabsx00:
This brings back the caching that I introduced a month back in and needed to revert because of inconsistent results. With this change, we get much improved performance while generating the same number of flows, and  `(source,sink)` pairs over runs. I do still suspect some areas in the code that could lead to inconsistencies, which I will look into next. In particular, I can imagine problems for overlapping flows.

Notes:
* I reworked the deduplication in the process, and it is a lot more aggressive now, meaning, we are intentionally returning only one result per `(source,sink)` pair
* The result table seems to not always be exactly the same, but it may not need to be.
* The number of held tasks can vary. This may also be perfectly fine, but it is worth inspecting.
* We should now be able to further simplify the TaskSolver as it works exclusively within single functions.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
@@ -18,27 +18,40 @@ import scala.jdk.CollectionConverters._
 import scala.util.{Failure, Success, Try}
 
 /** @param taskStack
-  *   The list of tasks that was solved to arrive at this task, including the current task, which is to be solved as the
-  *   last element of the list.
+  *   The list of tasks that was solved to arrive at this task, including the current task, which is to be solved. The
+  *   current task is the last element of the list.
   *
   * @param initialPath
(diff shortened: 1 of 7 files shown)


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
```diff
@@ -50,6 +52,9 @@ class Engine(context: EngineContext) {
 
   private val mainResultTable: ResultTable = new ResultTable
 
+  private val started: mutable.Buffer[ReachableByTask] = mutable.Buffer()
+  private var held: List[ReachableByTask]              = List()
```
max-leuthaeuser, line 56: Why is one a mutable Buffer and the other one a immutable List but a var?
Looks like they are used very similar down below.

fabsx00: Because I couldn't make up my mind, lol. I'll choose one.

max-leuthaeuser (approved):

PR #2130, opened 2022-12-22
Use X2CpgFrontend and X2CpgConfig in Pysrc frontend.

ml86:
Reuse already existing components. Most importantly use the same option
parser library and provide common API via X2CpgFrontend trait.

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/Py2CpgOnFileSystem.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/Py2CpgOnFileSystem.scala
@@ -1,49 +1,57 @@
 package io.joern.pysrc2cpg
 
-import io.joern.x2cpg.X2Cpg
+import io.joern.x2cpg.{X2Cpg, X2CpgConfig, X2CpgFrontend}
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.utils.IOUtils
 import org.slf4j.LoggerFactory
-import overflowdb.Graph
-
-import java.nio.charset.StandardCharsets
 import java.nio.file.attribute.BasicFileAttributes
 import java.nio.file.{FileVisitResult, Files, Path, Paths, SimpleFileVisitor}
 import scala.collection.mutable
+import scala.util.Try
 
-case class Py2CpgOnFileSystemConfig(outputFile: Path, inputDir: Path, ignoreVenvDir: Option[Path])
+case class Py2CpgOnFileSystemConfig(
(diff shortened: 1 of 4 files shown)


ml86:
@tuxology Little Christmas present for you. Let me know whether `Py2CpgOnFileSystem` now fullfills your needs.

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/Py2CpgOnFileSystem.scala
```diff
@@ -1,49 +1,57 @@
...
+  }
 
-case class Py2CpgOnFileSystemConfig(outputFile: Path, inputDir: Path, ignoreVenvDir: Option[Path])
+  override def withOutputPath(x: String): Py2CpgOnFileSystemConfig = {
+    copy(outputFile = Paths.get(x))
+  }
+}
 
-object Py2CpgOnFileSystem {
+class Py2CpgOnFileSystem extends X2CpgFrontend[Py2CpgOnFileSystemConfig] {
```
tuxology, line 27: Why do we need the class to be `Py2CpgOnFilesystem`? Can we just not move all this and make it part of Py2Cpg?

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/Py2CpgOnFileSystem.scala
```diff
@@ -1,49 +1,57 @@
...
   private val logger = LoggerFactory.getLogger(getClass)
 
   /** Entry point for files system based cpg generation from python code.
     * @param config
     *   Configuration for cpg generation.
     */
-  def buildCpg(config: Py2CpgOnFileSystemConfig): Cpg = {
+  override def createCpg(config: Py2CpgOnFileSystemConfig): Try[Cpg] = {
```
tuxology, line 34: Since now there is no `createCpgWithAllOverlays`, and just `createCpg`, what API should I use to apply the OSSdataflow once I get the cpg from here? Essentially wanted to do this: https://github.com/joernio/joern/pull/2124/files#diff-1bc22ed3515fa68269df0781c82f74256c1d899c8bfdaee309087a340b67cbacR51-R53 somehow via an API directly instead of

tuxology (commented):
Thanks a lot. Since we are using X2cpgFrontend now, this is nice. I left a few comments and requested small changes. Can we also standardize an API to get the added overlays? I was generally fine with variation of createCpgWithAlOverlays or createCpgWithOverlays

tuxology:
@ml86 I am fine with merging this as is, I'll add other overlays myself :+1:

fabsx00 (approved):

PR #2135, opened 2022-12-29
[dataflowengineoss]: Fix Cross-Tainting of Arguments to Semantic Defined Parameter

### LINKED ISSUE #2120: [dataflowengineoss] Cross-Parameter Tainting on Semantics that only "Self-Taint"
Given the code
```java
class Main{
    public static void main(String[] args) {
        int a = 1;
        int b = 2;
        int c = 3;
        foo(a, b, c);

        sink1(a);
        sink2(b);
        sink3(c);
    }

    public static void sink1(int x) {
     System.out.println(x);
    }

    public static void sink2(int x) {
     System.out.println(x);
    }

    public static void sink3(int x) {
     System.out.println(x);
    }
}
```
The foo call method full name is `Main.foo:<unresolvedSignature>(3)`. Adding the following semantic `FlowSemantic("Main.foo:<unresolvedSignature>(3)", List((1, 1), (2, 2), (3, 3)))` should essentially just mean that each parameter should taint itself. 

When running
```
def source = cpg.method("main").ast.isIdentifier.name("a").lineNumber(3)
def sink = cpg.call("sink.*").argument(1)
println(sink.reachableByFlows(source).p)
```
We get
```
List(__________________________________________________________________________________________________________________________________________________
| nodeType   | tracked      | lineNumber| method| file                                                                                            |
|=================================================================================================================================================|
| Identifier | int a = 1    | 4         | main  | /var/folders/8g/yy1yqblj6s15pmknbln5g1300000gn/T/x2cpgTestTmpDir10370922593745029661/Test0.java |
| Identifier | foo(a, b, c) | 7         | main  | /var/folders/8g/yy1yqblj6s15pmknbln5g1300000gn/T/x2cpgTestTmpDir10370922593745029661/Test0.java |
| Identifier | sink1(a)     | 9         | main  | /var/folders/8g/yy1yqblj6s15pmknbln5g1300000gn/T/x2cpgTestTmpDir10370922593745029661/Test0.java |
, __________________________________________________________________________________________________________________________________________________
| nodeType   | tracked      | lineNumber| method| file                                                                                            |
|=================================================================================================================================================|
| Identifier | int a = 1    | 4         | main  | /var/folders/8g/yy1yqblj6s15pmknbln5g1300000gn/T/x2cpgTestTmpDir10370922593745029661/Test0.java |
| Identifier | foo(a, b, c) | 7         | main  | /var/folders/8g/yy1yqblj6s15pmknbln5g1300000gn/T/x2cpgTestTmpDir10370922593745029661/Test0.java |
| Identifier | foo(a, b, c) | 7         | main  | /var/folders/8g/yy1yqblj6s15pmknbln5g1300000gn/T/x2cpgTestTmpDir10370922593745029661/Test0.java |
| Identifier | sink2(b)     | 10        | main  | /var/folders/8g/yy1yqblj6s15pmknbln5g1300000gn/T/x2cpgTestTmpDir10370922593745029661/Test0.java |
, __________________________________________________________________________________________________________________________________________________
| nodeType   | tracked      | lineNumber| method| file                                                                                            |
|=================================================================================================================================================|
| Identifier | int a = 1    | 4         | main  | /var/folders/8g/yy1yqblj6s15pmknbln5g1300000gn/T/x2cpgTestTmpDir10370922593745029661/Test0.java |
| Identifier | foo(a, b, c) | 7         | main  | /var/folders/8g/yy1yqblj6s15pmknbln5g1300000gn/T/x2cpgTestTmpDir10370922593745029661/Test0.java |
| Identifier | foo(a, b, c) | 7         | main  | /var/folders/8g/yy1yqblj6s15pmknbln5g1300000gn/T/x2cpgTestTmpDir10370922593745029661/Test0.java |
| Identifier | sink3(c)     | 11        | main  | /var/folders/8g/yy1yqblj6s15pmknbln5g1300000gn/T/x2cpgTestTmpDir10370922593745029661/Test0.java |
)
```
Which implies cross tainting. I can confirm the semantic loads correctly since if I load `FlowSemantic("Main.foo:<unresolvedSignature>(3)", List((1, 1)))` I only get 1 flow from `a -> sink1`.

DavidBakerEffendi:
- Added a special case where CFG nodes are in succession but are only so because they are arguments to the same parameter, not because the data flows across nodes

Fixes #2120

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/nodemethods/ExpressionMethods.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/nodemethods/ExpressionMethods.scala
@@ -1,7 +1,7 @@
 package io.joern.dataflowengineoss.language.nodemethods
 
-import io.shiftleft.codepropertygraph.generated.nodes.{Expression, Method}
 import io.joern.dataflowengineoss.semanticsloader.{FlowSemantic, Semantics}
+import io.shiftleft.codepropertygraph.generated.nodes.{Call, Expression, Method}
 import io.shiftleft.semanticcpg.language._
 import overflowdb.traversal.Traversal
 
@@ -21,6 +21,32 @@ class ExpressionMethods[NodeType <: Expression](val node: NodeType) extends AnyV
     s.isEmpty || s.exists { semantic =>
       semantic.mappings.exists { case (_, dstIndex) =>
         dstIndex == node.argumentIndex
+      }
+    }
+  }
+
+  /** Determines if this node and the given target node are arguments to the same call.
+    * @param other
+    *   the node to compare
+    * @return
+    *   true if these nodes are arguments to the same call, false if otherwise.
+    */
+  def isArgToSameCallWith(other: Expression): Boolean =
+    node.astParent.collectAll[Call].headOption.equals(other.astParent.collectAll[Call].headOption)
+
+  /** Determines if this node has a flow to the given target node in the defined semantics.
+    * @param tgt
+    *   the target node to check.
+    * @param semantics
+    *   the pre-defined flow semantics.
+    * @return
+    *   true if there is flow defined between the two nodes, false if otherwise.
+    */
+  def hasDefinedFlowTo(tgt: Expression)(implicit semantics: Semantics): Boolean = {
+    val s = semanticsForCallByArg.l
+    s.isEmpty || s.exists { semantic =>
+      semantic.mappings.exists { case (srcIndex, dstIndex) =>
+        srcIndex == node.argumentIndex && dstIndex == tgt.argumentIndex
       }
     }
   }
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi:
For more context, following the example described in the issue #2120, when the CFG follows the argument, we get `a->b->c`. Now, since the semantics are `(1, 1), (2, 2), (3, 3)`, the `ReachingDefPass` finds that all parameters are defined and all arguments are used. Thus, when iterating pairs of CFG nodes, we find that these arguments are valid edges since they are both defined and used. This is false, as they are not defined and used directly on each other.

This fix adds a check, that in summary:
- Are these CFG nodes simply arguments under the same call?
- Is the child defined and the parent?
- Then, do the semantics define a flow between these parameters?

If all three checks above are satisfied, then we define a "cross-taint" flow between these parameters.

fabsx00 (approved):

PR #2141, opened 2023-01-03
c2cpg: refactorings and clean-ups

max-leuthaeuser:
Put some helper node creation functions into AstNodeBuilder which simplifies AST node creation a lot. The Scala Int -> Java Integer mapping looks strange but is required as the NewNode interface requires Java Integers there. Switching to Scala Int also fixes some compiler warnings for c2cpg (type was inferred to Any).

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForTypesCreator.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForTypesCreator.scala
@@ -35,22 +35,12 @@ trait AstForTypesCreator { this: AstCreator =>
   }
 
   private def astForNamespaceDefinition(namespaceDefinition: ICPPASTNamespaceDefinition): Ast = {
-    val linenumber   = line(namespaceDefinition)
-    val columnnumber = column(namespaceDefinition)
-    val filename     = fileName(namespaceDefinition)
-
     val (name, fullname) =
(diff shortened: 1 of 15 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -62,82 +52,42 @@ trait AstForTypesCreator { this: AstCreator =>
...
-    val code = "namespace " + name + " = " + fullname
-    val cpgNamespace = NewNamespaceBlock()
-      .code(code)
-      .lineNumber(linenumber)
-      .columnNumber(columnnumber)
-      .filename(filename)
-      .name(name)
-      .fullName(fullname)
-
+    val code         = "namespace " + name + " = " + fullname
```
mpollmeier, line 62: better use string interpolation - more readable and safer
there's a few other places in this PR as well...
```suggestion
    val code         = s"namespace $name = $fullname"
```

mpollmeier (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForTypesCreator.scala
mpollmeier: looks like that's true for the c2cpg codebase - might be worth fixing it - intellij can do that semi-automatically for you...

PR #2143, opened 2023-01-03
[x2cpg]: DynamicCallLinker Excludes Non-Overriden Inheritted Methods

### LINKED ISSUE #2142: [x2cpg] DynamicCallLinker Excludes Non-Overriden Inheritted Methods
In the following example:
```java
class Foo {

  private int x = 0;

  public int foo(int a) {
    this.x = a;
    return this.x;
  }
}

class Bar extends Foo {

  public int bar(int b) {
    return foo(b);
  }

}
```

The call to `foo` within the method `Foo.bar` gives a full name of `Bar.foo:int(int)`. The call graph does not see `Bar.foo` as this is not explicitly overridden and thus isn't a concrete method and simply an alias to the superclass.

Desired solution:
* Change the `methodFullName` of calls to the superclass if no overridden method is present
* Include superclass methods in the `validM` value set for a call in the `DynamicCallLinker`

DavidBakerEffendi:
- If no methods are in the list of callees, attempts to first resolve the result by inspecting the superclass method names and signatures
- The method full name still makes use of the derived full name

Resolves #2142

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/DynamicCallLinker.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/DynamicCallLinker.scala
@@ -1,5 +1,6 @@
 package io.joern.x2cpg.passes.callgraph
 
+import io.joern.x2cpg.Defines.DynamicCallUnknownFallName
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.codepropertygraph.generated.nodes.{Call, Method, Type, TypeDecl}
 import io.shiftleft.codepropertygraph.generated.{DispatchTypes, EdgeTypes, PropertyNames}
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/DynamicCallLinker.scala
```diff
@@ -100,6 +104,30 @@ class DynamicCallLinker(cpg: Cpg) extends CpgPass(cpg) {
     }
   }
 
+  /** Recursively returns all the super-types of the given type declaration. Does not account for circular hierarchies.
```
fabsx00, line 107: If we do get a circular hierarchy here, does it just run forever? If so, I'd recommend we include a loop detector. Whenever you think something can't have loops, it ends up having loops. :D

DavidBakerEffendi: Will include that in the change

fabsx00 (approved):

PR #2149, opened 2023-01-04
Optimze `completeHeldTasks`

fabsx00:
The results produced when completing a task only depend on the entries in the main table for the fingerprint of the task. Therefore, we only have to recalculate results for a task when there are changes in those entries, as opposed to recalculating whenever any entry has changed.


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/Engine.scala
```diff
@@ -196,19 +196,27 @@ class Engine(context: EngineContext) {
...
+        }
+        .seq
+
+      changed = toProcess.map { t => t.fingerprint -> false }.toMap
 
       taskResultsPairs.foreach { case (t, resultsForTask, newResults) =>
         if (newResults.nonEmpty) {
           addCompletedTasksToMainTable(newResults.toList)
-          changed = true
+          newResults.foreach { case (fingerprint, _) =>
```
DavidBakerEffendi, line 217: Perhaps could be 
```scala
changed ++= newResults.map(_._1 -> true).toMap
```
But this just makes it from 3 lines to 1, arguably less readable

DavidBakerEffendi (approved):
Just left two suggestions

PR #2157, opened 2023-01-05
c2cpg: fixed macro expansion and CFG generation with nested blocks

max-leuthaeuser:
Nested macro expansions with blocks contained multiple nested blocks which led to broken CFGs. We fixed that by checking for blocks in asChildOfMacroCall.

Also: blocks are not allowed as CFG nodes within logical operators.


fabsx00 (approved):
I think for Javascript, we're using BLOCK nodes for object notation, so, I'm not sure whether BLOCKs can't occur as operands in logical operations.

max-leuthaeuser:
> I think for Javascript, we're using BLOCK nodes for object notation, so, I'm not sure whether BLOCKs can't occur as operands in logical operations.

I will check this 👍

PR #2167, opened 2023-01-06
Added null/ to .gitignore

max-leuthaeuser:
This is some Coursier cache folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ pdg-for-funcs.json
 joern-cli/workspace
 /workspace
 .bloop
+null
 /.bsp
 /gnupg-*
 /joern-cli/${sys:LOG_DIR}/


mpollmeier:
i don't have that... are you sure? how can i reproduce?

max-leuthaeuser:
Run sbt compile or test. Maybe it's a Windows related issue.

max-leuthaeuser:
See: https://github.com/sbt/sbt/issues/5206
Well.. Don't know exactly how to solve this issue.

max-leuthaeuser:
Ok, got it. It only happens on Windows after upgrading SBT.
So for future readers:
1) update SBT
2) run compile or test at least once
3) manually delete the `null/` folder created by SBT/Coursier.
4) run compile or test and notice that no new `null/` folder will be created.

mpollmeier:
Thanks for digging into this, good to know!
Btw I don't mind having `null/` in our gitignore, I just wanted to understand where this is coming from.

mpollmeier (approved):

PR #2178, opened 2023-01-16
[pysrc2cpg] Abstract Type Recovery Pass \w Python Implementation

DavidBakerEffendi:
This is inspired by abstract interpretation, but to make this fast, it is flow-insensitive and does not focus on the symbols assigned but rather just the possible types. e.g.

```python
#  foobar.py
from baz import Faz, Far

a = Faz()
a = Far()
```

Will give a symbol table entry:
`(foobar.py, a) => [baz.py:<module>.Faz, baz.py:<module>.Far]`

Description of these changes:

- Reads imports and adds type hints to identifiers assigned to these, either to `DYNAMIC_TYPE_HINT_FULL_NAME` or `TYPE_FULL_NAME`. This is determined if the identifier is at the assignment, or elsewhere. Additionally, if at the end of the recovery there is only 1 entry in `DYNAMIC_TYPE_HINT_FULL_NAME` then it is promoted to `TYPE_FULL_NAME`.
- The symbol table makes use of `(File, Identifier)` pairs as keys. This is vulnerable to shadowing, but this frontend does not link `REF` edges from children methods to parent modules, so we will make this approximation for now. Ideally, we would use `Declaration` nodes as keys to the symbol table.
- The call linker has been reworked to rely on the recovered type information and will attempt to link the new callees if they exist. `METHOD_FULL_NAME` is set if there is only one callee recovered.
- Steps in the type recovery pass are expressed as forkable and will fork up to a parent execution pool.

This is not a feature-complete type recovery. Missing features that aren't handled:
- Field accesses/loads
- Method return values in assignments
- Parameters
- Super types aren't considered, e.g. CHA 

Refer to the tests to see the current capabilities.


DavidBakerEffendi:
This currently doesn't "know" of types in child method scopes. Will look at the fix now...

DavidBakerEffendi:
To be robust to shadowing, I could make the symbol table key `(String, Seq[String], String) => (File, Seq[Method], Identifier)`. This way the method chain determines how far the identifier can be referenced, i.e. the shallower, the more methods it's in scope to.

DavidBakerEffendi:
@johannescoetzee see the above suggestion for symbol table keys. This may make this more precise and useful for other frontends too.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/PythonTypeRecovery.scala
```diff
@@ -0,0 +1,318 @@
...
+  */
+class PythonTypeRecovery(cpg: Cpg) extends CpgPass(cpg) {
+
+  private val symbolTable = new SymbolTable()
+  private val logger      = LogManager.getLogger(classOf[PythonTypeRecovery])
+
+  override def run(builder: DiffGraphBuilder): Unit =
+    try {
+      // Set aliases that point to imports for local and external methods/modules
+      def importCalls = cpg.call.where(_.nameExact("import"))
```
Liyw979, line 32: cpg.call.nameExact("import")

DavidBakerEffendi:
Latest commit adds support for Python literals

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/PythonTypeHintCallLinker.scala
```diff
@@ -0,0 +1,44 @@
...
+import overflowdb.traversal.Traversal
+
+/** Attempts to set the <code>methodFullName</code> and link to callees using the recovered type information from
+  * [[PythonTypeRecovery]]. Note that some methods may not be present as they could be external and have been
+  * dynamically discovered, thus the [[io.joern.x2cpg.passes.base.MethodStubCreator]] would have missed it.
+  *
+  * @param cpg
+  *   the target code property graph.
+  */
+class PythonTypeHintCallLinker(cpg: Cpg) extends CpgPass(cpg) {
```
fabsx00, line 17: Looks like we are processing all calls of the code base in a single thread. I wonder if we can find a good unit for parallelization here. For example, can we process one compilation unit at a time?

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/PythonTypeHintCallLinker.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  implicit private val resolver: NoResolve.type = NoResolve
+
+  override def run(builder: DiffGraphBuilder): Unit = {
+
+    def calls: Traversal[Call] = cpg.call
+      .filterNot(c => c.name.matches("^(<operator>|import).*"))
```
fabsx00, line 24: Would be good to comment here in which parts are language agnostic and which are language specific. I take it the `<operator>`-part is language agnostic, while the `import` is specific to the python frontend.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/PythonTypeRecovery.scala
```diff
@@ -0,0 +1,370 @@
...
+import scala.collection.concurrent.TrieMap
+import scala.util.Try
+
+/** Based on a flow-insensitive symbol-table-style approach. This does not accurately determine the difference between
+  * shadowed variables in the same file but this is due to REF edges not connecting children methods to parent scope
+  * (yet).
+  *
+  * TODO: Currently this just improves the type recovery of imported function and types.
+  */
+class PythonTypeRecovery(cpg: Cpg) extends CpgPass(cpg) {
```
fabsx00, line 25: Same comment as for the above pass: can we do this in parallel for multiple independent units?

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/PythonTypeRecovery.scala
```diff
@@ -0,0 +1,370 @@
...
+        symbolTable.put(k, symbolTable(k).filterNot(_.contains("__init__.py")))
+      }
+    }
+  }
+
+  /** Tasks responsible for populating the symbol table with import data.
+    * @param node
+    *   a node that references import information.
+    */
+  class SetImportTask(node: CfgNode) extends RecursiveTask[Unit] {
```
fabsx00, line 73: Ah, I see you're using RecursiveTask here for parallelism.

fabsx00 (changes requested):
Great work! Can you provide a more high-level overview of how the algorithm works and possibly include that in a longer comment? I think this is the kind of code that we'll be happy to have commented in a few months from now :)

DavidBakerEffendi:
@fabsx00 I've gone ahead and separated the class into an abstract and implementation class. This way the algorithmic stuff stays in the abstract classes' comments. I've used the HTML tags so that it generates in the ScalaDoc nicely. Here is an example of the description above `XTypeRecovery.scala`.

```
Based on a flow-insensitive symbol-table-style approach. This does not accurately determine the difference between shadowed variables in the same file but this is due to REF edges not connecting children methods to parent scope (yet).
The algorithm flows roughly as follows:

1. Scan the CPG for method signatures of methods which are in scope of each compilation unit, either by internally defined methods or by reading import signatures. This includes looking for aliases, e.g. import foo as bar. Store these details in the global table
2. TODO: While performing the above, note field/member assignments in the global table.
3. (Optionally) Prune these method signatures by checking their validity against the CPG.
4. Visit each compilation unit and:
  4.1 Visit assignments to populate where variables are assigned a value to extrapolate its type. Store these values in a    local symbol table.
  4.2 Find instances of where these variables are used and update their type information.
  4.3 If this variable is the receiver of a call, make sure to set the type of the call accordingly.
  
The global and local symbol tables use the SymbolTable class to track possible type information.
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeHintCallLinker.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import io.shiftleft.codepropertygraph.Cpg
+import io.shiftleft.codepropertygraph.generated.{EdgeTypes, PropertyNames}
+import io.shiftleft.codepropertygraph.generated.nodes.{Call, Method}
+import io.shiftleft.passes.CpgPass
+import io.shiftleft.semanticcpg.language._
+import overflowdb.traversal.Traversal
+
+/** Attempts to set the <code>methodFullName</code> and link to callees using the recovered type information from
+  * [[XTypeRecovery]]. Note that some methods may not be present as they could be external and have been dynamically
```
fabsx00, line 11: What is the effect of the method not being present here? Are you saying that you handle that condition?

DavidBakerEffendi: TLDR: I don't handle it, not sure if it's within the scope of this pass or not, but it could be.

Essentially the `MethodStubPass` has already happened at this point, yet we may find new external methods. So no new external method stubs are linked under `callee`, but the information about this new call may be stored under `methodFullName`. Ideally, we would be able to run `MethodStubPass` after this, or be able to create new stubs on the fly here.

For the sake of tagging from `dynamicTypeHintFullName`, `methodFullName`, and `typeFullName`, it's fine. For the sake of using `call.callee`, it would be an issue.

DavidBakerEffendi: PS: Similar situation with type stubs.

`MethodStubPass` could perhaps have an `object` function where, given a `builder` and some method info, could register a new method in the `builder`.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeHintCallLinker.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def calls: Traversal[Call] = cpg.call
+    .filterNot(c => c.name.startsWith("<operator>"))
+    .filter(c => calleeNames(c).nonEmpty)
+    .filter(_.callee.isEmpty)
+
+  def calleeNames(c: Call): Seq[String] =
+    (c.dynamicTypeHintFullName ++ Seq(c.typeFullName)).filterNot(_.equals("ANY")).map {
+      // Python call from  a type
+      case typ if typ.split("\\.").last.charAt(0).isUpper => s"$typ.${c.name}"
```
fabsx00, line 29: This may throw a `NoSuchElementException` when `typ` is empty, probably better to handle gracefully using lastOption.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeRecovery.scala
```diff
@@ -0,0 +1,354 @@
...
+  *
+  * @param node
+  *   a node that references import information.
+  */
+abstract class SetXProcedureDefTask(node: CfgNode) extends RecursiveTask[Unit] {
+
+  protected val logger: Logger = LogManager.getLogger(classOf[SetXProcedureDefTask])
+
+  override def compute(): Unit =
+    node match {
```
fabsx00, line 142: default case missing, which may result in match error at runtime

fabsx00 (approved):
Good to go.

PR #2184, opened 2023-01-17
Added support for the astgen macOS arm binary

max-leuthaeuser:


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/Environment.scala
```diff
@@ -5,8 +5,9 @@ import java.nio.file.Paths
 
 object Environment {
 
-  val IS_MAC: Boolean   = scala.util.Properties.isMac
-  val IS_LINUX: Boolean = scala.util.Properties.isLinux
+  val IS_MAC: Boolean     = scala.util.Properties.isMac
```
mpollmeier, line 8: if we change this to `IS_MAC_X86` (or similar) it'd be more specific, and the order of the conditional in `AstGenRUnner` wouldn't matter any more... just a thought

mpollmeier (approved):

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/Environment.scala
max-leuthaeuser: Yes, right. I changed that to something more explicit. Please have a look :-)

mpollmeier (approved):
Not sure if this causes any issues, and it's not really important, but if we followed scala's naming convention it would be `val OperatorSystem` and `val Architecture`.

max-leuthaeuser:
Right. The Scala naming convention suggests to use camel case with capital starting letter.
I took the chance to change that everywhere in jssrc2cpg.

mpollmeier (approved):

max-leuthaeuser:
@mpollmeier I removed the caching and it worked.

mpollmeier:
🎉

mpollmeier:
that might significantly slow down the PR builds though, because it'll download half of maven central every time...
maybe just exclude the ammonite directory from the cache?

mpollmeier:
your test probably did exactly that

PR #2196, opened 2023-01-18
shadow io.github.retronym.java9rtexport.Export with bugfix

mpollmeier:
will test this here first and then provide an upstream fix


max-leuthaeuser (approved):

mpollmeier:
this can be reverted once https://github.com/com-lihaoyi/Ammonite/pull/1318 is merged and released

PR #2210, opened 2023-01-25
Refactored `NodeBuilders.methodReturnNode`

xavierpinho:
Essentially 2 changes:

1. Moved `methodReturnNode` out of `AstCreator` (where methods return an `Ast`) and into `NodeBuilders` (where methods return a `Node`).
2. Set the default value of `code` to "RET". Only the C frontend seemed to be using it any differently (by using the type full name), but it seems like a needless distinction (?) as the frontend seems to work just as well.


fabsx00:
@max-leuthaeuser unless setting the code field to the type in the C frontend is necessary downstream, I'd suggest we merge this.

max-leuthaeuser:
Afaik we have a couple of dataflow tests for various frontends in codescience/securityprofile but they should not be affected.

max-leuthaeuser (approved):

Issue #2206, opened 2023-01-24
[Bug] Joern silently crasing with no crash logs


tuxology:
**Describe the bug**
Latest version of `joern` when installed from the install script is not booting up. There are multiple SLF4J logs on console but no detailed log of crash. Using Joern as a library post v1.1.359 also shows the same behavior when silent crashes and no logs are observed. Rebuilding from scratch using `sbt` and then testing does not show the same error however. Tested on `v1.1.403`

**To Reproduce**
Install and start joern using `sudo ./joern-install.sh --reinstall --interactive` and then start joern

```
$ joern
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
SLF4J: Class path contains SLF4J bindings targeting slf4j-api versions 1.7.x or earlier.
SLF4J: Ignoring binding found at [jar:file:/opt/joern/joern-cli/lib/org.apache.logging.log4j.log4j-slf4j-impl-2.19.0.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See https://www.slf4j.org/codes.html#ignoredBindings for an explanation.
Compiling /home/suchakra/(console)
```
End of log :point_up_2: 

**Expected behavior**
Joern starts normally and when integrated as a library, is able to print all logs. 

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Elementary OS v5.1
 - Joern Version : v1.1.403
 - Java version: 

```
openjdk version "11.0.17" 2022-10-18
OpenJDK Runtime Environment (build 11.0.17+8-post-Ubuntu-1ubuntu218.04)
OpenJDK 64-Bit Server VM (build 11.0.17+8-post-Ubuntu-1ubuntu218.04, mixed mode, sharing)
```

mpollmeier:
@tuxology does this also happen if you install without `sudo`?
I currently have covid, so I'm a bit limited atm, in many ways.

max-leuthaeuser:
Just tested it on my machine.
It worked without any problems (with and without `sudo`).

```
openjdk 11.0.18 2023-01-17
OpenJDK Runtime Environment (build 11.0.18+10)
OpenJDK 64-Bit Server VM (build 11.0.18+10, mixed mode)
```

fabsx00:
[...]

PR #2215, opened 2023-01-27
add SLF4J dependency

pr0me:
This should fix joern exiting after throwing the warning discussed in #2206 by adding a `slf4j` dependency. 
This has been now reported from multiple sides.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -39,8 +39,10 @@ lazy val aggregatedProjects: Seq[ProjectReference] = Seq(
   kotlin2cpg
 )
 
-ThisBuild / libraryDependencies +=
-  "org.apache.logging.log4j" % "log4j-slf4j-impl" % Versions.log4j % Test
+ThisBuild / libraryDependencies ++= Seq(
+  "org.apache.logging.log4j" % "log4j-slf4j-impl" % Versions.log4j % Test,
+  "org.slf4j" % "slf4j-simple" % "2.0.6"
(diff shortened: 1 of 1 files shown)


mpollmeier (changes requested):
Unfortunately it's not that easy. If we do this, then `slf4j-simple` becomes a joern dependency and everything downstream inherits it. That's a big no-no though, because downstream projects may likely already bring in other slf4j implementations (there's like 20...). And with this change it means they'll have two, and that's even worse than having exactly one...

mpollmeier:
To fix these warnings we need to add the dependency to the specific places that use it, e.g. the frontend that's being invoked in that ticket.

pr0me:
Okay I see, thanks for the heads-up!

If I recall correctly, I saw that error just on starting joern-shell, before any frontends are invoked. 
But I'll investigate more.

mpollmeier:
ouch, yes you're right... I'd say that's new.
I'll try to figure out when that started...

mpollmeier:
oh, looks like that's an slf4j v2 issue. some dependency upgraded slf4j to version 2.x and new the whole mess just got another dimension 😭 
gotta love this stuff, probably got enough material for a book by now. not a happy one though

anyway, thanks for the heads-up, i'll get this sorted

mpollmeier:
superseded by https://github.com/joernio/joern/pull/2215

Issue #2205, opened 2023-01-23
Improve astgen dependency management


iandinwoodie:
**Is your feature request related to a problem? Please describe.**

In my attempts to create a Homebrew formula to address https://github.com/joernio/joern/issues/1107 I've encountered the following behaviors:

1. The build system does not search the path for an `astgen` executable. Instead, it checks for the binaries only in the `jeorn-cli/frontends/jssrc2cpg/bin/astgen` directory.
    * This requires me to create a symlink from the `astgen` dependency found in the path to the `jeorn-cli/frontends/jssrc2cpg/bin/astgen` directory.
2. The build system downloads `astgen` binaries for all platforms, regardless of the user's platform, except for when a satisfactory file is found in the `jeorn-cli/frontends/jssrc2cpg/bin/astgen` directory.
    * This requires me to create placeholder files in `jeorn-cli/frontends/jssrc2cpg/bin/astgen` to prevent unnecessary downloads.
    * E.g., on macOS ARM I need to create `astgen-linux`, `astgen-macos`, and `astgen-win.exe` placeholder files.
3. The build system makes copies of all the files in the `jeorn-cli/frontends/jssrc2cpg/bin/astgen` directory.
    * This copies the symlinked `astgen` binary into the build directory (Note: it copies the actual binary, not the symlink).
        * For Homebrew purposes, this secret decouples `jeorn` from its `astgen` dependency formula.
        * Therefore, commands like `brew info jeorn` won't accurately represent the underlying state.
            * E.g., `brew uninstall --force astgen` would seemingly break `jeorn` according to Homebrew, but `jeorn` would continue operating fine because it is relying on it's own local copy of `astgen`.

**Describe the solution you'd like**

* To address point 1, the build system could be updated to first consider an `astgen` binary found in the path.
    * If there are concerns about the build system proceeding with an outdated version of `astgen`, then the build system could be updated to error if the output of `astgen --version` indicates a version older than the build system requires.
* To address point 2, the build system could be updated to only require `astgen` binaries relevant to the user's machine.
   * It's possible the current behavior exists so a single build results in a distributable that is platform-agnostic.
       * If this is the case, then perhaps the build system could be updated to accept an argument or environment variable to specify a platform-specific build when necessary.
* To address point 3, the build system could be updated to not copy `astgen` binaries to the build directory when the build is using an `astgen` binary found in the path.

**Describe alternatives you've considered**

[...]

PR #2216, opened 2023-01-27
Smarter astgen handling 

max-leuthaeuser:
This improves the astgen binary handling.
The following rules apply:
1) If astgen is found on the system it is used without downloading binaries at all iff the version installed matches or is higher.
2) sbt stage downloads all binaries so the distribution always works on all platforms like before. This does not happen iff rule 1 triggers.
3) compile only downloads the astgen binary for the system it's invoked on.

This is for: https://github.com/joernio/joern/issues/2205


iandinwoodie:
@max-leuthaeuser is there is any chance that item 1 could be expanded to:

> 1. If astgen is found on the system it is used without downloading binaries at all iff **a compatible version is found**.

where a compatible version satisfies both:

1. `found.major == required.major`
2. `found.minor >= required.minor`

If we require an exact match in version, then there is going to be an issue when it comes time to bump the version of the `joern` and `astgen` formulas. For example:

* If the `astgen` formula is bumped first, then `joern` would find a newer `astgen` in the PATH and will resort to fetching binaries until its own formula is bumped.
* If the `joern` formula is bumped first, then the `astgen` built by Homebrew will be older than required and the build system will again resort to fetching binaries until the `astgen` formula is bumped.

I do not believe the Homebrew maintainers allow for two formulas to be bumped in the same PR. If we instead check for a compatible version (as indicated by SemVer), then we would be able to bump the `astgen` formula and then the `joern` formula without ever resorting to fetching binaries.

The edge case here is when `found.major != required.major`. A possible solution for this is to add a means (e.g., command line option or environment variable) for us to fail the build if a compatible `astgen` version is not found. This means that the `joern` formula will fail until it is bumped whenever `astgen` has incompatible API changes, but I think that sounds reasonable.

***

An alternative solution is to implement item 1 as-is, but include the means to fail the build when an exact match is not found. I could then pin an exact `astgen` version in the `joern` formula to keep the two formula in sync. This way `joern` built with Homebrew either builds with a Homebrew built `astgen` or fails. The only drawback to this approach is that if you have `astgen 3.2.0` installed, but `joern` has `astgen 3.1.0` pinned, then you’re going to end up with both `astgen 3.1.0` and `astgen 3.2.0` installed even though `astgen 3.2.0` should have been sufficient (according to SemVer).

max-leuthaeuser:
@iandinwoodie Thanks for spotting this issue! You are right.

Astgen is using SemVer versioning but its sufficient to check for a version installed >= the version requested by joern. We do not need to distinguish between major/minor versions as I will never release a new astgen without having that specific new version tested within joern and updated there as well.

I updated this PR to do that comparison.

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala
```diff
@@ -6,26 +6,83 @@ import io.joern.jssrc2cpg.preprocessing.EjsPreprocessor
...
 object AstGenRunner {
+
+  private val logger = LoggerFactory.getLogger(getClass)
+
   case class AstGenRunnerResult(
     parsedFiles: List[(String, String)] = List.empty,
     skippedFiles: List[(String, String)] = List.empty
   )
+
+  private val ExecutableName = Environment.OperatingSystem match {
```
mpollmeier, line 29: style, here and in other places: the value name should be lowercase, i.e. `val executableName`. Only constants are uppercase by convention, e.g. `val MagicNumber = 42`

mpollmeier (approved):

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala
max-leuthaeuser: Right, will change that 👍

PR #2226, opened 2023-01-31
jssrc2cpg: fixed JS call linker for this expressions

max-leuthaeuser:
Fixes: https://discord.com/channels/832209896089976854/832209896089976857/1069951841456099338

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/JavascriptCallLinker.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/JavascriptCallLinker.scala
@@ -47,11 +47,11 @@ class JavascriptCallLinker(cpg: Cpg) extends CpgPass(cpg) {
         .argument(1)
         .flatMap {
           case assignee: Identifier if isStaticSingleAssignmentLocal(assignee) && assignee.method.name == ":program" =>
-            Some(assignee.name)
+            Option(assignee.name)
           case assignee: Call
               if assignee.methodFullName == Operators.fieldAccess &&
                 (JS_EXPORT_NAMES.contains(assignee.argument(1).code) || assignee.code.startsWith("_tmp_")) =>
-            Some(assignee.argument(2).asInstanceOf[FieldIdentifier].canonicalName)
+            Option(assignee.argument(2).asInstanceOf[FieldIdentifier].canonicalName)
           case _ => None
         }
         .headOption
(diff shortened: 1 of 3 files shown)


fabsx00 (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/JavascriptCallLinker.scala
```diff
@@ -137,17 +137,24 @@ class JavascriptCallLinker(cpg: Cpg) extends CpgPass(cpg) {
...
-    } else {
-      None
-    }
+  private def isArgumentFromExport(c: Call): Boolean =
+    JS_EXPORT_NAMES.contains(c.argument(1).code)
+
+  private def isArgumentFromThis(c: Call): Boolean = {
+    c.argument(1).code == "this" ||
+    // We end up here for calls like "this.x.y.foo()" that are lowered like "(_tmp = this.x.y).foo()"
+    c.argument(1).collectAll[Call].code.headOption.exists(_.contains(" this"))
```
maltek, line 146: that's too error-prone for my tastes. give me a few minutes to come up with something more specific

maltek: Thinking about this more, I have some reservations about this. For `this.foo()`, the assumption that it's referring to a `foo` method from the same file is quite safe, but for `this.foo.bar()` that's a larger stretch.

We can try it, but we need to keep this change in mind if we see some weird new FPs in the near future.

maltek: see here for an AST query to find the `this`: https://github.com/joernio/joern/commit/f6d00c3a060864692f7146e1f557226382c678af It's quite a monster though.

At this point I feel that if we go with this change, it'd be better to just skip the `this` check completely. There's no real difference between `this.foo().bar()` and `foo().bar()` - they are equally likely to refer to a `bar` method from the same file if that method exists.

max-leuthaeuser: Yes, it's quite ugly. Better we go without the check. Want to edit this PR here directly?

max-leuthaeuser: Thanks for that!
I merged your branch into this one here. Yes, the query is quite complicated but does the job and its tested. I guess its safer to do it this way for now.

PR #2228, opened 2023-02-01
fix logging dependencies setup

mpollmeier:
We declare slf4j-api as a regular dependency, since most logging
libraries have an implementation for it, and it is considered
a lowest common denominator for logging.

We choose log4j as the logging provider for this build, i.e. for tests
and `sbt stage`. `Optional` means "not transitive", but still included
in "stage/lib". I.e. downstream builds do not inherit it and can/must
make their own choices.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -39,8 +39,12 @@ lazy val aggregatedProjects: Seq[ProjectReference] = Seq(
   kotlin2cpg
 )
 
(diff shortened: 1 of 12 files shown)


mpollmeier:
grafik

grafik

max-leuthaeuser (approved):

PR #2231, opened 2023-02-01
jssrc2cpg: fixed JS call linker for simple field accesses with known type

max-leuthaeuser:


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/JavascriptCallLinker.scala
```diff
@@ -156,9 +156,19 @@ class JavascriptCallLinker(cpg: Cpg) extends CpgPass(cpg) {
...
 
-  private def fromFieldAccess(c: Call): Option[String] = c.methodFullName match {
-    case Operators.fieldAccess if isArgumentFromExport(c) || isFieldAccessWithThisBase(c) => Option(c.argument(2).code)
-    case _                                                                                => None
+  private def isFieldAccessWithType(call: Call): Boolean = {
+    // if the type of argument = 1 of the field access is known and a matching type decl member can be found
+    // TODO: make this work for for access chains, e.g., "a.b.c.foo()"
+    val fieldName = call.argument(2).code
+    val typeDeclCandidates =
+      call.argument(1).collectAll[Identifier].dynamicTypeHintFullName.flatMap(cpg.typeDecl.nameExact)
```
maltek, line 164: If we start using these in joern, we should really move the `DynamicTypeLinker` from CS to joern instead of adding (incompatible) ad-hoc queries like this one.

Also if we resolve this to a specific type here, then we should also return that type here - this is currently running in the context of https://github.com/joernio/joern/blob/f6d00c3a060864692f7146e1f557226382c678af/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/JavascriptCallLinker.scala#L95 which doesn't really make sense in a case where the type is already known.

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/CallLinkerPassTest.scala
```diff
@@ -5,11 +5,33 @@ import io.shiftleft.codepropertygraph.Cpg
...
+          |class Foo {
+          |  a() {}
+          |}
+          |
+          |const f = new Foo();
+          |f.a();
+          |""".stripMargin)
+
+      inside(cpg.identifier("f").l) { case List(constF, fieldF, baseF) =>
+        constF.dynamicTypeHintFullName shouldBe List("Foo")
```
maltek, line 23: This type hint is wrong. It's supposed to be a `fullName` of a `TypeDecl`, not a `name`.

maltek (changes requested):
I feel we can add infinite special cases like this and never get done. IMO, either we add proper type resolving machinery (like I think is currently happening for Python, or we have in CS) or we just don't bother...

max-leuthaeuser:
Yes, I guess you are right. Although, we can't move the call linking stuff from CS as this requires to have carret paths available which we do not have here at all.
The full name vs name issue is a separate thing. tsc wont give me the full names. Different story though.

PR #2241, opened 2023-02-03
PythonTypeRecovery: Dummy Return from External Libraries 

DavidBakerEffendi:
* Method return values from external libraries now have dummy values so that their calls can be matched with regex
* Added a safety check on method name building for method stub 
* Made sure to check imported type references that use a collection of imports


joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonTypeRecovery.scala
```diff
@@ -202,7 +202,7 @@ class RecoverForPythonFile(cpg: Cpg, cu: File, builder: DiffGraphBuilder, global
           // Case 2: The identifier is receiving a constructor invocation, thus is now an instance of the type
           setIdentifier(i, importedTypes.map(_.stripSuffix(s".${Defines.ConstructorMethodName}")))
         } else {
-          // TODO: This identifier should contain the type of the return value of 'c'
+          // TODO: This identifier should contain the type of the return value of 'c'. See CallCpgTests L204
```
fabsx00, line 205: Line numbers in comments will be outdated soon :D

fabsx00 (approved):

PR #2242, opened 2023-02-03
[x2cpg] Refactor some duplicated `root{Code,Type}` methods

xavierpinho:
This moves 2 duplicated methods out of PHP/Javasrc's frontends, viz. `rootCode` and `rootType`. 

It also fixes a very minor bug in PHP's `isset` `code` property due to an unbalanced parenthesis.

Aside: I'm not sure I've written fully idiomatic Scala (it's been a few years since the last time) here. I look forward for any kind of improvement suggestions you see fit.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/AstPropertiesUtil.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/AstPropertiesUtil.scala
@@ -1,0 +1,21 @@
+package io.joern.x2cpg.utils
+
+import io.joern.x2cpg.Ast
+import io.shiftleft.codepropertygraph.generated.PropertyNames
+
+object AstPropertiesUtil {
+
+  implicit class RootProperties(ast: Ast) {
+
+    private def rootProperty(propertyName: String): Option[String] = {
+      ast.root.flatMap(_.properties.get(propertyName).map(_.toString))
+    }
+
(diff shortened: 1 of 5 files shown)


fabsx00:
@max-leuthaeuser and @mpollmeier can comment re: "idiomatic scala".

fabsx00 (approved):

max-leuthaeuser:
It's good to have it in one spot now 👍 
I am not a big fan of implicit conversions as it brings some overhead both in compiling as well as at runtime (creates a wrapper class instance).
We could optimize it by exploiting value classes:
```
implicit class RootProperties(ast: Ast) extends AnyVal {
    def rootType ...
    ...
```
This compiles into a module load, followed by invokevirtual.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/AstPropertiesUtil.scala
```diff
@@ -0,0 +1,30 @@
+package io.joern.x2cpg.utils
+
+import io.joern.x2cpg.Ast
+import io.shiftleft.codepropertygraph.generated.PropertyNames
+
+object AstPropertiesUtil {
+
+  implicit class RootProperties(ast: Ast) {
```
mpollmeier, line 8:
```suggestion
  implicit class RootProperties(val ast: Ast) extends AnyVal {
```

mpollmeier (approved):

xavierpinho:
Thanks for the suggestions!

Issue #2217, opened 2023-01-28
dataflow issue related to return statement


Liyw979:
**Describe the bug**
dataflow issue related to ```return``` statement

**To Reproduce**
```
    "bug" in {
      implicit val cpg: Cpg = code(
        """
          |class Foo {
          |    public static int bar() {
          |        int a = 42;
          |        return a;
          |    }
          |}
          |""".stripMargin)

      def src = cpg.literal("42")
      def snk = cpg.identifier("a").lineNumber(5)
      snk.reachableByFlows(src).p.foreach(println)
    }
```

**Expected behavior**
shoule print a flow from ```return a``` to ```42```

**Screenshots**

**Desktop (please complete the following information):**
 - Windows
 - v1.1.1408
 - Java 11

Liyw979:
fixed

PR #2244, opened 2023-02-03
fix dataflow of return statement

Liyw979:
#2217


fabsx00 (approved):

fabsx00:
Thanks, good catch!

fabsx00:
Running into some invalid flows with this one, I believe. Will provide more details when I can. Let me revert it for now but get back to you very soon to bring this back in.

fabsx00:
This certainly looks correct: we do not want to add the edge to the return, we want to add it to the identifier, which in turn is connected to the return.

PR #2248, opened 2023-02-04
x2cpg: Introduces (do)whileAst methods in AstCreatorBase

xavierpinho:
Also adds code/line/column properties on javasrc2cpg' `astForDo` - figured I might as well.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
@@ -3,7 +3,7 @@ package io.joern.x2cpg
 import io.joern.x2cpg.passes.frontend.MetaDataPass
 import io.joern.x2cpg.utils.NodeBuilders.methodReturnNode
 import io.shiftleft.codepropertygraph.generated.nodes._
-import io.shiftleft.codepropertygraph.generated.ModifierTypes
+import io.shiftleft.codepropertygraph.generated.{ControlStructureTypes, ModifierTypes}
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
```diff
@@ -122,6 +122,42 @@ abstract class AstCreatorBase(filename: String) {
...
+    code: Option[String] = None,
+    lineNumber: Option[Integer] = None,
+    columnNumber: Option[Integer] = None
+  ): Ast = {
+    var whileNode = NewControlStructure()
+      .controlStructureType(ControlStructureTypes.WHILE)
+      .lineNumber(lineNumber)
+      .columnNumber(columnNumber)
+
+    if (code.isDefined)
```
xavierpinho, line 137: Ideally there would be a `code(Option[String])` method like we have for lines/columns.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
```diff
@@ -98,13 +98,13 @@ abstract class AstCreatorBase(filename: String) {
   def controlStructureAst(
     controlStructureNode: NewControlStructure,
     condition: Option[Ast],
-    children: List[Ast] = List(),
+    children: Seq[Ast] = Seq(),
```
xavierpinho, line 101: Some frontends work with `List`s and others with `Seq`s. Thus it seemed natural to simply generalize than to convert to `List` in a few specific places.

fabsx00 (approved):

PR #2252, opened 2023-02-08
WIP Python: handling of flows starting at member variables

fabsx00:


joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/MemberCpgTests.scala
```diff
@@ -0,0 +1,25 @@
...
+import io.joern.pysrc2cpg.Py2CpgTestContext
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.should.Matchers
+import io.shiftleft.semanticcpg.language._
+
+class MemberCpgTests extends AnyFreeSpec with Matchers {
+
+  lazy val cpg = Py2CpgTestContext.buildCpg("""
+      |class Foo():
+      |   x = 'foo'
```
DavidBakerEffendi, line 12: Could we also add a test for 


```python
class Foo():

  def __init__(self, x):
    self.x = x
```
Where `x` would be some member here?

DavidBakerEffendi: Or is this out of scope?

DavidBakerEffendi (approved):

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/MemberCpgTests.scala
fabsx00: Yeah, good point.

fabsx00: Deferring to follow-up PR.

PR #2255, opened 2023-02-09
pysrc2cpg: Adjust Recovered Method and Type Names

DavidBakerEffendi:
- Made changes to recover the types based on the "type<body>" style syntax to improve linkage
- Added value for MetaData("root") in pysrc2cpg
- Check if the method belongs to an internal file to make the `isExternal=false` if a stub ultimately needs to be created due to incorrect method full name


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeRecovery.scala
```diff
@@ -175,6 +175,7 @@ abstract class RecoverForXCompilationUnit[ComputationalUnit <: AstNode](
     // Prune import names if the methods exist in the CPG
     postVisitImports()
     // Populate local symbol table with assignments
+//    symbolTable.view.filterNot(_._2.exists(_.startsWith("built"))).foreach(println)
```
fabsx00, line 178: Stale debug comment

fabsx00 (approved):

PR #2265, opened 2023-02-12
Consistent callAST creation across frontends

fabsx00:
As I tried to use a single utility method for the creation of call ASTs across frontends, I noticed that there were in fact conflicting assumptions about how a call site should look. In particular, the function `createCallAst` in the javascript frontend and the `callAst` function used in the Java, PHP, and C/C++ frontends differ in that the former allows both a base and a receiver to be specified, while the later only supports a receiver. Moreover, in the Javascript frontend, the receiver always has an `argumentIndex` of `-1`, while in the other frontends, it has an `argumentIndex` of 0 and is treated as a parameter (sometimes).

After taking a more detailed look into the difference between base and receiver, it seemed that what we called "receiver" in Java is just the instance parameter, and more closely related to the concept of "base" than that of "receiver", although these seem to mostly be the same in Java anyway. In particular, in those frontends, we are assuming that the receiver is an argument, while that's not the case in Javascript.

To bring the two worlds together, I have now defined a `callAst` method that works for both. While for Java/PHP/C/C++, we only make use of `base` now, the Javascript frontend additionally makes use of `receiver`. We can now also assure that the `argumentIndex` for `receiver` is always `-1`, and that of `base` is always `0`. Furthermore, `base` is an argument, while receiver is not.

*Update* I found another method called `createCall` in the Python frontend, which again is quite similar but not the same. I will take a closer look here as well and see if it makes sense to use `callAst` for `pysrc2cpg` as well.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/AstCreationPassTests.scala
+++ b/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/AstCreationPassTests.scala
@@ -320,12 +320,10 @@ class AstCreationPassTests extends AbstractPassTest {
           lit.code shouldBe "10"
         }
 
-        inside(lambda2call.argument.l) { case List(lit: Literal) =>
-          lit.code shouldBe "10"
-        }
-        inside(lambda2call.receiver.l) { case List(ref: MethodRef) =>
+        inside(lambda2call.argument.l) { case List(ref: MethodRef, lit: Literal) =>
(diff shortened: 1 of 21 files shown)


DavidBakerEffendi (commented):
Now that `-1` is the receiver, does the data flow engine notation still use `-1` as the return value? Or will many of the existing semantics need to be adjusted?

fabsx00:
This is the call site, not the method stub, so, we should be good there. `parameterIndex` is untouched.

DavidBakerEffendi (approved):

Issue #2224, opened 2023-01-31
[XTypeRecoveryPass] Follow-Up Discovered Calls with [Type|Method] Stubs


DavidBakerEffendi:
**Is your feature request related to a problem? Please describe.**
`XTypeRecoveryPass` will often find method invocations that were not immediately clear by the time the initial `MethodStubPass` runs. This will lead to no methods to hold a `CALL` edge and potentially miss information one can achieve from the `.callee` step.

**Describe the solution you'd like**
To leverage the code in `MethodStubPass` to generate additional stubs for newly discovered calls that need to be linked in the `XTypeHintCallLinker`.

**Describe alternatives you've considered**
`MethodStubPass` could likely have a few functions moved to an object class whereby a builder is an argument.

**Additional context**
N/A

DavidBakerEffendi:
Partially fixed by #2232

DavidBakerEffendi:
Closed by #2282

PR #2282, opened 2023-02-20
MethodStubCreator: Link Method to TypeDecl if Present

DavidBakerEffendi:
This will link the stubbed TYPE_DECL to the stubbed METHOD node by an AST edge, if present.

The aim of this change will allow us to traverse the type node from a method stub. e.g.
```scala
cpg.call.callee(NoResolve).definingTypeDecl.fullName.l
```

Contributes to #2224


fabsx00 (approved):
Comments offline, apart from that, looks good to me.

PR #2284, opened 2023-02-20
Fix tracking of uninitialized member variables

fabsx00:
Previously, we were only able to track initialized members, e.g., tracking from source to sink in the following code would fail:
```
class Foo {
   int source;
  public void foo() {
     sink(source);
  }
}
```
while `int source = 1;` would work.


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
```diff
@@ -149,6 +142,19 @@ class SourceToStartingPoints(src: StoredNode) extends RecursiveTask[List[CfgNode
     }
   }
 
+  private def usagesForName(name: String, m: Method) = {
```
max-leuthaeuser, line 145: Could you add the return type here?
It's inferred yes, but adding it explicitly shortens compile time.

max-leuthaeuser (approved):

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
fabsx00: Sure, done.

DavidBakerEffendi (approved):

PR #2301, opened 2023-02-23
minor refactor to keep this java8 compatible for a downstream project...

mpollmeier:


maltek (approved):

mpollmeier:
@DavidBakerEffendi this broke one of our downstream builds that unfortunately requires to be run with java8 :( 

To be clear: this is no one's fault, just an unfortunate combination of things. We really shouldn't rely on java8, and at the same time I'm not comfortable with the idea of configuring scalac/javac with `target=jdk8` in joern

mpollmeier:
Scala compares Strings by value, not by reference, so this is all good.

maltek:
Not just strings, btw. Scala compiles `==` and `!=` to calls to the JVM `equals` method. (In the rare circumstance where you need reference equality it's called `.eq` in Scala.)

PR #2319, opened 2023-02-28
Fix missing passes if running with 'importCpg'

yzn12345:
Run language specific passes such as 'PythonTypeHintCallLinker' for importCpg just like it behaves for importCode.

### DIFF
--- a/console/src/main/scala/io/joern/console/Console.scala
+++ b/console/src/main/scala/io/joern/console/Console.scala
@@ -360,7 +360,10 @@ class Console[T <: Project](
 
     cpgOpt
       .filter(_.metaData.hasNext)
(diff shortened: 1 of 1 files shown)


fabsx00 (approved):

fabsx00:
Thank you, good catch!

fabsx00 (changes requested):
I'll have to look into this, because I'm not entirely sure the data flow passes are executed correctly now. The post processing passes sometimes rely on the data flow layer being present. Will look into it soon and then merge or fix.

fabsx00 (approved):

fabsx00:
I was wrong: the post-processing passes run after the base layer has been applied, but before data flow calculation.

PR #2334, opened 2023-03-03
Python: Added UT for function param tracing

rahul-privado:
Added a UT to trace flow of function params to sinks.


mpollmeier:
Thank you! If you run `sbt format` it should greenify

fabsx00 (approved):

PR #2348, opened 2023-03-07
pysrc2cpg: propagate type hints on parameters

fabsx00:


joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/passes/TypeRecoveryPassTests.scala
```diff
@@ -533,4 +533,17 @@ class TypeRecoveryPassTests extends PySrc2CpgFixture(withOssDataflow = false) {
...
+  "a type hint on a parameter" should {
+    lazy val cpg = code("""
+        |import sqlalchemy.orm as orm
+        |
+        |async def get_user_by_email(email: str, db: orm.Session):
+        |   return db.query(user_models.User).filter(user_models.User.email == email).first()
+        |""".stripMargin)
+    "be sufficient to resolve method full names at calls" in {
+      val List(call) = cpg.call("query").l
+      call.methodFullName.startsWith("sqlalchemy.orm") shouldBe true
```
DavidBakerEffendi, line 545: One thing I noticed now, and on line 505, is that these full names do not follow the same convention as the rest, which make sense of some kind of file path. Should we try make this consistent?

fabsx00: Can you elaborate? What would they need to look like to be consistent? Or do you maybe just want to update it?

DavidBakerEffendi: For example, one of the other tests recovers a type such as `flask_sqlalchemy.py:<module>.SQLAlchemy.SQLAlchemy<body>`. I suppose line 505 should technically have something that looks like `sqlalchemy/orm.py:<module>.Session.Session<body>.query`. This is to be consistent with how the frontend generates internal method full names.

fabsx00: I see, ok, that makes sense.

DavidBakerEffendi: Of course, this in itself is "best-guess" since the method could be exported via a `__init__.py` file and generated elsewhere. Much chaos as we can't confirm the location of the method as we could with internally defined methods. So if it's an external method, we could also decide to be a bit easier on ourselves and opt for a consistent notation that looks like what is on 505.

fabsx00: This is a tricky decision. Surely, we want to be consistent, but then again `sqlalchemy.orm.Session.query` looks much more like a method full name than `sqlalchemy/orm.py:<module>.Session.Session<body>.query`. In fact, the later exposes a lot of detail of our representation that is probably more confusing than helpful. I wonder if we should in fact rather try to go with the simpler representation in call cases. Maybe we merge this one as is and revisit once your larger refactoring is in?

DavidBakerEffendi: Agreed. I would prefer the `sqlalchemy.orm.Session.query` format - it is more consistent with C and Java frontends.

DavidBakerEffendi (approved):

Issue #2338, opened 2023-03-04
Increase the length of tracked variable or make it configurable


prabhu:
**Is your feature request related to a problem? Please describe.**
The variable is currently getting abbreviated at less than 20 characters reducing readability. Increase to 100 or make this configurable.

https://github.com/joernio/joern/blob/628becc052b0c9454209a92ee9db8c82a1c1147a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/Path.scala#LL46C88-L46C90

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

itsacoderepo:
Closing, because it is solved with env vars

PR #2366, opened 2023-03-12
Partial fix for #2338. Increases max width to 256

prabhu:


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/Path.scala
```diff
@@ -21,6 +21,8 @@ case class Path(elements: List[AstNode]) {
 
 object Path {
 
+  val MAX_TRACKED_WIDTH = 256
```
mpollmeier, line 24:
```suggestion
  val MaxTrackedWidth = 256
```

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/Path.scala
```diff
@@ -43,7 +45,7 @@ object Path {
                 s"$methodName($paramsPretty)"
               case _ => cfgNode.statement.repr
             }
-            val tracked = StringUtils.normalizeSpace(StringUtils.abbreviate(statement, 20))
+            val tracked = StringUtils.normalizeSpace(StringUtils.abbreviate(statement, MAX_TRACKED_WIDTH))
```
mpollmeier, line 48:
```suggestion
            val tracked = StringUtils.normalizeSpace(StringUtils.abbreviate(statement, MaxTrackedWidth))
```

mpollmeier (commented):
just my 2cents for scala naming convention. 
256 chars might be too long as a default for just that one column... maybe default to 40, make configurable, and document that? @fabsx00 thoughts?
grafik

fabsx00:
Yeah, the reason I haven't merged this yet is that it's pretty hard to say what a good default number is. No idea what this will look like.

mpollmeier:
Well I guess the best option would be a dynamic solution that takes the console width into account. Responsive UI for terminals. (One might even think that'd be a nice feature for https://github.com/mpollmeier/scala-repl-pp)

In the interim, I'd say it makes sense to increase the default, since most users will have a fairly wide screen:

```
importCode.c.fromString("""#include <stdio.h>
       #include <stdlib.h>
       #include <string.h>

       int main(int argc, char *argv[]) {
         if (argc > 1 && strcmp(argv[1], "42") == 0) {
           fprintf(stderr, "It depends!\n");
           exit(42);
         }
         printf("What is the meaning of life?\n");
         exit(0);
       }""")

def source = cpg.method.parameter
def sink = cpg.call.name("strcmp")
sink.reachableByFlows(source).p
```

current master:
grafik

this PR:
grafik



However, I wouldn't go as far as 256 chars, since that's almost never truncating anything:
grafik

I just rolled some dice and got `5` and `6`, so why not multiply those and set this value to `30`? 🤔

prabhu:
Any word wrap solution available?

mpollmeier:
@prabhu my comment above was probably a bit confusing. How about this: we increase the value to `30` (not 256) for now, and will revisit with a proper scalable/responsive renderer, once we're on Scala 3. Sounds good?

prabhu:
Thanks, @mpollmeier. Sounds good to me.

mpollmeier:
👍🏻 in that case, the easiest way forward would be if you accept my change suggestions above

mpollmeier (approved):

PR #2373, opened 2023-03-13
JavaScriptTypeRecovery: Initial Implementation Class

DavidBakerEffendi:
* Implemented `XTypeRecovery` classes for JavaScript
* Fixed bug in how import nodes were fetched from `XTypeRecovery` - added convenient traversal for this
* Carried tests over from `pysrc2cpg` and re-implemented a couple into JS/TS to test basic interprocedural type recovery


max-leuthaeuser (approved):
👍 
(only needs formatting)

fabsx00 (approved):

max-leuthaeuser:
We also have some type info already available via astgen / tsc (see e.g., `io.joern.jssrc2cpg.types.TSTypesTest`) when tested with io.joern.jssrc2cpg.passes.AbstractPassTest.AstFixture.apply(..., tsTypes = true).
It needs to be checked that they don't clash.

PR #2374, opened 2023-03-13
Ruby: Adding a skeleton lexer

xavierpinho:
This is only a starting point. Stuff like regex literals or heredocs are next. However, I figured this is enough of a lexer for us to start fleshing out the parser as well.


fabsx00 (approved):
Good stuff. As mentioned, one concern I have is the performance characteristics once we build a robust parser on top of antlr. In the past, we've looked into this issue and decided to go with javacc instead to get around some of these issues. I can imagine that one can achieve the same with ANTLR though. @ml86 is a good person to talk to about this.

PR #2376, opened 2023-03-14
jssrc2cpg: added support for Decorators

max-leuthaeuser:
Decorators are added as Annotations now at

- functions
- parameters
- classes

just like we do in javasrc2cpg.

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
@@ -11,6 +11,8 @@ import io.joern.x2cpg.datastructures.Stack._
 import io.shiftleft.codepropertygraph.generated.EdgeTypes
 import io.shiftleft.codepropertygraph.generated.nodes.{NewCall, NewImport}
 import io.shiftleft.codepropertygraph.generated.DispatchTypes
+import io.shiftleft.codepropertygraph.generated.nodes.NewAnnotationParameter
+import io.shiftleft.codepropertygraph.generated.nodes.NewAnnotationParameterAssign
 import ujson.Value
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
```diff
@@ -159,6 +161,63 @@ trait AstForDeclarationsCreator { this: AstCreator =>
     }
   }
 
+  protected def astsForDecorators(elem: BabelNodeInfo): Seq[Ast] = {
+    if (hasKey(elem.json, "decorators") && !elem.json("decorators").isNull) {
+      elem.json("decorators").arr.toList.map(d => astForDecorator(createBabelNodeInfo(d)))
+    } else Seq.empty
+  }
+
+  private def createAnnotationAssignmentAst(name: String, value: Value, code: String): Ast = {
```
fabsx00, line 170: This one looks like we may want to split it up into a part that depends on the `jssrc2cpg` specific parse tree and the part that deals with the creation of the CPG-AST. We could then make it available via `AstCreatorBase`. Could you add a comment about that if you agree?

fabsx00 (approved):

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForDeclarationsCreator.scala
max-leuthaeuser: Yes, makes sense. Do you want to do this right away with this PR here or as a separate refactoring?

fabsx00: Feel free to merge.

PR #2378, opened 2023-03-14
Extracted annotation AST creation to AstCreatorBase

max-leuthaeuser:
And applied that to jssrc2cpg and javasrc2cpg.


fabsx00 (approved):
Very nice, thanks!

PR #2388, opened 2023-03-16
jssrc2cpg: small fix for JavaScriptTypeRecovery

max-leuthaeuser:
(`.isInstanceOf[Some[Block]]` does not work due to type erasure)

And a small refactoring for its tests.


DavidBakerEffendi (approved):

DavidBakerEffendi:
Next ticket I need to also verify is `require` from relative paths - I don't know if they are handled correctly.

PR #2395, opened 2023-03-17
PythonTypeRecovery: Add additional constructor check to superficial heuristic

DavidBakerEffendi:
* Instead of simply checking based on Python style guides, follow-up with checking against the CPG if it is a normal function
* Move tests from DataFlowTests.scala to TypeRecoveryPassTests.scala as they do not track data flow but rather check type recovery related functionality


joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonTypeRecovery.scala
```diff
@@ -164,17 +164,23 @@ class RecoverForPythonFile(
       if (procedureName.contains("__init__.py")) procedureName
       else procedureName.replace(".py", s"${JFile.separator}__init__.py")
 
-    possibleCalleeNames(procedureName, isConstructor(expEntity), procedureName.contains("<var>"))
+    val isMaybeConstructor = expEntity.split("\\.").lastOption.exists(_.charAt(0).isUpper)
```
fabsx00, line 167: Does `charAt(0)` raise an exception if the string is empty?

fabsx00 (approved):

PR #2408, opened 2023-03-21
Add new javasrc2cpg type model and JarTypeReader

johannescoetzee:
This is the first PR in the project to move javasrc2cpg type solving away from the sometimes broken javaparser symbol solver. This introduces a new type model and the JarTypeReader which extracts the necessary information from JARs. None of this information is currently used, but PRs for that will follow.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/build.sbt
+++ b/joern-cli/frontends/javasrc2cpg/build.sbt
@@ -11,7 +11,9 @@ libraryDependencies ++= Seq(
   "org.gradle"              % "gradle-tooling-api"         % Versions.gradleTooling,
   "org.scalatest"          %% "scalatest"                  % Versions.scalatest % Test,
   "org.projectlombok"       % "lombok"                     % "1.18.26",
-  "org.scala-lang.modules" %% "scala-parallel-collections" % "1.0.4"
+  "org.scala-lang.modules" %% "scala-parallel-collections" % "1.0.4",
+  "org.scala-lang.modules" %% "scala-parser-combinators"   % "2.2.0",
+  "net.lingala.zip4j"       % "zip4j"                      % "2.11.5"
 )
 
 scalacOptions ++= Seq(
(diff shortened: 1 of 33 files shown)


joern-cli/frontends/javasrc2cpg/build.sbt
```diff
@@ -21,3 +23,35 @@ scalacOptions ++= Seq(
...
+  import net.lingala.zip4j.ZipFile
+  import net.lingala.zip4j.model.ZipParameters
+  import net.lingala.zip4j.model.enums.{CompressionLevel, CompressionMethod}
+  import java.nio.file.Paths
+
+  val pkgRoot = "io"
+  val testClassOutputPath = Paths.get("joern-cli", "frontends", "javasrc2cpg", "target", "scala-2.13", "test-classes")
+  val relativeTestCodePath = Paths.get(pkgRoot, "joern", "javasrc2cpg", "jartypereader", "testcode")
+
+  println("foo")
```
claudiuvursache, line 39:
```suggestion
```

johannescoetzee: Whoops!

mpollmeier: Johannes probably left it there to test if you are good enough to spot it. Really well done @claudiuvursache 💯 🏅

claudiuvursache:
I need some help understanding the reason for not using this work already...do the frontend tests pass? if yes, then why not switch directly? if not, then what are the open issues?

johannescoetzee:
> I need some help understanding the reason for not using this work already...do the frontend tests pass? if yes, then why not switch directly? if not, then what are the open issues?

This is the first part of a more research-y exploration of java types and the only purpose of this PR is to add the component that can read the type information we need from JARs. It is not yet the full type solver (and at this point it's still unclear what exactly that will look like). The plan is to create separate PRs for the various components that will eventually make up part of this, since that makes the components easier to review, creates a more structured breakdown and timeline of what is included in the solver and avoids the extra cost of maintaining a separate branch that's ever growing. 

So I guess the reason for not using this comes down to the scope of the PR vs the scope of the project as a whole. When we get to the point where the frontend gives the same results with the new types as with javasrc, then the scope of the project has been fulfilled. For now, however, the scope of this PR is just the JAR type reader which is working at least as far as shown by the unit tests included here.

claudiuvursache (approved):
Can't judge everything in this dense PR, but what I can judge LGTM

PR #2412, opened 2023-03-22
upgrade to JDK19

mpollmeier:
Core here is dropping the scalac flag `--target`. Previously I
thought that JDK19 doesn't support `--release=11`, but turns out they
only [dropped support for JDK7 in JDK20](https://bugs.openjdk.org/browse/JDK-8173605), so we can leave this as is for
now. `--target` is deprecated anyway, so this seems to be the best way
forward.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -11,11 +11,11 @@ jobs:
         with:
           fetch-depth: 1
       - uses: coursier/cache-action@v6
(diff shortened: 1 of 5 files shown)


xavierpinho (approved):

xavierpinho:
Nice! Tested (macOS with JDK 19) and no issues whatsoever 🎉

PR #2413, opened 2023-03-22
CfgNodeMethods: Handle TypeDecl Before Method during `.method` step

### LINKED ISSUE #2407: [Bug] NullPointerException in JavaScriptTypeRecovery Pass
The new JavaScriptTypeRecovery default overlay results in a non-recoverable crash and non-creation of a CPG on some samples.

**To Reproduce**
1. Import the attached Typescript project.
[ts_bug_sample.zip](https://github.com/joernio/joern/files/11029223/ts_bug_sample.zip)
2. Observe the following error:
```java
2023-03-21 13:53:35.489 ERROR CpgPassBase: Pass io.joern.jssrc2cpg.passes.JavaScriptTypeRecovery failed
java.util.concurrent.ExecutionException: java.lang.NullPointerException
	at java.util.concurrent.ForkJoinTask.get(ForkJoinTask.java:1006) ~[?:?]
	at io.joern.x2cpg.passes.frontend.XTypeRecovery.$anonfun$run$3(XTypeRecovery.scala:63) ~[io.joern.x2cpg-1.1.1536.jar:1.1.1536]
[...]
```

**Expectation:**
The CPG creation should finish correctly.
If not possible, the new pass should fail gracefully and explain why TypeRecovery could not be performed on the sample. 

**Desktop:**
 - OS: Linux pop-os 6.0.12-76060006-generic
 - Joern: v1.1.1536 
 - Java: javac 11.0.18

DavidBakerEffendi:
Since `jssrc2cpg` files are `:program` methods, if a type is dynamically declared under this, we have a TYPE_DECL<-AST-METHOD relationship.

If there is a decorator on top of the TYPE_DECL, children of the type decl's annotation have a CHILD<-CONTAINS-TYPE_DECL relationship. CHILD<-CONTAINS-TYPE_DECL is marked initially as invalid by CfgNodeMethods.walkUpContains but in this context it is valid.

Resolves #2407

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CfgNodeMethods.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CfgNodeMethods.scala
@@ -146,13 +146,16 @@ class CfgNodeMethods(val node: CfgNode) extends AnyVal with NodeExtension {
 
   private def walkUpContains(node: StoredNode): Method =
     node._containsIn.onlyChecked match {
-      case method: Method                                                => method
-      case typeDecl: TypeDecl if typeDecl.astParent.isInstanceOf[Method] =>
-        // For a language such as jssrc2cpg, types may be dynamically declared under procedures
-        typeDecl.astParent.asInstanceOf[Method]
-      case _: TypeDecl =>
-        // TODO - there are csharp CPGs that have typedecls here, which is invalid.
-        null
+      case method: Method => method
+      case typeDecl: TypeDecl =>
+        typeDecl.astParent match {
+          case method: Method =>
+            // For a language such as javascript, types may be dynamically declared under procedures
+            method
+          case _ =>
+            // there are csharp CPGs that have typedecls here, which is invalid.
+            null
+        }
     }
 
 }
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CfgNodeMethods.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CfgNodeMethods.scala
@@ -146,8 +146,11 @@ class CfgNodeMethods(val node: CfgNode) extends AnyVal with NodeExtension {
 
   private def walkUpContains(node: StoredNode): Method =
     node._containsIn.onlyChecked match {
-      case method: Method => method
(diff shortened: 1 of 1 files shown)


max-leuthaeuser (approved):

mpollmeier:
how about this as a slight variation? I also changed the comment
```scala
  private def walkUpContains(node: StoredNode): Method =
    node._containsIn.onlyChecked match {
      case method: Method =>
        method
      case typeDecl: TypeDecl =>
        // For a language such as javascript, types may be dynamically declared under procedures
        walkUpContains(typeDecl.astParent)
      case _: TypeDecl =>
        // TODO - there are csharp CPGs that have typedecls here, which is invalid.
        null
    }

```

DavidBakerEffendi:
> how about this as a slight variation? I also changed the comment
> 
> ```scala
>   private def walkUpContains(node: StoredNode): Method =
>     node._containsIn.onlyChecked match {
>       case method: Method =>
>         method
>       case typeDecl: TypeDecl =>
>         // For a language such as javascript, types may be dynamically declared under procedures
>         walkUpContains(typeDecl.astParent)
>       case _: TypeDecl =>
>         // TODO - there are csharp CPGs that have typedecls here, which is invalid.
>         null
>     }
> ```

Since there are two matching cases with `TypeDecl` only the first will be taken, should I still keep the if statement on the first one? e.g.
```scala
  private def walkUpContains(node: StoredNode): Method =
    node._containsIn.onlyChecked match {
      case method: Method =>
        method
      case typeDecl: TypeDecl if typeDecl.astParent.isInstanceOf[Method] =>
        // For a language such as javascript, types may be dynamically declared under procedures
        walkUpContains(typeDecl.astParent)
      case _: TypeDecl =>
        // TODO - there are csharp CPGs that have typedecls here, which is invalid.
        null
    }
 ```

**Comment by mpollmeier** [comment]
oops, sorry about that, yes you're right. so either your option from the last comment or this - whatever you prefer, I don't mind really:
```scala
  private def walkUpContains(node: StoredNode): Method =
    node._containsIn.onlyChecked match {
      case method: Method => method
      case typeDecl: TypeDecl =>
        typeDecl.astParent match {
          case method: Method => 
            // For a language such as javascript, types may be dynamically declared under procedures
            method
          case _ => 
            // there are csharp CPGs that have typedecls here, which is invalid.
            null
        }
    }
```

PR #2426, opened 2023-03-27
Expression as Child of TypeDecl under Method Bug

### LINKED ISSUE #2420: [Bug] ClassCastException in dataflowengineoss default pass using jssrc2cpg
When importing and creating CPGs for certain TypeScript projects, 
the default pass `dataflowengineoss` results in a `ClassCastException`, terminating early.

**To Reproduce**
1. Import one of the two attached TypeScript projects
[examples.zip](https://github.com/joernio/joern/files/11062318/examples.zip)

2. Observe the following error:
```
ERROR CpgPassBase: Pass io.joern.dataflowengineoss.passes.reachingdef.ReachingDefPass failed
java.lang.ClassCastException: null
	at jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[?:?]
[...]
```

**Expected behavior**
The dataflow pass should finish successfully.

 - Linux 5.15.0-1030-gcp 37-Ubuntu
 - Joern: v1.1.1544
 - Java: javac 11.0.18

DavidBakerEffendi:
Similar to [#2413](https://github.com/joernio/joern/pull/2413), if an expression is a child to a type declaration where the contains edge is the TypeDecl and not the Method then a class cast exception is thrown. This circumstance is common in `jssrc2cpg`.

Resolves #2420

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/generalizations/ExpressionTraversal.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/generalizations/ExpressionTraversal.scala
@@ -3,7 +3,7 @@ package io.shiftleft.semanticcpg.language.types.expressions.generalizations
 import io.shiftleft.codepropertygraph.generated.nodes._
 import io.shiftleft.codepropertygraph.generated.EdgeTypes
 import io.shiftleft.semanticcpg.language._
-import overflowdb.traversal.Traversal
+import overflowdb.traversal.{InitialTraversal, PathAwareTraversal, Traversal}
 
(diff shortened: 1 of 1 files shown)


DavidBakerEffendi:
@max-leuthaeuser Same code snippet in the issue that was similarly resolved should be sufficient: https://github.com/joernio/joern/issues/2407#issuecomment-1479505188

The reason this was not picked up earlier is that the test suite I ran it on was not a data-flow test suite. The expression is not a direct AST child to the type declaration I believe, it is more so that it is following `CONTAINS` edges and the first node is a type declaration not a method.

max-leuthaeuser (approved):

PR #2429, opened 2023-03-28
Implement python parser support for pattern matching.

ml86:
Pattern matching was introduced by python 3.10 and is now supported by
our python parser.
The python frontend still emits UNKNOWN node for pattern match
statements. A fine granular CPG representation will follow soon in a
separate PR.

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/pythonGrammar.jj
+++ b/joern-cli/frontends/pysrc2cpg/pythonGrammar.jj
@@ -94,6 +94,11 @@ public class PythonParser {
     }
     return stmt;
   }
+
+  void setCurrentToken(Token setTo) {
+    token = setTo;
+    jj_ntk = -1;
+  }
 }
 PARSER_END(PythonParser)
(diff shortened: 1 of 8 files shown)


joern-cli/frontends/pysrc2cpg/pythonGrammar.jj
```diff
@@ -727,48 +804,67 @@ ArrayList<istmt> statementsAndDedentOrEof(): {
...
-          // If something goes wrong trying to parse as python2
-          // statement, we can rely on the error recovery already for
-          // the initial parse error.
-        }
+    try {
+      token = reparseStartToken;
+      setCurrentToken(reparseStartToken);
+      Name name = name();
+
+      // We now test for parse alternatives which would either require a hugh
```
claudiuvursache, line 839:
```suggestion
      // We now test for parse alternatives which would either require a huge
```

claudiuvursache (approved):

PR #2435, opened 2023-03-29
XTypeRecoveryPass: Replace Global Table by Writing to CPG

DavidBakerEffendi:
* Write to CPG each iteration of the type recovery pass to get rid of global table. This should save some memory.
* Fix some bugs where `<var>` placeholder keys would get in the final output for type hints
* Created a wrapper pass `XTypeRecoveryPass` that can run multiple propagation passes internally
* Holding member nodes until the end to make sure the updates don't overwrite one another
* Fixed issue in `pysrc2cpg` where `Member` nodes would have `Seq("")` under `dynamicTypeHintFullName`

Next up is to optimize the cost per iteration given some heuristics and early stopping.

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/NodeBuilder.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/NodeBuilder.scala
@@ -65,15 +65,17 @@ class NodeBuilder(diffGraph: DiffGraphBuilder) {
     addNodeToDiff(typeRefNode)
   }
 
-  def memberNode(name: String, dynamicTypeHintFullName: Option[String] = None): nodes.NewMember = {
+  def memberNode(name: String): nodes.NewMember = {
     val memberNode = nodes
       .NewMember()
       .code(name)
       .name(name)
       .typeFullName(Constants.ANY)
-    dynamicTypeHintFullName.foreach(hint => memberNode.dynamicTypeHintFullName(hint :: Nil))
     addNodeToDiff(memberNode)
   }
+
+  def memberNode(name: String, dynamicTypeHintFullName: String): nodes.NewMember =
+    memberNode(name).dynamicTypeHintFullName(dynamicTypeHintFullName :: Nil)
 
   def bindingNode(): nodes.NewBinding = {
     val bindingNode = nodes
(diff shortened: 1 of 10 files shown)


joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonTypeRecovery.scala
```diff
@@ -287,8 +273,8 @@ class RecoverForPythonFile(
       // TODO: inheritsFromTypeFullName does not give full name in pysrc2cpg
       val baseTypeFullNames = cpg.typ.nameExact(baseTypes: _*).fullName.toSeq
       (parentTypes ++ baseTypeFullNames)
-        .map(_.concat(".<body>"))
-        .filterNot(t => t.toLowerCase.matches("(any|object)"))
+        .map(f => f + "." + f.split("\\.").last + "<body>")
```
fabsx00, line 276: This looks like it can throw a `NoSuchElementException` if `f` does not contain a `.`

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/PySrc2CpgFixture.scala
```diff
@@ -28,7 +28,9 @@ class PySrcTestCpg extends TestCpg with PythonFrontend {
     X2Cpg.applyDefaultOverlays(this)
     new ImportsPass(this).createAndApply()
     new DynamicTypeHintFullNamePass(this).createAndApply()
-    new PythonTypeRecovery(this).createAndApply()
+    val maxInt = 2
```
fabsx00, line 31: Looks like this isn't `maxInt` but rather the number of iterations?

fabsx00 (approved):

DavidBakerEffendi:
NB: Breaking change for third-party users is that `XTypeRecovery` needs to be replaced with `XTypeRecoveryPass`, e.g. `PythonTypeRecovery[Pass]`.

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/NodeBuilder.scala
```diff
@@ -65,13 +65,13 @@ class NodeBuilder(diffGraph: DiffGraphBuilder) {
     addNodeToDiff(typeRefNode)
   }
 
-  def memberNode(name: String, dynamicTypeHintFullName: String): nodes.NewMember = {
+  def memberNode(name: String, dynamicTypeHintFullName: Option[String] = None): nodes.NewMember = {
```
ml86, line 68: I would suggest providing two `memberNode` methods.
1. `memberNode(String, String): nodes.NewMember`
2. `memberNode(String): nodes.NewMember`

That way we dont have to wrap the second argument in `Option`/`Some`.

ml86 (approved):

PR #2444, opened 2023-03-31
Remove warning about handled unhandled php statements in namespace bodies

johannescoetzee:
Only specific types of statements were expected for class bodies, but when that code was shared with namespace bodies which can contain any statement, the fallback case wasn't cleaned up. The ASTs for these statements were created anyways, so this changes nothing about the functionality of the frontend while removing the warning for perfectly valid code.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/NamespaceTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/NamespaceTests.scala
@@ -1,8 +1,18 @@
 package io.joern.php2cpg.querying
 
 import io.joern.php2cpg.testfixtures.PhpCode2CpgFixture
+import io.shiftleft.semanticcpg.language._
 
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/NamespaceTests.scala
```diff
@@ -1,8 +1,18 @@
...
+  "namespaces should be able to contain statements as top-level AST children" in {
+    val cpg = code("""<?php
+        |namespace foo {
+        |  echo 0;
+        |}
+        |""".stripMargin)
 
-  // TODO
+    inside(cpg.namespaceBlock.name("foo").l) { case List(ns) =>
+      ns.astChildren.l.map(_.code) shouldBe List("echo 0")
```
maltek, line 15:
```suggestion
      ns.astChildren.code.l shouldBe List("echo 0")
```

not that it makes any difference at all, but that's the DSLy way to write this

maltek (approved):

PR #2459, opened 2023-04-04
XTypeRecovery: Major Performance Improvements and Regex Handling 

### LINKED ISSUE #2458: [Bug] DynamicTypeHintFullNamePass Injects Asterix Directly into Regex Pattern
**Describe the bug**
 Some imports in Python use an astrix (*) which is a special regex element. This is then added to a pattern and interpreted as a dangling astrix. This throws a regex-related exception.

**To Reproduce**

Importing [frappe](https://github.com/frappe/frappe) with PySrc2Cpg will trigger this.

**Expected behavior**
The import may not need to be resolved, but this should not crash CPG generation.

 - OS MacOs Ventura
 - Joern Version v1.1.1563
 - Java version 18

DavidBakerEffendi:
* Found `DynamicTypeHintFullNamePass` directly injected code into regex, made a note here to handle this better but, for now, simply using a `Try`
* Implemented early stopping if no changes were detected from previous iteration
* Found that whole graph traversals were being done where partial graph traversals were intended when propagating method and parameter types

Resolves #2458

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/JavaScriptTypeRecovery.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/JavaScriptTypeRecovery.scala
@@ -137,8 +137,8 @@ private class RecoverForJavaScriptFile(cpg: Cpg, cu: File, builder: DiffGraphBui
   }
 
   override protected def isField(i: Identifier): Boolean =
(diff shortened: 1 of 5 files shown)


max-leuthaeuser (approved):

maltek:
I've only skimmed the PR and am not sure how easy it would be to use it here - just in case you're not aware: there is a [`Pattern.quote`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#quote-java.lang.String-) for escaping values used in regexes.

PR #2464, opened 2023-04-05
Fix php namespace and method full names

johannescoetzee:
The main purpose of this PR is to establish fixed methodFullNames for method nodes. These deviate from the typical format seen in Joern front-ends for 2 main reasons:
- PHP uses `\` as a delimiter for namespaces 
- PHP does not support overloading via signatures, so signatures are omitted from the MFN.

The naming scheme also adopts different notation for functions, class methods and instance methods:
```php
// function `foo` in global namespace
foo

// function `foo` in namespace A\B
A\B\foo

// static method foo in class Foo in namespace A\B
A\B\Foo::foo

// instance method foo in class Foo in namespace A\B
A\B\Foo->foo
```

The `::` and `->` notation was chosen to mimic how these methods would be called (at least from external classes/code). More examples illustrating the naming scheme are shown in the unit tests. 

2 notable cases regarding method names:
- Constructors use the `Defines.ConstructorMethodName` (i.e. `<init>`), instead of the PHP default ` __construct`. The reasoning behind this is basically not knowing whether the closed source dataflow tracker treats the `<init>` name in any special way. If constructor names are only special at the policy level, then using `__construct` makes more sense.
- A `<clinit>` method (with the name again matching JVM since there's already precedent for that) is added for classes with constant or static variable intializers. This was part of a much earlier php2cpg PR and was added since I remember running into issues attempting to add arbitrary statements as direct children to TypeDecl nodes. Whether or not having this method is the correct choice is certainly up for debate, but while it is present, `<clinit>` is probably a fine name for it since there isn't a PHP equivalent, as far as I'm aware.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
@@ -165,21 +165,38 @@ class AstCreator(filename: String, phpAst: PhpFile, global: Global) extends AstC
     asts
   }
 
+  private def composeMethodFullName(
+    namespacePrefix: Option[String],
+    className: Option[String],
+    methodName: String,
+    isStatic: Boolean
+  ): String = {
+    val methodDelimiter = if (isStatic) StaticMethodDelimiter else InstanceMethodDelimiter
+
+    val nameWithClass = List(className, Some(methodName)).flatten.mkString(methodDelimiter)
+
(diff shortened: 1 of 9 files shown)


johannescoetzee:
The Windows test suite failed due to a php-composer setup issue. These errors are happening quite frequently now, so a follow-up TODO is to simplify the php-parser setup.

maltek (approved):
makes all sense to me :+1: .

I don't believe there's any special-casing for `<init>` and we could change that to match PHP conventions. (There's a method flag for constructors, so there's little reason to rely on the name for this.)

The backslashes will be a bit annoying to deal with in strings, regexes and policies but given that these names are shown to users in some circumstances and PHP code also uses backslashes it still seems like the best choice.

PR #2467, opened 2023-04-06
jssrc2cpg: fix handling of for-in/of loops with destructing

max-leuthaeuser:


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -442,4 +437,365 @@ trait AstForStatementsCreator { this: AstCreator =>
...
+    // end surrounding block:
+    scope.popScope()
+    localAstParentStack.pop()
+
+    val blockNodeChildren =
+      Seq(iteratorAssignmentAst, Ast(resultNode)) ++ loopVariableNodes.map(Ast(_)) :+ whileLoopAst.withChild(
+        whileLoopBlockAst
+      )
+    setArgumentIndices(blockNodeChildren)
+    Ast(blockNode).withChildren(blockNodeChildren)
```
claudiuvursache, line 605: you could use `blockAst` here

claudiuvursache (approved):

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForStatementsCreator.scala
max-leuthaeuser: Right, good point. The whole thing is target to some refactoring later on anyway. Then I will do this.

PR #2482, opened 2023-04-11
PythonTypeRecovery: Filter Type Hints by Existing Types

DavidBakerEffendi:
* Attempt to find instances of `TypeDecl` nodes that match the type hint path
* This fixes instances of partial paths that do not reflect the fully qualified name

This degrades performance slightly

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/DynamicTypeHintFullNamePass.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/DynamicTypeHintFullNamePass.scala
@@ -2,10 +2,12 @@ package io.joern.pysrc2cpg
 
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.codepropertygraph.generated.PropertyNames
+import io.shiftleft.codepropertygraph.generated.nodes.StoredNode
 import io.shiftleft.passes.CpgPass
 import io.shiftleft.semanticcpg.language._
 import overflowdb.BatchedUpdate
 
+import java.io.File
 import java.util.regex.Pattern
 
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/DynamicTypeHintFullNamePass.scala
```diff
@@ -47,12 +48,32 @@ class DynamicTypeHintFullNamePass(cpg: Cpg) extends CpgPass(cpg) {
...
+    importedEntity: String
+  ) = {
+    val typeFullName = typeHint.replaceFirst(Pattern.quote(alias), importedEntity)
+    val typeFilePath = typeFullName.replaceAll("\\.", File.separator)
+    val pythonicTypeFullName = typeFullName.split("\\.").lastOption match {
+      case Some(typeName) =>
+        typeFilePath.stripSuffix(s"${File.separator}$typeName").concat(s".py:<module>.$typeName")
+      case None => typeFullName
+    }
+    cpg.typeDecl.fullName(s".*$pythonicTypeFullName").l match {
```
fabsx00, line 73: `fullName` or `fullNameExact`?

fabsx00 (approved):

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/DynamicTypeHintFullNamePass.scala
DavidBakerEffendi: Definitely full name, since `app.foo.bar` is imported as `foo.bar` so this catches that, but I think Windows is failing since I'm injecting that variable into the regex raw.

PR #2492, opened 2023-04-12
pysrc2cpg: Inheritance of External Types

DavidBakerEffendi:
* Attempt to resolve the full path of an external type imported in a variety of ways
* Added a map to speed up retrieval
* Create corresponding type decl stubs for these imported types


joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/InheritanceFullNamePass.scala
```diff
@@ -1,39 +1,49 @@
...
-        val resolvedTypeDecls = resolveInheritedTypeFullName(t)
-        builder.setNodeProperty(t, PropertyNames.INHERITS_FROM_TYPE_FULL_NAME, resolvedTypeDecls.map(_.fullName))
-        cpg.typ
-          .fullNameExact(resolvedTypeDecls.fullName.toSeq: _*)
-          .foreach(tgt => builder.addEdge(t, tgt, EdgeTypes.INHERITS_FROM))
+        val resolvedTypeDecls = resolveInheritedTypeFullName(t, builder)
+        if (resolvedTypeDecls.nonEmpty) {
+          val fullNames = resolvedTypeDecls.map(_.fullName)
+          builder.setNodeProperty(t, PropertyNames.INHERITS_FROM_TYPE_FULL_NAME, fullNames)
+          cpg.typ
```
fabsx00, line 34: Seems like a potential bottleneck. Maybe initialize a map to increase lookup speed here?

fabsx00 (approved):

DavidBakerEffendi:
This pass has an additional bottleneck in that imports `foo.bar.Baz` can't be directly matched without regexing around `foo/bar.py:<module>.Baz` so it cannot change until those names are fixed. cc: @fabsx00

PR #2497, opened 2023-04-13
kotlin2cpg: create nodes for annotations on params

claudiuvursache:


claudiuvursache:
`x2cpg`'s `methodAst` now takes `Seq[Ast]` for the method parameters because these ASTs can have children e.g. ANNOTATION

maltek (approved):

PR #2501, opened 2023-04-13
pysrc2cpg: Removal of <body> type

DavidBakerEffendi:
There may be a reason for the presence of the `<body>` type, but this PR proposes to remove it as a way to simplify the `pysrc2cpg` frontend and aim for better consistency with other frontends. Open to discussion...

* Removed <body> type. The AST hierarchy now would look like `<module>->Foo<meta>` and `<module>->Foo`
* Methods are now directly under `Foo` alongside members
* Adjusted compensation for Python types in data-flow engine
* Adjusted tests that acknowledge <body>

Breaking change:
All Python method full names go from:
`Foo.Foo<body>.bar` to `Foo.bar`. See changed tests for more examples.


joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonTypeRecovery.scala
```diff
@@ -245,10 +233,10 @@ private class RecoverForPythonFile(cpg: Cpg, cu: File, builder: DiffGraphBuilder
     val constructorPaths = symbolTable
       .get(c)
       .map(_.stripSuffix(s"${pathSep}__init__"))
-      .map {
-        case t if t.endsWith("<body>") => t
-        case t                         => t.split(pathSep).lastOption.map(x => s"$t$pathSep$x<body>").getOrElse(t)
-      }
+//      .map {
```
fabsx00, line 236: remove uncommented code.

fabsx00 (approved):

PR #2506, opened 2023-04-13
Use joernio fork of php-parser for build robustness and char bug fix

johannescoetzee:
With this PR, php2cpg will use the joernio fork of PHP-parser at https://github.com/joernio/PHP-Parser instead of the standard php-parser release installed with composer. This comes with 2 benefits:
- the new version doesn't rely on PHP composer's autoloaders, so we can fetch the source directly from the release and avoid the step of installing the parser via composer. This has been quite flaky in the past, so the change should improve general setup stability
- the new version includes a fix for missing files found in wordpress, and any other codebase that uses single byte characters with values greater than 127 (in the case of wordpress, `\xFF`) 

This PR also adds a php test to distroTests.


joern-cli/frontends/php2cpg/build.sbt
```diff
@@ -6,6 +6,9 @@ name := "php2cpg"
 scalaVersion       := "2.13.8"
 crossScalaVersions := Seq("2.13.8", "3.2.2")
 
+lazy val phpParserVersion = settingKey[String]("php-parser version")
```
mpollmeier, line 9: do you ever plan to change this setting other than in this place, e.g. via `sbt phpParserVersion := "some-other-version" compile`? My guess would be 'no', in which case this doesn't need to be an sbt setting but rather a simple value. I.e. simply `val phpParserVersion = "4.15.4-charfix"` and drop the `= settingKey` line above. 
Same goes for `phpParserDlUrl`.

mpollmeier (approved):

itsacoderepo (approved):

joern-cli/frontends/php2cpg/build.sbt
johannescoetzee: That's a good point! I do think the answer is no, so I'll go simplify that.

mpollmeier:
@johannescoetzee this change removed `joern-cli/frontends/php2cpg/bin/installdeps.sh` which is stil referenced in https://github.com/joernio/joern/blob/b916a61/.github/workflows/upgrade-deps.yml#L23-L24 and therefor reddifies the 'upgrade-dependencies' runner: https://github.com/joernio/joern/actions/runs/4760723865/jobs/8461301739

I guess we can just drop that call?

PR #2508, opened 2023-04-14
c2cpg: fixed handling of relative paths as input CLI argument

max-leuthaeuser:
If c2cpg was invoked with a relative path to the input folder we kept it that way. That broke absolute to relative (and vice versa) path conversion later on ultimately resulting in absolute paths for CPG filenames.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/Main.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/Main.scala
@@ -32,8 +32,11 @@ final case class Config(
     }
   }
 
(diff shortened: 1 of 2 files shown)


maltek (approved):

maltek:
could we fix this in `toRelativePath` so the next frontend won't run into this same bug again?

max-leuthaeuser:
> could we fix this in `toRelativePath` so the next frontend won't run into this same bug again?

`toRelativePath` is in `c2cpg.utils.IOUtils` which is c2cpg specific. 
But I guess these helper methods should be placed into x2cpg with a later refactoring.

Issue #2502, opened 2023-04-13
[Bug] php2cpg: files should use relative paths


maltek:
```
> cpg.file.name.sorted
...
  "/home/mkraus/tmp/php/yii2/tests/framework/widgets/LinkSorterTest.php",
  "/home/mkraus/tmp/php/yii2/tests/framework/widgets/ListViewTest.php",
  "/home/mkraus/tmp/php/yii2/tests/framework/widgets/MenuTest.php",
...
ocular> cpg.metaData.root.head 
res44: String = "/home/mkraus/tmp/php/yii2"
```

Paths are supposed to be relative to `cpg.metaData.root`.

Issue #2504, opened 2023-04-13
[Bug] php2cpg: <global> methods should be namespaced by file name in the fullname


maltek:
Import any large directory tree and you'll end up with hundreds of methods all with a fullname `<global>`. Fullnames are supposed to be unique. The usual solution for this for synthetic functions is to put the filename (relative to the root) into the fullname.

PR #2509, opened 2023-04-14
Use relative filenames for php2cpg and prepend those to global namespace and methods

johannescoetzee:
Fixes https://github.com/joernio/joern/issues/2502 and https://github.com/joernio/joern/issues/2504

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
@@ -14,14 +14,14 @@ abstract class AstCreatorBase(filename: String) {
 
   /** Create a global namespace block for the given `filename`
     */
-  def globalNamespaceBlock(): NewNamespaceBlock = {
-    val absPath  = absolutePath(filename)
+  def globalNamespaceBlock(useAbsolutePath: Boolean = true): NewNamespaceBlock = {
+    val path     = if (useAbsolutePath) absolutePath(filename) else filename
     val name     = NamespaceTraversal.globalNamespaceName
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
```diff
@@ -14,14 +14,14 @@ abstract class AstCreatorBase(filename: String) {
 
   /** Create a global namespace block for the given `filename`
     */
-  def globalNamespaceBlock(): NewNamespaceBlock = {
-    val absPath  = absolutePath(filename)
+  def globalNamespaceBlock(useAbsolutePath: Boolean = true): NewNamespaceBlock = {
```
maltek, line 17: @max-leuthaeuser @johannescoetzee  the other uses of this helper seem to be `javasrc2cpg, `jssrc2cpg` and `c2cpg` - do you see any problem with changing its behavior to use relative paths? (Instead of adding the option for relative paths like this diff is currently doing.)

At the end of the day we shouldn't have absolute paths in those frontends either.

maltek (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
johannescoetzee: `javasrc2cpg` currently uses absolute paths as a default everywhere, so that just needs to be fixed in general. As far as this particular helper goes, however, I don't see any issue with using the relative path by default in javasrc.

max-leuthaeuser (approved):

PR #2514, opened 2023-04-17
javasrc2cpg: add tests for custom FlowSemantic

evilpan:
As discussed in Discord, we may need the abilities to customize dataflow engine, and these tests could be a start point.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/antlr4/io/joern/rubysrc2cpg/parser/RubyParser.g4
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/antlr4/io/joern/rubysrc2cpg/parser/RubyParser.g4
@@ -208,9 +208,9 @@ indexingArguments
 
 argumentsWithParentheses
     :   LPAREN wsOrNl* RPAREN
-    |   LPAREN arguments RPAREN
-    |   LPAREN expressions WS* COMMA wsOrNl* chainedCommandWithDoBlock wsOrNl* RPAREN
-    |   LPAREN chainedCommandWithDoBlock RPAREN
+    |   LPAREN wsOrNl* arguments wsOrNl* RPAREN
+    |   LPAREN wsOrNl* expressions WS* COMMA wsOrNl* chainedCommandWithDoBlock wsOrNl* RPAREN
+    |   LPAREN wsOrNl* chainedCommandWithDoBlock wsOrNl* RPAREN
     ;
 
 expressions
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/parser/InvocationWithParenthesesTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/parser/InvocationWithParenthesesTests.scala
@@ -1,0 +1,95 @@
+package io.joern.rubysrc2cpg.parser
+
+class InvocationWithParenthesesTests extends RubyParserAbstractTest {
+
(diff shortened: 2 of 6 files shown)


DavidBakerEffendi:
I have been working on something similar for a while that has been shelved for now, but I would love to get a discussion started around it. I currently have a DSL in a private fork that can generate semantics according to the DSL too:
```java
import java.sql.*;

/**
 * A JDBC SELECT (JDBC query) example program.
 */
class Query1 {

    public static void foo(String url) {
        try {
            Connection conn = DriverManager.getConnection(url,"","");
            Statement stmt = conn.createStatement();
            ResultSet rs;

            rs = stmt.executeQuery("SELECT Lname FROM Customers WHERE Snum = 2001");
            while ( rs.next() ) {
                String lastName = rs.getString("Lname");
                System.out.println(lastName);
            }
            conn.close();
        } catch (Exception e) {
            System.err.println("Got an exception!");
            System.err.println(e.getMessage());
        }
    }
}
```
Would generate
```scala 
​​"Query1.foo:void(java.lang.String)".external(
  _.p(1) ->> (
   _.t("java.sql.DriverManager").c("getConnection").a(1) ->> (
    _.t("java.sql.Connection").c("createStatement").a(0) -> 
     (_.t("java.sql.Statement").c("executeQuery").a(0)) ->> (
      _.t("java.sql.ResultSet").c("getString").a(0) -> (_.t("java.io.PrintStream").c("println").a(1)),
      _.t("java.sql.ResultSet").c("next").a(0),
     ),
    _.t("java.sql.Connection").c("close").a(0),
    ),
   _.p(1)
  )
)
```

The only snag is that is it quite verbose, but that is the tradeoff to create a more detailed summary I suppose. Right now, it doesn't support adding these to the behaviour of the data-flow engine, but perhaps feedback and discussion should start, and then we merge things bit by bit?

Another example:
```
foo(a, b, d):
  a.f = d
  return b.f
```
Becomes
```scala
"foo".internal(
	_.p(3) -> (_.p(1).fstore("f")),
	_.p(2).fload("f") -> (_.r)
)
```

evilpan:
For me, the main usage scenario of `FlowSemantic` is to model external (library/framework) methods with or without source code, such as methods in the `java.lang.*` or `android.*` package, or to skip the data flow analysis process for some complex methods. 

What's the use case for generating verbose semantics of certain method?

DavidBakerEffendi:
Mostly, since the semantics of one method may rely on the semantics of another, or may access a field which would affect whether a return would be tainted or not. The current syntax can only offer an approximation in these cases.

DavidBakerEffendi:
For this ticket, I can help introduce a means to add additional flows via a simple API that can be used during a Joern session, perhaps?

evilpan:
I agree, the current flow semantic syntax is too simple. Previously I was considering add wildcard/regex matching to the methodName. In my opinion, it would be always a good idea to add more flexible ways to configure the data-flow engine.

DavidBakerEffendi:
@evilpan This commit adds an entrypoint for custom semantics in the test suite via `OssDataFlow` https://github.com/joernio/joern/commit/b916a61db9c8ea25939683fa99b46fca4423426c. See in this line how to extend a test suite to pass these down: https://github.com/joernio/joern/commit/b916a61db9c8ea25939683fa99b46fca4423426c#diff-fc5ea0120cecc9b2da8786941350c33035fb0e202b836bc5d9507192c06a5cb4R21

evilpan:
That's neat! I'll try to add it `JavaSrc2CpgTestContext` later.

PR #2526, opened 2023-04-18
kotlin2cpg: lower anonymous objects in `val`s

claudiuvursache:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/passes/KtPsiToAst.scala
```diff
@@ -1536,7 +1542,54 @@ trait KtPsiToAst {
 
       val initAst = callAst(initCallNode, argAsts, Option(initReceiverAst))
       Seq(localAst, assignmentCallAst, initAst)
+    } else if (hasRHSObjectLiteral) {
+      val typedExpr = expr.getDelegateExpressionOrInitializer.asInstanceOf[KtObjectLiteralExpression]
+
+      val containingDeclaration = expr.getParent.getParent.asInstanceOf[KtNamedFunction] // TODO: fix!!!
```
max-leuthaeuser, line 1548: Yes, this looks bad (two times .getParent and a cast)
Could you make this more fail-safe?

max-leuthaeuser (approved):

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/passes/KtPsiToAst.scala
claudiuvursache: If it's too much for you too, will do

PR #2527, opened 2023-04-18
Use blockAst in c2cpg and jssrc2cpg

max-leuthaeuser:


claudiuvursache (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -375,34 +375,31 @@ trait AstCreatorHelper { this: AstCreator =>
...
-
     val calls = withIndex(d.getDesignators) { (des, o) =>
       val callNode = newCallNode(d, op, op, DispatchTypes.STATIC_DISPATCH, o)
       val left     = astForNode(des)
       val right    = astForNode(d.getOperand)
       callAst(callNode, List(left, right))
     }
     scope.popScope()
-    Ast(b).withChildren(calls)
+    blockAst(blockNode, calls.toList)
```
johannescoetzee, line 402: It doesn't look like actually uses the fact that the arguments are lists, so perhaps it would be cleaner to change the parameter type to expect a seq instead of converting to a list here.

johannescoetzee (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
max-leuthaeuser: That's actually an issue with `withIndex` here as it returns a Seq.
In most places we already have Lists.
But anyway, we are using a wild mixture of Lists and Seqs across all frontends for sub ASTs.
Guess that's a target for a larger refactoring later on.

PR #2528, opened 2023-04-18
JavaTypeRecoveryPass: Introducing `XTypeRecoveryPass` for Java

DavidBakerEffendi:
* This aids the current type recovery by helping propagate types across boundaries and having optional dummy types
* Fixed issue where `Call(typeFullName)` was being used to hold the type of the call base/receiver instead of return value.
* Fixed assumption on `PythonTypeHintCallLinker` where if there were non-dummy types it would pick the first one instead of checking if there is exactly one before setting it to `methodFullName`

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/TypeInferenceTests.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/TypeInferenceTests.scala
@@ -6,6 +6,8 @@ import io.shiftleft.codepropertygraph.generated.DispatchTypes
 import io.shiftleft.codepropertygraph.generated.nodes.{Identifier, Literal}
 import io.shiftleft.semanticcpg.language._
 
+import java.io.File
+
 class NewTypeInferenceTests extends JavaSrcCode2CpgFixture {
 
   "methodFullNames for unresolved methods in source" should {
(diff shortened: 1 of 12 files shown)


joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/TypeInferenceTests.scala
```diff
@@ -197,6 +199,58 @@ class NewTypeInferenceTests extends JavaSrcCode2CpgFixture {
...
+        |	}
+        |}
+        |""".stripMargin,
+      Seq("net", "javaguides", "hibernate", "NamedQueryExample.java").mkString(File.separator)
+    )
+
+    "should be resolved using dummy return values" in {
+      val Some(getResultList) = cpg.call("getResultList").headOption
+      // Changes the below from <unresolvedNamespace>.getResultList:<unresolvedSignature>(0) to:
+      getResultList.methodFullName shouldBe "org.hibernate.Session.createNamedQuery:<unresolvedSignature>(2).<returnValue>.getResultList"
```
johannescoetzee, line 244: For Java, it could be useful to preserve the signature of `getResultList`. I'm not sure this would actually be useful at this level of uncertainty, but it is information that could potentially disambiguate fullNames.

DavidBakerEffendi: @johannescoetzee I see, so this would become more like `org.hibernate.Session.createNamedQuery:<unresolvedSignature>(2).<returnValue>.getResultList:<unresolvedSignature>(0)` in this case?

johannescoetzee: Indeed, although as I mentioned above, I'm not sure this would be useful information. By the time we get to this point, I don't imagine we'd be able to say much more than "this call goes to library X" in any case, so treating `getResultList:<unresolvedSignature>(0)` and `getResultList:<unresolvedSignature>(1)` as distinct is unlikely to tell us anything useful.

Issue #2547, opened 2023-04-21
jssrc: ignore transpiled files if original source files get scanned


maltek:
There are repos like https://github.com/google/closure-templates/tree/master/javascript that have transpiled files and original source files next to each other. Scanning that with standard settings will result in both of these files in the CPG.

We can consider that user error, but maybe we can add some simple logic to drop files with a source map when the original source file is scanned at the same time?

max-leuthaeuser:
The project you linked there does not contain source map files.
Did you encounter projects with source maps files too?

Otherwise, there would be no way of determine if a file is actually result of some transpilation.

PR #2549, opened 2023-04-21
jssrc2cpg: added simple detection of transpiled files

max-leuthaeuser:
Fixes: https://github.com/joernio/joern/issues/2547

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala
@@ -205,10 +205,10 @@ class AstGenRunner(config: Config) {
 
   private def processEjsFiles(in: File, out: File, ejsFiles: List[String]): Try[Seq[String]] = {
     val tmpJsFiles = ejsFiles.map { ejsFilePath =>
-      val ejsFile           = File(ejsFilePath)
-      val maybeOriginalFile = File(ejsFilePath.stripSuffix(".ejs") + ".js")
-      if (isTranspiledFile(maybeOriginalFile.pathAsString)) {
-        maybeOriginalFile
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala
```diff
@@ -156,42 +155,70 @@ class AstGenRunner(config: Config) {
...
-      val sourceFileContent = IOUtils.readLinesInFile(ejsFile.path).mkString("", ls, ls)
-      val preprocessContent = new EjsPreprocessor().preprocess(sourceFileContent)
-      (out / in.relativize(ejsFile).toString).parent.createDirectoryIfNotExists(createParents = true)
-      val newEjsFile = ejsFile.copyTo(out / in.relativize(ejsFile).toString)
-      val jsFile     = newEjsFile.changeExtensionTo(".js").writeText(preprocessContent)
-      newEjsFile.createFile().writeText(sourceFileContent)
-      jsFile
+      val maybeOriginalFile = File(ejsFilePath.stripSuffix(".ejs") + ".js")
+      if (isTranspiledFile(maybeOriginalFile.pathAsString)) {
+        maybeOriginalFile
```
maltek, line 211: If I understand things right, then this reuses a transpiled `.js` file if it exists for a `.ejs` template instead of doing our own transpilation? Makes sense to me, but shouldn't the variable referring to the `.js` file than be named sth like `maybeTranspiledFile`?

maltek (approved):
cool, yes this is about what I was thinking :+1:

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala
max-leuthaeuser: Right, it should 👍

PR #2550, opened 2023-04-21
(wip) remove Traversal: adapt to changes (now it compiles)

bbrehm:
Tests don't compile so far. Will either continue next week or give to @mpollmeier to improve upon.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name                     := "joern"
 ThisBuild / organization := "io.joern"
 ThisBuild / scalaVersion := "2.13.8"
 
-val cpgVersion = "1.3.598"
+val cpgVersion = "1.3.598+1-103c8f04"
 
 lazy val joerncli          = Projects.joerncli
 lazy val querydb           = Projects.querydb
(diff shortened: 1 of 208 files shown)


bbrehm:
Tests now compile as well, but some still fail (not done yet). The diff is somewhat giant, though. 

I'd like to also get rid of the whole joern scripts stuff -- that currently errors out with ammonite telling me that something is wrong, but not _what_ is wrong. 

We already removed all scripting nonsense from codescience, any objections to removing it here as well? (the existing scripts would be turned into ordinary scala files).

Now that we can see how it feels to get rid of Traversal, what do you think @fabsx00 @ml86 @mpollmeier ?

The most annoying things are: 
1. `.head` and `.headOption` become `.next()` and `.nextOption()` on Iterator. 
2. There is a tendency to be clueless about what types one operates on. Hence sprinkled `.l` on things that are already lists. 
3. There is the annoying issue of conclicting imports (we talked about that at length with @mpollmeier @ml86 ).

mpollmeier (commented):
Looks good to me so far. Yes, in pure numbers this PR is big, but since it's mostly just about `.head` -> `.next` I don't have a problem with that, i.e. conceptually it's not big. 

I'll merge the scala3 branches shortly, which might simplify things around scripts and it fixes the "issue of conclicting imports" you mentioned above. And yes, big 👍🏻 for integrating those scripts into the regular build rather than script files.

PR #2552, opened 2023-04-21
Add option to specify php-parser path via an envvar

johannescoetzee:
This is useful (and perhaps necessary) for sl integration.


maltek (approved):

maltek:
could maybe downgrade those log messages to debug

PR #2553, opened 2023-04-21
Reuse more logic from `x2cpg`

ursachec:


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/Defines.scala
```diff
@@ -4,4 +4,5 @@ object Defines {
   val anyTypeName: String            = "ANY"
   val voidTypeName: String           = "void"
   val qualifiedNameSeparator: String = "::"
+  val empty                          = "<empty>"
```
maltek, line 7: you're also definining this as `io.joern.x2cpg.Defines.Empty` in this PR

maltek (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/Defines.scala
ursachec: ups, forgot to remove that one, thanks!

Issue #2561, opened 2023-04-24
[Bug] c2cpg: missing buffer argument for placement new


maltek:
```
joern> importCode.cpp.fromString("""
       void a() {
           char buf[80];
           new (buf) string("hi");
       }
       """)
joern> cpg.call.code.p           
res68: List[String] = List("new (buf) string(\"hi\")")
joern> cpg.call.argument.code.l 
res71: List[String] = List("string", "\"hi\"")
```

Note that there is no argument for `buf`. Can we add it as a 3rd parameter to `<operator>.new`? (It's a different order than in the source code, but it's more important that the arguments that are always there always have the same argument index.)

PR #2566, opened 2023-04-24
c2cpg: added missing argument for placement in NewExpression

max-leuthaeuser:
Fixes: https://github.com/joernio/joern/issues/2561


maltek (approved):
I just realized that the ideal representation would split up the arguments for the constructor and the arguments of the operator, instead of mixing both in the same call.

But I guess because we can't resolve those constructors calls in most cases anyway, it's not urgent right now.

Issue #2570, opened 2023-04-24
[Bug] c2cpg: pointer and array type inconsistencies


maltek:
pointer types (should all be `char*` imo):

```
joern> importCode.cpp.fromString("""
       struct x { char * z; };
       char *a(char *y) {
           char *x;
       }
       """)
joern> (cpg.member.name("z").typeFullName ++ cpg.parameter.name("y").typeFullName ++ cpg.local.name("x").typeFullName ++ cpg.method.name("a").methodReturn.typeFullName).l 
res157: List[String] = List("char", "char", "char", "char*")
```

array types (should all be one of `char[]` or `char[1]` imo, matter of definition):

```
joern> importCode.cpp.fromString("""
       struct x { char z[1]; };
       void a(char y[1]) {
           char x[1];
       }
       """) 
joern> (cpg.member.name("z").typeFullName ++ cpg.parameter.name("y").typeFullName ++ cpg.local.name("x").typeFullName).l 
res161: List[String] = List("char", "char[]", "char[1]")
```

Playing around with https://cdecl.org/ gives even weirder edge cases, e.g. "declare x as function returning pointer to array 5 of pointer to function returning char":

```
joern> workspace.reset ; importCode.cpp.fromString("""
       char (*(*x())[5])();
       """) 
joern> cpg.method.nameNot("<global>").l 
res166: List[Method] = List(
  Method(
    id = 7L,
    astParentFullName = "tmp.cpp:<global>",
    astParentType = "TYPE_DECL",
    code = "char  ()",                                 !
    columnNumber = Some(value = 6),
    columnNumberEnd = Some(value = 19),
    filename = "tmp.cpp",
    fullName = "",                                     !
    hash = None,
    isExternal = false,
    lineNumber = Some(value = 2),
    lineNumberEnd = Some(value = 2),
    name = "",                                         !
    order = 1,
    signature = "char ()"                              !
  )
)
joern> cpg.method.nameNot("<global>").methodReturn.typeFullName.l 
res167: List[String] = List("char")
```

PR #2574, opened 2023-04-25
c2cpg: fixed pointer and array type inconsistencies

max-leuthaeuser:
Fixes: https://github.com/joernio/joern/issues/2570


joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/AstCreationPassTests.scala
```diff
@@ -1953,6 +1950,40 @@ class AstCreationPassTests extends AbstractPassTest {
...
+    "be consistent with pointer types" in AstFixture("""
+        |struct x { char * z; };
+        |char *a(char *y) {
+        |  char *x;
+        |}
+      """.stripMargin) { cpg =>
+      cpg.member.name("z").typeFullName.head shouldBe "char"
+      cpg.parameter.name("y").typeFullName.head shouldBe "char"
+      cpg.local.name("x").typeFullName.head shouldBe "char"
+      cpg.method.name("a").methodReturn.typeFullName.head shouldBe "char"
```
max-leuthaeuser, line 1975: We do not append the pointer * to the type fullname.
1) we do not have a suitable representation for pointer types (type decls)
2) we won't find the corresponding typedecl at the backend without custom/adapted logic which would be C/C++ specific
3) does not have any impact on DF analysis
4) you can always have a look into the code field (e.g., locals) to identify if its a pointer (for AST-based queries)

maltek: I would much prefer having the `*` here. I want to do things like find all parameters that are pointers. Until we have a proper way to represent structural types in the CPG, checking whether the type name ends with a `*` would be a decent approximation for that. Parsing the code field would be much more annoying because then I'd have to figure out how to strip the non-type information from that code field (the name is usually in the end, except when it isn't...).

In the backend I think without any work we'd just end up creating stub typedecls for all these pointer types. Those won't have the information that `char*` is in any way related to `char`. While that would be nice to have, we're already in the exact same situation for e.g. `char[1]` or `signed char` and I'm not aware of it being much of a problem.

max-leuthaeuser:
```
class Foo { ... }

Foo *a() {
  return ...;
}
```
would result in a return type `Foo*` which we simply do not have. We have only `Foo`.

maltek: How is this different for `Foo[1]`? Won't the `TypeDeclStubCreator` just create it on-demand?

max-leuthaeuser: `Foo[1]` is actually a different type (an array of `Foo`).
Yes, the `TypeDeclStubCreator` would create it but we have no link between `Foo*` and `Foo` then for tracking.
Btw: we already had this discussion here in the Joern issues a few times (see Fabs comments).

max-leuthaeuser: Remaining pointer inconsistencies are resolved with the latest commit here.
Please have a look.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/MethodRefLinker.scala
```diff
@@ -78,24 +82,36 @@ object MethodRefLinker {
         srcNode
           .propertyOption(key)
           .filter { dstFullName =>
-            !srcNode.propertyDefaultValue(dstFullNameKey).equals(dstFullName)
+            !srcNode.propertyDefaultValue(dstFullNameKey).equals(dereferenceTypeFullName(dstFullName))
```
maltek, line 85:
```suggestion
            srcNode.propertyDefaultValue(dstFullNameKey) != dereferenceTypeFullName(dstFullName)
```

since we're already changing this line let's make it more scala-y

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/MethodRefLinker.scala
```diff
@@ -56,6 +54,12 @@ object MethodRefLinker {
   def nodesWithFullName(cpg: Cpg, x: String): mutable.Seq[NodeRef[_ <: NodeDb]] =
     cpg.graph.indexManager.lookup(PropertyNames.FULL_NAME, x).asScala
 
+  /** Types from C/C++ can be annotated with * to indicate being a reference. As our CPG schema currently lacks a
+    * separate field for that information the * is part of the type full name and needs to be removed when linking.
+    */
+  private def dereferenceTypeFullName(fullName: String): String =
```
maltek, line 60: this would feel better to me if we moved it into an `if language == newc` or a c-specific subclass

maltek (approved):
awesome, thank you :pray:

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/MethodRefLinker.scala
max-leuthaeuser: Good point.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/MethodRefLinker.scala
max-leuthaeuser: Done.

Issue #2579, opened 2023-04-25
[Bug] Embedded JS Template (.ejs) Files break JSSRC Frontend


pr0me:
**Describe the bug**
If `.ejs` files are present while importing a Javascript project in joern, the jssrc2cpg frontend breaks.  
Apparently it tries to import the files which involves some internal renaming / file extension assumptions, resulting in trying to open a non-existent file.  
Subsequently, no CPG is generated.

**To Reproduce**
1. run `importCode` in joern on the attached example. 
The example includes on `.js` file as language frontend inference will not work if _only_ `.ejs` files are present.
[example.zip](https://github.com/joernio/joern/files/11321623/example.zip)
2. Observe the following error:
```
WARN AstGenRunner$: 	- failed to parse '/path/to/example/bare-end.js': 'Unexpected token (26:7)'

java.io.FileNotFoundException: /path/to/example/bare-end.js (No such file or directory)
	at java.base/java.io.FileInputStream.open0(Native Method)
	at java.base/java.io.FileInputStream.open(FileInputStream.java:219)
	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)
	at scala.io.Source$.fromFile(Source.scala:94)
	at scala.io.Source$.fromFile(Source.scala:79)
	at io.shiftleft.utils.IOUtils$.readLinesInFile(IOUtils.scala:77)
	at io.joern.jssrc2cpg.passes.AstCreationPass.$anonfun$finish$1(AstCreationPass.scala:38)
```
<details><summary>Full Stack Trace</summary>
<p>

```java
2023-04-25 12:28:25.647 WARN AstGenRunner$: 	- failed to parse '/home/pr0me/Documents/code/phd/paper-typeclassifier/type-inference-models/evaluation/lambdanet_ds/test_set/maquette/example/bare-end.js': 'Unexpected token (26:7)'
java.io.FileNotFoundException: /home/pr0me/Documents/code/phd/paper-typeclassifier/type-inference-models/evaluation/lambdanet_ds/test_set/maquette/example/bare-end.js (No such file or directory)
	at java.base/java.io.FileInputStream.open0(Native Method)
	at java.base/java.io.FileInputStream.open(FileInputStream.java:219)
	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)
	at scala.io.Source$.fromFile(Source.scala:94)
	at scala.io.Source$.fromFile(Source.scala:79)
[...]

PR #2581, opened 2023-04-25
jssrc2cpg: fixed EJS preprocessing for special tags in script block

max-leuthaeuser:
Fixes: https://github.com/joernio/joern/issues/2579


ursachec (approved):

ursachec:
I think it would be nice to have a test case for special parsing fixes like these

max-leuthaeuser:
The EJS preprocessing is a nightmare already and I am looking for alternatives. Thats why I did not want to spend more time here. 🙂

Issue #2601, opened 2023-04-27
[Bug] Usage of CRLF crashes jssrc2cpg


xavierpinho:
**Describe the bug**
Trying to scan a file using CRLF (0x0d 0x0a) for newlines crashes jssrc2cpg.

**To Reproduce**
1. Scan the attached file via jssrc2cpg.sh: [example.zip](https://github.com/joernio/joern/files/11345139/example.zip)
2. Observe the following crash
```
java.util.NoSuchElementException: None.get
        at scala.None$.get(Option.scala:627) ~[org.scala-lang.scala-library-2.13.8.jar:?]
        at scala.None$.get(Option.scala:626) ~[org.scala-lang.scala-library-2.13.8.jar:?]
        at io.joern.jssrc2cpg.astcreation.AstCreatorHelper.getLineOfSource(AstCreatorHelper.scala:128) ~[io.joern.jssrc2cpg-1.1.1639.jar:1.1.1639]
        at io.joern.jssrc2cpg.astcreation.AstCreatorHelper.$anonfun$lineEnd$1(AstCreatorHelper.scala:120) ~[io.joern.jssrc2cpg-1.1.1639.jar:1.1.1639]
        at io.joern.jssrc2cpg.astcreation.AstCreatorHelper.$anonfun$lineEnd$1$adapted(AstCreatorHelper.scala:120) ~[io.joern.jssrc2cpg-1.1.1639.jar:1.1.1639]
        at scala.Option.map(Option.scala:242) ~[org.scala-lang.scala-library-2.13.8.jar:?]
        at io.joern.jssrc2cpg.astcreation.AstCreatorHelper.lineEnd(AstCreatorHelper.scala:120) ~[io.joern.jssrc2cpg-1.1.1639.jar:1.1.1639]
        at io.joern.jssrc2cpg.astcreation.AstCreatorHelper.lineEnd$(AstCreatorHelper.scala:120) ~[io.joern.jssrc2cpg-1.1.1639.jar:1.1.1639]
        at io.joern.jssrc2cpg.astcreation.AstCreator.lineEnd(AstCreator.scala:27) ~[io.joern.jssrc2cpg-1.1.1639.jar:1.1.1639]
        at io.joern.jssrc2cpg.astcreation.AstCreatorHelper.createBabelNodeInfo(AstCreatorHelper.scala:33) ~[io.joern.jssrc2cpg-1.1.1639.jar:1.1.1639]
```

**Expected behavior**

[...]

PR #2605, opened 2023-04-27
jssrc2cpg: Replace IOUtils.readLinesInFile with IOUtils.readEntireFile

xavierpinho:
Is there any catch on doing this? EDIT: I suppose BOM wouldn't be handled. Will rework that.

Fixes: #2601

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name                     := "joern"
 ThisBuild / organization := "io.joern"
 ThisBuild / scalaVersion := "2.13.8"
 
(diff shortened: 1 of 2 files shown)


max-leuthaeuser:
> EDIT: I suppose BOM wouldn't be handled. Will rework that.

Yes, exactly. IOUtils.readLinesInFile does quite a lot w.r.t. to BOM handling and strange UTF8 encoding.

maltek (approved):

PR #2608, opened 2023-04-27
Container images based on almalinux for joern

prabhu:
Almalinux 9 and 8 images for joern. Almalinux uses RedHat UBI as the source and is enterprise friendly.

The workflow uses a schedule to publish the images with the tag `nightly`, `master`, and `latest` based on version tags.

After approving this PR, please visit the package settings and make all the packages public. GitHub would create packages as private by default and send a billing notice to upgrade.

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -1,0 +1,76 @@
+name: Upload Container image
+
+on:
+  schedule:
+  - cron: "0 */12 * * *"
+  push:
+    branches:
+     - master
+    tags:
+      - 'v*'
+  workflow_dispatch:
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      packages: write
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            ghcr.io/joernio/joern
+
+      - name: Build and push Docker images
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: ci/Dockerfile.alma
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha,scope=joern
+          cache-to: type=gha,mode=max,scope=joern
+          provenance: true
+          sbom: true
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            ghcr.io/joernio/joern-alma8
+
+      - name: Build and push Docker images
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: ci/Dockerfile.alma8
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha,scope=joern-alma8
+          cache-to: type=gha,mode=max,scope=joern-alma8
+          provenance: true
+          sbom: true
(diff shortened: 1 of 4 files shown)


.github/workflows/containers.yml
```diff
@@ -0,0 +1,76 @@
+name: Upload Container image
+
+on:
+  schedule:
+  - cron: "0 */12 * * *"
+  push:
+    branches:
+     - master
+    tags:
+      - 'v*'
```
mpollmeier, line 10: Looks like this would get triggered too often, I'd say either 'master' or (better) the 'v' tag should be sufficient. Running this every 12 hours sounds 'no matter what' sounds a bit much...

.github/workflows/containers.yml
```diff
@@ -0,0 +1,76 @@
...
+  workflow_dispatch:
```
mpollmeier, line 11: what's this? can be dropped?

mpollmeier:
> publish the images with the tag nightly, master, and latest based on version tags.

Can you elaborate please? I don't quite understand, based on the proposed changes...

mpollmeier:
Why did you add Dockerfiles for alma8 and 9? Why not just 9?

mpollmeier:
maybe add a readme section (one-liner) on how to run the dockerfile?

mpollmeier:
when I build and run it locally, I get a rather strange encoding - probably not a problem but was just wondering if you know...
image

.github/workflows/containers.yml
prabhu: workflow_dispatch is useful for running this workflow manually. Useful for troubleshooting or forcing the update against master.

prabhu:
> Why did you add Dockerfiles for alma8 and 9? Why not just 9? Almalinux 9 requires the CPU to support SSE4.2. For kvm64 VM, use the Almalinux 8 version works.

prabhu:
> when I build and run it locally, I get a rather strange encoding - probably not a problem but was just wondering if you know...
> image

For this one, we need to pass some xterm env variable or some such to make docker support unicode characters. I forgot the values. If you find this we can get this added to the file.

.github/workflows/containers.yml
prabhu: The schedule will only run against the master branch. Push against master is required to test the master branch prior to creating a release.

prabhu:
> > when I build and run it locally, I get a rather strange encoding - probably not a problem but was just wondering if you know...
> > image
> 
> For this one, we need to pass some xterm env variable or some such to make docker support unicode characters. I forgot the values. If you find this we can get this added to the file.

Hurray, had to install couple of glibc packages and set the env variables.

```
❯ docker run --rm -it -v /tmp:/tmp -v $(pwd):/app:rw -w /app -t ghcr.io/joernio/joern
Compiling /app/(console)

     ██╗ ██████╗ ███████╗██████╗ ███╗   ██╗
     ██║██╔═══██╗██╔════╝██╔══██╗████╗  ██║
     ██║██║   ██║█████╗  ██████╔╝██╔██╗ ██║
██   ██║██║   ██║██╔══╝  ██╔══██╗██║╚██╗██║
╚█████╔╝╚██████╔╝███████╗██║  ██║██║ ╚████║
 ╚════╝  ╚═════╝ ╚══════╝╚═╝  ╚═╝╚═╝  ╚═══╝
Version: 1.1.1647
Type `help` or `browse(help)` to begin

joern>
```

.github/workflows/containers.yml
mpollmeier: 👍🏻

prabhu:
@mpollmeier Anything else needed to get this merged?

mpollmeier (approved):

mpollmeier:
@itsacoderepo can you please use your admin rights to configure the package settings as per the PR description?

mpollmeier:
@prabhu Niko updated the settings as per the PR description - can you please check that it works?

mpollmeier:
@prabhu looks like the gh action config needs some fixes
https://github.com/joernio/joern/actions/runs/4912708026

```
Invalid workflow file: .github/workflows/containers.yml#L58
The workflow is not valid. .github/workflows/containers.yml (Line: 58, Col: 13): The identifier 'meta' may not be used more than once within the same scope.
```

prabhu:
@mpollmeier PR to fix the workflow issue - https://github.com/joernio/joern/pull/2669

mpollmeier:
the fix was merged, now we're running into
```
 ERROR: failed to solve: failed to push ghcr.io/joernio/joern-alma8:master: failed commit on ref "manifest-sha256:5ae9ee18db13afddad7616a6eb3e1c03eb28b7dea6f718f0766864382c3e3944": unexpected status: 403 Forbidden
Error: buildx failed with: ERROR: failed to solve: failed to push ghcr.io/joernio/joern-alma8:master: failed commit on ref "manifest-sha256:5ae9ee18db13afddad7616a6eb3e1c03eb28b7dea6f718f0766864382c3e3944": unexpected status: 403 Forbidden
```
https://github.com/joernio/joern/actions/runs/4914514555/jobs/8775893593

I manually disabled the workflow for now (on https://github.com/joernio/joern/actions/workflows/containers.yml)

prabhu:
@mpollmeier could you rerun the workflow? Sometimes github might fail for weird reasons.

mpollmeier:
it's green now

Issue #2612, opened 2023-04-28
jssrc2cpg: Support for Dynamic Declared Members


DavidBakerEffendi:
**Is your feature request related to a problem? Please describe.**
In the code below, the fields dynamically declared under the constructor are not populated as members:
```javascript
class ClassA {
  constructor(param1, param2) {
    this.e = param1;
    this.f = param2;
  }
}
```
`cpg.typeDecl.nameExact("ClassA").member.l` will not reflect any members.

During debugging, under `AstForExpressionsCreator::astForMemberExpression` I do see these coming through and these are labelled as field accesses, but the corresponding members aren't created. 

Modifying the `JsClassesAstCreationPassTest::have member with initialization in TYPE_DECL for ClassA` to look like:
```javascript
class ClassA {
  a = 1
  b = "foo"
  static c = true
  static d
  static {
    this.d = false
  }

  constructor(param1, param2) {
    this.e = param1;
    this.f = param2;
  }
}
```
I can observe `List(a, b, c, d)`. I'm just not quite sure where to register `e` and `f` to become members.

cc @max-leuthaeuser 

**Describe the solution you'd like**
I would like `this.foo <operator>.assignment (Literal|Identifier|MethodRef|TypeRef)` calls within a type to be registered as members under the type.

**Additional context**
If you could point me to how to register the members then I can go ahead and do it myself, but I am slightly thrown off atm.

PR #2619, opened 2023-05-02
jssrc2cpg: added support for dynamically declared members

max-leuthaeuser:
Class members that are dynamically initialized in a class constructor are added as class member now. This init assignment call itself stays in the constructor block.

Fixes: https://github.com/joernio/joern/issues/2612


DavidBakerEffendi (approved):
Thanks! Looks good

PR #2622, opened 2023-05-02
x2cpg: rename node creation fns

ursachec:
The idea is to have two consistent ways of generating nodes:
1. The `AstNodeBuilder` trait which provides fns with signatures similar to: `aTypeOfNode(PROP1, PROP2, PROP3)`
2. The `NodeBuilders` object which provides fns with signatures similar to: `newATypeOfNode(PROP1, PROP2, PROP3, LINE, COLUMN)`

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
@@ -5,7 +5,7 @@ import io.shiftleft.codepropertygraph.generated.nodes.{ExpressionNew, NewNode}
 import io.shiftleft.codepropertygraph.generated.{DispatchTypes, Operators}
 import io.joern.x2cpg.Ast
 import io.joern.x2cpg.SourceFiles
-import io.joern.x2cpg.utils.NodeBuilders.dependencyNode
+import io.joern.x2cpg.utils.NodeBuilders.newDependencyNode
 import io.shiftleft.codepropertygraph.generated.EdgeTypes
 import io.shiftleft.utils.IOUtils
 import org.apache.commons.lang.StringUtils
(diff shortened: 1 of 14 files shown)


max-leuthaeuser (approved):

johannescoetzee (approved):

johannescoetzee:
I'm not sure of the best way to do this, but adding some way of enforcing this or documentation stating that this is the intent would be useful for keeping this consistent as more nodes are added.

xavierpinho (approved):
Left me wondering: should we next aim to have only `newATypeOfNode(PROP1, PROP2, PROP3, LINE, COLUMN)` kind of methods?

ursachec:
@johannescoetzee I've added a comment

johannescoetzee:
> Left me wondering: should we next aim to have only `newATypeOfNode(PROP1, PROP2, PROP3, LINE, COLUMN)` kind of methods?

Moving everything into the `AstNodeBuilder` trait would also be cleaner in this case.

ursachec:
@xavierpinho Why? In the case of the trait, all the frontends use a consistent invocation _and_ one doesn't need to set the _line_/_column_

ursachec:
> > Left me wondering: should we next aim to have only `newATypeOfNode(PROP1, PROP2, PROP3, LINE, COLUMN)` kind of methods?
> 
> Moving everything into the `AstNodeBuilder` trait would also be cleaner in this case.

true, nice idea!

johannescoetzee:
> Why? In the case of the trait, all the frontends use a consistent invocation and one doesn't need to set the line/column

They could perhaps have line and column as an optional parameter with `None` default. That way we can keep the existing behaviour unchanged, while also not having to worry about 2 different signature patterns (except perhaps for nodes which can't actually have a line and column set)

ursachec:
After having tried a few different options locally, I think it's best to keep this PR as is, remove as much of the individual NodeBuilders as possible, and only then look for further refactoring possibilities

PR #2623, opened 2023-05-02
jssrc2cpg: remove unncessary fn

ursachec:


ursachec:
the fn is identical to the one found in `x2cpg`
```
  
  /** For a given return node and arguments, create an AST that represents the return instruction. The main purpose of
    * this method is to automatically assign the correct argument indices.
    */
  def returnAst(returnNode: NewReturn, arguments: Seq[Ast] = List()): Ast = {
    setArgumentIndices(arguments)
    Ast(returnNode)
      .withChildren(arguments)
      .withArgEdges(returnNode, arguments.flatMap(_.root))
  }
  ```

PR #2626, opened 2023-05-02
Switch PHP types from empty to ANY and register them.

johannescoetzee:
This fixes a type related crash in a WIP closed-source pass.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
@@ -1086,7 +1086,7 @@ class AstCreator(filename: String, phpAst: PhpFile, global: Global) extends AstC
       case None =>
         // With variable variables, it's possible to use a valid variable without having an obvious assignment to it.
         // If a name is unknown at this point, assume it's a local that had a value assigned in some way at some point.
-        val local = NewLocal().name(identifier.name).code(s"$$${identifier.code}")
+        val local = NewLocal().name(identifier.name).code(s"$$${identifier.code}").typeFullName(identifier.typeFullName)
         scope.addToScope(local.name, local)
         diffGraph.addEdge(identifier, local, EdgeTypes.REF)
     }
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
```diff
@@ -1079,7 +1086,7 @@ class AstCreator(filename: String, phpAst: PhpFile, global: Global) extends AstC
       case None =>
         // With variable variables, it's possible to use a valid variable without having an obvious assignment to it.
         // If a name is unknown at this point, assume it's a local that had a value assigned in some way at some point.
-        val local = NewLocal().name(identifier.name).code(s"$$${identifier.code}").typeFullName(identifier.typeFullName)
+        val local = NewLocal().name(identifier.name).code(s"$$${identifier.code}")
```
maltek, line 1089: Why are we no longer setting a type here? (Not saying it's wrong, genuinely wondering.)

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/NodeBuilders.scala
```diff
@@ -107,7 +107,7 @@ object NodeBuilders {
       .code(code)
       .signature("")
       .dispatchType(DispatchTypes.STATIC_DISPATCH)
-      .typeFullName(typeFullName.getOrElse(PropertyDefaults.TypeFullName))
+      .typeFullName(typeFullName.getOrElse("ANY"))
```
maltek, line 110: Is there no generic constant for this? (It's used across most of our frontends, so there probably should be. But this doesn't have to block this PR.)

maltek (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/NodeBuilders.scala
johannescoetzee: Surprisingly no (at least not that I could find from a quick search). I agree it would be a good idea to have a single constant for this though.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
johannescoetzee: That's a good catch. There should be something there, so I suspect I removed that to replace it with something else (since we'd need to check that the typeFullName exists or use ANY otherwise) and forgot to do the replacement. I'll go through the PR to make sure I didn't miss another.

PR #2637, opened 2023-05-03
c2cpg: generate stable names for header file content

max-leuthaeuser:


mpollmeier:
tested with sptests locally: the test assertions (`anonymous_enum_0:<clinit>` and friends) are stable now between runs 🎉

mpollmeier:
tested with sptests locally: the test assertions (`anonymous_enum_0:<clinit>` and friends) are stable now between runs 🎉

mpollmeier (approved):

PR #2638, opened 2023-05-03
jssrc2cpg: ThisExpression & TypeOf Imports

DavidBakerEffendi:
There have been interesting type hints given for `this` keywords, e.g.
* `import("relative/or/absolute/path")` (where it originally would be `typeof import(...`)
* `this`
Whenever `this` is equalled to an import, it is always that of the `:program` type. Similarly, when the type hint for `this` is `this` I give the `:program` type (if the `dynamicInstanceTypeStack` is empty)

These observations can be found by querying `this` identifiers under [OpenWrap](https://github.com/PubMatic/OpenWrap)


max-leuthaeuser (approved):
Thanks! Very nice.

PR #2640, opened 2023-05-03
Use relative filenames in javasrc2cpg and fix name bug when using current working dir

johannescoetzee:


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
```diff
@@ -137,11 +137,8 @@ class JavaSrc2Cpg extends X2CpgFrontend[Config] {
...
 
     val analysisAstsMap = analysisSources.par.flatMap { sourceFilename =>
-      val originalFilename = sourceFilename.replaceAll(
-        // Pattern.quote used to escape Windows paths
-        escapeBackslash(sourceDirectories.analysisSourceDir),
-        escapeBackslash(config.inputPath)
-      )
+      val originalFilename =
+        Paths.get(sourceDirectories.analysisSourceDir).relativize(Paths.get(sourceFilename)).toString
       val sourceFileInfo  = SourceFileInfo(sourceFilename, originalFilename)
```
maltek, line 142: there's another instance of `SourceFileInfo` getting created a few lines below and in kt2cpg `JavasrcInterop`, that should probably get the same treatment? (Github doesn't let me comment there because it hasn't changed...)

maltek (approved):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
johannescoetzee: The instance a few lines below doesn't need to be changed since that path isn't actually used for anything. I did miss the one in JavasrcInterop, however, so that needs to be updated. I'll refactor this to reduce the code duplication for all the cases though.

PR #2649, opened 2023-05-04
c2cpg: remove the global lock

max-leuthaeuser:
**Background info:**

Eclipse CDT copies (physically and transitively) all of the content from included header files into the resulting Eclipse CDT AST. Before this PR, we simply took that as is and generated a CPG for all of that.

**Implications:**

Thus, we had to take care of duplicated content (e.g., if the same header file is included into multiple different *.c files). We did that by checking the contents filename, line, and column number. That needed to happen inside a global lock as we are iterating over all input files in parallel. To no one's surprise, this was the big performance bottleneck in c2cpg. Generating stable names also proved difficult.

**Approach taken in this PR:**

We now filter out all nodes with a different origin filename than the one associated to the AST we are currently traversing. That's easy because Eclipse CDT copies all the content from included header files to a top-level location (`IASTTranslationUnit#getDeclarations`). With that, Eclipse CDT is still able to resolve macros, types, and all the other interesting stuff that comes with header files.

**Benefits of this PR:**

1) Massive performance increase as there is no global lock anymore.
2) Generating stable names is much easier as we do not have to check for duplicated content (which used to shift around when iterating files in parallel).
3) We can now enable scanning for system header files by default. Without this PR here a simple `#include<stdio.h>` results in half of your systems C header files being part of the CPG. Now that stuff is only used for CDTs internal type resolving and macro resolution which is exactly what we want.

**TODO:**

Needs more testing on real-world applications. VLC or Firefox source code is perhaps a valuable target, especially w.r.t. performance and memory consumption.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/C2Cpg.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/C2Cpg.scala
@@ -21,7 +21,7 @@ class C2Cpg extends X2CpgFrontend[Config] {
       new AstCreationPass(cpg, AstCreationPass.SourceFiles, config, report).createAndApply()
       new AstCreationPass(cpg, AstCreationPass.HeaderFiles, config, report).createAndApply()
       new TypeNodePass(CGlobal.typesSeen(), cpg).createAndApply()
-      new HeaderContentPass(cpg, config).createAndApply()
+      new HeaderContentPass(cpg).createAndApply()
       report.print()
     }
   }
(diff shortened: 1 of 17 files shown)


fabsx00:
Nice! Yeah, please test on a bunch of real world examples and see if we can in fact enable scanning of system headers by default again.

max-leuthaeuser:
Results for CPG generation (c2cpg standalone, `-Xmx48G`, **with** system header files)
- vlc: 162mb CPG, 2.802.919 nodes, 1:21 mins

Results for CPG generation (c2cpg standalone, `-Xmx48G`, **without** system header files)
- vlc: 125mb CPG, 2.152.722 nodes, 0:26 mins

Without this PR (and with the same maximum heap settings) c2cpg did not even finish on my machine when running c2cpg with system header files. It ran into eternal GC collection and eventually got killed.

Nevertheless, I will switch off the inclusion of system header files by default like it was before so we do not change the default behavior to not confuse people.

fabsx00 (approved):

PR #2655, opened 2023-05-05
javasrc2cpg: Handling Generic Types + Type Arguments

DavidBakerEffendi:
* Fixed bug where `MethodReturn` type will be `ANY` for generic types due to type arguments being added to the search
* Created a prototype Type->TypeArgument tree for representing nested type arguments
* Added `evalType` edges to `Ast` class

Currently, my `Type-AST->TypeArgument(-AST>TypeArgument)*` edges are not working... I think this needs to be `Type-REF>TypeDecl->TypeParameter(-AST>TypeParameter)*`? 

I suppose `List<Long>` and `List<Integer>` get their own `TypeDecl` for this kind of schema? Right now, there is only `List`

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
@@ -1,6 +1,6 @@
 package io.joern.javasrc2cpg.passes
 
-import com.github.javaparser.ast.`type`.TypeParameter
+import com.github.javaparser.ast.`type`.{ClassOrInterfaceType, Type, TypeParameter}
 import com.github.javaparser.ast.{CompilationUnit, Node, NodeList, PackageDeclaration}
 import com.github.javaparser.ast.body.{
   AnnotationDeclaration,
@@ -145,8 +145,11 @@ import io.shiftleft.codepropertygraph.generated.nodes.{
   NewNamespaceBlock,
   NewNode,
   NewReturn,
+  NewType,
+  NewTypeArgument,
   NewTypeDecl,
-  NewTypeRef
+  NewTypeRef,
+  NewUnknown
 }
 import io.joern.x2cpg.{Ast, AstCreatorBase, Defines}
 import io.joern.x2cpg.datastructures.Global
@@ -698,7 +701,7 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
 
     val modifiers = List(newModifierNode(ModifierTypes.CONSTRUCTOR), newModifierNode(ModifierTypes.PUBLIC))
 
-    methodAstWithAnnotations(constructorNode, Seq(thisAst), bodyAst, returnNode, modifiers)
+    methodAstWithAnnotations(constructorNode, Seq(thisAst), bodyAst, Ast(returnNode), modifiers)
   }
 
   private def astForEnumEntry(entry: EnumConstantDeclaration): Ast = {
@@ -820,7 +823,7 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
       constructorNode,
       thisAst :: parameterAsts,
       bodyAst,
-      methodReturn,
+      Ast(methodReturn),
       modifiers,
       annotationAsts
     )
@@ -1003,7 +1006,15 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
     val expectedReturnType = Try(symbolSolver.toResolvedType(methodDeclaration.getType, classOf[ResolvedType])).toOption
     val returnTypeFullName = expectedReturnType
       .flatMap(typeInfoCalc.fullName)
-      .orElse(scopeStack.lookupVariableType(methodDeclaration.getTypeAsString, wildcardFallback = true))
+      .orElse(
+        scopeStack.lookupVariableType(methodDeclaration.getTypeAsString.takeWhile(_ != '<'), wildcardFallback = true)
+      )
+      .orElse(Option(s"${Defines.UnresolvedNamespace}.${methodDeclaration.getTypeAsString}"))
+    val typeNode = methodDeclaration.getType match {
+      case x: ClassOrInterfaceType if x.getTypeArguments.isPresent =>
+        astForGenericType(x)
+      case _ => Ast() // This will be created by some TypePass
+    }
 
     scopeStack.pushNewScope(MethodScope(ExpectedType(returnTypeFullName, expectedReturnType)))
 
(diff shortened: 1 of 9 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -1003,7 +1006,15 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
     val expectedReturnType = Try(symbolSolver.toResolvedType(methodDeclaration.getType, classOf[ResolvedType])).toOption
     val returnTypeFullName = expectedReturnType
       .flatMap(typeInfoCalc.fullName)
-      .orElse(scopeStack.lookupVariableType(methodDeclaration.getTypeAsString, wildcardFallback = true))
+      .orElse(
+        scopeStack.lookupVariableType(methodDeclaration.getTypeAsString.takeWhile(_ != '<'), wildcardFallback = true)
+      )
+      .orElse(Option(s"${Defines.UnresolvedNamespace}.${methodDeclaration.getTypeAsString}"))
```
johannescoetzee, line 1012: Should the generic type parameters be included in the unresolved case? For consistency with the rest, it feels like stripping them here would be good to.

I'm not saying this is the way to go though. I'm legitimately asking.

DavidBakerEffendi: That is a good idea, I should add that for consistency

DavidBakerEffendi:
Okay, I have a plan:

Java will simply give types with fully qualified type arguments (if possible), i.e. instead of `java.util.List` we have `java.util.List<java.lang.Long>` and then this will create a unique `Type` and `TypeDecl`. Then after `TypeUsagePass` we have, say, a `JavaGenericTypeHandlerPass` that can modify the structure of what it detects as generics to build the tree out correctly with `TypeParameter` and `TypeArgument` nodes [as intended](https://github.com/DavidBakerEffendi/codepropertygraph/blob/master/schema/src/main/scala/io/shiftleft/codepropertygraph/schema/Type.scala).

@johannescoetzee  does this sound more sane?

johannescoetzee:
Regarding the type edges, the [TypeUsagePass is perhaps a good reference](https://github.com/joernio/joern/tree/24a556b4563570be79e7676c00e9cfd9a8de1a50/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/base/TypeUsagePass.scala) (at least as a start). That creates Type->REF->TypeDecl edges, along with the eval types to match.

DavidBakerEffendi:
TypeUsagePass seems to use the linking util pretty blindly, I don't think polymorphic types will be handled well here. Seems like `TypeDecl->TypeParameter` e.g. `List<Comparable>` or whatever, but if I spawn a `Type->TypeArgument` where I have `List<Long>` and `List<Integer>` then I don't think the linking util will check that `Long` and `Integer` are both using `Comparable` as a base class.

But this is a future issue and probably more appropriate for internal generic classes in any case

johannescoetzee:
> Okay, I have a plan:

@DavidBakerEffendi  Fabs and I discussed this a while ago in the context of inheritsFrom and decided against including generic types in the typeFullName. This was a while ago, so I can't remember exactly why we decided this. I think it was for policy compatibility with the closed source java frontend. Since our closed source policies don't include the generic types, we omitted those from javasrc  *fullNames as well.

It might be possible to add a workaround on the closed source side, but for now I wouldn't go with this approach.

johannescoetzee:
> TypeUsagePass seems to use the linking util pretty blindly, I don't think polymorphic types will be handled well here. Seems like `TypeDecl->TypeParameter` e.g. `List<Comparable>` or whatever, but if I spawn a `Type->TypeArgument` where I have `List<Long>` and `List<Integer>` then I don't think the linking util will check that `Long` and `Integer` are both using `Comparable` as a base class.
> 
> But this is a future issue and probably more appropriate for internal generic classes in any case

Yeah, it doesn't handle polymorphic types well at all. I was just suggesting it as a reference for the currently expected edge types. There's definitely still a lot of room for improvement for how we handle generics in general.

johannescoetzee:
This is perhaps a conversation that @ml86 should be a part of, since he'll have a much better idea of what the implications of this are for the closed source dataflow tracker. I'm not sure how important uniqueness of types is, but I know uniqueness of methodFullNames is a topic that's been brought up.

ml86:
TypeArguments and TypeParameter nodes where added in the early days of the CPG as placeholders, but so far have never been implemented by any frontend. This is also the reason why there are basically no querying facilities, we just never needed them.
Trying to now add TypeArgument and TypeParamter nodes in javasrc2cpg would not disturb production as long as the full names and general structure is unchanged but i would very much expect this to be non straight forward since corresponding node and edge definitions have back than been written up quite quickly. So we will likely need to adjust some of the definitions as we try to implement it.

Changing the full names and structure is pretty much out of question since we need to keep compatibility to the internal byte code frontend.

I am curious why you need the generic types in the CPG? After all in the JVM world the type erased type names are what counts.

DavidBakerEffendi:
If the schema is up for change, then I think I have an okay solution and propose something with a diagram sometime.

DavidBakerEffendi:
A bit rudimentary diagram, but since the description of Type nodes are that they are an instance of a TypeDecl, then we can keep instances of them that have TypeArgument nodes that can be queried. These can all point to the same TypeDecl as usual.

Then I've also kept the default Type node that would be generated by TypeUsagePass. 

image

DavidBakerEffendi:
@ml86 and I had a chat about this one offline. The correct way to use the schema was explained, so I'll be able to handle that, but where we implement this comes with two options, where the main goal is to avoid duplication. It was also mentioned that `Type(fullName)` can be whatever as long as every unique `Type` node points to the correct single `TypeDecl`. E.g. `Type(java.util.List<java.lang.String>` and `Type(java.util.List<java.lang.Integer>` must point to `TypeDecl(java.util.List<T>)`. 

With that, here are two proposed options:

1) We register and log the operation for building the generic type structure "in-line" i.e., within the base passes. This avoids memory issues and makes this a non-optional feature. `TypeUsagePass` is what currently reads `typeFullName` properties of nodes generated by the `AST` and use these to generate de-duplicated `Type` nodes and do all the linking. This pass would need to be modified to understand generics.

2) This is perhaps simpler but comes with some memory overhead, but we accumulate all the nodes that evaluate to types with generic arguments where we have those details and persist this information in a separate pass, say, before `TypeUsagePass`. I believe `TypeUsagePass` is idempotent so whichever types are not accumulated and written, this pass will handle the rest. This is close to what is already in this PR with `CodeTree`. To negate the memory, I would propose to change the global structure to map `CodeTree -> Set[NodeRef]` whereas it now stands, it is `NodeRef -> CodeTree`. This means that we can easily add an off-switch to disable this process and leave `TypeUsagePass` as it is. 

3) Some combination of 1) and 2) where we accumulate the generic types and have the implementation in `TypeUsagePass`.

`CodeTree` is a nice structure as it can generate the respective language dependent type since the syntactic sugar is overridable - so can be useful as a `X2Cpg` data-structure.

DavidBakerEffendi:
Example for `List`
graphviz

PR #2660, opened 2023-05-05
remove traversal second attempt

bbrehm:
Second attempt to remove traversal, this time with smaller diff (because we retain head/headOption/last/lastOption). Tests pass locally with appropriate locally published upstream.


Liyw979:
Hi, Why do you remove traversals here?

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
```diff
@@ -43,7 +42,7 @@ class SourceTravsToStartingPointsTask[NodeType](sourceTravs: Traversal[NodeType]
     val sources: List[StoredNode] = sourceTravs
       .flatMap(_.toList)
       .collect { case n: StoredNode => n }
-      .dedup
+      .distinct
```
fabsx00, line 45: Looks like we're breaking all queries containing `dedup` here. Wondering if we could just introduce an alias.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
```diff
@@ -126,14 +119,14 @@ class LambdaTests extends JavaSrcCode2CpgFixture {
           val fallbackLocal = cpg.method.name(".*lambda.*").local.name("fallback").head
           fallbackClosureBinding.closureBindingId shouldBe fallbackLocal.closureBindingId
 
-          fallbackClosureBinding.outE.collectAll[Ref].map(_.inNode()).l match {
+          fallbackClosureBinding.outE.asScala.collectAll[Ref].map(_.inNode()).l match {
```
fabsx00, line 122: Can we introduce `collectAll` on `outE` and friends?

fabsx00:
@bbrehm are you on this?

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
mpollmeier: How about an extension method `collectAll` on `java.util.Iterator`? I guess that's what fabs had in mind here.

PR #2668, opened 2023-05-08
pysrc2cpg: Arg Index for Named Arguments

DavidBakerEffendi:
While according to the comments, arg index is ignored for named arguments, but the data-flow engine relies on argument index regardless for defining semantics. This PR adds the index to enable data-flow engine users to define the semantics of the flows.

Note: While named arguments can move parameters around out-of-order, perhaps the data-flow engine should match the semantic arguments on the method stub and not the call site.


DavidBakerEffendi:
I suppose if we have external method calls, then we won't know the true parameter order either. But perhaps we should populate the argument index to at least have the option to define semantics?

DavidBakerEffendi:
Spoke with @ml86 about this. We decided that the better long term solution would be to enable named parameter support on the data-flow tracker's side. If this turns out to be super non-trivial, then we can come back to this.

PR #2682, opened 2023-05-09
kotlin2cpg: lower inline ctor calls

ursachec:


joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallsToConstructorTests.scala
```diff
@@ -238,4 +238,54 @@ class CallsToConstructorTests extends KotlinCode2CpgFixture(withOssDataflow = fa
...
+  "CPG for code with ctor call as argument of another call" should {
+    lazy val cpg = code("""|fun main() {
+         |    val l = listOf(java.lang.Error("err"))
+         |    println(l)
+         |}
+         |""".stripMargin)
+
+    "contain a correctly lowered representation" in {
+      val List(c: Call) = cpg.call.code("listOf.*").l
+      c.methodFullName shouldBe "kotlin.collections.listOf:java.util.List(java.lang.Object)"
```
johannescoetzee, line 251: Is setting the parameter type to `java.lang.Object` instead of "java.lang.Error` intentional?

ursachec: yes, but I'm going through the fullnames again next to fix some issues the closed-source dataflow engine has with lambdas, so these might change

PR #2685, opened 2023-05-09
c2cpg: no duplicate method stubs for forward declarations

max-leuthaeuser:


max-leuthaeuser:
Yes, that would be the downside of this simple solution here as we are only matching signatures (where the fullname would differ).

PR #2686, opened 2023-05-09
XTypeHintCallLinker: Separated Speculated Methods Stubs

### LINKED ISSUE #2673: Separate Type Recovery Stubs in a Different Namespace
Normal method stubs are generated when after the base AST pass, which only sets method full names based on mostly factual information. The stubs generated by `XTypeRecovery` are speculatively built on heuristics, which may lead to invalid methods at times - where the value is mostly knowing what namespace the method is a part of. It would be good to separate these.

Some kind of namespace such as `speculative` might be good for methods generated by `XTypeRecovery` that were not already in the CPG after the base passes.

**Additional context**
Spoke offline with @ml86 about this. Relates to issues introduced by[ this PR](https://github.com/joernio/joern/pull/2282/files) where type stubs are generated speculatively from method full name in `X2Cpg` which is not sound.

DavidBakerEffendi:
MethodStubs generated by the type recovery that end up not matching up to an existing method (or associate with a type that does not exist) will be associated with a special namespace.

Resolves #2673

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeHintCallLinker.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeHintCallLinker.scala
@@ -2,8 +2,8 @@ package io.joern.x2cpg.passes.frontend
 
 import io.joern.x2cpg.passes.base.MethodStubCreator
 import io.shiftleft.codepropertygraph.Cpg
-import io.shiftleft.codepropertygraph.generated.nodes.{Call, Method, MethodBase, NewMethod}
-import io.shiftleft.codepropertygraph.generated.{EdgeTypes, PropertyNames}
+import io.shiftleft.codepropertygraph.generated.nodes._
+import io.shiftleft.codepropertygraph.generated.{EdgeTypes, NodeTypes, PropertyNames}
 import io.shiftleft.passes.CpgPass
 import io.shiftleft.proto.cpg.Cpg.DispatchTypes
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeHintCallLinker.scala
```diff
@@ -104,16 +105,58 @@ abstract class XTypeHintCallLinker(cpg: Cpg) extends CpgPass(cpg) {
...
+        stub
+          .astParentFullName(typeDecl.fullName)
+          .astParentType(NodeTypes.TYPE_DECL)
+        builder.addEdge(typeDecl, stub, EdgeTypes.AST)
+      }
+      stub
+    } else {
+      stub
+    }
+  }
```
ml86, line 138: I would recommend rewriting this to a method which calculated `astParentFullName` and `astParentType` not based on `stub: NewMethod` but rather `fullName: Stirng`. That way we can directly call `createMethodStub` with the approriate values and dont need to overwrite the after stub creation.

ml86 (changes requested):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeHintCallLinker.scala
DavidBakerEffendi: @ml86 made the adjustment

Issue #2590, opened 2023-04-26
[Bug] php2cpg: some special methods not linked to files


maltek:
at least `<global>`, `__construct` and `<clinit>` methods seem to only have their `filename` property set to `<empty>` and missing a `SOURCE_FILE` edge from the methods to the corresponding file nodes.

PR #2688, opened 2023-05-09
Add filenames to methods that were missing them

johannescoetzee:
Resolves https://github.com/joernio/joern/issues/2590


maltek (approved):
also in `astForStaticAndConstInits`, please

Issue #2602, opened 2023-04-27
[Bug] php2cpg: constructor calls and methods have mismatched fullnames


maltek:
constructor calls still have a `methodFullName` like `FooClass.__construct:<unresolvedSignature>(4)` that does not match the simple `FooClass.__construct` fullname on method nodes.

PR #2702, opened 2023-05-11
Switch to new node builders in php and remove signatures from call names

johannescoetzee:
The main change in this is switching to using the node builders in `AstNodeBuilder` throughout php2cpg's AstCreator. I've also done a bit of refactoring to be able to provide a sensible `PhpNode` argument for the new `*Node` methods.

This also removes signatures from the `methodFullName` for calls, making them more consistent with the naming scheme used for method nodes. This also resolves https://github.com/joernio/joern/issues/2602.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
@@ -52,21 +52,20 @@ class AstCreator(filename: String, phpAst: PhpFile, global: Global)
 
     scope.pushNewScope(namespaceBlock)
 
-    val globalTypeDecl = NewTypeDecl()
-      .name(namespaceBlock.name)
-      .fullName(namespaceBlock.fullName)
-      .astParentFullName(namespaceBlock.fullName)
-      .code(namespaceBlock.code)
-      .filename(filename)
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
```diff
@@ -147,16 +146,19 @@ class AstCreator(filename: String, phpAst: PhpFile, global: Global)
...
+    val thisNode = parameterInNode(
+      originNode,
+      name = NameConstants.This,
+      code = NameConstants.This,
+      index = 0,
+      isVariadic = false,
+      evaluationStrategy = EvaluationStrategies.BY_SHARING,
+      typeFullName = typeFullName
+    ).dynamicTypeHintFullName(typeFullName :: Nil)
+    // TODO Add dynamicTypeHintFullName to parameterInNode param list
```
johannescoetzee, line 161: I left this as a TODO for this PR since it's already getting big and this change would affect other frontends as well. Was this intentionally omitted? If so, then keeping the code as is and just removing the TODO would be the way to go.

ursachec: I didn't add that parameter because I wanted to keep the PRs with as minimal as possible, but I don't see why not add it as an optional parameter in the near future

PR #2707, opened 2023-05-12
jssrc2cpg: bump astgen version

max-leuthaeuser:
Also: small cleanups


joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/io/CodeDumperTest.scala
```diff
@@ -59,10 +59,6 @@ class CodeDumperTest extends JsSrc2CpgSuite {
       )
     }
 
-    "methodCode should return nothing on invalid filename" in {
```
max-leuthaeuser, line 62: I removed it because it tests CoreDumper functionality and not jssrc2pg itself.

ursachec (approved):

PR #2718, opened 2023-05-15
Make BindingTable independent from JavaParser library types.

ml86:


ursachec (approved):

fabsx00:
I guess we have discussed my stance on `util` directories at length :D

PR #2720, opened 2023-05-15
Ghidra 10.3 support

prabhu:
Not tested yet. Please share a build to test this or share some instructions.


mpollmeier (approved):
Looks good. For testing you can build it locally with `sbt stage`, and then run with `./joern` in the project dir. I just did and it seem works just fine.

PR #2722, opened 2023-05-16
ossdataflowengine: fix edge case when tracking from receivers

fabsx00:
This problem became apparent after https://github.com/joernio/joern/pull/2680 was merged: when using a receiver as a source, and we manage to track back all the way to an invocation on that receiver, we do not consider the call on that receiver to be a source as well. In fact, it's somewhat debatable whether we should. The reason I believe we should is that if we are unable to resolve the method, we will actually jump back to the receiver, while if we can resolve it, we don't. In order to be consistent regardless of whether the method can be resolved or not, this change is necessary.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
@@ -82,7 +82,7 @@ class SourceToStartingPoints(src: StoredNode) extends RecursiveTask[List[CfgNode
           List(x).collectAll[CfgNode].toList ++ x.refsTo.collectAll[Local].flatMap(sourceToStartingPoints)
         ) ++ x.refsTo.capturedByMethodRef.referencedMethod.flatMap(m => usagesForName(x.name, m))
       case x: Call =>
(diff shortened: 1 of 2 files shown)


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
```diff
@@ -81,6 +81,8 @@ class SourceToStartingPoints(src: StoredNode) extends RecursiveTask[List[CfgNode
         withFieldAndIndexAccesses(
           List(x).collectAll[CfgNode].toList ++ x.refsTo.collectAll[Local].flatMap(sourceToStartingPoints)
         ) ++ x.refsTo.capturedByMethodRef.referencedMethod.flatMap(m => usagesForName(x.name, m))
+      case x: Call =>
+        (x._receiverIn.l ++ List(x)).collect { case y: CfgNode => y }
```
DavidBakerEffendi, line 85:
```suggestion
        (x._receiverIn.l :+ x).collect { case y: CfgNode => y }
```

DavidBakerEffendi (approved):
Just one more Scala-centric style suggestion

ml86 (approved):

Issue #2710, opened 2023-05-12
[Bug] php2cpg: global variables missing AST edges


maltek:
```
joern> workspace.reset ; importCode.php.fromString("""<? $a = 1;""") 
res0_1: Cpg = Cpg (Graph [25 nodes])

joern> cpg.local.head 
res9: Local = Local(
  id = 4L,
  closureBindingId = None,
  code = "$a",
  columnNumber = None,
  dynamicTypeHintFullName = ArraySeq(),
  lineNumber = Some(value = 1),
  name = "a",
  order = -1,
  typeFullName = "ANY"
)

joern> cpg.local.head.astParent 
java.util.NoSuchElementException
  io.shiftleft.Implicits$JavaIteratorDeco$.onlyChecked$extension(Implicits.scala:43)
  io.shiftleft.semanticcpg.language.nodemethods.AstNodeMethods$.astParent$extension(AstNodeMethods.scala:69)
  ammonite.$sess.cmd7$.<clinit>(cmd7.sc:1)

joern> cpg.local.head.location 
java.util.NoSuchElementException: next on empty iterator
  scala.collection.Iterator$$anon$19.next(Iterator.scala:966)
  scala.collection.Iterator$$anon$19.next(Iterator.scala:964)
  scala.collection.Iterator$$anon$10.next(Iterator.scala:611)
  scala.collection.IterableOps.head(Iterable.scala:222)
  scala.collection.IterableOps.head$(Iterable.scala:222)
  overflowdb.traversal.Traversal.head(Traversal.scala:16)
  io.shiftleft.semanticcpg.language.nodemethods.LocalMethods$.location$extension(LocalMethods.scala:10)
  ammonite.$sess.cmd8$.<clinit>(cmd8.sc:1)
```

PR #2727, opened 2023-05-16
Refactor php2cpg file AST creation

johannescoetzee:
With these changes, we re-use the existing `astForClassLike` and `astForMethodDecl` methods to create the ASTs for the fake global method and typeDecl.

This also fixes https://github.com/joernio/joern/issues/2710

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
@@ -23,14 +23,16 @@ import io.shiftleft.passes.IntervalKeyPool
 import io.shiftleft.semanticcpg.language.types.structure.NamespaceTraversal
 import org.slf4j.LoggerFactory
 import overflowdb.BatchedUpdate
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
```diff
@@ -793,12 +807,11 @@ class AstCreator(filename: String, phpAst: PhpFile, global: Global)
     astForMethodDecl(constructorDecl, fieldInits, isConstructor = true)
   }
 
-  private def prependNamespacePrefix(name: String): String = {
+  private def prependNamespacePrefix(name: String, includeGlobalFullName: Boolean = false): String = {
```
maltek, line 810: seems this new parameter is unused?

maltek (approved):

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
johannescoetzee: Good catch!

Issue #2703, opened 2023-05-11
php2cpg: Ignore .git, vendor and test directories


johannescoetzee:
We're currently generating the CPG using all php sources in the project directory, which includes a lot of essentially useless code. Filter out the .git, vendor and test directories by default, and add an exclude option to override this behaviour.

PR #2733, opened 2023-05-17
Add config to TestCpg and filter php files

johannescoetzee:


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/Main.scala
```diff
@@ -28,7 +29,11 @@ private object Frontend {
       programName("php2cpg"),
       opt[String]("php-ini")
         .action((x, c) => c.copy(phpIni = Some(x)))
-        .text("php.ini path used by php-parser. Defaults to php.ini shipped with Joern.")
+        .text("php.ini path used by php-parser. Defaults to php.ini shipped with Joern."),
+      opt[Seq[String]]("exclude-overrides")
```
maltek, line 33: can we split this in two? I think from an end-user's perspective they either want to disable the default exclusions or add additional exclusions, but rarely both at the same time.

johannescoetzee:
Changing to WIP pending https://github.com/joernio/joern/pull/2741

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/FileTests.scala
```diff
@@ -0,0 +1,54 @@
...
+      "tests/sub/e.php",
+      "vendor/autoload.php",
+      "misc/g.php",
+      ".git/h.php",
+      "sub/tests/i.php",
+      "nottests/j.php",
+      "<unknown>"
+    ).map(PathFilter.standardisePath)
+  }
+}
```
maltek, line 54: with other frontends we've seen bugs with filtering for untested combinations of absolute and relative paths for a) the input path and b) the excluded path. Can you add some tests for here to ensure that works as intended in all cases?

maltek (approved):

max-leuthaeuser:
Sorry, I did not see this PR.
Yeah, https://github.com/joernio/joern/pull/2741 should make filtering easier.

johannescoetzee:
> Sorry, I did not see this PR. Yeah, #2741 should make filtering easier.

No worries! The extracted filtering is definitely the better way to go

gacevedo:
@max-leuthaeuser with @johannescoetzee on vacation, could you please refactor this PR to make use of https://github.com/joernio/joern/pull/2741? Needed to fix https://github.com/joernio/joern/issues/2703. Thank you!

gacevedo:
> @max-leuthaeuser with @johannescoetzee on vacation, could you please refactor this PR to make use of #2741? Needed to fix #2703. Thank you!

Never mind. Looks like the PR was reverted.

PR #2736, opened 2023-05-18
[ossdataflowengine] code clean up and remove redundant check

Liyw979:
From comment in `TaskSolver.scala`:
> This method stays within the method (intra-procedural analysis) and terminates at method parameters and at output
> arguments

Since the variable `path` in `TaskSolver` contains only nodes within the same method, 
`isArgOrRetOfMethodWeCameFrom` will always return false.


fabsx00:
Hm... you may be right about that: it used to be the case that nodes on the path can be from other methods, but I believe that was refactored some time in January when I was doing the work on performance improvements of the data flow engine. Give me a bit of time to look at this in detail.

PR #2737, opened 2023-05-18
Disables python inheritance pass

prabhu:


fabsx00:
@DavidBakerEffendi will look into the performance issues on Monday, so, closing this one.

PR #2749, opened 2023-05-22
Performance Improvement: Inheritance and Type Recovery

DavidBakerEffendi:
Both passes initially started out as small, rudimentary passes but have become more and more powerful in what they offer and, as such, how expensive they are. They were designed on `CpgPass` which is ideally for small tasks, but these were now moved to `ConcurrentWriterPass` which exploits the independent task nature of how these passes operate. 

* Inheritance pass caches its expensive regex matching strategy and now uses `ConcurrentWriterPass`
* `XTypeRecoveryPass` now uses `ConcurrentWriterPass` but it could perhaps be re-designed around this better further down the line

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/JavaScriptTypeRecovery.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/JavaScriptTypeRecovery.scala
@@ -20,19 +20,18 @@ class JavaScriptTypeRecoveryPass(cpg: Cpg, config: XTypeRecoveryConfig = XTypeRe
 }
 
 private class JavaScriptTypeRecovery(cpg: Cpg, state: XTypeRecoveryState) extends XTypeRecovery[File](cpg, state) {
-  override def compilationUnit: Traversal[File] = cpg.file
+  override def generateParts: Array[File] = cpg.file.toArray
 
-  override def generateRecoveryForCompilationUnitTask(
-    unit: File,
-    builder: DiffGraphBuilder
-  ): RecoverForXCompilationUnit[File] =
-    new RecoverForJavaScriptFile(
-      cpg,
-      unit,
-      builder,
-      state.copy(config =
-        state.config.copy(enabledDummyTypes = state.isFinalIteration && state.config.enabledDummyTypes)
-      )
+  override def runOnPart(builder: DiffGraphBuilder, unit: File): Unit =
+    changeTracker.addOne(
+      new RecoverForJavaScriptFile(
+        cpg,
+        unit,
+        builder,
+        state.copy(config =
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/JavaScriptTypeRecovery.scala
```diff
@@ -20,19 +20,18 @@ class JavaScriptTypeRecoveryPass(cpg: Cpg, config: XTypeRecoveryConfig = XTypeRe
...
-      state.copy(config =
-        state.config.copy(enabledDummyTypes = state.isFinalIteration && state.config.enabledDummyTypes)
-      )
+  override def runOnPart(builder: DiffGraphBuilder, unit: File): Unit =
+    changeTracker.addOne(
+      new RecoverForJavaScriptFile(
+        cpg,
+        unit,
+        builder,
+        state.copy(config =
```
fabsx00, line 31: Use a `val` here for increased readability

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonTypeRecovery.scala
```diff
@@ -23,19 +23,18 @@ class PythonTypeRecoveryPass(cpg: Cpg, config: XTypeRecoveryConfig = XTypeRecove
...
-      )
+  override def generateParts: Array[File] = cpg.file.toArray
+
+  override def runOnPart(builder: DiffGraphBuilder, unit: File): Unit =
+    changeTracker.addOne(
+      new RecoverForPythonFile(
+        cpg,
+        unit,
+        builder,
+        state.copy(config =
```
fabsx00, line 34: Use a val here for increased readability

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeRecovery.scala
```diff
@@ -110,35 +111,18 @@ abstract class XTypeRecoveryPass[CompilationUnitType <: AstNode](
...
+    extends ConcurrentWriterCpgPass[CompilationUnitType](cpg) {
 
-  override def run(builder: DiffGraphBuilder): Unit = {
-    val changesWereMade = compilationUnit
-      .map(unit => generateRecoveryForCompilationUnitTask(unit, builder).fork())
-      .map(_.get())
+  protected def changeTracker: ArrayBuffer[Boolean] = mutable.ArrayBuffer.empty[Boolean]
+
+  override def finish(): Unit = {
+    val changesWereMade = changeTracker
```
fabsx00, line 120: `changesMade` is shorter and conveys the same.

fabsx00 (approved):

PR #2751, opened 2023-05-22
Revert "x2cpg: extracted file exclusion by regex and file path (#2741)"

ursachec:
This reverts commit 152842f026d7704e0ee55ea4c22cb22ebbdebe3a.


ursachec:
The output path is not set correctly in certain scenarios:
Works:
```
$ mkdir -p /tmp/k1 && echo 'fun main() = println("hi")' >> /tmp/k1/main.kt
$ ./kotlin2cpg.sh /tmp/k1 -o xxx.bin.zip
$ ls xxx.bin.zip | wc -l                
1
```

Does not work:
```
$ mkdir -p /tmp/k1 && echo 'fun main() = println("hi")' >> /tmp/k1/main.kt
$ ./kotlin2cpg.sh /tmp/k1 -o xxx.bin.zip --classpath /tmp/k1
$ ls xxx.bin.zip | wc -l
ls: cannot access 'xxx.bin.zip': No such file or directory
0

```

max-leuthaeuser:
I am not sure if that's actually an issue with the PR itself or the command line parser from x2cpg. I can't check this as I am off for two weeks from right now on.

max-leuthaeuser (approved):

johannescoetzee:
> I am not sure if that's actually an issue with the PR itself or the command line parser from x2cpg. I can't check this as I am off for two weeks from right now on.

@max-leuthaeuser I looked into this a bit since I'm still interested in these changes for php2cpg, and I think the issue is that the `copy` method generated by the frontend-specific config case class doesn't copy the values defined in the parent trait. I'm still working on fixing that, but just letting you know to avoid duplicate work.

max-leuthaeuser:
I think you are totally right there. Of course values from the parent trait are not respected within the case class's copy. 
Let me know whether you find a fix, otherwise I can have a look once I am back (next Wednesday).

PR #2754, opened 2023-05-22
Slicing Module

DavidBakerEffendi:
* Moved CPG slicing logic into its own module
* Modified data-flow slicing to output JSON and be more friendly to deserialization into other languages and libraries
* If CPG detected does not have base/data-flow overlays, then these are applied

CC: @prabhu


fabsx00:
The  directory structure looks a bit off: if `slicing` is it's down sub project, then it should have a top-level directory like `ossdataflowengine`. The only exception to this right now are the frontends, which are in `joern-cli/frontends`, something that I believe we should change at some point.

DavidBakerEffendi:
@fabsx00 The long term for this is to be able to call the slicer in a pass, e.g. under `dave/joernti` I have `JoernTI` under `x2cpg` as well as a `SliceBasedTypeInferencePass` which is optional and also under `x2cpg`. Though I can imagine slicing to naturally fit under `dataflowengineoss` and it not be a module.

PR #2757, opened 2023-05-23
trim querydb dependencies: fix testDistro.sh

mpollmeier:
Querydb shipped a number of frontends in it's regular classpath, which
is causing problems with the usual suspects log4j/slf4j (see error below).
The main culprit is probably ghidra which ships an older version of log4j
in a fat jar.

```
Loading base CPG from:
/home/mp/Projects/shiftleft/joern/workspace/c/cpg.bin.tmp
java.lang.NoSuchMethodError: 'java.lang.ClassLoader[]
org.apache.logging.log4j.util.LoaderUtil.getClassLoaders()'
        at
org.apache.logging.log4j.core.impl.ThreadContextDataInjector.getServiceProviders(ThreadContextDataInjector.java:83)
        at
org.apache.logging.log4j.core.impl.ThreadContextDataInjector.initServiceProviders(ThreadContextDataInjector.java:73)
        at
org.apache.logging.log4j.core.impl.ThreadContextDataInjector.getProviders(ThreadContextDataInjector.java:285)
```

### DIFF
--- a/querydb/build.sbt
+++ b/querydb/build.sbt
@@ -5,9 +5,6 @@ enablePlugins(JavaAppPackaging)
 dependsOn(
   Projects.console,
   Projects.macros,
-  Projects.ghidra2cpg,
-  Projects.javasrc2cpg,
-  Projects.kotlin2cpg,
   Projects.ghidra2cpg  % "test->test",
   Projects.javasrc2cpg % "test->test",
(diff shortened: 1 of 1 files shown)


ursachec (approved):

ursachec:
would like to be more helpful, but the only answer i have is "not sure" @mpollmeier

PR #2595, opened 2023-04-27
WIP: Scala 3

mpollmeier:
The long awaited upgrade to Scala 3. 
Ready for review now, beta testing (both in house and in the community) is ongoing. 
Scala `3.3.0-RC4` will "most likely" be re-released as `3.3.0` in a few weeks. 

Depends on https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1702

PR #2758, opened 2023-05-24
querydb: cleanup/simplify dependency tree, move classes into querydb namespace

mpollmeier:
* console shouldn't depend on macros
* move querydb classes into querydb namespace: q, Query, QueryDatabase


mpollmeier:
only bring in after we throw out scala2: https://github.com/joernio/joern/pull/2595

PR #2761, opened 2023-05-24
RubyLexer: basic support for regex literals

xavierpinho:


joern-cli/frontends/rubysrc2cpg/src/main/java/io/joern/rubysrc2cpg/parser/RubyLexerBase.java
```diff
@@ -0,0 +1,98 @@
...
+        Token token = super.nextToken();
+        if (token.getChannel() == Token.DEFAULT_CHANNEL){
+            previousToken = token;
+        }
+        return token;
+    }
+
+    /** To be invoked when encountering `/`, deciding if it should emit a
+     * `REGULAR_EXPRESSION_START` token. */
+    protected boolean couldBeRegexStart(){
```
maltek, line 34: just a nit, but since this method is used to determine whether we are at a regex start or not I would expect the method to be named something like `isStartOfRegex`.

maltek (approved):

PR #2770, opened 2023-05-25
XTypeRecovery: Improved Import/Existing JS Type Recovery

DavidBakerEffendi:
* Improved patterns that match a variety of `export` patterns
* Improved handling of type prepopulation when type names are found in `typeFullName`
* Fixed bug where some nodes would slip through setType* patterns and not have their types set


fabsx00 (approved):
I am overall a bit worried about the language-specific aspects of the type propagation growing in size, but let's get this in to resolve immediate problems and then look into any refactorings we can do to further reduce code size.

DavidBakerEffendi:
@fabsx00 It very often comes down to how imports/exports are resolved and handled. Perhaps a next big step would be to extend the `Import` node to actually link to the method/type/field it is importing, and then the type recovery pass could accept the info from `Import` nodes as-is.

This perhaps speaks to the scope of `XTypeRecovery` being an import resolver and type propagation algorithm getting a bit messy.

PR #2771, opened 2023-05-25
github actions: master build and use -latest runners

mpollmeier:
Now that we 'only' release once a day we should run tests on each
merge to master, just to be safe. There's a few paths how things get
merged to master that are red, e.g. two PRs that are merged at the
same time that bring in incompatible changes. Or simply someone using
admin rights to merge.

Also use the -latest runners to avoid being in the 'no runner
available because they've been sunsetted' situation again.


mpollmeier:
grafik
no idea where this comes from, those are the ones that are being dropped in this very PR...

PR #2777, opened 2023-05-30
jssrc2cpg: Default Parameters Show Default in .code

DavidBakerEffendi:
Right now, while default parameter semantics are modelled in the CPG, there is no easy query for "Does this parameter have a default value?".

Now ideally, there would be some `MethodParameterIn(defaultValue: Option[String])` or `MethodParameterIn -DEFAULT-> Literal` construct in the schema, but this PR is a quick-and-dirty way in the meantime to easily show a parameter is assigned a default.


fabsx00:
Which field are we changing here?

DavidBakerEffendi:
@fabsx00 only the `.code` property. Currently, for parameters with defaults, `name` == `code`.

fabsx00:
Alright.

Issue #2732, opened 2023-05-17
[Bug] php2cpg: identifiers missing in AST for static method calls and foreach


maltek:
```
joern> workspace.reset ; importCode.php.fromString("""<?
       \Doctrine\ORM\Tools\Console\ConsoleRunner::run($helperSet);
       """) 
creating workspace directory: /home/mkraus/src/joern/workspace
Creating project `console3776893619653733582` for code at `/tmp/console3776893619653733582`
=======================================================================================================
Invoking CPG generator in a separate process. Note that the new process will consume additional memory.
If you are importing a large codebase (and/or running into memory issues), please try the following:
1) exit joern
2) invoke the frontend: /home/mkraus/src/joern/joern-cli/target/universal/stage/php2cpg -J-Xmx7828m /tmp/console3776893619653733582 -o /home/mkraus/src/joern/workspace/console3776893619653733582/cpg.bin.zip
3) start joern, import the cpg: `importCpg("path/to/cpg")`
=======================================================================================================

moving cpg.bin.zip to cpg.bin because it is already a database file
Creating working copy of CPG to be safe
Loading base CPG from: /home/mkraus/src/joern/workspace/console3776893619653733582/cpg.bin.tmp
Code successfully imported. You can now query it using `cpg`.
For an overview of all imported code, type `workspace`.
Adding default overlays to base CPG
The graph has been modified. You may want to use the `save` command to persist changes to disk.  All changes will also be saved collectively on exit
The graph has been modified. You may want to use the `save` command to persist changes to disk.  All changes will also be saved collectively on exit
res5_1: Cpg = Cpg (Graph [25 nodes])

joern> cpg.identifier.filter{ i => scala.util.Try { i.astParent }.isFailure }.l 
res6: List[Identifier] = List(
  Identifier(
    id = 5L,
    argumentIndex = -1,
    argumentName = None,
    code = "Doctrine\\ORM\\Tools\\Console\\ConsoleRunner",
    columnNumber = None,
    dynamicTypeHintFullName = ArraySeq(),
    lineNumber = Some(value = 2),
    name = "Doctrine\\ORM\\Tools\\Console\\ConsoleRunner",
    order = -1,
    typeFullName = "ANY"
  )
)
```

```
joern> workspace.reset ; importCode.php.fromString("""<?
           foreach ($GLOBALS as $foo) {
           }
       """) 
creating workspace directory: /home/mkraus/src/joern/workspace
[...]

PR #2782, opened 2023-05-30
Add php2cpg local creation and ClosureRefEdge passes

johannescoetzee:
Since PHP doesn't require variable declarations, php2cpg would create a Local node on the first use of a new identifier in scope and add a ref edge from the identifier to the local. This approach works well if the identifier in question is added to the CPG, but this isn't always the case. This led to a situation where identifier nodes are added to the CPG (due to the ref edge being added), but not to the AST, leaving them without AST parents. 

This PR removes most Local creation from the AstCreator and creates them in a follow-up pass instead. This also means that locals captured by closures must be referenced separately, since the locals being referenced may not exist yet, so the ClosureRefPass adds these.

Resolves https://github.com/joernio/joern/issues/2732

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/ClosureRefPass.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/ClosureRefPass.scala
@@ -36,12 +36,8 @@ class ClosureRefPass(cpg: Cpg) extends ConcurrentWriterCpgPass[ClosureBinding](c
     }
   }
 
-  private def hasMethod(in: Traversal[AstNode]): Traversal[AstNode] = {
-    in.where(_.collectAll[Method])
-  }
-
   private def getMethod(methodRef: MethodRef): Option[Method] = {
-    methodRef.repeat(_.astParent)(_.until(hasMethod).emit(hasMethod)).collectAll[Method].headOption
+    methodRef.repeat(_.astParent)(_.until(_.isMethod).emit(_.isMethod)).isMethod.headOption
   }
 
   private def addRefToCapturedNode(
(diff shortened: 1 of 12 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/ClosureRefPass.scala
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+
+  private def hasMethod(in: Traversal[AstNode]): Traversal[AstNode] = {
+    in.where(_.collectAll[Method])
+  }
+
+  private def getMethod(methodRef: MethodRef): Option[Method] = {
+    methodRef.repeat(_.astParent)(_.until(hasMethod).emit(hasMethod)).collectAll[Method].headOption
+  }
```
maltek, line 45: I believe this can be simplified to

```suggestion
  private def getMethod(methodRef: MethodRef): Option[Method] = {
    methodRef.repeat(_.astParent)(_.until(_.isMethod).emit(_.isMethod)).isMethod.headOption
  }
```

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/ClosureRefPass.scala
```diff
@@ -0,0 +1,69 @@
...
+    method match {
+      case None =>
+        logger.warn(s"No parent method for methodRef for ${closureBinding.closureBindingId}. REF edge will be missing")
+
+      case Some(method) =>
+        closureBinding.closureOriginalName.foreach { name =>
+          val maybeCaptured =
+            method.parameter
+              .find(_.name == name)
+              .orElse(method.ast.collectAll[Local].find(_.name == name))
```
maltek, line 61: `.ast` also recurses down into nested methods inside the current one. I think this may need a manual `.repeat` step that stops at those.

johannescoetzee: Good point! I'll update that

PR #2785, opened 2023-05-31
Reintroduce Python Compatibility Changes

DavidBakerEffendi:
* Reverted [commit 4691b4a](https://github.com/joernio/joern/commit/4691b4a54b23d23d4e6aa3193bb703cfb7398b57)
* Added tests that were resulting in some failure
* Added support for string constant type hints (idk how but this is valid Python)
* Using `ForkJoinParallelCpgPass` for `DynamicTypeHintFullNamePass`
* Making sure that `DynamicTypeHintFullNamePass` considers types within the current scope


ml86:
Is this the minimally adjusted version of commit 4691b4a to make the type propagation happy? Do we need the switch to `ForkJoinParallelCpgPass` in this PR? I am asking because i would like to diff this PR and my initial version of it.

DavidBakerEffendi:
@ml86 Wouldn't say "minimally" but it does fix some assumptions on the type propagation to make it "happier". The `ForkJoinParallelCpgPass` is not necessary, but the other changes on `DynamicTypeHintFullNamePass` are necessary.

ml86 (approved):
Look good. Tests pass.

PR #2795, opened 2023-06-02
Correctly Account for Method Names in TS Interfaces

DavidBakerEffendi:
For example
```typescript
interface Symbol {
    /** Returns a string representation of an object. */
    toString(): string;

    /** Returns the primitive value of the specified object. */
    valueOf(): symbol;
}
```
The names of members for `Symbol` would be the full code. With this change, it would (more) correctly use the identifier name for `Member.name`.


max-leuthaeuser (approved):

max-leuthaeuser:
Thanks! Could you add an unit test for that?

PR #2804, opened 2023-06-02
Johannes/excludes fix

johannescoetzee:
This PR re-introduces the changes in https://github.com/joernio/joern/pull/2741 and fixes the bug that resulted in the revert in https://github.com/joernio/joern/pull/2751.

The issue in the original PR was a result of the frontend-specific case classes extending the common config with its own fields. The `copy` method didn't copy these as well, so the fix here is to just manually update the parent fields by calling `copy(...).withInheritedFields(this)`, instead of just `copy`. This pattern presents some opportunity for errors to slip in when adding new options, but does work as is with some simple unit tests added to try to catch these.

This also removes the `--ignore-path` option from `kotlin2cpg` since that's replaced by the `--exclude` option, but if customers are relying on this, I'll not remove it instead.


ursachec:
:+1: from me on changing `--ignore-path` to `--exclude` for _kotlin2cpg_

ursachec (approved):

max-leuthaeuser:
There is a solution to the copy problem w.r.t. superclass fields but I am not sure if we want to go that route: https://github.com/milessabin/shapeless/blob/main/examples/src/main/scala/shapeless/examples/basecopy.scala (open families).

max-leuthaeuser (approved):

PR #2811, opened 2023-06-05
rubysrc2cpg: refactor to use x2cpg's DynamicCallUnknownName

xavierpinho:
In a follow-up, `MethodFullNames.OperatorPrefix` shall also be removed.


DavidBakerEffendi (approved):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1199,7 +1198,7 @@ class AstCreator(filename: String, global: Global)
       .name(name)
       .methodFullName(name)
       .signature(localIdentifier.getText())
-      .typeFullName(MethodFullNames.UnknownFullName)
+      .typeFullName(DynamicCallUnknownFullName)
```
ursachec, line 1201: in order to be consistent with a frontend like _pythonsrc_, I'd change the TYPE_FULL_NAME for call like these to _"ANY"_

PR #2812, opened 2023-06-05
rubysrc2Cpg: Call node code field set

rahul-privado:
### Description

This code change sets  code field for call nodes.

### Observations/Testing done
The code being set is the most code that is returned by `ctx.getText` for the closest `ctx`. Please comment on whether this is fine and what the expectations from code are.

```
class Person
  attr_accessor :name, :age

  def initialize(name, age)
    @name = name
    @age = age
  end

  def greet
    puts "Hello, my name is #{@name} and I am #{@age} years old."
  end

  def have_birthday
    @age += 1
    puts "Happy birthday! You are now #{@age} years old."
  end
end

p = Person. new
p.greet
```

Code field for calls
```
joern> cpg.call.map( x => (x.name, x.code, x.lineNumber)).l 
res3: List[(String, String, Option[Integer])] = List(
  ("attr_accessor", "attr_accessor", Some(value = 2)),
  ("=", "=", Some(value = 5)),
  ("=", "=", Some(value = 6)),
  ("puts", "puts", Some(value = 10)),
  ("+=", "+=", Some(value = 14)),
  ("puts", "puts", Some(value = 15)),
  ("=", "=", Some(value = 19)),
  ("new", "new", Some(value = 19)),
  ("greet", "greet", Some(value = 20))
)
```


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/passes/ast/SimpleAstCreationPassTest.scala
```diff
@@ -112,4 +112,103 @@ class SimpleAstCreationPassTest extends RubyCode2CpgFixture {
       encoding.columnNumber shouldBe Some(5)
     }
   }
+
+  "Code field for simple fragments" should {
```
xavierpinho, line 116: As these tests are pretty much the same (in terms of assertions) as the ones in the previous group (line 9), I would suggest to add them there, instead of creating a new group ("Code field for simple fragments"). In other words, it's a needless distinction, right?

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/passes/ast/SimpleAstCreationPassTest.scala
```diff
@@ -112,4 +112,103 @@ class SimpleAstCreationPassTest extends RubyCode2CpgFixture {
...
+
+    "have correct code for a single left had side call" in {
```
xavierpinho, line 118: Moreover, once moved to the previous group, tests should ideally be renamed as well (as we're not testing .code exclusively), to keep things aligned.

xavierpinho (approved):
Looks good. But please consider moving tests around and maybe renaming them to keep similar stuff looking similar.

PR #2816, opened 2023-06-06
Added methodFullName support for external dependency

ankit-privado:
Added Support to resolve the `methodFullName` when external dependency present in ruby file. 

Created one passes:
1. `AstPackagePass`: This pass collect all external dependency files ending with `.rb` extension. Parsing and storing its method info into `packageTable`. 

Added `enableDependencyDownload` flag, If provided, then only gem dependencies will be downloaded. Note, this is troublesome to get working on Windows due to some permissions issue so, for now, this is only available on MacOS and Linux.

Here `packageTable` is like cache for storing the method and require info. Furthermore, we only search for the method name in the module mentioned in the file.


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/RubyMethodFullNameTests.scala
```diff
@@ -0,0 +1,51 @@
...
+        |
+        |val v = Help::InnerHelp.new
+        |puts v.printValue("any")
+        |""".stripMargin)
+
+    "recognise all call node" in {
+      cpg.call.name("printValue").l.size shouldBe 1
+    }
+
+    "recognise methodFullName for call node" in {
```
khemrajrathore, line 47: test case missing

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/packageTable.scala
```diff
@@ -0,0 +1,42 @@
...
+  override def hashCode: Int = {
+    methodName.hashCode
+  }
+}
+
+case class PackageContext(moduleName: String, packageTable: PackageTable)
+
+class PackageTable() {
+
+  private val methodTableSet = mutable.HashSet[MethodTableModel]()
```
khemrajrathore, line 22: This DS should be a Map as we are doing `get`, `contains` on `methodName`. So the key to the map will be `methodName`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/packageTable.scala
```diff
@@ -0,0 +1,42 @@
+package io.joern.rubysrc2cpg.utils
+
+import scala.collection.mutable
+
+case class MethodTableModel(moduleName: String, methodName: String, parentClassPath: String) {
```
khemrajrathore, line 5: The Data structures defined in this class `packageTable.scala` have redundant properties, can we optimize the schema?

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1903,16 +1906,13 @@ class AstCreator(filename: String, global: Global)
...
-              .filter(node => node.isInstanceOf[NewLiteral])
-              .head
-              .asInstanceOf[NewLiteral]
-              .code
-          println(s"AST to be created for imported file ${importedFile}")
-
+          val importedNode =
+            argsAsts.head.nodes.filter(node => node.isInstanceOf[NewLiteral]).head.asInstanceOf[NewLiteral]
+          println(s"AST to be created for imported file ${importedNode.code}")
+          Seq(callAst(callNode.code(importedNode.code), Seq(Ast(importedNode))))
```
khemrajrathore, line 1912: Why do we create CallAst for importedNodes? we should create ImportNodes right?

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/packageTable.scala
```diff
@@ -0,0 +1,42 @@
...
+      methodTableSet.add(packageMethod)
+    }
+  }
+
+  def get(methodName: String): Set[MethodTableModel] = {
+    methodTableSet.filter(method => method.methodName == methodName).toSet
+  }
+
+  def contains(methodName: String): Boolean = {
+    methodTableSet.contains(MethodTableModel("", methodName, ""))
```
khemrajrathore, line 36: the moduleName and parentClassPath won't be empty always right?

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/RubyMethodFullNameTests.scala
```diff
@@ -0,0 +1,51 @@
+package io.joern.rubysrc2cpg.querying
+
+import io.joern.rubysrc2cpg.testfixtures.RubyCode2CpgFixture
+import io.shiftleft.semanticcpg.language._
+
+class RubyMethodFullNameTests extends RubyCode2CpgFixture {
```
khemrajrathore, line 6: Can we add more test cases? Ex - 
- class defined in a package, 
- For the call Node is there a method with the same name and methodFullName

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/RubyMethodFullNameTests.scala
ankit-privado: Yes sure, will add.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/packageTable.scala
ankit-privado: Yes changed to HashMap data structure to store method info.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/packageTable.scala
ankit-privado: yes resolved after changing to hashMap.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/packageTable.scala
ankit-privado: yes removed the code.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
ankit-privado: Not sure about this but i think to parse the complicated require statement like: `require File.join(__dir__, 'file.rb')`

@rahul-privado can you help?

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/RubyMethodFullNameTests.scala
ankit-privado: Yes sure will add.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
rahul-privado: Makes sense. Did not know about them.

@ankit-privado Can you change this and use `newImportNode()`

ankit-privado: yes sure @rahul-privado

ankit-privado: done.

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/RubyMethodFullNameTests.scala
ankit-privado: done

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
```diff
@@ -1,26 +1,58 @@
...
+    Try("gem env gempath".!!) match {
+      case Success(gemPath) =>
+        gemPath.split(":").toList
+      case Failure(exception) =>
+        logger.error(s"Error While fetching gem path: ${exception.getMessage}")
+        List.empty
+    }
+  }
+
+  private def downloadDependency(inputPath: String): Unit = {
```
xavierpinho, line 48: This worries me: I don't think we should be messing with the user's environment in any capacity when scanning. At best, I'd put this behind a feature flag. Ideally, we would live without it.

xavierpinho: If I understand things through, we're downloading and scanning the dependencies together with the project in order to resolve full-names, right? We have a similar challenge in JS with imports. In Ruby it's arguably worse since `require`/`include` is but a "statement" (in JS we'd have some variable assigned to its result or some aliasing).
cc: @DavidBakerEffendi, @max-leuthaeuser  

In this approach, couldn't it happen we also need to download and scan the original dependencies' dependencies and so on until closure?

I'd frame this approach as an incremental improvement available through some feature flag, and assume we haven't scanned the dependencies by default.

DavidBakerEffendi: +1 to making it disabled by default, we might introduce a CWE through Joern otherwise. And yes, there would also result in some kind of recursive download situation. Ideally, there is an API we can use to gather all the dependencies and their transitive dependencies to do 1 download and force Joern into a whole-program analysis.

But beware, whole-program analysis is slow and resource intensive.

prabhu: @ankit-privado As a tangent, I'm working on an API server to fetch these infos by wrapping the joern frontends. There might some opportunities to collaborate since your optional pass could use this service that runs locally or remotely or use a cli.

ankit-privado: Done @DavidBakerEffendi @xavierpinho, added flag to download the dependency, by default it will not download anything.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1945,4 +1936,37 @@ class AstCreator(filename: String, global: Global)
   def astForYieldWithOptionalArgumentPrimaryContext(ctx: YieldWithOptionalArgumentPrimaryContext): Seq[Ast] = {
     astForYieldWithOptionalArgumentContext(ctx.yieldWithOptionalArgument())
   }
+
+  def resolveRequireOrLoadPath(argsAst: Seq[Ast]): Seq[Ast] = {
+    val importedNode =
+      argsAst.head.nodes.filter(node => node.isInstanceOf[NewLiteral]).head.asInstanceOf[NewLiteral]
```
DavidBakerEffendi, line 1942: I would suggest being a lot more defensive here, e.g. 
```
argsAst.head.nodes.collect[Literal].headOption
```

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1945,4 +1936,37 @@ class AstCreator(filename: String, global: Global)
...
+    val pathValue = importedNode.code.replaceAll("'", "").replaceAll("\"", "")
+    val result = pathValue match {
+      case path if Files.isRegularFile(Paths.get(path)) =>
+        path
+      case path if Files.isRegularFile(Paths.get(path + ".rb")) =>
```
DavidBakerEffendi, line 1947:
```suggestion
      case path if Files.isRegularFile(Paths.get(s"${path}.rb")) =>
```

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1945,4 +1936,37 @@ class AstCreator(filename: String, global: Global)
...
+      case _ =>
+        pathValue
+    }
+    packageStack.append(result)
+    astForImportNode(importedNode.code)
+  }
+
+  def resolveRelativePath(currentFile: String, argsAst: Seq[Ast]): Seq[Ast] = {
+    val importedNode =
+      argsAst.head.nodes.filter(node => node.isInstanceOf[NewLiteral]).head.asInstanceOf[NewLiteral]
```
DavidBakerEffendi, line 1958: Similarly, I would recommend being more defensive here:
```
argsAst.head.nodes.collect[Literal].headOption
```

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1945,4 +1936,37 @@ class AstCreator(filename: String, global: Global)
...
+    val pathValue        = importedNode.code.replaceAll("'", "").replaceAll("\"", "")
+    val updatedPath      = if (pathValue.endsWith(".rb")) pathValue else s"$pathValue.rb"
+    val currentDirectory = new java.io.File(currentFile).getParentFile
```
DavidBakerEffendi, line 1961: Could you preferably use the `better.files.File` API? It is more idiomatic Scala style

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1945,4 +1936,37 @@ class AstCreator(filename: String, global: Global)
...
+    val file             = new java.io.File(currentDirectory, updatedPath)
+    packageStack.append(file.getAbsolutePath)
+    astForImportNode(importedNode.code)
+  }
+
+  def astForImportNode(code: String): Seq[Ast] = {
```
DavidBakerEffendi, line 1967: Will this be fully implemented later? The most important properties of this node are `importedAs` and `importedEntity`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/AstPackagePass.scala
```diff
@@ -0,0 +1,40 @@
...
+    extends ConcurrentWriterCpgPass[String](cpg) {
+  override def generateParts(): Array[String] = getRubyDependenciesFile(inputPath)
+
+  override def runOnPart(diffGraph: DiffGraphBuilder, filePath: String): Unit = {
+    Try(new AstCreator(filePath, global, PackageContext(resolveModuleNameFromPath(filePath), packageTable)).createAst())
+  }
+
+  private def getRubyDependenciesFile(inputPath: String): Array[String] = {
+    Files
+      .walk(Paths.get(inputPath))
```
DavidBakerEffendi, line 25: If you use `better.files.File` there is an `extension` property which may be safer to compare with instead of `toString.endsWith`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/AstPackagePass.scala
```diff
@@ -0,0 +1,40 @@
...
+      .filter(_.toString.endsWith(".rb"))
+      .map(_.toString)
+      .toArray
+      .map(_.asInstanceOf[String])
```
DavidBakerEffendi, line 29: Seems like an unsafe case, do we need it? If so, then preferrably use `collect`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/AstPackagePass.scala
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  private def resolveModuleNameFromPath(path: String): String = {
+    if (path.contains(tempExtDir)) {
+      val moduleNameRegex = """\/gems\/([^\/]+)\/lib\/.*""".r
+      moduleNameRegex.findFirstMatchIn(path).map(_.group(1)).getOrElse("").split("/").last.split("-").head
```
DavidBakerEffendi, line 35: Would the `/` work on Windows? If this is related to file paths, then maybe use `java.io.File.separator` instead.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/PackageTable.scala
```diff
@@ -0,0 +1,36 @@
...
+      if (methodTableMap.contains(module)) {
+        methodTableMap(module)
+          .filter(_.methodName == methodName)
+          .foreach(method => {
+            finalMethodName.addOne(s"$module.${method.parentClassPath}$methodName:<unresolvedSignature>")
+          })
+      }
+    })
+    if (finalMethodName.isEmpty) {
+      List("<unknowfullname>")
```
DavidBakerEffendi, line 31: Typo here, rather use `Defines.DynamicCallUnknownFullName`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/PackageTable.scala
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  def getMethodFullNameUsingName(packageUsed: List[String], methodName: String): List[String] = {
+    val finalMethodName = ListBuffer[String]()
+    packageUsed.foreach(module => {
+      if (methodTableMap.contains(module)) {
+        methodTableMap(module)
+          .filter(_.methodName == methodName)
+          .foreach(method => {
+            finalMethodName.addOne(s"$module.${method.parentClassPath}$methodName:<unresolvedSignature>")
```
DavidBakerEffendi, line 26: Use `Defines.UnresolvedSignature`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/AstCreationPass.scala
```diff
@@ -1,19 +1,18 @@
...
-import io.joern.x2cpg.passes.frontend.TypeNodePass
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.passes.ConcurrentWriterCpgPass
 import org.slf4j.LoggerFactory
 
-import java.util.concurrent.ConcurrentHashMap
 import scala.jdk.CollectionConverters.EnumerationHasAsScala
 
-class AstCreationPass(inputPath: String, cpg: Cpg) extends ConcurrentWriterCpgPass[String](cpg) {
+class AstCreationPass(inputPath: String, cpg: Cpg, global: Global, packageTable: PackageTable, tempDir: String)
```
rahul-privado, line 12: Please pass a flat here instead of string that determines whether the AST should be applied to the diffgraph

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/testfixtures/RubyCode2CpgFixture.scala
```diff
@@ -12,7 +12,7 @@ trait RubyFrontend extends LanguageFrontend {
   override val fileSuffix: String = ".rb"
 
   override def execute(sourceCodeFile: File): Cpg = {
-    implicit val defaultConfig: Config = Config()
+    implicit val defaultConfig: Config = Config(true)
```
DavidBakerEffendi, line 15: Rather, make this configurable and not true by default for all test suites. May require adding a `def enableDependencyDownload` or changing `RubyFrontend` to an abstract class or something

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1407,6 +1410,9 @@ class AstCreator(filename: String, global: Global)
       .filename(filename)
     callNode.methodFullName(classPath + callNode.name)
 
+    val classType = if (classStack.isEmpty) "Standalone" else classStack.head
```
rahul-privado, line 1413: Use `classStack.top`. Just more intuituve

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1089,8 +1093,7 @@ class AstCreator(filename: String, global: Global)
     val column         = localIdentifier.getSymbol().getCharPositionInLine()
     val line           = localIdentifier.getSymbol().getLine()
     val name           = getActualMethodName(localIdentifier.getText)
-    val methodFullName = s"$filename:$name"
-
+    val methodFullName = packageContext.packageTable.getMethodFullNameUsingName(packageStack.toList, code).head
```
rahul-privado, line 1096: Use `name` and not `code` to lookup for the methodfullname
The first argument to this method should be the classname as we discussed and not the package name. It will be used when type recovery is done. For now, nothing will be passed.

rahul-privado: The class stack cannot be considered the package.I would suggest passing only the method name argument till we figure out what should be the exact content of the class

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
```diff
@@ -1,26 +1,61 @@
...
 
-import scala.util.Try
+import java.io.File
+import java.nio.file.{Files, Paths}
+import scala.sys.process._
+import scala.util.{Failure, Success, Try}
 
 class RubySrc2Cpg extends X2CpgFrontend[Config] {
 
+  private val tempDependencyDir = "tempExtDependency"
```
DavidBakerEffendi, line 20: Rather use `java.io.File.createTempFile` or `better.files.File.newTemporaryDirectory()`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
```diff
@@ -1,26 +1,61 @@
...
   override def createCpg(config: Config): Try[Cpg] = {
     withNewEmptyCpg(config.outputPath, config: Config) { (cpg, config) =>
+      val packageTableInfo = new PackageTable()
       new MetaDataPass(cpg, Languages.RUBYSRC, config.inputPath).createAndApply()
       new ConfigPass(cpg, config.inputPath).createAndApply()
-      val astCreationPass = new AstCreationPass(config.inputPath, cpg)
+      if (config.enableDependencyDownload) {
+        downloadDependency(config.inputPath)
+        new AstPackagePass(cpg, tempDependencyDir, global, packageTableInfo, config.inputPath).createAndApply()
+        deleteRecursively(new File(s"${config.inputPath}/$tempDependencyDir"))
```
DavidBakerEffendi, line 34: Do not implement your own one. `better.files.File` has a `delete` function that is recursive for directories.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
DavidBakerEffendi, line 1096: Are we always guaranteed that the list is non-zero? Else use `headOption`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
```diff
@@ -1,26 +1,61 @@
...
+      Try(s"bundle install --gemfile=$inputPath/Gemfile --path=$tempDependencyDir".!!) match {
+        case Success(bundleOutput) =>
+          logger.info(s"Dependency installed successfully: $bundleOutput")
+        case Failure(exception) =>
+          logger.error(s"Error while downloading dependency: ${exception.getMessage}")
+      }
+    }
+  }
+
+  private def deleteRecursively(file: File): Unit = {
```
DavidBakerEffendi, line 53: As alluded to above, this can be deleted as this exists under `better.files.File.delete`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1866,28 +1872,13 @@ class AstCreator(filename: String, global: Global)
       val callNodes = methodIdentifierAsts.head.nodes.filter(node => node.isInstanceOf[NewCall])
```
DavidBakerEffendi, line 1872: Rather use `collect`

DavidBakerEffendi: If you use `collect` then `callNodes` will already be type cast to a collection of `Call` and you don't have to do the type casting on line 1874

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/RubyMethodFullNameTests.scala
```diff
@@ -0,0 +1,91 @@
...
+    )
+      .moreCode(
+        """
+          |class Outer
+          | def printValue()
+          |   puts "print"
+          | end
+          |end
+          |""".stripMargin,
+        "util/help.rb"
```
DavidBakerEffendi, line 76: This will break for windows. Use `Seq("util", "help").mkString(java.io.File.separator)`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
ankit-privado: Yes using the `collectFirst` instead.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
ankit-privado: yes this will be fully implemented later.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
ankit-privado: @rahul-privado Yes yes, this `packageStack` is only for optimisation (use in narrowing down the searching methodName in imports defined in the file). we still matching the method by its name.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
```diff
@@ -1,26 +1,51 @@
...
       astCreationPass.createAndApply()
       new TypeNodePass(astCreationPass.allUsedTypes(), cpg).createAndApply()
     }
   }
+
+  private def downloadDependency(inputPath: String, tempDir: String): Unit = {
+    if (File(s"$inputPath/Gemfile").exists) {
+      Try(s"bundle install --gemfile=$inputPath/Gemfile --path=$tempDir".!!) match {
+        case Success(bundleOutput) =>
+          println("lolo " + bundleOutput)
```
DavidBakerEffendi, line 44: Remove print

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
```diff
@@ -1,26 +1,51 @@
...
     withNewEmptyCpg(config.outputPath, config: Config) { (cpg, config) =>
+      val packageTableInfo = new PackageTable()
       new MetaDataPass(cpg, Languages.RUBYSRC, config.inputPath).createAndApply()
       new ConfigPass(cpg, config.inputPath).createAndApply()
-      val astCreationPass = new AstCreationPass(config.inputPath, cpg)
+      if (config.enableDependencyDownload) {
+        val tempDir = File.newTemporaryDirectory()
+        downloadDependency(config.inputPath, tempDir.toString())
+        new AstPackagePass(cpg, tempDir.toString(), global, packageTableInfo, config.inputPath).createAndApply()
+        tempDir.delete()
```
DavidBakerEffendi, line 32: Perhaps best to have this under a `try-finally` to ensure the file is deleted

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
```diff
@@ -1,26 +1,51 @@
...
+      }
+      val astCreationPass = new AstCreationPass(config.inputPath, cpg, global, packageTableInfo)
       astCreationPass.createAndApply()
       new TypeNodePass(astCreationPass.allUsedTypes(), cpg).createAndApply()
     }
   }
+
+  private def downloadDependency(inputPath: String, tempDir: String): Unit = {
+    if (File(s"$inputPath/Gemfile").exists) {
```
DavidBakerEffendi, line 41: Use OS-agnostic notation, you can do this as follows:
```suggestion
    if (File(inputPath) / "Gemfile").exists) {
```

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1863,31 +1869,16 @@ class AstCreator(filename: String, global: Global)
       methodNameAsIdentifierStack.push(methodIdentifierAsts.head)
       val argsAsts = astForArgumentsWithoutParenthesesContext(ctx.argumentsWithoutParentheses())
 
-      val callNodes = methodIdentifierAsts.head.nodes.filter(node => node.isInstanceOf[NewCall])
+      val callNodes = methodIdentifierAsts.head.nodes.collect(_.asInstanceOf[NewCall])
```
DavidBakerEffendi, line 1872:
```suggestion
      val callNodes = methodIdentifierAsts.head.nodes.collectAll[NewCall]
```

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1945,4 +1936,38 @@ class AstCreator(filename: String, global: Global)
   def astForYieldWithOptionalArgumentPrimaryContext(ctx: YieldWithOptionalArgumentPrimaryContext): Seq[Ast] = {
     astForYieldWithOptionalArgumentContext(ctx.yieldWithOptionalArgument())
   }
+
+  def resolveRequireOrLoadPath(argsAst: Seq[Ast]): Seq[Ast] = {
+    val importedNode = argsAst.head.nodes.collectFirst(_.asInstanceOf[NewLiteral]).get
```
DavidBakerEffendi, line 1941:
```suggestion
    argsAst.head.nodes.collectFirst { case x: NewLiteral => x }.foreach { importedNode =>
```

DavidBakerEffendi: Wrap this code a `foreach` so that you don't have to call an unsafe `.get` (which may return an error if there is no `NewLiteral`.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/AstPackagePass.scala
```diff
@@ -0,0 +1,37 @@
...
+    if (currentDir.exists) {
+      currentDir.listRecursively.filter(_.extension.exists(_ == ".rb")).map(_.path.toString).toArray
+    } else {
+      Array.empty
+    }
+  }
+
+  private def resolveModuleNameFromPath(path: String): String = {
+    if (path.contains(tempExtDir)) {
+      val moduleNameRegex = s"${java.io.File.separator}gems${java.io.File.separator}([^${java.io.File.separator}]+)${java.io.File.separator}lib${java.io.File.separator}.*".r
```
DavidBakerEffendi, line 31: Does this work? This should de-duplicate the number of times `java.io.File.separator` is written
```suggestion
      val moduleNameRegex = Seq("gems", "([^", "]+)", "lib", ".*").mkString(java.io.File.separator).r
```

DavidBakerEffendi (changes requested):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
ankit-privado: yes done.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
ankit-privado: there is no option as `collectAll`, only `collect` and `collectFirst` are there.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
ankit-privado: yes not using `.get` changed the logic.

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/testfixtures/RubyCode2CpgFixture.scala
ankit-privado: done.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/AstPackagePass.scala
DavidBakerEffendi: Is this perhaps why it's not working for Windows?

ankit-privado: yes i changed to `java.io.file.sperateor` but its still failing due to some bundle install issue in window https://github.com/ankit-privado/joern/actions/runs/5339001828/jobs/9677138386#step:7:2971

DavidBakerEffendi:
If windows fails again, I can perhaps take a look on my Windows machine

DavidBakerEffendi (approved):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
```diff
@@ -25,7 +27,7 @@ class RubySrc2Cpg extends X2CpgFrontend[Config] {
       val packageTableInfo = new PackageTable()
       new MetaDataPass(cpg, Languages.RUBYSRC, config.inputPath).createAndApply()
       new ConfigPass(cpg, config.inputPath).createAndApply()
-      if (config.enableDependencyDownload) {
+      if (config.enableDependencyDownload && !sys.props.getOrElse("os.name", "").toLowerCase.contains("win")) {
```
max-leuthaeuser, line 30: `scala.util.Properties.isWin`

DavidBakerEffendi: @ankit-privado One last change 😉

DavidBakerEffendi:
@ankit-privado I just resolved a merge conflict, non-zero chance that the formatting won't pass

ankit-privado:
@DavidBakerEffendi fixed the issue, we can merge this if looking good to everyone.

PR #2821, opened 2023-06-07
jssrc2cpg: added support for TypeCastExpression

max-leuthaeuser:


johannescoetzee (approved):
A unit test is failing with  
```
[info] - have correct types for empty method with rest parameter *** FAILED ***
[info]   "ANY[]" was not equal to "ANY[[]]" (TSTypesTest.scala:50)
```

but that looks like a minor issue. Approved pending an update of the test or brackets in the name

PR #2831, opened 2023-06-07
Fix - Unresolved methodFullName when base class function returns external object

khemrajrathore:
When a super class calls a method that is only implemented in a base class, and this call returns an externally imported call's return value, will attempt to resolve it on the spot.


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -917,9 +917,17 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
   }
 
   private def expressionReturnTypeFullName(expr: Expression): Option[String] = {
-    tryWithSafeStackOverflow(expr.calculateResolvedType()).toOption
-      .flatMap(typeInfoCalc.fullName)
-      .orElse(exprNameFromStack(expr))
+    val resolvedTypeOption = tryWithSafeStackOverflow(expr.calculateResolvedType()) match {
```
DavidBakerEffendi, line 920: Could this not be part of the `flatMap` that already existed?
e.g. 
```scala
tryWithSafeStackOverflow(expr.calculateResolvedType()).flatMap {
     case Failure(ex) =>
         ex match {
           // If ast parser fails to resolve type, try resolving locally by using name
           case symbolException: UnsolvedSymbolException =>
             scopeStack.lookupVariableType(symbolException.getName)
           case _ => None
         }
       case Success(resolvedType) => typeInfoCalc.fullName(resolvedType)
}.orElse(exprNameFromStack(expr))

khemrajrathore: FlatMap when applied on a Try, the statement inside of it only gets executed for Success.
In this case, inside flatMap, I directly get `resolvedType`

DavidBakerEffendi: Ok make sense

DavidBakerEffendi:
@fabsx00 I'm happy with this, perhaps you can give it a set of eyes, then we can go ahead and merge

DavidBakerEffendi (approved):

khemrajrathore:
@fabsx00 can we get this merged?

PR #2833, opened 2023-06-08
dotgenerator: display Identifier in ControlStructure

max-leuthaeuser:
Structures like `int a; if(a) { foo(); }` are actually something like `int a; if(a!=null) { foo(); }`. Hence `a` is the condition in the CFG and should be displayed even if it's just an Identifier. This PR fixes that.

See: https://discord.com/channels/832209896089976854/832214243230744626/1115821653956513884

Also: some minor clean-ups.


mahaloz:
Ty again!

fabsx00 (approved):

PR #2835, opened 2023-06-08
jssrc2cpg: added method signatures

max-leuthaeuser:
tsc is quite strong when it comes to inferring function and parameter types. Thus, we can set the signatures now.

See: https://discord.com/channels/832209896089976854/832214243230744626/1116319958272593950


max-leuthaeuser:
Please do not merge yet. Needs testing with the closed source DF tracker.

max-leuthaeuser:
There are a couple of method signature checks, e.g., in:
- dataflowengine\engine\CallResolver.scala
- dataflowengine\engine\Misc.scala
- dataflowengine\passes\languagespecific\javascript\RestoreTranspiledClasses.scala
So we can't do this here as easily as I thought.

PR #2838, opened 2023-06-08
rubysrc2cpg: Scope resolution operator and more tests

rahul-privado:
1. Scope resolution operator `::` further implementation and corrections
2. Rescue and associated implementation
3. Associated unit tests
4. More data flow tests. One is not working and needs to be fixed and hence has been marked as ignore


xavierpinho (approved):
In the forthcoming PRs, I'd suggest we stop using `dotAst` as a test and replace it with explicit assertions. But let's keep the momentum :)

PR #2846, opened 2023-06-09
rubysrc2cpg: Better method/variable identification and ChainedInvocationPrimaryContext fixes

rahul-privado:
1. Fixes and better implementation for method/variable to be correctly identified
2. Corrected implementation for `astForChainedInvocationPrimaryContext()`
3. Associated unit tests


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/passes/ast/SimpleAstCreationPassTest.scala
```diff
@@ -505,5 +505,47 @@ class SimpleAstCreationPassTest extends RubyCode2CpgFixture {
       identifierNode.lineNumber shouldBe Some(3)
       identifierNode.columnNumber shouldBe Some(0)
     }
+
+    "have correct structure for a addition expression with space before addition" in {
+      val cpg            = code("x + y")
+      val List(callNode) = cpg.call.name(Operators.addition).l
+      callNode.code shouldBe "x+y"
```
xavierpinho, line 512: Not something critical, but would it be possible to have the whitespace in `.code` as well?

xavierpinho (approved):

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/passes/ast/SimpleAstCreationPassTest.scala
rahul-privado: The .code part has been built by stitching together individual parts. While I would agree that the whitespace would be accurate, I am not sure whether we could actually stitch it back that accurately.
Let me look further into it. If it is possible without making the code complicated, I will do it in the next PR

PR #2847, opened 2023-06-09
rubysrc2cpg: split and refactor query tests

tuxology:
Plan is to split the IdentifierTests in logical groups (like ControlStructureTests, AssignmentTests, NamespaceTests etc.), keeping all original tests for now. For now, inside each of these files, the test cases would for now be identifier and literal checks primarily. We will then gradually add the actual checks for control structure conditions, assignment LHS/RHS etc. to create more coverage.


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
```diff
@@ -0,0 +1,248 @@
...
+      cpg.identifier
+        .name("x")
+        .l
+        .size shouldBe 2
+    }
+
+    "recognise all call nodes" in {
+      cpg.call
+        .name("num1")
+        .l
```
fabsx00, line 234: `.l.size` : you can just say `.size`

fabsx00 (approved):

Issue #2820, opened 2023-06-06
[Bug] joern-parse crash with "Failure: null"


h3xstream:
**Describe the bug**

joern-parse crash with no detail regarding the failure.

**To Reproduce**

```
$ ./bin/joern/joern-cli/joern-parse --language java xxxxxxxxx.jar
Parsing code at: xxxxxxxxx.jar - language: `java`
[+] Running language frontend
=======================================================================================================
Invoking CPG generator in a separate process. Note that the new process will consume additional memory.
If you are importing a large codebase (and/or running into memory issues), please try the following:
1) exit joern
2) invoke the frontend: /home/philippe/bin/joern/joern-cli/jimple2cpg -J-Xmx3142m xxxxxxxxx.jar --output cpg.bin
3) start joern, import the cpg: `importCpg("path/to/cpg")`
=======================================================================================================

[+] Applying default overlays
Failure: null
```

**Expected behavior**

An exception description, a stacktrace

Continue to build the CPG even their is an unexpected NPE. (possibly a null pointer exception)

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04 (WSL)
 - Joern Version : Latest
 - Java version: openjdk 11.0.19 2023-04-18

Source:
https://github.com/joernio/joern/blob/46fa84574cfaba9ad582ab23483c8c99fb9c4fbc/joern-cli/src/main/scala/io/joern/joerncli/JoernParse.scala#L23

mpollmeier:
I can't reproduce it :(

Might be specific to that jar? Did you test other jars? Can you share the jar, or reduce/modify it to a sharable piece that still trigger the error?

h3xstream:
The issue is that exception details are not provided to be able to troubleshoot.

You can throw any type of exception to reproduce it.

mpollmeier:
Very true, I'll get that sorted. Thanks for reporting!

h3xstream:
For reference, exceptions such as NullPointerException or OutOfMemoryError do not have message detail. Adding the class name of the exception and the stacktrace would be enough to get a better idea about the failure cause.

DavidBakerEffendi:
In terms of `XTypeRecoveryPass` created nodes without attached `CONTAINS` edges, I'll push a fix today. Hopefully this resolves a certain class of these issues.

DavidBakerEffendi:
#2843 Fixes a class of this problem caused by `XTypeRecoveryPass`

mpollmeier:
[...]

PR #2848, opened 2023-06-10
joern-parse exception handling: report root causes

mpollmeier:
prior to this, the exception handling threw away useful information
like stack traces, and was a mix of different styles, including
`Either` and regular `try` blocks. Using `Try` everywhere makes it
more readable and report all information users might want.

fixes https://github.com/joernio/joern/issues/2820

note: this will make the output more verbose - the stack trace was previously hidden..


DavidBakerEffendi:
Looks good, though seems like `JoernSlice` needs to be accommodated with these new changes

mpollmeier:
true that - fixed now

DavidBakerEffendi (approved):

PR #2856, opened 2023-06-11
Allowing for a jRuby backend

xavierpinho:
* Allowing the option to use jRuby's parser in case the on-going ANTLR-based parser performs too poorly.
* Extracts `gem <name>, <version>` dependency declarations from `Gemfile` files.


prabhu:
@xavierpinho If the intention is to only parse Gemfile we don't need a whole new dependency for that. Below is the parsing implementation for Gemfile and Gemfile.lock

https://github.com/CycloneDX/cdxgen/blob/master/utils.js#L2628
https://github.com/CycloneDX/cdxgen/blob/master/utils.js#L2628

prabhu:
@xavierpinho Glad to hear that. To further elaborate, I would love to bundle ruby support with [atom](https://github.com/AppThreat/atom) once it's ready. Since atom is published on npmjs and pypi, I only have 100MB to pack and the existing frontends included have already used up 80MB. So any jars we can exclude would help me as well as joern which is also bloated.

PR #2861, opened 2023-06-12
rubysrc2cpg: refactoring astForAliasStatement

xavierpinho:


ursachec:
I'm not familiar with Ruby's syntax, and I don't know if you already had a conversation with someone about how best to represent aliases - but on a first glance, I'd say that the frontend should create an ast for the alias (e.g. a method stub if that works)

ursachec (approved):
:+1: on the refactoring

xavierpinho:
Oh absolutely - just discussing that with @rahul-privado. The plan is: place `alias` statements in the AST and have a pass later on replacing aliased method call fullnames respectively. Right now, the resolving is done during the AST which can be a mess.

ursachec:
@xavierpinho I'm just throwing ideas out there - but I think it might be possible to remove the need for a separate pass if you stub a method for the alias for which its body consists of a single expression - a call of the original method.

xavierpinho:
> @xavierpinho I'm just throwing ideas out there - but I think it might be possible to remove the need for a separate pass if you stub a method for the alias for which its body consists of a single expression - a call of the original method.

That would be neat, yes! Thanks for the suggestion :) Alas, `undef` is a party pooper:
```ruby
def myMethod(arg); puts arg; end
alias :myAlias :myMethod
myAlias(0) # works
undef :myAlias
myAlias(0) # doesn't work anymore
```
But this is just semantics: hopefully snippets like these are not common!

xavierpinho:
Good stuff! 👍 I think it's worth trying that first, for sure :)

PR #2869, opened 2023-06-13
pysrc2cpg: File & Config Reading Consistency

DavidBakerEffendi:
* Have `pysrc2cpg` read files using `SourceFiles` in order to leverage `ignoreFilesRegex`
* Converted Java's `ConfigFileCreationPass` as an abstract class (`XConfigFileCreationPass`) and implemented a version of Java & Python
* Added ability for Python to ingest TOML, INI, and YAML files as configuration file nodes along with the usual `requirements.txt`
* Pass the `config.inputPath` to Python's frontend a bit more directly to correctly set `cpg.metaData.root`
* Made some adjustments to the part in the type recovery that uses the `root` argument


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XConfigFileCreationPass.scala
```diff
@@ -0,0 +1,75 @@
...
+
+        case _ => Array.empty
+      }
+    }
+
+  override def runOnPart(diffGraph: DiffGraphBuilder, file: File): Unit = {
+    Try(IOUtils.readLinesInFile(file.path)) match {
+      case Success(fileContentsLines) =>
+        val name       = configFileName(file)
+        val content    = fileContentsLines.mkString("\n")
```
xavierpinho, line 47: Having been bitten by this in the past (https://github.com/joernio/joern/issues/2601#issuecomment-1526054382), viz. files with mixed line break conventions, if you don't really care for lines, there's an alternative API (https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1703) that reads the entire file.

DavidBakerEffendi: Good idea, thanks!

xavierpinho (approved):

ml86 (approved):

PR #2870, opened 2023-06-13
rubysrc2cpg: Dataflow unit tests

rahul-privado:
Some dataflow tests


xavierpinho (approved):
I think we could be more explicit in our test titles. For instance, "through expressions" is not particularly clear: perhaps split that test into smaller ones that use only arithmetic/relational/range/etc. operators

rahul-privado:
> I think we could be more explicit in our test titles. For instance, "through expressions" is not particularly clear: perhaps split that test into smaller ones that use only arithmetic/relational/range/etc. operators

Idea behind several expressions in this is to make sure they work when cascaded.

tuxology:
@xavierpinho I am working on last ignored test case here. Will make changes as you suggested in a subsequent PR. Merging this for now

PR #2871, opened 2023-06-13
jssrc2cpg: Include Members for Dynamic Exported Types

DavidBakerEffendi:
Accommodates an edge case where objects exported as types did not have members. I hope this doesn't create members for everything that is defined as an object in this way, but perhaps that's not the worst thing.

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForTypesCreator.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForTypesCreator.scala
@@ -60,6 +60,11 @@ trait AstForTypesCreator { this: AstCreator =>
           .getOrElse(Ast())
       }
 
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -60,6 +60,11 @@ trait AstForTypesCreator { this: AstCreator =>
           .getOrElse(Ast())
       }
 
+    val allClassMembers = classMembersForTypeAlias(alias).toList
+    // adding all class methods / functions and uninitialized, non-static members
+    allClassMembers
+      .filter(member => isClassMethodOrUninitializedMember(member) && !isStaticMember(member))
+      .map(m => astForClassMember(m, aliasTypeDeclNode))
```
max-leuthaeuser, line 67: `.map` could be `.foreach` right?

max-leuthaeuser (approved):

PR #2881, opened 2023-06-15
Fix for astgen version on Windows

max-leuthaeuser:
Replaces: https://github.com/joernio/joern/pull/2879


prabhu:
Thank you. Have verified this fix locally using powershell and the fix worked correctly. To verify this,

- Use just powershell and not WSL
- Have only one copy of astgen.exe in a folder like `C:\Tools` and manually delete all copies of `astgen-win.exe` from within the joern folders
- Run joern cli and import any javascript code like juice-shop

In the master branch, it would error out with the message astgen not found. In the fix branch, the import would work correctly. It would be nice to have another person replicate and verify this fix.

ursachec (approved):

PR #2887, opened 2023-06-15
Update testDistro.sh

iAnonymous3000:


testDistro.sh
```diff
@@ -1,17 +1,34 @@
 #!/bin/bash
-set -e #stop on error
-# set -x #verbose on
 
+set -e # Stop on error
+set -u # Treat unset variables as an error
+# set -x # Uncomment this for debugging
+
+# Define log file path
+LOG_FILE="/var/log/my_script.log"
```
mpollmeier, line 8:
```suggestion
LOG_FILE="target/testDistro.log"
```

mpollmeier: context: 
1) the file name you suggested is very generic, 
2) users can't typically write there:
```
tee: /var/log/my_script.log: Permission denied
```

mpollmeier (changes requested):
apart from the one comment I really like this, thank you for sharing!

testDistro.sh
mpollmeier: you marked this as resolved but forgot to make the change... would you mind doing that so we can get this merged? thanks!

mpollmeier:
superseded by https://github.com/joernio/joern/pull/3505

PR #2899, opened 2023-06-19
Initial gosrc2cpg project structure

pandurangpatil:
1. All the required connecting dots to invoke the go frontend from the Joern.
shell.
2. At present, it only creates an empty CPG and return that CPG.

### DIFF
--- a/console/src/main/scala/io/joern/console/cpgcreation/GoCpgGenerator.scala
+++ b/console/src/main/scala/io/joern/console/cpgcreation/GoCpgGenerator.scala
@@ -8,15 +8,41 @@ import scala.util.Try
 /** Language frontend for Go code. Translates Go source code into Code Property Graphs.
   */
 case class GoCpgGenerator(config: FrontendConfig, rootPath: Path) extends CpgGenerator {
-  private lazy val command: Path = if (isWin) rootPath.resolve("gosrc2cpg.bat") else rootPath.resolve("gosrc2cpg")
+  private lazy val goSrc2CpgCommand: Path =
+    if (isWin) rootPath.resolve("gosrc2cpg.bat") else rootPath.resolve("gosrc2cpg")
 
+  /** Generate a CPG for the given input path. Returns the output path, or None, if no CPG was generated.
+    */
   override def generate(inputPath: String, outputPath: String): Try[String] = {
-    val arguments = List(inputPath) ++ Seq("-o", outputPath) ++ config.cmdLineParams
-    runShellCommand(command.toString, arguments).map(_ => outputPath)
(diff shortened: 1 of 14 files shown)


console/src/main/scala/io/joern/console/cpgcreation/GoCpgGenerator.scala
```diff
@@ -8,22 +8,15 @@ import scala.util.Try
 /** Language frontend for Go code. Translates Go source code into Code Property Graphs.
   */
 case class GoCpgGenerator(config: FrontendConfig, rootPath: Path) extends CpgGenerator {
```
fabsx00, line 10: Since this is `gosrc2cpg` as opposed to `go2cpg`, can you create a corresponding `GoSrcCpgGenerator.scala` instead? That way, the old `go2cpg` (commercial) does not need to be adapted.

pandurangpatil: Should I rename the same class `GoCpgGenerator` to `GoSrcCpgGenerator` or create a new one? If I create new one, only one of them will be invoked when `go` source code is found.

fabsx00: Yes, pelase create a new one because the old one is used by commercial code not visible in joern at the moment.

pandurangpatil: @fabsx00 I tired to create new `GoSrcCpgGenerator` and tried to give first priority to `GoCpgGenerator` first and if it fails then invoke `GoSrcCpgGenerator`. But I couldn't come up with a cleaner way to do it out of the generator. Rather I changed the `GoCpgGenerator` to handle both the use case and check if `go2cpg` script is available, if yes then invoke `go2cpg` frontend. In its absence, it will invoke `gosrc2cpg` frontend.

joern-cli/frontends/gosrc2cpg/src/test/scala/dummy.txt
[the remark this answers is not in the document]
pandurangpatil: done

console/src/main/scala/io/joern/console/cpgcreation/GoCpgGenerator.scala
pandurangpatil: I have verified adding dummy `go2cpg.sh` locally and see if it gets invoked. And it works as expected.

fabsx00: That's also a great solution, thank you.

fabsx00 (approved):

PR #2907, opened 2023-06-20
Unifying Import Resolution

DavidBakerEffendi:
Import resolution until now has been integrated into the type recovery pass. The type recovery pass is an iterate-and-propagate algorithm, where import resolution would be run every iteration for every file. Since the type recovery operates on a fork-join concurrency (since inter-file writing is involved), whereas the import resolution alone is mostly read-only, there is room for optimization. The following changes are made:
* Introduced the `XImportPass` class, which unifies the two `ImportPass` classes from `jssrc2cpg` and `pysrc2cpg`, and implements them using the `ConcurrentWriterCpgPass`.
* Introduced a `XImportResolutionPass` class, which uses the `Tag` node feature to resolve the entities that the `ImportNode`s refer to. The values of the `Tag` node are serialized versions of a `ResolvedImport` trait which is implemented under `XImportResolutionPass` as a number of classes which map the CPG entities being imported (or indicates when they're unknown or are external). `XImportResolutionPass` implements `ConcurrentWriterCpgPass`.
* Removed the import recovery from the `jssrc2cpg` and `pysrc2cpg`'s `XTypeRecovery` implementations and instead added import parsing of these `Tag` nodes in `XTypeRecovery` themselves.

The result is a fast, parallel, import resolution that is run exactly once and a much simpler type recovery that focuses only on the context of the compilation unit for the iterate-and-propagate recovery algorithm.

Breaking changes: `jssrc2cpg.passes.ImportResolutionPass` and `pysrc2cpg.ImportResolutionPass` need to be included after their respective `ImportPass`.

### DIFF
--- a/console/src/main/scala/io/joern/console/cpgcreation/PythonSrcCpgGenerator.scala
+++ b/console/src/main/scala/io/joern/console/cpgcreation/PythonSrcCpgGenerator.scala
@@ -28,6 +28,7 @@ case class PythonSrcCpgGenerator(config: FrontendConfig, rootPath: Path) extends
 
   override def applyPostProcessingPasses(cpg: Cpg): Cpg = {
     new ImportsPass(cpg).createAndApply()
+    new ImportResolverPass(cpg).createAndApply()
     new DynamicTypeHintFullNamePass(cpg).createAndApply()
     new PythonInheritanceNamePass(cpg).createAndApply()
     new PythonTypeRecoveryPass(cpg, XTypeRecoveryConfig(enabledDummyTypes = !pyConfig.forall(_.disableDummyTypes)))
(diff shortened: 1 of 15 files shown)


fabsx00 (approved):
My overall impression here is that the meat of the code is still in the individual versions of the ImportResolverPass while the base class does not do too much. I still think this is a step in the right direction.

johannescoetzee:
My comments about javasrc are all from the perspective of wanting to  drop the JavaParser type solver and use X2Cpg type recovery instead, so they're not quite relevant yet but will hopefully be soon.

DavidBakerEffendi:
@johannescoetzee I agree with having the call be optional, unfortunately the `Import` node doesn't allow for `Tag` nodes to connect - but this can change once this is supported in the `codepropertygraph` schema. These tags definitely add to what kind of properties the `ImportNode` may need, e.g. a description of the entity in the CPG or what it may be if it's external.

Definitely looking forward to developing this a bit further!

PR #2908, opened 2023-06-20
Fix - Dataflows for Java Lambda

khemrajrathore:
The PR fixes the following 
- Flows which originate from inside of lambda and sink is outside of lambda
- The above is achieved by linking the MethodRef of lambda with the corresponding MethodReturn node in dataflow engine


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/TaskCreator.scala
```diff
@@ -93,49 +101,71 @@ class TaskCreator(context: EngineContext) {
...
-            ReachableByTask(taskStack, newPath)
-          }
-        } else {
-          returnStatements.map { returnStatement =>
-            val newPath = Vector(PathElement(methodReturn, result.callSiteStack)) ++ path
-            val taskStack =
-              result.taskStack :+ TaskFingerprint(returnStatement, call :: result.callSiteStack, callDepth + 1)
-            ReachableByTask(taskStack, newPath)
+      outArg.toList.flatMap {
+        case call: Call =>
```
fabsx00, line 105: This first part looks to me as if the `.collect` was just turned into an equivalent `.flatMap`. Can you explain if that is the only change up to line 130?

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/TaskCreator.scala
```diff
@@ -93,49 +101,71 @@ class TaskCreator(context: EngineContext) {
...
-          List[MethodParameterOut]()
-        } else {
-          argToOutputParams(arg).l
-        }
-        outParams
-          .filterNot(_.method.isExternal)
-          .map { p =>
-            val newStack = arg.inCall.headOption.map { x => x :: result.callSiteStack }.getOrElse(result.callSiteStack)
-            ReachableByTask(result.taskStack :+ TaskFingerprint(p, newStack, callDepth + 1), path)
+      args.toList.flatMap {
```
fabsx00, line 133: From here to line 150, are there any changes or is this just formatting?

fabsx00 (commented):
Looks good: just seems like there are a lot of mere format changes. Can you revise this slightly so that the semantics of this change become more apparent?

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/TaskCreator.scala
khemrajrathore: Yes, only formatting and handling of `case _ => Vector.empty` for nonCall nodes

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/TaskCreator.scala
khemrajrathore: Yes, only formatting and handling of `case _ => Vector.empty` for nonExpression nodes

fabsx00 (approved):

PR #2910, opened 2023-06-20
c2cpg: exit early on old JREs

max-leuthaeuser:
We now exit with 1 if the JRE c2cpg runs on is < JRE 17.


fabsx00:
Ok, where does the JRE 17 requirement come from?

max-leuthaeuser:
https://github.com/ShiftLeftSecurity/codescience/pull/7075

fabsx00 (approved):

PR #2915, opened 2023-06-21
Update -  addition of DDG edge for capturedIdentifiers

khemrajrathore:


DavidBakerEffendi:
Ah, this is nice, I think this probably means that we can remove some code in `SourcesToStartingPoints`, e.g. [this line that goes down captured declarations](https://github.com/joernio/joern/blob/451d5519f58a01a37a3b7a21e11412789a74e34f/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala#LL82C38-L82C38)

fabsx00:
Can you provide a brief description of what's happening here?

khemrajrathore:
The Idea here is to create DDG edge for identifiers to their captured CounterParts.

Ex - 
```
List<String> userNamesList = new ArrayList<>();  // LNo 1
userPayload.forEach(item -> {                                  // LNo 2
   sink3(userNamesList);                                           // LNo 3
   sink4(userPayload);                                                // LNo 4
});
sink1(userNamesList);                                                // LNo 6
       ```
       
Here we create following REACHING_DEF edges 
userNamesList(1) ----> userNamesList(3)
userNamesList(3) -----> userNamesList(6)

File: joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala [hunk 1]
```diff
@@ -678,14 +678,12 @@ class DataflowTest extends DataFlowCodeToCpgSuite {
 
     "literal to captured closure" in {
       val literalSource = cpg.literal.codeExact("\"https://test-api-service.com\"").l
-      literalSource.size shouldBe 1
```
DavidBakerEffendi, line 681: I think we should keep the `literalSource.size` check, since it helps us debug the test if it fails

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala
```diff
@@ -172,10 +172,27 @@ class DdgGenerator(semantics: Semantics) {
...
+                val firstUsage = usage.head // First usage of identifier
+                val lastUsage  = usage.last // Last usage of identifier
+                if (
+                  identifier.lineNumber.isDefined && firstUsage.lineNumber.isDefined && (identifier.lineNumber.get <= firstUsage.lineNumber.get)
+                )
+                  Some(identifier, firstUsage)
+                else if (
+                  identifier.lineNumber.isDefined && lastUsage.lineNumber.isDefined && (identifier.lineNumber.get >= lastUsage.lineNumber.get)
+                )
+                  Some(lastUsage, identifier)
```
DavidBakerEffendi, line 187: Why is the order swapped here? Should it not be `Some(identifier, lastUsage)`?

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala
```diff
@@ -172,10 +172,27 @@ class DdgGenerator(semantics: Semantics) {
       val identifierDestPairs =
         method._identifierViaContainsOut
           .flatMap { identifier =>
-            identifierToFirstUsages(identifier).map(usage => (identifier, usage))
+            val firstAndLastUsageByMethod = identifierToFirstUsages(identifier).groupBy(_.method)
+            firstAndLastUsageByMethod.values.flatMap { usage =>
+              if (usage.nonEmpty) {
```
DavidBakerEffendi, line 177: Perhaps instead of all these if-statements calls to `isDefined` we could use some filtering and pattern matching.
```scala
firstAndLastUsageByMethod
   .values
   .filter(_.nonEmpty)
   .map(x => (x.head, x.last))
   .flatMap { case (firstUsage, lastUsage) => 
    (identifier.lineNumber, firstUsage.lineNumber, lastUsage.lineNumber) match {
       case (Some(iNo), Some(fNo), _) if iNo <= fNo => Some(identifier, firstUsage)
       case (Some(iNo), _, Some(lNo)) if iNo >= lNo => Some(identifier, lastUsage)
       case _ => None
    }
 }

**Review by DavidBakerEffendi — CHANGES_REQUESTED** [review]

File: dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala (hunk 2 above)
Comment (khemrajrathore) [reply dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala:187]: having the edge from `lastUsage` to `identifier` empowers us to get the logical flow, as the final flow should be from a linenumber which is less than the node with a higher linenumber

File: joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala (hunk 1 above)
Comment (khemrajrathore) [reply joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala:681]: Ok, will update this

File: dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala (hunk 2 above)
Comment (DavidBakerEffendi) [reply dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala:187]: Ah, right now I understand, sorry seems I had a misunderstanding of the output

File: dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala (hunk 3 above)
Comment (DavidBakerEffendi) [reply dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala:177]: Updated this to reflect the original ordering

File: dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala [hunk 4]
```diff
@@ -172,10 +172,20 @@ class DdgGenerator(semantics: Semantics) {
...
+              .map(x => (x.head, x.last))
+              .flatMap { case (firstUsage, lastUsage) =>
+                (identifier.lineNumber, firstUsage.lineNumber, lastUsage.lineNumber) match {
+                  case (Some(iNo), Some(fNo), _) if iNo <= fNo => Some(identifier, firstUsage)
+                  case (Some(iNo), _, Some(lNo)) if iNo >= lNo => Some(lastUsage, identifier)
+                  case _                                       => None
+                }
+              }
           }
+          .toSet
```
DavidBakerEffendi, line 187: If the output is going to be a list in any case, best to use `distinct` instead of creating a set then a list.

khemrajrathore: good catch

khemrajrathore: @DavidBakerEffendi Have updated the PR

DavidBakerEffendi (approved):
@fabsx00 maybe you should give some eyes on it, but LGTM

Issue #2900, opened 2023-06-19
[Bug] unresolved calls have methodFullName = "<empty>"


maltek:
```
joern> workspace.reset ; importCode.php.fromString(
               """<?php                                                                        
                 |$a->b();  
                 |""".stripMargin)
joern> cpg.call.methodFullName("<empty>").head 
res47: Call = Call(
  id = 7L,
  argumentIndex = -1,
  argumentName = None,
  code = "$a->b()",
  columnNumber = None,
  dispatchType = "DYNAMIC_DISPATCH",
  dynamicTypeHintFullName = ArraySeq(),
  lineNumber = Some(value = 2),
  methodFullName = "<empty>",
  name = "b",
  order = 2,
  signature = "<unresolvedSignature>(0)",
  typeFullName = "ANY"
)
```

we end up with bogus `<empty>` method stubs because of this:

```
joern> cpg.method.fullName("<empty>").l 
res49: List[Method] = List(
  Method(
    id = 20L,
    astParentFullName = "<global>",
    astParentType = "NAMESPACE_BLOCK",
    code = "<empty>",
    columnNumber = None,
    columnNumberEnd = None,
    filename = "<empty>",
    fullName = "<empty>",
    hash = None,
    isExternal = true,
    lineNumber = None,
    lineNumberEnd = None,
    name = "b",
    order = 0,
    signature = "<unresolvedSignature>(0)"
  )
)
```

PR #2938, opened 2023-06-26
php2cpg: use sensible unresolved call names

johannescoetzee:
Fixes https://github.com/joernio/joern/issues/2900

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
@@ -89,7 +89,7 @@ class CallTests extends PhpCode2CpgFixture {
 
     inside(cpg.call.l) { case List(fooCall) =>
       fooCall.name shouldBe "foo"
-      fooCall.methodFullName shouldBe s"<unresolvedNamespace>\\$$f->foo"
+      fooCall.methodFullName shouldBe """<unresolvedNamespace>\$f->foo"""
       fooCall.dispatchType shouldBe DispatchTypes.DYNAMIC_DISPATCH
       fooCall.lineNumber shouldBe Some(2)
       fooCall.code shouldBe "$f->foo($x)"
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -89,7 +89,7 @@ class CallTests extends PhpCode2CpgFixture {
 
     inside(cpg.call.l) { case List(fooCall) =>
       fooCall.name shouldBe "foo"
-      fooCall.methodFullName shouldBe s"<empty>"
+      fooCall.methodFullName shouldBe s"<unresolvedNamespace>\\$$f->foo"
```
maltek, line 92: I like writing strings like this with `"""` so that there is no need to escape anything:

```suggestion
      fooCall.methodFullName shouldBe """<unresolvedNamespace>\$f->foo"""
```

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -109,7 +109,7 @@ class CallTests extends PhpCode2CpgFixture {
 
     inside(cpg.call.filter(_.name != Operators.fieldAccess).l) { case List(fooCall) =>
       fooCall.name shouldBe "$foo"
-      fooCall.methodFullName shouldBe s"<empty>"
+      fooCall.methodFullName shouldBe s"<unresolvedNamespace>\\$$$$f->$$foo"
```
maltek, line 112:
```suggestion
      fooCall.methodFullName shouldBe """<unresolvedNamespace>\$$f->$foo"""
```

maltek (approved):
When opening the issue I was just thinking of changing the `<empty>` to an actual empty string (like we do in JavaScript). This is probably more helpful short-term, though I suspect we'll have to change it again once we try to get the dataflow engine to work for these calls.

Issue #2866, opened 2023-06-12
php2cpg: model ::class references as TypeRef nodes


maltek:
currently it's a field access of a field named `class` of an identifier named after the class:

```
joern> workspace.reset ; importCode.php.fromString("""<?
       namespace A { class Class {} }
       $x = A\Class::class;
       """) 
joern> cpg.assignment.source.ast.l 
res8: List[AstNode] = List(
  Call(
    id = 10L,
    argumentIndex = 2,
    argumentName = None,
    code = "A\\A\\Class::class",
    columnNumber = None,
    dispatchType = "STATIC_DISPATCH",
    dynamicTypeHintFullName = ArraySeq(),
    lineNumber = Some(value = 3),
    methodFullName = "<operator>.fieldAccess",
    name = "<operator>.fieldAccess",
    order = 2,
    signature = "",
    typeFullName = "ANY"
  ),
  Identifier(
    id = 11L,
    argumentIndex = 1,
    argumentName = None,
    code = "A\\A\\Class",
    columnNumber = None,
    dynamicTypeHintFullName = ArraySeq(),
    lineNumber = Some(value = 3),
    name = "A\\A\\Class",
    order = 1,
    typeFullName = "ANY"
  ),
  FieldIdentifier(
    id = 12L,
    argumentIndex = 2,
    argumentName = None,
    canonicalName = "class",
    code = "class",
    columnNumber = None,
    lineNumber = Some(value = 3),
    order = 2
  )
)
```

Issue #2822, opened 2023-06-07
php2cpg: model true/false as literals instead of field access


maltek:
I don't believe it breaks anything, but `true/false` currently being field accesses seems weird to me.

PR #2940, opened 2023-06-26
Fix php magic constants that were represented as field accesses

johannescoetzee:
Fixes https://github.com/joernio/joern/issues/2866
Fixes https://github.com/joernio/joern/issues/2822


maltek (approved):
I know I didn't complain about this before, but `null` could use the same treatment as `true/false`. Clearly no need to block this PR, but maybe something to also sneak in.

(Also, this will break our PHP tests in CS. Will be easy enough to adjust, so just FYI.)

PR #2948, opened 2023-06-27
jssrc2cpg: partially revert dynamic type hints

max-leuthaeuser:
- joernio/joern@eba81a2 - adds highly questionable type hints for 'this' outside explicit classes - reverted this
- joernio/joern@ea2046f - CS can't handle dynamic type hints coming from TS in this case here - reverted this partially

To make progress on: https://github.com/ShiftLeftSecurity/codescience/pull/7001

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstCreator.scala
@@ -108,9 +108,7 @@ class AstCreator(
     scope.pushNewMethodScope(fullName, name, blockNode, None)
     localAstParentStack.push(blockNode)
 
-    val thisParam =
-      parameterInNode(astNodeInfo, "this", "this", 0, false, EvaluationStrategies.BY_VALUE)
-        .dynamicTypeHintFullName(typeHintForThisExpression())
+    val thisParam = parameterInNode(astNodeInfo, "this", "this", 0, isVariadic = false, EvaluationStrategies.BY_VALUE)
     scope.addVariable("this", thisParam, MethodScope)
 
     val methodChildren = astsForFile(astNodeInfo)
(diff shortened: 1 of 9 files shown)


max-leuthaeuser:
@maltek Once that is in, I will regenerate sptests expectations after jssrc2cpg-internal has been bumped for another comparison.

fabsx00 (changes requested):
Would like to know what @DavidBakerEffendi has to say about this one.

DavidBakerEffendi:
So with this PR, it brings back an instance where a lot of the `this` identifiers have `this` as a dynamic type hint, where the minority of `this` identifiers have the respective `:program` or class. What is the reason for this? e.g. from this branch we see
```json
  {
    "dynamicTypeHintFullName":[
      "this"
    ],
    "name":"this",
    "code":"this",
    "typeFullName":"ANY",
    "columnNumber":6,
    "order":1,
    "_label":"IDENTIFIER",
    "argumentIndex":1,
    "lineNumber":68,
    "id":4707
  }
```

Unfortunately, the PRs being reverted didn't come with accompanying tests, so the pipeline is deceptively green, but the test I was working with at the time looked something like:
```javascript
// test.js::program

function foo() {};

var refThis = this;
refThis.foo(); 
```
Since `this` had no sensible type, the type recovery didn't easily resolve what `refThis` could be, so the `refThis.foo()` call edge was missing.

Should I instead add the respective types in the type recovery?

max-leuthaeuser:
Maybe @maltek can elaborate but from what I understood here was:
- eba81a2: breaks for some JS patterns and sets wrong type hints as often as it sets the correct ones. That breaks linking in CS and, ultimately, sptests. Either we use this revert here or extract it into a separate pass that only runs in joern but not in CS.
- ea2046f: CS can't handle most of the dynamic type hints from TS and, thus, results in the same problems as eba81a2.

DavidBakerEffendi:
Well I think `this` is obviously not a valid type hint, so it would be a regression to bring that back. Which makes me wonder if reverting is the correct move here? 

I am happy to help remediate a fix that solves the bug you mentioned about `this` being incorrect in certain causes, but perhaps it can simply be done in another PR which only adds types to these where the confidence is high.

As for "CS can't handle most of the dynamic type hints from TS", this sounds like a bug in CS? Can't CS remove these type hints? But perhaps more elaboration is required around this point.

max-leuthaeuser:
Closing this for now. We will disable the usage of tsc types in sptests, cs test, and for SL.
That ensures cs won't break for strange TS types until we find a better solution.

PR #2952, opened 2023-06-27
add - Node addition for Imports

khemrajrathore:
This PR covers the following for the Go frontend
- Read the JSON output of the parser and map it to Scala objects
- Add Import nodes to CPG, by reading the import statements from the JSON output
- Handle various combinations of import
- Move import node creation utility to X2Cpg from C2Cpg


fabsx00 (approved):
Looks good! Can you edit the PR description to provide a better description? Apart from that, this looks good.

PR #2963, opened 2023-06-29
rubysrc2cpg: Use XConfigFileCreationPass

xavierpinho:


xavierpinho:
Do you think it would be a good idea for `XConfigFileCreationPass` to expose `rootDir`? In this particular example, it would allow us to remove the `inputPath` parameter from rubysrc2cpg's `ConfigFileCreationPass`?

DavidBakerEffendi (approved):
@xavierpinho Funny you mention that, I think `cpg.metaData.root` should be that path? But I think I had it as an explicit parameter because of the funky way PySrc used to load files. If you find a way to improve that then I say go ahead.

PR #2964, opened 2023-06-29
[rubysrc2cpg] Fix Class Stack & Method Full Names

DavidBakerEffendi:
* The class stack & scope are not being used properly, while this does not address the scope issue, this does start addressing the class stack issue
* Filename was used in the place of what a relative file name should be used, thus method full names were incorrect
* Method full names and calls also always assumed they were at the base of the file, and did not include the `:program` class (it was originally missing from the stack)
* Added the `NaiveCallLinker` to the post-processing pass to get some call linking going
* Creating very basic type declaration nodes for now, but this is to be revisited in another PR, as it cannot proceed until this work is merged
* Removed `<unresolvedSignature>` from dependency call resolver as there is no overloading in Ruby

### DIFF
--- a/console/build.sbt
+++ b/console/build.sbt
@@ -14,6 +14,7 @@ dependsOn(
   Projects.javasrc2cpg,
   Projects.jssrc2cpg,
   Projects.pysrc2cpg,
+  Projects.rubysrc2cpg,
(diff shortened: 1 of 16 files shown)


fabsx00 (approved):

xavierpinho (approved):
🎉

PR #2975, opened 2023-07-02
[gosrc2cpg] Refactoring of code and removed part of duplicate code

pandurangpatil:
1. Made changes to return Seq[Ast] by every Ast creator function. This
will facilitate reusable methods that can follow consistent return types.
Where ever caller expects a single value to be returned, we can always
extract the value by retrieving the head of the sequence.
2. Changed the places where empty Ast() nodes are returned and instead
returned empty Seq().


pandurangpatil:
Closing this PR by creating a new PR for the same change with a clean branch created from the master. Refer this PR - https://github.com/joernio/joern/pull/2981

PR #2971, opened 2023-06-30
Revert "jssrc2cpg: Handle Destructured Params (#2883)"

max-leuthaeuser:
This reverts commit b5336999a69a25cccc69531629af660e263d9999.

@DavidBakerEffendi Had to revert this as it creates identifier nodes not being included into the AST. That crashes many passes later on. Check e.g., tarpit-nodejs

PR #2979, opened 2023-07-03
[jssrc2cpg] Fixed dangling Cfg Nodes From Destruct Params

DavidBakerEffendi:
Related to [#2971](https://github.com/joernio/joern/pull/2971), prevents the case where the destructuring of parameter operations are not linked to the parent block node correctly.


max-leuthaeuser (approved):

max-leuthaeuser:
Thanks a lot!

PR #2960, opened 2023-06-28
[gosrc2cpg] Formation of fully qualified namespace from package name and Initial base AST nodes along with namespace blocks.

pandurangpatil:
**Formation of fully qualified namespace from package name**

1. Updated goastparser version to pickup the changes to read go.mod file metadata.
2. Made changes to separately identify the `go.mod.json` file
3. Used circa library to unmarshal JSON to scala entity to read `.mod.json`
4. Started added NamespaceBlock unit test, with one working unit test
5. Prepared the utility which will form the namespace based on the information available along with respective unit tests covering majority of the possibilities.

a. If there is no `go.mod` file and `<root project path>/test.go` (package `main`) => `main`
b. `<root project path>/first/second/test.go` (package `main`) => `/first/second/main`
c. `<root project path>/test.go` (package `main`) => `main`
d. `go.mod` (module `jorn.io/trial`) and `<root project path>/foo.go`(package `trial`) => `jorn.io/trial`
e. `go.mod` (module `jorn.io/trial`) and `<root project path>/foo.go` (package `foo`) => `jorn.io/trial>foo`
f. `go.mod` (module `jorn.io/trial`) and `<root project path>/first/foo.go` (package `first`) => `jorn.io/trial/first`
g. `go.mod` (module `jorn.io/trial`) and `<root project path>/first/foo.go` (package `bar`) => `jorn.io/trial/first>bar`

if you pay attention to `e` and `g` use case where package name is not matching with the wrapped folder. Then in that case go makes use of alias mechanism like

If I consider `g` use case and try to import that package somewhere else I will be doing it as follows

`import bar "jorn.io/trial/first"`. I am representing this alias as `jorn.io/trial/first>bar` to form the namespace. If anyone has any different suggestions add your comments.

**Initial base AST nodes along with namespace blocks.**
1. Incorporated initial root AST nodes using the package full name generated using the utility.
2. Added utility method in `x2cpg` to create `method return node`. Have added a respective comment in c2cpg where I tried to use the method from `x2cpg` and tried removing it from `c2pg`. However it occurred to me, it has been getting used from multiple places in `c2cpg` module. Hence I left with TODO
3. Addressed all the review comments by David.
4. There is still scope for few more unit tests, I will try to cover them in next PR.

PR #2980, opened 2023-07-03
simplify versions: use the same circeVersion as globally (0.14.5)

mpollmeier:
fixup for https://github.com/joernio/joern/pull/2960


mpollmeier:
cc @pandurangpatil

DavidBakerEffendi (approved):
I was wondering if we had something like this. Good to know!

PR #2982, opened 2023-07-03
Add javasrc2cpg option to get JDK type information from jars/jmods.

johannescoetzee:
Also adds a `.withConfig` method to `TestCpg` with an example of usage in javasrc to make testing various configurations easier. If this PR is merge, I'll create a follow-up PR to make use of that for other frontends as well.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JdkJarTypeSolver.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JdkJarTypeSolver.scala
@@ -91,11 +91,14 @@ class JdkJarTypeSolver private (jdkPath: String) extends TypeSolver {
   }
 
   private def addPathToClassPool(archivePath: String): Try[ClassPath] = {
-    if (archivePath.isJarPath) Try(classPool.appendClassPath(archivePath))
-    else if (archivePath.isJmodPath)
+    if (archivePath.isJarPath) {
+      Try(classPool.appendClassPath(archivePath))
+    } else if (archivePath.isJmodPath) {
       val classPath = new JmodClassPath(archivePath)
       Try(classPool.appendClassPath(classPath))
-    else Failure(new IllegalArgumentException("$archivePath is not a path to a jar/jmod"))
+    } else {
+      Failure(new IllegalArgumentException("$archivePath is not a path to a jar/jmod"))
+    }
   }
 
   def withJars(archivePaths: Seq[String]): JdkJarTypeSolver = {
@@ -105,7 +108,7 @@ class JdkJarTypeSolver private (jdkPath: String) extends TypeSolver {
 
   def addArchives(archivePaths: Seq[String]): Unit = {
     archivePaths.foreach { archivePath =>
-      Try(addPathToClassPool(archivePath)) match {
+      addPathToClassPool(archivePath) match {
         case Success(_) => registerKnownClassesForJar(archivePath)
 
         case Failure(e) =>
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JdkJarTypeSolver.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JdkJarTypeSolver.scala
@@ -20,6 +20,7 @@ import io.joern.x2cpg.SourceFiles
 import javassist.CtClass
 import com.github.javaparser.symbolsolver.javassistmodel.JavassistFactory
 import javassist.NotFoundException
+import javassist.ClassPath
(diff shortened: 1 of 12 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JdkJarTypeSolver.scala
```diff
@@ -0,0 +1,171 @@
...
+import com.github.javaparser.symbolsolver.javassistmodel.JavassistFactory
+import javassist.NotFoundException
+
+class JdkJarTypeSolver private (jdkPath: String) extends TypeSolver {
+
+  private val logger = LoggerFactory.getLogger(this.getClass())
+
+  private var parent: Option[TypeSolver] = None
+  private val classPool                  = new NonCachingClassPool()
+  private val knownClassNames            = mutable.Map[String, String]()
```
ml86, line 30: Please give this a proper name and a description. Since the type is `Map[String, String]` this is not clear from just the types and the current name would be more fit for something like a `Set`.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/ClassLoaderTypeTests.scala
```diff
@@ -0,0 +1,98 @@
+package io.joern.javasrc2cpg.querying
+
+import io.joern.javasrc2cpg.Config
+import io.joern.javasrc2cpg.testfixtures.JavaSrcCode2CpgFixture
+import io.shiftleft.semanticcpg.language._
+import io.joern.x2cpg.utils.ExternalCommand
+import org.scalatest.Ignore
+
+@Ignore
+class ClassLoaderTypeTests extends JavaSrcCode2CpgFixture {
```
ml86, line 10: In addition to these not active tests, i would like to see a always running test which takes `System.getProperty("java.home")` and feeds it to `withJdkPath`. You can than check for some types which have Java 8 as minimum requirement so that the test can succeed with all expected Java versions. (Java 8 until most recent).

ml86 (changes requested):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JdkJarTypeSolver.scala
```diff
@@ -0,0 +1,171 @@
...
+        classPool.appendClassPath(jarPath)
+
+      case jmodPath =>
+        classPool.appendClassPath(new JdkArchiveClassPath(jmodPath))
+    }
+  }
+
+  private def withJars(jarPaths: Seq[String]): JdkJarTypeSolver = {
+    jarPaths.foreach { jarPath =>
+      Try(classPool.appendClassPath(new JdkArchiveClassPath(jarPath))) match {
```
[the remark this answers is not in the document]
johannescoetzee: We don't need to, but I did it this way since the `JdkArchiveClassPath` handles the JDK jars as well (and will any other jars that don't have classes in a `classes.*` package. I changed this for the latest iteration, however, since we could use this class for all jar types in future. The new class path is now `JmodClassPath` to make it clearer.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JdkJarTypeSolver.scala
```diff
@@ -84,26 +90,28 @@ class JdkJarTypeSolver private (jdkPath: String) extends TypeSolver {
...
+    this
   }
 
-  private def withJars(jarPaths: Seq[String]): JdkJarTypeSolver = {
-    jarPaths.foreach { jarPath =>
-      Try(classPool.appendClassPath(new JdkArchiveClassPath(jarPath))) match {
-        case Success(_) => registerKnownClassesForJar(jarPath)
+  def addArchives(archivePaths: Seq[String]): Unit = {
+    archivePaths.foreach { archivePath =>
+      Try(addPathToClassPool(archivePath)) match {
```
ml86, line 108: This `Try()` wrapping seems unnecessary as `addPathToClassPool` already returns a `Try`. If the wrapping is only to catch possible exceptions coming from `new JmodClassPath(...` in line 96, than i suggest you do the transition from exception to `Try` there.

ml86 (approved):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JdkJarTypeSolver.scala
johannescoetzee: Good point. Updating the PR

PR #2986, opened 2023-07-03
GoSrc2Cpg : If Else statement Ast handling

khemrajrathore:
The PR covers the following
- Basic Ast handling for IF-Else block
- Nested If-else block
- Implemented `code` method which return code snippet for a Node

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
@@ -14,7 +14,7 @@ trait AstCreatorHelper { this: AstCreator =>
   private val MinCodeLength: Int = 50
 
   protected def createParserNodeInfo(json: Value): ParserNodeInfo = {
-    val c     = shortenCode(code(json))
+    val c     = shortenCode(code(json).toOption.getOrElse(""))
     val ln    = line(json)
     val cn    = column(json)
     val lnEnd = lineEndNo(json)
@@ -24,7 +24,7 @@ trait AstCreatorHelper { this: AstCreator =>
   }
 
   private def nodeType(node: Value): ParserNode = fromString(node(ParserKeys.NodeType).str)
-  protected def code(node: Value): String = {
+  protected def code(node: Value): Try[String] = Try {
 
     val lineNumber    = line(node).get
     val colNumber     = column(node).get - 1
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
@@ -26,19 +26,18 @@ trait AstCreatorHelper { this: AstCreator =>
   private def nodeType(node: Value): ParserNode = fromString(node(ParserKeys.NodeType).str)
   protected def code(node: Value): String = {
 
-    val lineNumber = line(node).get
-    val colNumber = column(node).get - 1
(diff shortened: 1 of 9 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -25,9 +24,26 @@ trait AstCreatorHelper { this: AstCreator =>
...
-    // TODO Need to build a utlity which returns code for a node
-    ""
+  protected def code(node: Value): Try[String] = Try {
+
+    val lineNumber    = line(node).get
+    val colNumber     = column(node).get - 1
+    val lineEndNumber = lineEndNo(node).get
+    val colEndNumber  = columnEndNo(node).get - 1
+
+    var nodeCode = ""
```
DavidBakerEffendi, line 34: I would discourage the use of mutable variables. `If-else` also returns values. So for this I'd also perhapse encourage a `StringBuilder` since those are faster at building strings. Also, there is an `Iterator` functional style of conditionally iterating which is preferable as it doesn't use mutable variables. See below:
```scala
    if (lineNumber == lineEndNumber) {
       lineNumberMapping(lineNumber).substring(colNumber, colEndNumber)
     } else {
       val sb = new mutable.StringBuilder()
       sb.append(lineNumberMapping(lineNumber).substring(colNumber))
       Iterator
         .from(lineNumber + 1)
         .takeWhile(currentLineNumber => currentLineNumber < lineEndNumber)
         .foreach(currentLineNumber => sb.append(s"\n${lineNumberMapping(currentLineNumber)}"))
       sb.toString()
     }
```

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -4,26 +4,34 @@ import io.joern.gosrc2cpg.parser.ParserAst._
...
 import scala.util.Try
 
 trait AstForStatementsCreator { this: AstCreator =>
-  def astForBlockStatement(blockStmt: ParserNodeInfo): Seq[Ast] = {
-
-    val newBlockNode = blockNode(blockStmt)
+  def astForBlockStatement(blockStmt: ParserNodeInfo, order: Int = -1): Ast = {
+    val newBlockNode = blockNode(blockStmt, Defines.empty, Defines.voidTypeName).order(order).argumentIndex(order)
     scope.pushNewScope(newBlockNode)
+    var currOrder = 1
```
DavidBakerEffendi, line 15: Instead of using `currOrder`, use the `withIndex` method exposed by `X2Cpg`'s `AstCreatorBase`. This is similar to `zipWithIndex` but will start at 1.

DavidBakerEffendi (changes requested):

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
khemrajrathore: Thanks, will make this change

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForStatementsCreator.scala
khemrajrathore: @DavidBakerEffendi We cannot use `withIndex` here as we are updating `currOrder` based on Ast length from inside Expression

khemrajrathore:
@DavidBakerEffendi Have made the requested change

DavidBakerEffendi (approved):

PR #2989, opened 2023-07-03
Revert "kotlin2cpg: bump version of embedded compiler (#2984)"

ursachec:
This reverts commit ce05d56e26d57f37b958d65ccfb1dff8b350cb58.


ursachec:
The GitHub Actions quota for Git LFS is already exceeded and the price is really high given that we only have one file over 50 MB (at 70MB) which needs it

ursachec:
Bringing the change for the kotlin compiler version bump without enabling git lfs in a separate pr

xavierpinho (approved):

PR #2992, opened 2023-07-04
rubysrc2cpg: More test case and fixes

rahul-privado:
1. Lot of new test cases
2. Fixes for some test cases. Added several new disabled test cases for later fixing
3. Fixes for crashes while loading repositories
4. Comments in the code to track test case status of code that is not covered under successful cases

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
@@ -254,11 +254,12 @@ class AstCreator(
   def astForMultipleRightHandSideContext(ctx: MultipleRightHandSideContext): Seq[Ast] = {
     if (ctx == null) return Seq(Ast())
 
-    val exprAsts = if (ctx.expressionOrCommands() != null) {
-      ctx.expressionOrCommands().expressionOrCommand().asScala.flatMap(astForExpressionOrCommand).toSeq
-    } else {
-      Seq()
-    }
+    val expCmd = ctx.expressionOrCommands()
+    val exprAsts = Option(expCmd) match
+      case Some(expCmd) =>
+        expCmd.expressionOrCommand().asScala.flatMap(astForExpressionOrCommand).toSeq
+      case None =>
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -213,7 +213,7 @@ class AstCreator(
       val xAsts = astForPrimaryContext(ctx.primary())
       val localVar = {
         if (ctx.LOCAL_VARIABLE_IDENTIFIER() != null) {
-          ctx.LOCAL_VARIABLE_IDENTIFIER()
+          ctx.LOCAL_VARIABLE_IDENTIFIER() // failing dataflow test case
```
xavierpinho, line 216: This (and similar) comment needs to be further elaborated. Right now it doesn't mean much.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -254,7 +254,11 @@ class AstCreator(
   def astForMultipleRightHandSideContext(ctx: MultipleRightHandSideContext): Seq[Ast] = {
     if (ctx == null) return Seq(Ast())
 
-    val exprAsts = ctx.expressionOrCommands().expressionOrCommand().asScala.flatMap(astForExpressionOrCommand).toSeq
+    val exprAsts = if (ctx.expressionOrCommands() != null) {
```
xavierpinho, line 257: It would be more idiomatic to use `Option` instead of a null-comparison. There are quite a few examples in the code base. Please check look them up.

xavierpinho (changes requested):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
rahul-privado: Done

PR #2996, opened 2023-07-04
Enable withConfig for rubysrc2cpg

johannescoetzee:
https://github.com/joernio/joern/pull/2982 added a `withConfig` method to testCpgs to set the config on a per-testcase basis without having to wire class parameters through the various fixtures. This PR enables this for rubysrc2cpg.


xavierpinho (approved):
Nice!

PR #3006, opened 2023-07-04
Replace javasrc2cpg ReflectionTypeSolver with JdkJarTypeSolver at java home

johannescoetzee:
Using the `ReflectionTypeSolver` and the `JdkJarTypeSolver` with jars from the current runtime java home give similar, but not identical results. This is problematic for getting consistent results, particularly when updating expectations for SP tests. This PR switches to the `JdkJarTypeSolver` as the single "java stdlib" type solver for the sake of consistency.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
@@ -50,4 +50,22 @@ object JavaSrc2Cpg {
       new JavaTypeHintCallLinker(cpg)
     )
   }
+
+  def showEnv(): Unit = {
+    val value =
+      JavaSrcEnvVar.values.foreach { envVar =>
+        val currentValue = Option(System.getenv(envVar.name)).getOrElse("<unset>")
+        println(s"${envVar.name}:")
+        println(s"- Description  : ${envVar.description}")
+        println(s"- Current value: $currentValue")
+      }
+  }
+
+  enum JavaSrcEnvVar(val name: String, val description: String) {
+    case JdkPath
+        extends JavaSrcEnvVar(
+          "JAVASRC_JDK_PATH",
+          "Path to the JDK home used for retrieving type information about builtin Java types."
+        )
+  }
 }
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
@@ -56,8 +56,8 @@ object JavaSrc2Cpg {
       JavaSrcEnvVar.values.foreach { envVar =>
         val currentValue = Option(System.getenv(envVar.name)).getOrElse("<unset>")
         println(s"${envVar.name}:")
-        println(s"- Description  : ${envVar.description}")
-        println(s"- Current value: $currentValue")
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
```diff
@@ -90,15 +90,14 @@ class AstCreationPass(config: Config, cpg: Cpg, preCreatedAsts: Option[SplitJpAs
...
-    config.jdkPath match {
-      case Some(path) =>
-        val jdkJarTypeSolver = JdkJarTypeSolver.fromJdkPath(path)
-        combinedTypeSolver.add(jdkJarTypeSolver)
-
-      case None =>
-        combinedTypeSolver.add(new CachingReflectionTypeSolver())
+    val jdkPath = config.jdkPath.getOrElse {
+      val javaHome = System.getProperty("java.home")
+      logger.debug("No explicit jdkPath set in config, so using system java.home at $javaHome")
```
ml86, line 95: That is definitely worth logging as INFO

ml86 (approved):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
```diff
@@ -50,4 +50,22 @@ object JavaSrc2Cpg {
...
+    val value =
+      JavaSrcEnvVar.values.foreach { envVar =>
+        val currentValue = Option(System.getenv(envVar.name)).getOrElse("<unset>")
+        println(s"${envVar.name}:")
+        println(s"- Description  : ${envVar.description}")
+        println(s"- Current value: $currentValue")
+      }
+  }
+
+  enum JavaSrcEnvVar(val name: String, val description: String) {
```
johannescoetzee, line 64: This is perhaps excessive for a single environment variabl, but I also think that, if environment variables are going to be used for configuration, they should be documented somehow and easily discoverable. I plan to extend this idea to the rest of joern (I know at least php2cpg uses some environment variables as well), but put this here since it's currently limited to javasrc.

johannescoetzee:
Unit tests are running into `OutOfMemory` errors, so that's something to investigate before merging.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/Main.scala
```diff
@@ -84,14 +89,25 @@ private object Frontend {
...
       opt[String]("jdk-path")
         .action((path, c) => c.withJdkPath(path))
-        .text("JDK used for resolving builtin Java types. If not set, current classpath will be used")
+        .text("JDK used for resolving builtin Java types. If not set, current classpath will be used"),
+      opt[Unit]("show-env")
+        .action((_, c) => c.withShowEnv(true))
+        // This should really be a print-and-exit but, with the current scopt setup, input paths
+        // are still required, so for now `javasrc2cpg --show-env <inputs>` is less confusing
+        // than `javasrc2cpg --show-env <dummy input to keep scopt happy>`
+        .text("print information about environment variables used by javasrc2cpg prior to analysis")
```
ml86, line 98: Add to the comment that this prevents the frontend from doing its actual task.

ml86 (approved):

johannescoetzee:
Running just the `javasrc2cpg` test suite locally doesn't seem to be using much heap memory at all (less than 160M), so I suspect we're very close to the limit and _slightly_ increased memory use from this was just enough to push us over. This theory is still to be confirmed.

ml86 (approved):

PR #3016, opened 2023-07-04
[gosrc2cpg] - Method node partial structure

pandurangpatil:
Method node partial structure
1. Handled different combination parameters while creating the signature
for the method.

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/test/scala/io/joern/go2cpg/passes/ast/MethodTests.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/test/scala/io/joern/go2cpg/passes/ast/MethodTests.scala
@@ -1,7 +1,9 @@
 package io.joern.go2cpg.passes.ast
 
 import io.joern.go2cpg.testfixtures.GoCodeToCpgSuite
-import io.shiftleft.semanticcpg.language._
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/gosrc2cpg/src/test/scala/io/joern/go2cpg/passes/ast/MethodTests.scala
```diff
@@ -0,0 +1,136 @@
...
+        |""".stripMargin,
+      "go.mod"
+    ).moreCode(
+      """
+        |package fpkg
+        |type Sample struct {
+        |	name int
+        |}
+        |""".stripMargin,
+      "fpkg/lib.go"
```
DavidBakerEffendi, line 110:
```suggestion
      Seq("fpkg", "lib.go").mkString(File.separator)
```

DavidBakerEffendi (approved):

joern-cli/frontends/gosrc2cpg/src/test/scala/io/joern/go2cpg/passes/ast/MethodTests.scala
pandurangpatil: done

PR #3027, opened 2023-07-05
[benchmarks] IntelliJ Warning on T for `Array.ofDim`

DavidBakerEffendi:
Seems that the type could not be implicitly inferred. for `ofDim` so this has been added explicitly. Seems to only happen on IntelliJ and not cause issues on SBT via terminal. Possibly this is coming from Scala 3 migration.

### DIFF
--- a/benchmarks/src/test/scala/io/joern/benchmarks/BenchmarkTags.scala
+++ b/benchmarks/src/test/scala/io/joern/benchmarks/BenchmarkTags.scala
@@ -95,8 +95,8 @@ object BenchmarkTags {
   }
 
   private def getTotalTests: (Int, Int, Int, Int, Int) = {
-    val impl = confusionMatrix.getOrElse(ImplicitFlows.name, Array.ofDim(4))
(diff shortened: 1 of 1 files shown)


mpollmeier:
Yes very likely that this is coming from the Scala 3 migration, or rather Intellij's nonperfect support for it. It definitely helps (and doesn't harm) to merge this. 

Just to double check: for me (with intellij 2023.1.3 and scala plugin version 2023.1.565) it does derive the type correctly to `Array[Int]`, however it complains about a missing TypeTag - is that the same for you?
grafik

mpollmeier (approved):

DavidBakerEffendi:
@mpollmeier My Scala plugin is 2023.1.19 from what I see but I have the exact same phenomena as you've shown above

DavidBakerEffendi:
Yes the IntellIj support has been bad for Scala 3 imo

mpollmeier:
even if there's no difference for this specific issue, it's worth changing to the nightly plugin builds IMO...
grafik

Issue #3022, opened 2023-07-05
[Bug] Ruby parsing issue


rahul-privado:
**Describe the bug**
Ruby parser fails for the below code
Please keep the conversation English! Make it easy for us and share your input code or a minified example. 

**To Reproduce**
Load [this](https://github.com/redis/redis-rb/blob/master/lib/redis/distributed.rb) code in the Ruby parser and see it failing.
Once this has been fixed please import the entire repository to verify that it does not produce any more parser issues

**Expected behavior**
Parsing should pass

**Screenshots**
NA
**Desktop (please complete the following information):**
 NA

**Additional context**

ankit-privado:
@xavierpinho still facing parser issue in [this](https://github.com/redis/redis-rb/blob/master/lib/redis/distributed.rb) file (using intellij ANTLR plugin).

xavierpinho:
For future reference, bear in mind that the ANTLR plugin is not to be 100% trusted, as it doesn't take into account custom predicates/semantic actions.

Nevertheless, just double-checked and you're right: the following is not properly handled:

```ruby
"#{@command.to_s.upcase} cannot be used in Redis::Distributed because the keys involved need " \
          "to be on the same server or because we cannot guarantee that the operation will be atomic."
``` 

Missing is the support for recognising strings separated by `\`.

Let's continue work in #3081

PR #3030, opened 2023-07-05
rubysrc2cpg: Fix missing double splatting argument/parameter

xavierpinho:
Found in #3022


johannescoetzee (changes requested):
A unit test to show the change would be nice (unless a test was already failing and is now fixed)

xavierpinho:
> A unit test to show the change would be nice (unless a test was already failing and is now fixed)

Of course, my bad.

johannescoetzee (approved):

PR #3031, opened 2023-07-05
Throw an exception when source paths don't exist or no source files are found

johannescoetzee:
__Motivation:__ We've run into issues in the past where invalid source directories were given, but `*2cpg` failed silently. These silent failures are difficult to debug (especially with customer submissions) and are most likely due to user error. This PR adds 2 failure modes when finding source files:
* __The input path doesn't exist or isn't readable:__ this is most likely due to a typo in the input path. This should definitely be an exception that terminates execution since this does not match user intent.
* ~__No source files could be found at the given input paths:__ This could be an acceptable error (for example when finding config files), but in general will not be. The current implementation always throws an exception (which can then be handled by the caller if an empty source list is acceptable), but adding an option to allow an empty sources list (with the default still being to not allow them) is potentially a better option.~

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
@@ -8,6 +8,10 @@ import java.nio.file.Paths
 
 object SourceFiles {
 
+  case object InvalidInputPathsException extends Exception("Invalid source input paths")
+
+  case object NoSourceFilesFoundException extends Exception("No source files found at input paths")
+
   private val logger = LoggerFactory.getLogger(getClass)
 
   private def isIgnoredByFileList(filePath: String, config: X2CpgConfig[_]): Boolean = {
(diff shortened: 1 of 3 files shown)


johannescoetzee:
Another option is to have the `determine` methods return `Try[List[String]]`. This is the better functional approach and will force all callers to handle the success/failure explicitly. This does come at the cost of being a more obtrusive change, however.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
```diff
@@ -70,17 +74,60 @@ object SourceFiles {
...
+  private def assertAllExist(files: Set[File]): Unit = {
+    val (existant, nonExistant) = files.partition(_.isReadable)
+    val nonReadable             = existant.filterNot(_.isReadable)
+
+    if (nonExistant.nonEmpty || nonReadable.nonEmpty) {
+      logErrorWithPaths("Source input paths do not exist", nonExistant.map(_.canonicalPath))
+
+      logErrorWithPaths("Source input paths exist, but are not readable", nonReadable.map(_.canonicalPath))
+
+      throw InvalidInputPathsException
```
ml86, line 105: Please reuse existing Java exceptions if possible. `NoSuchFileException` should be fine here.

ml86 (approved):
I like it, but this has the potential to break stuff if people used this for possibly non existing input path in the past. I think we should fix those cases but please wait for @fabsx00 approval to merge.

fabsx00 (approved):

PR #3033, opened 2023-07-05
[rubysrc2cpg] JRuby implementation on parsing External Dependency File

ankit-privado:
Previously we were using `AstCreation` pass to construct lookup map from external dependency files. Now using `JRuby`  parser to do the same with high accuracy (but has limited use cases).

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/ExternalDependenciesResolver.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/ExternalDependenciesResolver.scala
@@ -7,7 +7,6 @@ import io.joern.x2cpg.utils.ExternalCommand
 import io.shiftleft.codepropertygraph.Cpg
 import org.slf4j.LoggerFactory
 
-import scala.io.Source
 import sys.process.*
 import scala.util.{Failure, Success}
 
@@ -37,7 +36,7 @@ object ExternalDependenciesResolver {
   }
 
   private def fetchDependencyList(inputPath: String): List[(String, String)] = {
-    val gemFileContent = Source.fromFile(s"$inputPath${java.io.File.separator}Gemfile").mkString
+    val gemFileContent = File(s"$inputPath${java.io.File.separator}Gemfile").contentAsString
     val gemRegex       = """gem ['"]([^'"]+)['"](?:,\s*['"]([^'"]+)['"])?""".r
 
     gemRegex


pandurangpatil:
> Why not use the ANTLR parser? Isn't it capable of parsing those? It would be one less dependency.

While testing the opensource repositories, the resulting of method full names were almost zero while using ANTLR Parser. We were facing multiple errors. We tried Jruby on trial basis, we could be able to do the lookup more accurately and able to resolve method full names for at least 25% of call nodes for the three repositories we tried.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/AstPackagePass.scala
```diff
@@ -1,28 +1,56 @@
...
+      case ex: Exception =>
+        println(s"Error while parsing $moduleName module File ($filePath): ${ex.getMessage}")
+    }
+  }
+
+  private def fetchMethodInfoFromNode(node: Node, currentNameSpace: ListBuffer[String], moduleName: String): Unit = {
+    if (node != null) {
+      node.getNodeType match {
+        case NodeType.CLASSNODE | NodeType.MODULENODE =>
+          val classOrModuleName = node.childNodes().get(0).asInstanceOf[Colon2Node].getName.toString
```
khemrajrathore, line 34: get(0) doesn't seem safe, handle this code

xavierpinho:
> While testing the open-source repositories, the resulting method full names were almost zero while using ANTLR Parser. We were facing multiple errors. We tried Jruby on trial basis, we could be able to do the lookup more accurately and able to resolve method full names for at least 25% of call nodes for the three repositories we tried.

Were you able to figure out why? The ANTLR parser isn't finished, so please kindly report any issues you encounter. Also, the AstCreator is still very much in its infancy, with lots of features to be properly handled still.

A few weeks ago I also tried integrating JRuby, but didn't find the AST that great for our purposes, as it's more geared towards compilation/interpretation. So, e.g. we wouldn't be able to fill in `code` properties.

I'd propose we first try to understand why the current ANTLR parser isn't enough and then decide if we should bring in alternatives. 

cc: @fabsx00

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/PackageTable.scala
```diff
@@ -1,26 +1,49 @@
...
+    moduleMethodSet.add(packageMethod)
   }
 
-  def getMethodFullNameUsingName(packageUsed: List[String], methodName: String): List[String] =
-    packageUsed
-      .filter(methodTableMap.contains)
-      .flatMap(module =>
-        methodTableMap(module)
+  def getMethodFullNameUsingName(packageUsed: List[String], methodName: String): Option[String] = {
+    if (isExcluded(methodName)) {
```
khemrajrathore, line 27: The result of the if block is not used anywhere

ankit-privado: done.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/utils/PackageTable.scala
```diff
@@ -1,26 +1,49 @@
...
           .filter(_.methodName == methodName)
-          .map(method => s"$module::program:${method.parentClassPath}$methodName")
-      )
+          .foreach(method => {
+            finalMethodName.addOne(s"$module::program:${method.parentClassPath}:$methodName")
+          })
+      }
+    })
+
+    if (finalMethodName.size == 1) {
```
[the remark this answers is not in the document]
ankit-privado: done.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/AstPackagePass.scala
ankit-privado: done.

ankit-privado:
> > While testing the open-source repositories, the resulting method full names were almost zero while using ANTLR Parser. We were facing multiple errors. We tried Jruby on trial basis, we could be able to do the lookup more accurately and able to resolve method full names for at least 25% of call nodes for the three repositories we tried.
> 
> Were you able to figure out why? The ANTLR parser isn't finished, so please kindly report any issues you encounter. Also, the AstCreator is still very much in its infancy, with lots of features to be properly handled still.
> 
> A few weeks ago I also tried integrating JRuby, but didn't find the AST that great for our purposes, as it's more geared towards compilation/interpretation. So, e.g. we wouldn't be able to fill in `code` properties.
> 
> I'd propose we first try to understand why the current ANTLR parser isn't enough and then decide if we should bring in alternatives.
> 
> cc: @fabsx00

@xavierpinho Agree with you, As our use case is to fetch method info for lookups only for external dependency files. We're also constantly working on raising ANTLR parser issues as they come up. Could it be an option to switch to ANTLR parser later when everything sorted?

DavidBakerEffendi:
I'll also have a chat with Fabs to see what the best step here would be, if separation make sense then we could also look at having a JRuby pass as an optional pass.

xavierpinho:
> I'll also have a chat with Fabs to see what the best step here would be, if separation make sense then we could also look at having a JRuby pass as an optional pass.

That was on my mind as well. Ruby seems to be going through a renaissance period nowadays, with new parsers (e.g. YARP) having come up since we started this frontend. It could be interesting to have alternative parsers to choose from.

joern-cli/frontends/rubysrc2cpg/build.sbt
```diff
@@ -8,7 +8,8 @@ dependsOn(Projects.dataflowengineoss, Projects.x2cpg % "compile->compile;test->t
 libraryDependencies ++= Seq(
   "io.shiftleft"  %% "codepropertygraph" % Versions.cpg,
   "org.scalatest" %% "scalatest"         % Versions.scalatest % Test,
-  "org.antlr"      % "antlr4-runtime"    % Versions.antlr
+  "org.antlr"      % "antlr4-runtime"    % Versions.antlr,
+  "org.jruby"      % "jruby-complete"    % "9.2.20.0"
```
xavierpinho, line 12: You should be able to use `jruby-base` instead. And perhaps it would be a good idea to move "9.2.20.0" into `Versions` as well.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/AstPackagePass.scala
```diff
@@ -1,28 +1,60 @@
...
+      node.childNodes().forEach(childNode => fetchMethodInfoFromNode(childNode, currentNameSpace, moduleName))
+    }
+  }
+
+  private def processRubyDependencyFile(inputPath: String, moduleName: String): Unit = {
+    val currentFile = File(inputPath)
+    if (currentFile.exists) {
+      val rubyFile = new java.io.File(inputPath)
+      val rubyCode = scala.io.Source.fromFile(rubyFile).mkString
+      val rootNode = rubyInstance.parseEval(rubyCode, inputPath, null, 0)
```
xavierpinho, line 55: We can pass the file directly to JRuby.

```suggestion
      val rootNode = JRubyBasedParser.parseFile(inputPath)
```

Assuming we have something like this object:

```scala
package io.joern.rubysrc2cpg.parser.jruby

import org.jruby.ast.Node
import org.jruby.{Ruby, RubyInstanceConfig}

import java.io.FileInputStream

object JRubyBasedParser {
  private final val jruby = Ruby.newInstance(new RubyInstanceConfig(true))

  def parseFile(filename: String): Node =
    jruby.parseFile(new FileInputStream(filename), filename, jruby.getCurrentContext.getCurrentScope)

  def parseString(code: String, filename: String): Node = {
    jruby.parse(code, filename, jruby.getCurrentContext.getCurrentScope, 1, true)
  }
}
```

joern-cli/frontends/rubysrc2cpg/build.sbt
ankit-privado: Done.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/AstPackagePass.scala
ankit-privado: done.

PR #3035, opened 2023-07-05
rubySrc2Cpg : Linking import node with respective call nodes

khemrajrathore:
The PR contains the following
- Support for linking import nodes with call nodes
- Test Cases
- Currently have made this change directly in the AstCreation pass, should this be done in the Import pass separately?


tuxology:
I feel it is fine to do it as part of AST creation - this is relatively isolated.

tuxology (approved):

tuxology (approved):

DavidBakerEffendi (approved):

PR #3047, opened 2023-07-06
rubysrc2cpg: fix more method param issues

tuxology:
Fixes a minor issue and a test. We may need to remove multiple method return tests themselves since they don't seem valid. Will consult @pandurangpatil @DavidBakerEffendi and update this

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
@@ -1022,7 +1022,7 @@ class AstCreator(
           .lineNumber(varSymbol.getLine)
           .typeFullName(Defines.Any)
           .columnNumber(varSymbol.getCharPositionInLine)
-        if (arrayParameter != null) {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1019,6 +1019,9 @@ class AstCreator(
           .code(varSymbol.getText)
           .lineNumber(varSymbol.getLine)
           .columnNumber(varSymbol.getCharPositionInLine)
+        if (arrayParameter != null) {
```
DavidBakerEffendi, line 1022: Could we perhaps wrap all nullable values in `Option` type? Or implement an implicit class that automatically does this

tuxology: Done. But there are so many places we have to fix it then 😢

DavidBakerEffendi: Lol yeah fair enough

PR #3062, opened 2023-07-07
unit test for identifier

pandurangpatil:
1. `identifier` to `method parameter in` traversal
2. `identifier` to `local` traversal
3. data flows tests for flow from identifer to call node inside while, until and if block

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/passes/ast/ControlStructureTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/passes/ast/ControlStructureTests.scala
@@ -1,0 +1,132 @@
+package io.joern.rubysrc2cpg.passes.ast
+
+import io.joern.rubysrc2cpg.testfixtures.RubyCode2CpgFixture
+import io.shiftleft.codepropertygraph.Cpg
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/passes/ast/ControlStructureTests.scala
```diff
@@ -0,0 +1,132 @@
...
+    "data flow test" ignore {
+      val src  = cpg.identifier("counter").lineNumber(3).l
+      val sink = cpg.call("puts").l
+      sink.reachableByFlows(src).size shouldBe 1
+    }
+  }
+
+  "if statement use case 1" should {
+    val cpg = code("""
+        |def method()
```
rahul-privado, line 109: This test is incorrect since there is no invocation of `method()`. The correct test case would be
```
def method()
        counter = 5 
        if counter > 1 
        puts "Counter: #{counter}"
        end 
end

method
```

And it works.

Please make sure the test has valid Ruby code that works on the shell

pandurangpatil: I tried as per your suggestion test are still failing.

PR #3067, opened 2023-07-07
[rubysrc2cpg] Generate TypeDecl Members

### LINKED ISSUE #3007: [rubysrc2cpg] Generate members for fields defined under type declarations
Fields defined under classes do not generate member nodes under the type decl nodes. See the test under `ast/TypeDeclAstCreationPassTest:53` and `ast/TypeDeclAstCreationPassTest:76`. Perhaps easier to make the types for this `ANY` due to mutability of variables.

DavidBakerEffendi:
Using basic detection of `@` prefixes to determine class members. It's sorta hacky but there is a bit of a lack of existing structure for determining a class body outside of a normal body :(

Resolves #3007


xavierpinho (approved):
Looks good! Just one thing to be aware of (for later) is that class fields are prefixed with `@@`

PR #3076, opened 2023-07-10
[rubysrc2cpg] Add AST tests for basic operations

dbMundada:


DavidBakerEffendi (approved):

rahul-privado:
@DavidBakerEffendi @dbMundada 
The modelling in Ruby is not like Python. These test cases assume that the modelling will be like Python. I do not see them as valid test cases.

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/passes/ast/AssignCpgTests.scala
```diff
@@ -0,0 +1,176 @@
...
+import io.shiftleft.semanticcpg.language.types.structure.NamespaceTraversal
+
+class AssignCpgTests extends RubyCode2CpgFixture {
+  "single target assign" should {
+    val cpg = code("""x = 2""".stripMargin)
+
+    // TODO: .code property need to be fixed
+    "test assignment node properties" ignore {
+      val assignCall = cpg.call.methodFullName(Operators.assignment).head
+      assignCall.code shouldBe "x = 2" // "="
```
rahul-privado, line 15: `.code `property is given by the parser. The value in it is as per what the parser gives. The cannot be compared directly to other frontends

DavidBakerEffendi: but this is a low priority

PR #3078, opened 2023-07-11
[jimple2cpg] Consistent & Recursive JAR/WAR Unpacking

badly-drawn-wizards:
Add consistent recursive jar unpacking.

Previously jimple2cpg would have two distinct
code paths when importing a directory containing a
jar (which would unpack the jar) compared to importing
a jar directly (which would load it with soot).

This led to strange names starting with `META-INF` for method nodes when importing a jar directly.

This change does the following
- Always use the unpacking behaviour for jars
- Remove jimple from the source files being processed
  (It can be added back if needed downstream)
- Keep the .apk/.dex behaviour if imported directly (but not if contained
in a directory). This maintains the existing behaviour.
- Checks for zip-slips along the way


.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
DavidBakerEffendi, line 1: Probably should be removed from this/added to `.gitignore`

DavidBakerEffendi: Similarly, the `flake.[nix|lock]` should also be removed and probably added to the `.gitignore`

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/Jimple2Cpg.scala
```diff
@@ -75,101 +52,77 @@ class Jimple2Cpg extends X2CpgFrontend[Config] {
...
+  /** Load all class files from archives or directories recursively
+    * @return
+    *   The list of extracted class files and the components of their fully qualified class names
+    */
+  private def loadClassFiles(src: File, tmpDir: File): List[ClassFile] = {
+    val archiveFileExtensions = Set(".jar", ".war", ".zip")
+    extractClassesInPackageLayout(
+      src,
+      tmpDir,
+      isClass = e => e.extension.contains(".class"),
```
DavidBakerEffendi, line 64: You can use `extension.exists(_ == ".class")` which should take advantage of the equality check's performance. Strings are check by value in Scala.

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
```diff
@@ -1,158 +1,136 @@
...
-        val rootVisitor = new ClassPathVisitor()
-        cr.accept(rootVisitor, SKIP_CODE)
+    def shouldExtract(e: Entry) = e.maybeRegularFile() && (isArchive(e) || isSource(e))
+    unfoldArchives(
+      src,
+      {
+        case f if isSource(Entry(f)) =>
+          Left(ClassFile(f))
+        case f if f.isDirectory() =>
+          val files = f.listRecursively.filter(!_.isDirectory).toList
```
DavidBakerEffendi, line 53:
```suggestion
          val files = f.listRecursively.filterNot(_.isDirectory).toList
```

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
```diff
@@ -1,158 +1,136 @@
...
+          Right(files)
+        case f if isArchive(Entry(f)) =>
+          val xTmp = File.newTemporaryDirectory("extract-archive-", parent = Some(tmpDir))
+          Right(Try(f.unzipTo(xTmp, e => shouldExtract(Entry(f)))) match {
```
DavidBakerEffendi, line 57: Maybe for readability, break this into a statement, e.g. 
```scala
val unzipDir = Try....
Right(unzipDir)
```

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
```diff
@@ -1,158 +1,136 @@
...
-                  Option.empty
-              }
-            })
-            .toList
-        )
+  object ClassFile {
+    def getPackagePathFromByteCode(fis: FileInputStream): Option[String] = {
+      val cr                   = new ClassReader(fis)
+      var path: Option[String] = None
+      sealed class ClassNameVisitor extends ClassVisitor(Opcodes.ASM9) {
```
DavidBakerEffendi, line 73: I think my design for the initial `ClassNameVisitor` was pretty bad. How about moving `path` inside the class as a field:
```scala
sealed class ClassNameVisitor extends ClassVisitor(Opcodes.ASM9) {
    var path: Option[String] = None
    
    override def visit(...) { ... }
}

val rootVisitor = new ClassNameVisitor()
cr.accept(rootVisitor, SKIP_CODE)
rootVisitor.path
```
Just reduces the capturing of `path` in a separate closure in some way - also cleaner.

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
```diff
@@ -1,158 +1,136 @@
...
-        List(sourceCodeDir)
-      } else {
-        SourceFiles.determine(sourceCodeDir, archiveFileExtensions)
-      }
-    archives.flatMap { x => unzipArchive(new ZipFile(x), sourceCodeDir) }
+    def getPackagePathFromByteCode(file: File): Option[String] =
+      Try(file.fileInputStream.apply(getPackagePathFromByteCode))
+        .recover {
+          case e: Throwable => {
+            logger.warn(s"Error reading class file ${file.canonicalPath}", e)
```
DavidBakerEffendi, line 94:
```suggestion
            logger.error(s"Error reading class file ${file.canonicalPath}", e)
```

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
```diff
@@ -1,158 +1,136 @@
...
+            None
+          }
+        }
+        .getOrElse(None)
   }
+  sealed class ClassFile(val file: File, val packagePath: Option[String]) {
+    def this(file: File) = this(file, ClassFile.getPackagePathFromByteCode(file))
+
+    // Test that the path separator is always unix
+    val components: Option[Array[String]] = packagePath.map(_.split("/"))
```
DavidBakerEffendi, line 104: This is precarious, but let's see what it is, I wonder if the ASM docs have anything to say here, but the fastest will probably be to see what the Windows tests say.

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
```diff
@@ -1,158 +1,136 @@
...
+            logger.warn(s"Error reading class file ${file.canonicalPath}", e)
+            None
+          }
+        }
+        .getOrElse(None)
   }
+  sealed class ClassFile(val file: File, val packagePath: Option[String]) {
+    def this(file: File) = this(file, ClassFile.getPackagePathFromByteCode(file))
+
+    // Test that the path separator is always unix
```
DavidBakerEffendi, line 103:
```suggestion
    // TODO: Test that the path separator is always unix
```
I would usually suggest adding a TODO here

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
```diff
@@ -1,158 +1,136 @@
...
 
-  /** Removes all files in the temporary unpacking directory.
-    */
-  def clean(): Unit = {
-    FileUtils.deleteDirectory(getUnpackingDir.toFile)
+    val fqcn: Option[String] = components.map(_.mkString("."))
+    def moveToPackageLayoutIn(destDir: File): Option[ClassFile] =
+      packagePath
+        .map { path =>
+          val destClass = File(destDir, path + ".class")
```
DavidBakerEffendi, line 110: I would use String interpolation where possible, and better files has a nice method for children files.
```suggestion
          val destClass = destDir / s"$path.class"
```

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
```diff
@@ -1,158 +1,136 @@
...
+    def this(file: File) = this(file, ClassFile.getPackagePathFromByteCode(file))
+
+    // Test that the path separator is always unix
+    val components: Option[Array[String]] = packagePath.map(_.split("/"))
 
-  /** Removes all files in the temporary unpacking directory.
-    */
-  def clean(): Unit = {
-    FileUtils.deleteDirectory(getUnpackingDir.toFile)
+    val fqcn: Option[String] = components.map(_.mkString("."))
```
DavidBakerEffendi, line 106: I've changed my mind about this `fqcn` thing, I think it'll be best as the full wording
```suggestion
    val fullyQualifiedClassName: Option[String] = components.map(_.mkString("."))
```

flake.lock
```diff
@@ -0,0 +1,61 @@
+{
```
fabsx00, line 1: This should probably not be checked in (?)

flake.nix
```diff
@@ -0,0 +1,21 @@
+{
```
fabsx00, line 1: This should probably not be checked in (?)

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
badly-drawn-wizards: I hate that this function exists, but will use it if it is available.

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
badly-drawn-wizards: Windows tests have passed.

DavidBakerEffendi: Then we can remove this comment 👍

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
```diff
@@ -1,158 +1,137 @@
...
+      src,
+      {
+        case f if isSource(Entry(f)) =>
+          Left(ClassFile(f))
+        case f if f.isDirectory() =>
+          val files = f.listRecursively.filterNot(_.isDirectory).toList
+          Right(files)
+        case f if isArchive(Entry(f)) =>
+          val xTmp = File.newTemporaryDirectory("extract-archive-", parent = Some(tmpDir))
+          val unzipDirs = Try(f.unzipTo(xTmp, e => shouldExtract(Entry(f)))) match {
```
DavidBakerEffendi, line 57: One thing @fabsx00 pointed out to me is that we should be aware of zip-slip vulnerabilities, e.g. if a zip entry contains a `../` path to get you to write outside the destination output. So perhaps find a way to normalize or filter out these kinds of paths that contain `..` in them

joern-cli/frontends/jimple2cpg/src/test/scala/io/joern/jimple2cpg/unpacking/JarUnpacking.scala
```diff
@@ -1,57 +1,70 @@
...
-    val List(bar) = cpg.typeDecl.fullNameExact("pac.Bar").l
-    bar.name shouldBe "Bar"
-
-    cpg.method.filterNot(_.isExternal).fullName.toSetMutable shouldBe Set(
-      "Foo.<init>:void()",
-      "Foo.add:int(int,int)",
-      "pac.Bar.sub:int(int,int)",
-      "pac.Bar.<init>:void()"
-    )
+    for ((name, cpg) <- validCpgs) {
```
badly-drawn-wizards, line 49: Wasn't sure how the include `name` in ScalaTest.

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/util/ProgramHandlingUtil.scala
```diff
@@ -1,158 +1,137 @@
...
-    * @param files
-    *   the class files to move.
-    * @return
-    *   the list of class files at their new locations.
-    */
-  def moveClassFiles(files: List[String]): List[String] = {
-    var destPath: Option[String] = None
+    def this(file: File) = this(Left(file))
+    def this(entry: ZipEntry) = this(Right(entry))
+    private def file: File            = entry.fold(identity, e => File(e.getName))
```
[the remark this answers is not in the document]
badly-drawn-wizards: Handled with `isZipSlip` with an added test case (which fails without it).

joern-cli/frontends/jimple2cpg/src/test/scala/io/joern/jimple2cpg/unpacking/JarUnpacking.scala
DavidBakerEffendi: That is fine, this kind of info isn't really embedded in the CPG. In source code frontends, you'd get the project root under `cpg.metaData.root`

DavidBakerEffendi (approved):

joern-cli/frontends/jimple2cpg/src/test/scala/io/joern/jimple2cpg/unpacking/JarUnpacking.scala
badly-drawn-wizards: I meant more about having test failure attribution. I don't know if there is such a way to define sub tests or something like that

fabsx00 (approved):

PR #3113, opened 2023-07-14
rubysrc2cpg: Implicit return handling in blocks

rahul-privado:
1. Got implicit returns working when the last method statement is not just an expression but contains blocks like if-else
2. Could not do this for the last statement of begin blocks since there is an existing issue with the data flow engine. The below Java code does not return flows due to that issue. Raised it with @fabsx00 separately.
3. Enabled and corrected some other test cases that are working.

Java code that does not produce data flows.
```
public class Main {
  public static int foo(String args) {
    try {
      System.out.println("in begin");
      return 0 ;
    } catch (Exception e) {
      System.out.println("Something went wrong." + args);
    }
  }
  public static void main(String[ ] args) {
       String x = "1";
       foo(x);
  }
}
```

If we set `source -> x` and `sink -> println()` we do not get data flows in the below code. However, if we remove `return 0`, we do get data flows.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
@@ -71,6 +71,11 @@ class AstCreator(
   protected val relativeFilename: String =
     projectRoot.map(filename.stripPrefix).map(_.stripPrefix(JFile.separator)).getOrElse(filename)
 
+  protected var processingLastMethodStatement = false
+  protected var blockIdCounter                = 1
+  protected var currentBlockId                = 0
+  protected val blockChildHash                = mutable.HashMap[Int, Int]()
+
   protected def createIdentifierWithScope(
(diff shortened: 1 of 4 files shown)


xavierpinho:
Inserting `return` statements with just the AST to work with can be tricky. I was thinking if we could this in the CFG directly. Since I'm not that familiar with the codebase in that regard yet, maybe @DavidBakerEffendi or @max-leuthaeuser can share their thoughts on this?

Otherwise the PR looks good - but I would recommend adding a few explanatory comments for those extra parameters and how they play in.

rahul-privado:
> Inserting `return` statements with just the AST to work with can be tricky. I was thinking if we could this in the CFG pass directly. Since I'm not that familiar with the codebase in that regard yet, maybe @DavidBakerEffendi or @max-leuthaeuser can share their thoughts on this?
> 
> Otherwise the PR looks good - but I would recommend adding a few explanatory comments for those extra parameters and how they play in.

_Rahul_: Yes this is tricky indeed but we pretty much have it figured out in this PR. This PR will unblock us for other changes and we could always have a separate PR that perhaps does it in the CFG. We could eliminate this change then.

xavierpinho (approved):

PR #3115, opened 2023-07-14
[joern-slice] Parallelism Config & Tests Refactor

DavidBakerEffendi:
* Added a parallelism config for the slices and set the test suite to single threaded execution
* Data-flow and usage slices both now use `Executors.newWorkStealingPool` when `parallelism > 1`
* Moved the slice tests to the specific language the slicing is being tested on

cc @prabhu


prabhu:
Could you test with https://github.com/apolloconfig/apollo? It ran for 15 hours for me, and then I decided to kill it.

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/slicing/DataFlowSlicing.scala
```diff
@@ -13,30 +15,42 @@ object DataFlowSlicing {
...
-        val sliceNodes      = sinks.iterator.repeat(_.ddgIn)(_.maxDepth(config.sliceDepth).emit).dedup.l
-        val sliceNodesIdSet = sliceNodes.id.toSet
-        // Lazily set up the rest if the filters are satisfied
-        lazy val sliceEdges = sliceNodes
-          .flatMap(_.outE)
-          .filter(x => sliceNodesIdSet.contains(x.inNode().id()))
-          .map { e => SliceEdge(e.outNode().id(), e.inNode().id(), e.label()) }
-          .toSet
-        lazy val slice = Option(DataFlowSlice(sliceNodes.map(cfgNodeToSliceNode).toSet, sliceEdges))
+    }).iterator.method.withMethodNameFilter.withMethodParameterFilter.withMethodAnnotationFilter.call
```
prabhu, line 26: This filter is still slower than `where(c => c.callee.isExternal)`. In addition, we need to filter trivial stdlib calls such as `java.lang`

```
c.typeFullName.startsWith("java."))
```

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/slicing/DataFlowSlicing.scala
```diff
@@ -13,30 +15,42 @@ object DataFlowSlicing {
...
 
-        // Filtering
-        sliceNodes match {
-          case Nil                                                                     => None
-          case _ if config.mustEndAtExternalMethod && !sinksEndAtExternalMethod(sinks) => None
-          case _                                                                       => slice
-        }
+    override def call(): Option[DataFlowSlice] = {
+      val sinks           = config.sinkPatternFilter.map(filter => c.argument.code(filter).l).getOrElse(c.argument.l)
+      val sliceNodes      = sinks.iterator.repeat(_.ddgIn)(_.maxDepth(config.sliceDepth).emit).dedup.l
```
prabhu, line 36: Can we remove `.iterator` from here?

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/slicing/DataFlowSlicing.scala
```diff
@@ -13,30 +15,42 @@ object DataFlowSlicing {
...
+        .flatMap(_.outE)
+        .filter(x => sliceNodesIdSet.contains(x.inNode().id()))
+        .map { e => SliceEdge(e.outNode().id(), e.inNode().id(), e.label()) }
+        .toSet
+      lazy val slice = Option(DataFlowSlice(sliceNodes.map(cfgNodeToSliceNode).toSet, sliceEdges))
+
+      // Filtering
+      sliceNodes match {
+        case Nil                                                                     => None
+        case _ if config.mustEndAtExternalMethod && !sinksEndAtExternalMethod(sinks) => None
```
prabhu, line 49: I'm finding this block to be slow and could be removed if we do `.callee.isExternal` at the top level.

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/slicing/UsageSlicing.scala
```diff
@@ -370,7 +370,7 @@ object UsageSlicing {
   /** Adds extensions to extract all declaration nodes from a method body.
     */
   implicit class MethodDataSourceExt(trav: Iterator[Method]) {
-    def declaration: Iterator[Declaration] = trav.local ++ trav.parameter
+    def declaration: Iterator[Declaration] = trav.ast.collectAll[Declaration]
```
prabhu, line 373: This style of traversal using `declaration` is missing CALL nodes in python where some methods get imported and called directly without variable assignment.

Eg:

```python
from django.http import JsonResponse

def bar():
    return JsonReponse("foo")
```

DavidBakerEffendi: That is intentional, the usage slice is to describe usages of identifiers

PR #3133, opened 2023-07-18
rubysrc2cpg: anonymous class support

tuxology:
- [ ] Add TypeDecl node for anonymous class (`Class.new`)
- [ ] Arguments in `new` method should be used as base class (`MyException = Class.new(Exception)`)


DavidBakerEffendi:
Closing as this is handled in the new frontend

Issue #3040, opened 2023-07-06
[Bug] Ruby Parsing issue


ankit-privado:
Describe the bug

Ruby parser fails for the below code
Please keep the conversation English! Make it easy for us and share your input code or a minified example.

To Reproduce
Load [this](https://github.com/odinsride/olubalance/blob/6fe35f3a84ab25dc0b6822f96a3a660174cace00/app/controllers/transactions_controller.rb#L8) code in the Ruby parser and see it failing.

Expected behavior
Parsing should pass

Screenshots
NA

Desktop (please complete the following information):
NA

Additional context
NA

ankit-privado:
@xavierpinho you can also refer this sample code: https://github.com/zinagithub/blogger/blob/23287b51baa0ea3c1758334a6d5fb83611e057cf/config/spring.rb#L2

ankit-privado:
@xavierpinho Would it be appropriate to close this issue and create separate tickets for all Ruby % shortcuts? This will allow anyone to work on each task individually.

xavierpinho:
> @xavierpinho Would it be appropriate to close this issue and create separate tickets for all Ruby % shortcuts? This will allow anyone to work on each task individually.

I prefer to give this one preference and close the others by mentioning they are duplicates of this one. What matters are code samples, and they are not lost that way.

ankit-privado:
yes sure, not yet separated. thanks

pandurangpatil:
@xavierpinho what items are pending on this ticket?

xavierpinho:
@pandurangpatil I believe all except `%s` are supported in the `ruby-dev` branch at the moment. `%s` is (mutatis mutandis) the same as `%w`, I just didn't notice it was missing until yesterday.

PR #3146, opened 2023-07-19
rubysrc2cpg: Crash fix and zero keyword

rahul-privado:
1. Defensive check to avoid crash due to https://github.com/joernio/joern/issues/3040
2. Support for `zero` keyword
3. Test case correction and addition
4. 
@xavierpinho please review


xavierpinho:
Sorry, but I have to reject:

* `zero` is not a keyword.
* The samples you have in your unit-tests already parse?! Also, no explanation given for  the change at https://github.com/joernio/joern/pull/3146/files#diff-fd09805d84cab774eba6536d7823b98ed2cc999829f23351f1431889f569b5c9R2013

PR #3152, opened 2023-07-19
rubysrc2cpg: remove base classname from name

tuxology:
We don't need baseclass name in `name`. We have that in `fullName` already.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
@@ -31,7 +31,7 @@ trait AstForTypesCreator { this: AstCreator =>
       None
     }
 
-    val className = ctx.className(baseClassName)
+    var className = ctx.className(baseClassName)
     if (className != Defines.Any) {
       classStack.push(className)
       val fullName = classStack.reverse.mkString(pathSep)
@@ -40,6 +40,11 @@ trait AstForTypesCreator { this: AstCreator =>
 
       if (classStack.nonEmpty) {
         classStack.pop()
+      }
+
+      /* If we have basename, we will have X.Y as classname. We only need Y for "name" though */
+      if (baseClassName != null) {
+        className = Option(className.split("\\.").last).getOrElse(className)
       }
 
       val typeDeclNode = NewTypeDecl()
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/passes/ast/TypeDeclAstCreationPassTest.scala
```diff
@@ -31,6 +31,19 @@ class TypeDeclAstCreationPassTest extends RubyCode2CpgFixture {
...
+    "populate class name correctly for a derived class" in {
+      val cpg = code("""
+          |module ApplicationCable
+          |  class Channel < ActionCable::Channel::Base
+          |  end
+          |end
+          |""".stripMargin)
+      val List(myClass) = cpg.typeDecl.nameExact("Channel").l
+      myClass.name shouldBe "Channel"
+      myClass.fullName shouldBe "Test0.rb::program.ApplicationCable.ActionCable.Channel"
```
pandurangpatil, line 44: @tuxology I think full name also should be `Test0.rb::program.ApplicationCable.Channel` instead of `Test0.rb::program.ApplicationCable.ActionCable.Channel`

tuxology: @pandurangpatil @fabsx00 I removed the base class from the names, but I don't know how how this impacts type hierarchy

pandurangpatil: @tuxology basename doesn't become part of the fullname of the class at least in the contest of Java.

khemrajrathore: As per my understanding - For a given class its `name` and `fullName` should only correspond to self. If it extends another class, the base class fullName should be present in `inheritsFromFullName`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -154,9 +159,8 @@ trait AstForTypesCreator { this: AstCreator =>
         getClassNameScopedConstantReferenceContext(ctx.scopedConstantReference())
       else
         Option(ctx.CONSTANT_IDENTIFIER()).map(_.getText) match {
-          case Some(className) if baseClassName.isDefined => s"${baseClassName.get}.$className"
```
khemrajrathore, line 157: @tuxology Can we not just remove this line, instead of splitting on `.`?

tuxology: We removed this line already. I can remove the splitting code as well - it was left there just in case.

khemrajrathore: I believe If we just remove this line, this should fall in place. We should not need the splitting logic

Issue #2099, opened 2022-12-13
php - incorrect data dependency


d1tto:
When I do inter-procedural taint analysis on PHP code, I find that there are data dependencies that seemed incorrect. The PHP code is shown below.
```php
<?php
function getString() {
    return "111";
}
function runShell($cmd) {
    system($cmd);
}
$str = getString();
runShell($str);
?> 
```
The `system` function in the function `runShell` is dependent on the argument `$cmd`. But the PDG generated by joern does not seem to reflect this dependency.
```
digraph "runShell" {  
"17" [label = <(METHOD,runShell)<SUB>7</SUB>> ]
"20" [label = <(METHOD_RETURN,ANY)<SUB>7</SUB>> ]
"4" [label = <(PARAM,$cmd)<SUB>7</SUB>> ]
"19" [label = <(system,system($cmd))<SUB>8</SUB>> ]
  "19" -> "20"  [ label = "DDG: $cmd"] 
  "19" -> "20"  [ label = "DDG: system($cmd)"] 
  "17" -> "4"  [ label = "DDG: "] 
  "17" -> "19"  [ label = "DDG: "] 
}
```
This incorrect data dependency led to inter-procedural taint analysis fail.
When I use joern to parse similar C code, joern can produce the correct data dependencies
```c
#include <stdio.h>

char *getString() {
    return "111";
}

void runShell(char *cmd) {
    system(cmd);
}

int main() {
    char *str = getString();
    runShell(str);
}
```
```
digraph "runShell" {  
"13" [label = <(METHOD,runShell)<SUB>7</SUB>> ]
"18" [label = <(METHOD_RETURN,void)<SUB>7</SUB>> ]
"14" [label = <(PARAM,char *cmd)<SUB>7</SUB>> ]
"16" [label = <(system,system(cmd))<SUB>8</SUB>> ]
  "16" -> "18"  [ label = "DDG: cmd"] 
  "16" -> "18"  [ label = "DDG: system(cmd)"] 
  "13" -> "14"  [ label = "DDG: "] 
  "14" -> "16"  [ label = "DDG: cmd"] 
  "13" -> "16"  [ label = "DDG: "] 
}
```

d1tto:
The problem seems to be related to the `$` in front of the PHP variable name.
[...]

PR #3183, opened 2023-07-21
Try to fix [issue2099](https://github.com/joernio/joern/issues/2099)

d1tto:
In PHP frontend, `param.name` has no `$`, while `param.code` has `$`, which leads to false data dependency.


d1tto:
> Nice find! There has been very little testing of the integration between the PHP frontend that is currently under development and the open-source data flow engine, so, I'm not surprised larger issues like this still exist. Thanks for digging into the code to fix them!

I think this is a very urgent bug that needs to be fixed. My solution may have some issues. I hope you can fix it

johannescoetzee:
@d1tto Thanks for your PR! I've opened an alternative at https://github.com/joernio/joern/pull/3399 that doesn't use the parameter `code` field for the comparison since this will most likely contain default values in addition to the parameter name in the future.

johannescoetzee:
Closing in favour of https://github.com/joernio/joern/pull/3399. Please let me know if I missed a case with the name conversions that can still trigger this issue.

Issue #3164, opened 2023-07-20
[Bug] Crash/regression due to parser change


rahul-privado:
**Describe the bug**
[This](https://github.com/joernio/joern/pull/3131) commit is causing a crash while loading [this](https://github.com/mfssaura/HospitalManagementProject/blob/master/app/models/user.rb) file.

**To Reproduce**
The precise code fragment that causes the crash is 
```
class << self
end

```
and the top few lines of the crash stack are
```
java.lang.NullPointerException: Cannot invoke "io.joern.rubysrc2cpg.parser.RubyParser$ClassOrModuleReferenceContext.scopedConstantReference()" because the return value of "io.joern.rubysrc2cpg.astcreation.AstForTypesCreator$ClassOrModuleReferenceContextExt.ctx()" is null
	at io.joern.rubysrc2cpg.astcreation.AstForTypesCreator$ClassOrModuleReferenceContextExt.hasScopedConstantReference(AstForTypesCreator.scala:150)
	at io.joern.rubysrc2cpg.astcreation.AstForTypesCreator$ClassOrModuleReferenceContextExt.classOrModuleName(AstForTypesCreator.scala:153)
	at io.joern.rubysrc2cpg.astcreation.AstForTypesCreator$ClassDefinitionPrimaryContextExt.className$$anonfun$2(AstForTypesCreator.scala:144)
	at scala.Option.map(Option.scala:242)
	at io.joern.rubysrc2cpg.astcreation.AstForTypesCreator$ClassDefinitionPrimaryContextExt.className(AstForTypesCreator.scala:144)
	at io.joern.rubysrc2cpg.astcreation.AstForTypesCreator.astForClassDeclaration(AstForTypesCreator.scala:34)
	at io.joern.rubysrc2cpg.astcreation.AstForTypesCreator.astForClassDeclaration$(AstForTypesCreator.scala:17)
	at io.joern.rubysrc2cpg.astcreation.AstCreator.astForClassDeclaration(AstCreator.scala:23)
	at io.joern.rubysrc2cpg.astcreation.AstCreator.astForPrimaryContext(AstCreator.scala:328)
	at io.joern.rubysrc2cpg.astcreation.AstCreator.astForExpressionContext(AstCreator.scala:370)
	at io.joern.rubysrc2cpg.astcreation.AstForStatementsCreator.astForExpressionOrCommand(AstForStatementsCreator.scala:212)
[...]

PR #3188, opened 2023-07-21
rubysrc2cpg: Workaround for << operator in class definition

rahul-privado:
This fixes https://github.com/joernio/joern/issues/3164

Tagging @xavierpinho

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
@@ -265,7 +265,7 @@ class MiscTests extends RubyCode2CpgFixture {
         |end
         |""".stripMargin)
 
-    "recognise all identifier and call nodes" in {
+    "recognise all typedecl nodes" in {
       cpg.typeDecl.name("self").size shouldBe 1
     }
   }
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
@@ -259,17 +259,6 @@ class MiscTests extends RubyCode2CpgFixture {
     }
   }
 
-  "CPG for code with singleton object of self" should {
-    val cpg = code("""
-        |class << self
-        |end
-        |""".stripMargin)
-
-    "recognise all typedecl nodes" in {
-      cpg.typeDecl.name("self").size shouldBe 1
-    }
-  }
-
   "CPG for code with singleton object of some class" ignore {
     val cpg = code("""
         |class << "some_class"
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
@@ -258,4 +258,15 @@ class MiscTests extends RubyCode2CpgFixture {
         .size shouldBe 1
     }
   }
+
+  "CPG for code with singleton class" should {
+    val cpg = code("""
+        |class << self
+        |end
+        |""".stripMargin)
+
+    "recognise all identifier and call nodes" in {
+      cpg.typeDecl.name("self").size shouldBe 1
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
```diff
@@ -258,4 +258,15 @@ class MiscTests extends RubyCode2CpgFixture {
         .size shouldBe 1
     }
   }
+
+  "CPG for code with singleton class" should {
```
xavierpinho, line 262: I'm afraid the support for `class << self` is not that straightforward, as it's typically used to build class methods. Please take a moment to explore alternatives to this representation and we can discuss their pros.

rahul-privado: Can you help me with 1-2 examples of where this would not work?

xavierpinho: Off the top of my head, it seems that an empty `class << self .. end` can just be disregarded, since its purpose is to add members to the outer class and there's nothing to add. When it's non-empty, I'm not sure, but probably its members should be moved to the outer class. But as as I said, this is me thinking out loud, so please do some investigation to understand this feature and to decide which representation is best.

pandurangpatil: @xavierpinho can this be merged? at least this can fix few error handling. We will uncover more scenarios driven by testing, we can prioritise fixing more common issues first. Merging this will enable us to test those repos again for other issues.

cc: @tuxology

xavierpinho: @pandurangpatil I'd propose to merge just the null handling, then. (Going back to the AstCreator clean-up discussion.)

pandurangpatil: @xavierpinho Rewriting `className` to return an `Option[String]` is one part of it. @rahul-privado will get the implementation updated.

Could you provide specific other improvements/handling that you want to change? Give specific suggestion, so that we can incorporate those suggestions and merge this PR with updated comments.

rahul-privado: @xavierpinho 
I have put in code to handle this better but not completely. This will avoid the crash though it is insufficient to make it work properly. Can we push this in so that it unblocks us?
I have added a disabled test case and a comment to put it in the radar. I will continue pursuing this.

xavierpinho: Thanks!

> Could you provide specific other improvements/handling that you want to change? Give specific suggestion, so that we can incorporate those suggestions and merge this PR with updated comments.

Right now I'd need to dive into this feature to provide concrete suggestions, besides the ones I already mentioned previously here. If @rahul-privado has a translation plan for this, let's hear it! Otherwise, I'd just say we don't handle this case besides the null handling (i.e. it'd be as if the `class << self ... end` portion didn't exist.)

xavierpinho: > @xavierpinho I have put in code to handle this better but not completely. This will avoid the crash though it is insufficient to make it work properly. Can we push this in so that it unblocks us? I have added a disabled test case and a comment to put it in the radar. I will continue pursuing this.

To be honest, I don't particularly like having a `self` class declared (cf. [test](https://github.com/joernio/joern/pull/3188/files#diff-8b764a636f592aa5d5298074cd5c1784e3553ba4ce4e5abff511cd4a1fd47cc5R268)). This isn't a class declaration and worse we can end up with many of those. If we don't have a general solution, I'd propose:

* `getName` returns None for `class << self`
* The `getName` caller interprets this `None` result and returns an empty `Seq[Ast]`

rahul-privado: Changed `className()` as per your suggestion. I do not have a concrete way to solve this problem right now and need more time to figure it out.

xavierpinho (approved):

Issue #3203, opened 2023-07-23
[Bug][rubysrc2cpg] - block parameter without name is throwing error


pandurangpatil:
**Describe the bug**

The following code sample fails to parse

```
    def build(&)
        another_method_call(&)
    end
```

However, the below sample works.

```
    def build(&a)
        another_method_call(a)
    end
```

pandurangpatil:
@karan-batavia as you resolved this one - https://github.com/joernio/joern/issues/3168. So tagging you for this ticket.

cc: @xavierpinho

karan-batavia:
@xavierpinho , this looks straightforward. I'll take this up before #3189.

karan-batavia:
Closed by #3204.

xavierpinho:
> Closed by #3204.

I believe we still need to support ` another_method_call(&)` from the original sample though. I'm not sure how this should be modelled yet.

karan-batavia:
> > Closed by #3204.
> 
> I believe we still need to support ` another_method_call(&)` from the original sample though. I'm not sure how this should be modelled yet. For the parser though, this should be enough:
> 
> ```antlr
> blockArgument
>     :   AMP expression?
>     ;
> ```

Ah, yes. I didn't take this into account. Argument name could be optional too. Will look into this.

PR #3204, opened 2023-07-23
[rubysrc2cpg] [WIP] Added support for proc parameters with no name

karan-batavia:
Support for #3203.


karan-batavia:
@xavierpinho, couldn't figure out how the AST would look like for this change, so marked as draft. Will convert it once I fix that. You can review the parser changes till then.

karan-batavia:
@xavierpinho, please review. 
The proc parameters with no name will be handled using `param_<number>` - for both the code and the name field. Same with other types of parameters.

xavierpinho (approved):
Thanks! We shall still need to address how yields are handled, but that can be future work.

PR #3207, opened 2023-07-23
rubysrc2cpg: handle empty arrays

tuxology:
Handles cases like `x.y = [ ]` where a new empty literal of `NilClass` type is introduced when there is empty array in RHS

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
@@ -508,7 +508,12 @@ class AstCreator(
   }
 
   def astForArrayConstructorPrimaryContext(ctx: ArrayConstructorPrimaryContext): Seq[Ast] = {
-    astForIndexingArgumentsContext(ctx.arrayConstructor().indexingArguments())
+    if (ctx.getText == "[]") {
+      /* we might have empty array, so create empty literal with [] */
+      Seq(Ast(literalNode(ctx, ctx.getText, Defines.NilClass)))
+    } else {
+      astForIndexingArgumentsContext(ctx.arrayConstructor().indexingArguments())
+    }
   }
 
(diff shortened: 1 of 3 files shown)


tuxology:
@xavierpinho Is this a correct approach to handle this case? Technically, in Ruby this is not `nil` but is `empty` and/or `blank`. So should we introduce a new Global Type `EmptyClass` and then create empty literal with that class?

tuxology:
> I'd say we could handle this like we already do for non-empty array literals (`[0, 1, 2]`), just without any elements.

@xavierpinho So currently, we have something like this for `[ 0, 1 ]` etc

image


However, I think since `arrayInitializer` is not a terminal node, we can't place it there with empty array. Let me see how other frontends are doing this and get back to you. Btw, if we see an array with single element like `[0]`, we just assign its value to LHS.

xavierpinho:
> However, I think since `arrayInitializer` is not a terminal node, we can't place it there with empty array. Let me see how other frontends are doing this and get back to you. 

Let me know what you find -- on first sight I don't think it should be a problem, but maybe I'm overlooking something. The terminals would be the tokens:
image

> Btw, if we see an array with single element like `[0]`, we just assign its value to LHS.

If it is a "multiple assignment", that would be right (like in `x, = [0]`,`x` should be `0`). If it's a single assignment (like in `x = [0]`, `x` should be `[0]`), then it needs some tweaking (in a future PR.)

tuxology:
@xavierpinho Contuing what we already have in Ruby so far, I created an `arrayInitializer` node` with no child as follows: 

image

**Note:**  I observed that in Python for same empty array, we have `listLiteral` operator node as follows:

image

For JS, we have `Array.factory()` call node

image

But we have not used these in Ruby and instead we've just used `arrayInitializer`. Not sure what is the right way for us at this point - but at my current approach is uniform. I think this closes multiple issues in a lot of ruby code we were seeing where we had empty array assignments like `x = []` @fabsx00 If you have insight whether this is correct way to model or we should use something like `listLiteral` instead, let me know. I am not sure about the impact of this on dataflow later on.

xavierpinho:
@tuxology That looks very good to me. Thanks for bringing into the discussion an overview of how some other frontends do this!

xavierpinho (approved):

PR #3220, opened 2023-07-24
rubysrc2cpg: Fix for assignment linked method identifier

rahul-privado:
Tagging @xavierpinho

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
@@ -270,6 +270,7 @@ class MiscTests extends RubyCode2CpgFixture {
     }
   }
 
+  // TODO obj.foo="arg" should be interpreted as obj.foo("arg"). code change pending
   "CPG for code with method ending with =" should {
     val cpg = code("""
         |class MyClass
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
@@ -269,4 +269,27 @@ class MiscTests extends RubyCode2CpgFixture {
       cpg.typeDecl.name("some_class").size shouldBe 1
     }
   }
+
+  "CPG for code with method ending with =" should {
+    val cpg = code(
+      """
+        |class MyClass
+        | def foo=(value)
+        | puts value
+        | end
+        |end
+        |
+        |obj = MyClass.new
+        |obj.foo="arg"
+        |""".stripMargin)
+
+    "recognise all call nodes" in {
+      cpg.call.name("puts").size shouldBe 1
+      cpg.call.name("<operator>.fieldAccess").size shouldBe 1
+    }
+
+    "recognise all method nodes" in {
+      cpg.method.name("foo").size shouldBe 1
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
```diff
@@ -269,4 +269,26 @@ class MiscTests extends RubyCode2CpgFixture {
...
+        | end
+        |end
+        |
+        |obj = MyClass.new
+        |obj.foo="arg"
+        |""".stripMargin)
+
+    "recognise all call nodes" in {
+      cpg.call.name("puts").size shouldBe 1
+      cpg.call.name("<operator>.fieldAccess").size shouldBe 1
```
xavierpinho, line 287: I'm ok accepting right now, but the idea behind a method whose name ends in `=` is that `x.foo=arg` becomes `x.foo(arg)`, so the string `"arg"` should taint `puts`.

Please document this so that we know where to look for later on.

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
```diff
@@ -269,4 +269,26 @@ class MiscTests extends RubyCode2CpgFixture {
...
+    }
+
+    "recognise all method nodes" in {
+      cpg.method.name("foo").size shouldBe 1
```
xavierpinho, line 291: Why not name it `foo=`? I think one use-case when using Joern would be to find assignment methods, and for that we could build a query to filter method names ending in `=`.

xavierpinho (changes requested):

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
pandurangpatil: @xavierpinho as far I think, `=` doesn't become part of the method name. When we call the method, we would just make use of `foo` and not `foo=`. I guess name should be `foo` only, having said that in order to handle the use case that you are pointing at, we need to introduce another property which will indicate if the method has been defined with `=` operator.

rahul-privado: @pandurangpatil 
It is a assignment method and so we can choose to represent it so.

@xavierpinho 
I have included = in the name. That avoids the crash and unblocks us for further testing. However, more changes will be needed to have the right call node to make the data flow work and I intend to do that in a separate PR

xavierpinho: @rahul-privado Thanks! Please add some comment or unit-test explaining that `x.foo=arg` should actually mean `x.foo(arg)` in these cases, and we can merge this.

rahul-privado: Added comments

xavierpinho (approved):

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MiscTests.scala
pandurangpatil: @rahul-privado It looks like Ruby considers them as setter methods. New learning for me, 

@xavierpinho sorry for the misunderstanding. I didn't go trough the entire context, I mistook it for the Ruby 3 syntax to define method with `=` operator.

PR #3230, opened 2023-07-25
Fine-Tune Type Prop Queries & Added Iterations CLI Opt

### LINKED ISSUE #3211: [Bug] Potential memory leak caused by PythonTypeRecovery
**Describe the bug**
Memory usage goes up sigificantly for the second type recovery pass. `cloc` shows the code base has ~80k loc.

**To Reproduce**
1. Source code:
[mmdetection.zip](https://github.com/joernio/joern/files/12142985/mmdetection.zip)
2. importCode.python("")
3. iteration num attribute is set to 2
4. Initial run of type recovery uses 2.6gb ram
5. Second run causes gc monitor clearing references
image

**Screenshots**

 - Win
 - Joern 2.0.1

DavidBakerEffendi:
* Optimized queries and data structures here and there, was able to reduce the memory consumption but it only defers a huge allocation later.
* Added CLI opt for JS and Python `--type-prop-iterations` where one can set how many type propagation iterations run. 2 onwards typically involves interprocedural propagation. e.g. `./joern-parse .../../Workspace/test-repos/python/open-mmlab --frontend-args --type-prop-iterations 1` finishes for me but only performs intraprocedural type propagation.

@yzn12345 The type propagation is pretty rough around the edges and likely can be rewritten to be a bit more precise and performant but I hope this helps generate python CPGs with some type recovery!

Somewhat resolves #3211


max-leuthaeuser:
Can we error out (or at least show a warning) if someone enters an invalid value for `typePropagationIterations` (<1 or something very large)? Maybe as `OParser` validation?

DavidBakerEffendi:
@max-leuthaeuser I've made a type propagation config trait and re-usable OParser configs now so there is less duplication and central parsing options. I've opted for `x <= 0` to disable type propagation (with a warning) and warn for a `x > 5`.

max-leuthaeuser (approved):

Issue #3270, opened 2023-07-27
[Bug] RubyParser: Incorrect hierarchy with the methodDefinition production


rahul-privado:
**Describe the bug**
The methodDefinition production produces an incorrect parse tree for a certain permutation of methods.

**To Reproduce**
The below code produces an incorrect parse tree. The parse tree is incorrect since the method` foo2() `comes into the body of method `foo1()`
```
module SomeModule
def foo1
    return unless true
end
def foo2=(arg)
end
end
```

The below code which is the same with just the order of the method swapped generates a correct parse tree. The parse tree is correct because the methods are at the same level.
```
module SomeModule
def foo2=(arg)
end
def foo1
    return unless true
end
end
```

**Expected behavior**
Below is a correct parse tree
![Correct parse tree](https://github.com/joernio/joern/assets/119306799/8c2264e5-adf1-4069-8b57-dc9d2ec88d80)

Below is the incorrect parse tree.
![Incorrect parse tree](https://github.com/joernio/joern/assets/119306799/79d42f21-1d88-4514-af5c-f6bc4f0fb140)

1. The methods should be at the same level
2. This should be irrespective of the order in which the methods have been written in the module.

**Screenshots**

**Desktop (please complete the following information):**
 - OS
 - Joern Version 
 - Java version

**Additional context**

PR #3272, opened 2023-07-27
rubysrc2cpg: Fix for missing END in methodDefinition

rahul-privado:
This fixes https://github.com/joernio/joern/issues/3270


xavierpinho:
Can't add `END`, otherwise endless methods are lost. I'll handle it later today. Thanks for finding this issue.

PR #3329, opened 2023-07-27
[rubysec2cpg] Additional skipFileRegex parameter in importCode 

ankit-privado:
Now we can directly skip the file while using joern-cli. We will incorporate it for other frontends incrementally over the period of time.

Example: 
```
 importCode(inputPath = "path", skipFileRegex = "regex")
```

@fabsx00 @xavierpinho can you pls review this?

### DIFF
--- a/console/src/main/scala/io/joern/console/cpgcreation/CCpgGenerator.scala
+++ b/console/src/main/scala/io/joern/console/cpgcreation/CCpgGenerator.scala
@@ -13,7 +13,7 @@ case class CCpgGenerator(config: FrontendConfig, rootPath: Path) extends CpgGene
 
   /** Generate a CPG for the given input path. Returns the output path, or None, if no CPG was generated.
     */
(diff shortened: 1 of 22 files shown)


max-leuthaeuser:
We already have the `args` parameter for `importCode.apply` and `importcode.<lang>.fromString` (https://github.com/joernio/joern/blob/master/console/src/main/scala/io/joern/console/cpgcreation/ImportCode.scala#L81 and https://github.com/joernio/joern/blob/master/console/src/main/scala/io/joern/console/cpgcreation/ImportCode.scala#L94) which passes args to the respective frontend.
Frontends should use https://github.com/joernio/joern/blob/master/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/X2Cpg.scala#L22C13-L22C13 and https://github.com/joernio/joern/blob/master/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala#L64 respectively.

ankit-privado:
Thank you @max-leuthaeuser

PR #3357, opened 2023-07-28
rubysrc2cpg: Option based null check to avoid crash in indexing arguments

rahul-privado:
1. Fix for crash in indexing arguments handling by putting a Option/null check
2. Test for the same
3. Removed defensive code that was put in to handle this issue
@xavierpinho


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -766,7 +762,9 @@ class AstCreator(
 
   def astForIndexingExpressionPrimaryContext(ctx: IndexingExpressionPrimaryContext): Seq[Ast] = {
     val lhsExpressionAst = astForPrimaryContext(ctx.primary())
-    val rhsExpressionAst = astForIndexingArgumentsContext(ctx.indexingArguments())
+    val rhsExpressionAst = Option(ctx.indexingArguments()) match
```
xavierpinho, line 765:
```suggestion
    val rhsExpressionAst = Option(ctx.indexingArguments).map(astForIndexingArgumentsContext).getOrElse(Seq())
```

xavierpinho (approved):

PR #3358, opened 2023-07-28
[rubysrc2cpg] Enhancing TypeDecl implementation

khemrajrathore:
The PR contains the following
- For encountered TypeDecl create corresponding TypeRef assignments similar to how we do for python, so that types for the class instance usage within the same files is resolved
- For instance methods create a `this` parameter as the parameter with index 0, which help in tracking and linking dataflow for instance variables
- For method parameters refactored the current code to be short and precise along with adding corresponding `index` and `order` values
- Fixed 2 existing ignored test cases
cc - @fabsx00 @DavidBakerEffendi

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
@@ -173,6 +173,8 @@ class AstCreator(
     val undef                   = "<operator>.undef"
     val superKeyword            = "<operator>.super"
     val stringConcatenation     = "<operator>.stringConcatenation"
+    val formattedString         = "<operator>.formatString"
+    val formattedValue          = "<operator>.formatValue"
   }
   private def getOperatorName(token: Token): String = token.getType match {
     case ASSIGNMENT_OPERATOR => Operators.assignment
@@ -308,7 +310,19 @@ class AstCreator(
       .interpolatedStringSequence()
       .asScala
       .flatMap(inter => {
-        astForStatements(inter.compoundStatement().statements(), false, false)
+        Seq(
+          Ast(
+            NewCall()
+              .code(inter.getText)
+              .name(RubyOperators.formattedValue)
+              .methodFullName(RubyOperators.formattedValue)
+              .lineNumber(line(ctx))
+              .columnNumber(column(ctx))
+              .typeFullName(Defines.Any)
+              .dispatchType(DispatchTypes.STATIC_DISPATCH)
+          )
+        ) ++
+          astForStatements(inter.compoundStatement().statements(), false, false)
       })
       .toSeq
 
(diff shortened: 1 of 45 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -36,6 +86,10 @@ trait AstCreatorHelper { this: AstCreator =>
     usedVariableNames.put(variableName, counter)
     currentVariableName
   }
+
+  protected def addReceiverEdge(dstNode: nodes.NewNode, srcNode: nodes.NewNode): Unit = {
```
DavidBakerEffendi, line 90: Something like this exists in the `Ast()` class already

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -1314,9 +1317,25 @@ class AstCreator(
...
+    val astMethodName = astForMethodNamePartContext(ctx.methodNamePart())
+    val callNode      = astMethodName.head.nodes.filter(node => node.isInstanceOf[NewCall]).head.asInstanceOf[NewCall]
+
+    // Create thisParameter if this is an instance method
+    // TODO may need to revisit to make this more robust
+    val astMethodParamSeq = ctx.methodNamePart() match {
+      case _: SimpleMethodNamePartContext if !classStack.top.endsWith(":program") =>
+        val thisParameterNode = createMethodParameterIn(
+          "this",
+          typeFullName = callNode.methodFullName,
```
DavidBakerEffendi, line 1329: Does this call node refer to the `:program` method?

DavidBakerEffendi (changes requested):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
khemrajrathore: Will remove this code as anyways we are not using this function anywhere in ruby

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
khemrajrathore: No, the call node refers to methodDefinition. We use the information from the callNode and discard it

Basically here for any methodDefinition encountered we are categorizing the whether they are class method or instance method

khemrajrathore:
@DavidBakerEffendi Have made the requested changes

DavidBakerEffendi (approved):

PR #3401, opened 2023-08-02
rubysrc2cpg: handle private_class_method

tuxology:
In Ruby we can define a private class method in a way where the argument to the `private_class_method` function can itself be a function definition

```
private_class_method def foo(a) 
  1
end
```

Here, we model the argument of `private_class_method` function itself as an identifier and add the foo method definition to the main `:program` block (is that ideal?)

This is the resulting AST for above case 👇 
image

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
@@ -220,7 +220,8 @@ class AstCreator(
         Seq(Ast()),
         blockAst(
           blockNode,
-          statementAsts.toList ++ builtInMethodAst ++ methodRefAssignmentAsts ++ typeRefAssignmentAst
+          statementAsts.toList ++ builtInMethodAst ++ methodRefAssignmentAsts ++ typeRefAssignmentAst ++ methodDefInArgument
+            .map(ast => ast)
         ),
         methodRetNode
       )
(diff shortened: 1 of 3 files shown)


tuxology:
Also, realized there are many such methods we will have to take care of eventually. Eg. https://ruby-doc.org/3.2.2/Module.html#method-i-private_class_method

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -277,7 +277,7 @@ trait AstForStatementsCreator {
         resolveRequireOrLoadPath(argsAsts, callNode)
       } else if (callNode.name == "require_relative") {
         resolveRelativePath(filename, argsAsts, callNode)
-      } else if (callNode.name == "attr_accessor") {
+      } else if (callNode.name == "attr_accessor" || callNode.name == "private_class_method") {
```
DavidBakerEffendi, line 280: Perhaps we can create a set of these "special" symbols and just use a `.contains` check as there are likely more to come.

DavidBakerEffendi (changes requested):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreator.scala
```diff
@@ -196,7 +196,10 @@ class AstCreator(
       methodAst(
         programMethod,
         Seq(Ast()),
-        blockAst(blockNode, statementAsts.toList ++ builtInMethodAst ++ methodRefAssignmentAsts),
+        blockAst(
+          blockNode,
+          statementAsts.toList ++ builtInMethodAst ++ methodRefAssignmentAsts ++ methodDefInArgument.map(ast => ast)
```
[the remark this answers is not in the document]
tuxology: Ah, I just changed this to a ListBuffer. I was just testing there.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
tuxology: Yes, this is my thinking as well. I am prepping a list based on usages I am seeing and then will do `.contains`. Lets wait a few more hours as I scour usages like this. I am also planning to add a dataflow test for this.

DavidBakerEffendi (approved):

PR #3411, opened 2023-08-02
kotlin2cpg: handle destructuring for array access rhs

ursachec:


xavierpinho (approved):

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/KtPsiToAst.scala
```diff
@@ -911,6 +912,10 @@ trait KtPsiToAst {
       case _: KtCallExpression => typeInfoProvider.isConstructorCall(rhsCall).getOrElse(false)
       case _                   => false
     }
+    val isArrayAccess = expr.getInitializer match {
```
xavierpinho, line 915: Just for my own edification, `expr.getInitializer.isInstanceOf[KtCallExpression]` would mean the same, right?

ursachec: I suppose you mean `expr.getInitializer.isInstanceOf[KtArrayAccessExpression]`. if so, yes, that would be the same, and it's shorter so changing...

PR #3418, opened 2023-08-03
rubysrc2cpg: add more special prefix functions

tuxology:
Following up on #3401 adding more "special" method prefixes. In addition one more test that needs fixing a bit where multiple prefixes are used. Will be handled in subsequent PR


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -270,14 +270,25 @@ trait AstForStatementsCreator {
       methodDefInArgument.addOne(ast)
     }
 
+    val prefixMethods = List(
```
DavidBakerEffendi, line 273: Faster lookups with a set
```suggestion
    val prefixMethods = Set(
```

DavidBakerEffendi (approved):
Just 1 comment

PR #3419, opened 2023-08-03
[gosrc2cpg] Support for break, continue, goto

khemrajrathore:
- Support for Break, continue, goto
- Supporting Ast Test cases for the same
- Ast test case for Nested For Loop
- Refactoring to have `BaseExpr`, `BasePrimitive`, `BaseStmt` base classes for more easier grouping

cc - @DavidBakerEffendi

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForExpressionCreator.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForExpressionCreator.scala
@@ -10,12 +10,11 @@ import io.shiftleft.codepropertygraph.generated.{DispatchTypes, Operators}
 trait AstForExpressionCreator { this: AstCreator =>
   def astsForExpression(expr: ParserNodeInfo): Seq[Ast] = {
     expr.node match {
-      case BinaryExpr  => astForBinaryExpr(expr)
-      case StarExpr    => astForStarExpr(expr)
-      case UnaryExpr   => astForUnaryExpr(expr)
-      case ParenExpr   => astsForExpression(createParserNodeInfo(expr.json(ParserKeys.X)))
-      case _: BaseExpr => Seq(Ast())
-      case _           => Seq(Ast())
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForExpressionCreator.scala
```diff
@@ -8,13 +8,13 @@ import io.shiftleft.codepropertygraph.generated.nodes.NewCall
...
-  def astForExpression(expr: ParserNodeInfo): Seq[Ast] = {
+  def astsForExpression(expr: ParserNodeInfo): Seq[Ast] = {
     expr.node match {
       case BinaryExpr => astForBinaryExpr(expr)
       case StarExpr   => astForStarExpr(expr)
       case UnaryExpr  => astForUnaryExpr(expr)
-      case ParenExpr  => astForExpression(createParserNodeInfo(expr.json(ParserKeys.X)))
-      case _          => Seq()
+      case ParenExpr  => astsForExpression(createParserNodeInfo(expr.json(ParserKeys.X)))
+      case _          => Seq(Ast())
```
DavidBakerEffendi, line 17: The following should probably be fine? If not, it's not an issue really
```suggestion
      case _          => Seq.empty[Ast]
```

DavidBakerEffendi (approved):

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForExpressionCreator.scala
khemrajrathore: @DavidBakerEffendi One observation is if we use `Seq()` or `Seq.empty[Ast]` an empty Block node is created, which is not required

DavidBakerEffendi: That is interesting, lol. But cool will merge this

PR #3421, opened 2023-08-03
[rubysrc2cpg] Cowboy Parser Hardening

DavidBakerEffendi:
When an unexpected symbol is encountered, The Ruby grammar and parser seem to be prone to bombing out to `ProgramContext` and the default error strategy continues until `<EOF>`.

This is not great for a variety of reasons, but this is understandable as Ruby is a wild west language in terms of understanding which context we're currently in, so I propose an equally wild west solution - inspired by a questionable Discord conversation between @xavierpinho and I:

* Add an error listener and keep track of the line on which an error occurred.
* If the parser wrapper detects that an error occurred, remove that line and reparse. Continue until completion.

A more elegant, two-phase parsing strategy has been suggested by @fabsx00, but for the best interest of having a robust parser as soon as possible this should do for now...


xavierpinho (approved):
In a conversation with @ml86 today, he also made a pertinent observation: unless we keep track of which lines have been removed and adjust accordingly, we'll have wrong line number information from that line onwards.

DavidBakerEffendi:
@xavierpinho @ml86 Yeah @fabsx00 suggested just making those lines blank, will adjust accordingly

PR #3441, opened 2023-08-07
[rubysrc2cpg] Reverting Workarounds Related to Parsing

pandurangpatil:
Reverting the changes, @DavidBakerEffendi recently made to handle the situation where the parser was getting into a loop for the specific use cases of the file. For now we were able to proceed by removing this file.

1. It has introduced inconsistent behaviour in the scan results.
2. It has also increased the source code parsing time.
3. When I tried with the latest master by reverting these changes and with this change. I observed CPG of bigger size, along with more files being processed, and more sources and sinks being tagged.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/Main.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/Main.scala
@@ -5,11 +5,8 @@ import io.joern.x2cpg.passes.frontend.{TypeRecoveryParserConfig, XTypeRecovery}
 import io.joern.x2cpg.{X2CpgConfig, X2CpgMain}
 import scopt.OParser
 
-final case class Config(
-  enableDependencyDownload: Boolean = false,
-  antlrCacheMemLimit: Double = 0.6d,
-  parserTimeoutMs: Long = 5000
-) extends X2CpgConfig[Config]
+final case class Config(enableDependencyDownload: Boolean = false, antlrCacheMemLimit: Double = 0.6d)
+    extends X2CpgConfig[Config]
     with TypeRecoveryParserConfig[Config] {
 
   def withEnableDependencyDownload(value: Boolean): Config = {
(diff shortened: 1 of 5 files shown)


xavierpinho:
> 3\. When I tried with the latest master by reverting these changes and with this change. I observed CPG of bigger size, along with more files being processed, and more sources and sinks being tagged.

This observation intrigues me. Could it just be a matter of setting a large enough timeout?

pandurangpatil:
> This observation intrigues me. Could it just be a matter of setting a large enough timeout?

I tried increasing the timeout to 50000 ms. Observation remains the same, it adds more time to scan the sources.

DavidBakerEffendi:
Also, likely that the issue comes from the loop of:
1. Parse
2. If Error occurs, blank the line, go to previous step
3. Finish

The reparsing on every line that an error occurs on can happen many times for an erroneous file, so I can understand where the slowdown comes from.

pandurangpatil:
> Also, likely that the issue comes from the loop of:
> 
> 1. Parse
> 2. If Error occurs, blank the line, go to previous step
> 3. Finish
> 
> The reparsing on every line that an error occurs on can happen many times for an erroneous file, so I can understand where the slowdown comes from.

@DavidBakerEffendi Yea that's right.

pandurangpatil:
The proper fix is two-phase parsing, as @fabsx00 suggested.

DavidBakerEffendi:
> The proper fix is two-phase parsing, as @fabsx00 suggested.

Unfortunately, I cannot imagine that this is going to be easy to implement in Ruby, since unlike C, methods can happen anywhere. This solution also involves modifying the grammar.

DavidBakerEffendi (approved):

pandurangpatil:
@xavierpinho most important issue with this is inconsistent results in subsequent scans. You are not sure of the results.

DavidBakerEffendi:
@pandurangpatil you will need to revert the line separator fixes from this commit on the simple ast pass tests: https://github.com/joernio/joern/commit/533b32b081947cee99a9e642b061c9016b0ed1f3#diff-45b026c78f0405d131600a5e173a4dc262cdaae9ad115dfae83deb1ac6f1a938

pandurangpatil:
> @pandurangpatil you will need to revert the line separator fixes from this commit on the simple ast pass tests: [533b32b#diff-45b026c78f0405d131600a5e173a4dc262cdaae9ad115dfae83deb1ac6f1a938](https://github.com/joernio/joern/commit/533b32b081947cee99a9e642b061c9016b0ed1f3#diff-45b026c78f0405d131600a5e173a4dc262cdaae9ad115dfae83deb1ac6f1a938)

The error was pointing at `an attempted fix` unit test from `UnknowConstructPass`. I ignored that test for now.

PR #3448, opened 2023-08-07
[rubysrc2cpg] Array Entries With Newlines

DavidBakerEffendi:
This PR fixes new lines being interpreted as array initializer arguments and strips any trailing newlines from the string literal.


xavierpinho (approved):

DavidBakerEffendi:
@xavierpinho Looks like https://github.com/joernio/joern/pull/3447 may have caused the failure of the test here (i've rebased with master)

xavierpinho:
> @xavierpinho Looks like #3447 may have caused the failure of the test here (i've rebased with master)

@DavidBakerEffendi  Indeed, I forgot to include the leading/trailing separators in the rule for `%w` in #3447. This has been fixed in #3454, with the upside of not requiring any changes to AstCreator now.

DavidBakerEffendi:
@xavierpinho Following the fix in the last comment, I'm simply pushing the unit test to catch the case if this is ever re-introduced.

PR #3450, opened 2023-08-08
Steps.browse command: open `less` pager

mpollmeier:
similar to `browse()` from ammonite:
```
joern> new Steps(Iterator(1, 2, 3)).browse
```

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Steps.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Steps.scala
@@ -2,14 +2,14 @@ package io.shiftleft.semanticcpg.language
 
 import io.shiftleft.codepropertygraph.generated.nodes.{AbstractNode, NewNode, StoredNode}
 import org.json4s.native.Serialization.{write, writePretty}
-import org.json4s.{CustomSerializer, Extraction}
-import org.json4s.Formats
-import overflowdb.traversal._
+import org.json4s.{CustomSerializer, Extraction, Formats}
+import overflowdb.traversal.*
 import overflowdb.traversal.help.Doc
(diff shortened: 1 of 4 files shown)


maltek (approved):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Steps.scala
```diff
@@ -50,6 +50,11 @@ class Steps[A](val traversal: Traversal[A]) extends AnyVal {
   def p(implicit show: Show[A] = Show.default): List[String] =
     traversal.toList.map(show.apply)
 
+  @Doc(info = "execute this traversal and show the pretty-printed results in `less`")
+  // uses scala-repl-pp's `#|^` operator which let's `less` inherit stdin and stdout
+  def browse(implicit show: Show[A] = Show.default): Unit =
+    traversal.map(show.apply) #|^ "less"
```
maltek, line 56: I think this will break on Windows :(

max-leuthaeuser: On Windows we have `more` available in the standard installation.

mpollmeier: It will, but this never worked on windows...
For reference, here's how ammonite did it: https://github.com/com-lihaoyi/Ammonite/blob/main/amm/repl/api/src/main/scala/ammonite/runtime/tools/Tools.scala#L218

That's not to say that it'd be nice if it did work on windows. I tried with `more` which apparently is available on most windows setups, but somehow the jvm doesn't find the command. Will give it another try later - if you have any additional ideas/thoughts, let me know

PR #3451, opened 2023-08-09
rubysrc2cpg: fix dataflows through constructor with blocks

tuxology:
Fixes issues where we were not able to get flows from the literal to `.new` arguments: 

```class Foo
  API_HOST = 'https://graph.facebook.com'

  private

  def bar(payload)
    Faraday.new(API_HOST) do |x|
     x
    end
  end
end
```


DavidBakerEffendi:
Closing as this is handled in the new frontend

PR #3456, opened 2023-08-09
rubysrc2cg: Fix for packing LHS with multiple RHS

rahul-privado:
Fix for case in which packing LHS is the rightmost AST. Code cleanup to avoid code duplication.

@xavierpinho


xavierpinho (changes requested):
Can you please add some unit-test(s) that check the synthetic call nodes (i.e. their structure)? I think that's important to have to make sure future refactorings don't break any assumptions introduced here

rahul-privado:
> Can you please add some unit-test(s) that check the synthetic call nodes (i.e. their structure)? I think that's important to have to make sure future refactorings don't break any assumptions introduced here

Done

xavierpinho (approved):

PR #3457, opened 2023-08-09
Reduce javasrc2cpg memory usage by removing unused method bodies and tokens from resolved types

johannescoetzee:
# Main changes

## File parsing
It changes how java source files are read. Instead of pre-parsing all source files and using the resulting JavaParser compilation units for both the type solver and AstCreator, only the source filenames are now shared and files are parsed separately for AstCreation and type solving. This means that files will be parsed twice, but this is necessary for (2) and cleans up the file handling logic a lot through the new `SourceParser` class.
## Type size reduction
Method bodies are removed from the type declarations contained in JavaParser resolved types and parser tokens are not kept when parsing files for type information. These tokens are necessary for ast creation since they're used to find location information, but are not used for finding type names. These tokens were a very large contributor to the overall type cache memory use, so not storing them explains the largest portion of the memory savings from this PR.


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreator.scala
```diff
@@ -1898,7 +1898,7 @@ class AstCreator(filename: String, javaParserAst: CompilationUnit, global: Globa
       // it shouldn't cause issues since resolvedType is only used where the extra type
       // information not available in typeName is necessary.
       val typeName     = expressionReturnTypeFullName(value)
-      val resolvedType = Try(value.calculateResolvedType()).toOption
+      val resolvedType = tryWithSafeStackOverflow(value.calculateResolvedType()).toOption
```
johannescoetzee, line 1901: I found these new stack overflows while working on this change, but it seems unrelated (and is rather a consequence of switching to the JdkJarTypeSolver). Figuring out why these occur and fixing them is a TODO, but I included the patch in here as a temporary remedy.

ml86 (approved):

PR #3497, opened 2023-08-14
Early AST Validation

DavidBakerEffendi:
As a way to enable faster feedback and better debugging for frontend development, the `Ast` class has been extended with an optional schema validation flag that will throw an exception at the exact point where a schema violation occurs. This config has been added to `X2Cpg` as a CLI option and propagated throughout the frontends as an implicit enum.

Notes:
* We could instead log a warning with an un-thrown exception
* I have noticed that the schema inheritance has permitted a lot of potentially unwanted flexibility, e.g. `AstNode` permits `CALL -(AST)-> NAMESPACE_BLOCK` which seems weird.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name                     := "joern"
 ThisBuild / organization := "io.joern"
 ThisBuild / scalaVersion := "3.3.0"
 
-val cpgVersion = "1.4.1"
+val cpgVersion = "1.4.9+2-4ee586a9"
 
 lazy val joerncli          = Projects.joerncli
 lazy val querydb           = Projects.querydb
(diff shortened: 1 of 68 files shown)


DavidBakerEffendi:
For reviewers: Most of the implementation is under the `x2cpg` package, the rest is propagating the flag along the frontends.

max-leuthaeuser:
I guess you are planning for more types of schema validation? Otherwise, a boolean flag would be sufficient.

mpollmeier:
> a boolean flag would be sufficient.

The main reason we went for an enum here is that it's passed via an implicit. Having an implicit Boolean in scope is not a good idea, since it's not very specific and might be applied in unintended ways.

mpollmeier (approved):

PR #3500, opened 2023-08-14
Speed up javasrc type inference pass

johannescoetzee:
The focus of this PR is just performance, so it speeds up the pass without modifying the logic (which could do with a revision as well). With that being said, the new version takes the pass runtime from ~160s to <10s when creating a CPG for elasticsearch on my machine.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInferencePass.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInferencePass.scala
@@ -17,7 +17,7 @@ import io.joern.x2cpg.Defines.UnresolvedNamespace
 class TypeInferencePass(cpg: Cpg) extends ConcurrentWriterCpgPass[(String, List[Call])](cpg) {
 
   private val cache = new GuavaCache(CacheBuilder.newBuilder().build[String, Option[Method]]())
-  private val resolvedMethodIndex = cpg.method.internal
+  private val resolvedMethodIndex = cpg.method
     .filterNot(_.fullName.startsWith(Defines.UnresolvedNamespace))
     .filterNot(_.signature.startsWith(Defines.UnresolvedSignature))
     .groupBy(_.name)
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInferencePass.scala
```diff
@@ -12,73 +12,76 @@ import io.shiftleft.semanticcpg.language._
...
-    getReplacementMethod(call).foreach { method =>
-      diffGraph.setNodeProperty(call, PropertyNames.MethodFullName, method.fullName)
-      diffGraph.setNodeProperty(call, PropertyNames.Signature, method.signature)
-      diffGraph.setNodeProperty(call, PropertyNames.TypeFullName, method.methodReturn.typeFullName)
+
+  override def runOnPart(diffGraph: DiffGraphBuilder, part: (String, List[Call])): Unit = {
+    part match {
+      case (_, Nil)      => // This should never happen
+      case (name, calls) =>
+        // Since call full names match, the replacement method for each will be the same
```
ml86, line 77: The full names match but only because they contain "unknown signature" so why is it the case that we are allowed to treat the call sites the same?

johannescoetzee: Ah, I did miss the fact that this skips argument type checking. I'll update the PR to handle this better.

ml86 (approved):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/TypeInferencePass.scala
```diff
@@ -5,80 +5,91 @@ import com.google.common.cache.CacheBuilder
...
+      (method.parameter.size == (call.argument.size - argSizeMod))
+
+    lazy val argTypesMatch = doArgumentTypesMatch(method: Method, call: Call, skipCallThis = argSizeMod == 1)
+
+    lazy val typeDeclMatches = (callNameParts.typeDecl == methodNameParts.typeDecl)
+
+    parameterSizesMatch && argTypesMatch && typeDeclMatches
+  }
+
+  def doArgumentTypesMatch(method: Method, call: Call, skipCallThis: Boolean): Boolean = {
```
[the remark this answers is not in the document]
johannescoetzee: Will do!

PR #3525, opened 2023-08-21
[gosrc2cpg] - Handling method node creation along with parameter and return node

pandurangpatil:
1. Added handling for setting astParentFullName and astParentType
2. Binding node connecting Method and parent TypeDecl node.
3. Basic support for parameters with primitives along with variable arguments.
4. Refactored some of the unit tests to accommodate parameter tests.
5. Changed the way we create a namespace when the package name defined is different than the folder name. With the updated change, we consider the folder name the package name. As this is the way `Go` treats it.
6. Refactored some unit tests.
7. Handling of Evaluation Strategy

Did the following code refactor

1. Remove ThreadLocal for `fullyQualifiedPackage` as it was unnecessary in the first place (For every file being processed, a new AstCreator object is being created)
2. Did some refactoring of code to move some constants at commonplace.
3. Used common `getTypeFullName` method while constructing `ParameterIn` method nodes.
4. Refactored Import nodes to consider the last path element of the import node as the default alias and set it for `importedAs`. As this is how it's been treated in `Go`. Along with this change, updated respective unit tests for the same.

Intial type full name with some other refactor
1. Did some code refactoring in unit tests to use formatted string format instead of string concatenation.
2. Incorporated common Type Full Name identification mechanism along with mapping built-in types.
3. Updated respective unit tests with updated Type Fullname.

Handling more combinations of types

Following combinations of use cases handled.

1. Pointer to primitive
2. Array of primitive pointer
3. Pointer of an array of primitive
4. Variable argument of primitive pointer
5. Pointer to struct
6. Array of struct pointer
8. Pointer of an array of struct
9. Variable argument of struct pointer
10. Struct type argument from third-party dependency.
11. Method arguments with interface{} and any types along with array, pointer and variable argument of interface{}

TODO: 
1. Method return node handling


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -2,64 +2,116 @@ package io.joern.gosrc2cpg.astcreation
...
+    method: NewMethod,
+    methodName: String,
+    methodFullName: String,
+    signature: String
+  ): Ast = {
+
+    val parentNode: NewTypeDecl = methodAstParentStack.collectFirst { case t: NewTypeDecl => t }.getOrElse {
+      // TODO: Need to add respective Unit test to test this possibility, as looks to me as dead code. Replicated it from 'c2cpg' by referring AstForFunctionsCreator.
+      val astParentType     = methodAstParentStack.head.label
+      val astParentFullName = methodAstParentStack.head.properties("FULL_NAME").toString
```
DavidBakerEffendi, line 27:
```suggestion
      val astParentFullName = methodAstParentStack.head.properties(PropertyNames.FULL_NAME).toString
```
May need to import `import io.shiftleft.codepropertygraph.generated.PropertyNames`

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -91,16 +85,97 @@ trait AstCreatorHelper { this: AstCreator =>
...
+  }
+
+  protected def processTypeInfo(jsonNode: Value): (String, String, Boolean, String) = {
+    val nodeInfo = createParserNodeInfo(jsonNode)
+    nodeInfo.node match {
+      case ArrayType =>
+        val (fullName, typeNameForcode, evaluationStrategy) = internalStarExpHandler(
+          createParserNodeInfo(jsonNode.obj(ParserKeys.Elt))
+        )
+        (s"[]${fullName}", s"[]${typeNameForcode}", false, evaluationStrategy)
```
DavidBakerEffendi, line 165: You don't need the braces here
```suggestion
        (s"[]$fullName", s"[]$typeNameForcode", false, evaluationStrategy)
```

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -2,64 +2,116 @@ package io.joern.gosrc2cpg.astcreation
...
-    val fullname = s"${fullyQualifiedPackage.get()}.${name}"
+    val fullname = s"${fullyQualifiedPackage}.${name}"
     // TODO: handle multiple return type or tuple (int, int)
     val returnType     = getReturnType(funcDecl.json(ParserKeys.Type)).headOption.getOrElse("")
     val templateParams = ""
-    val methodType     = createParserNodeInfo(funcDecl.json(ParserKeys.Type))
+    val params         = createParserNodeInfo(funcDecl.json(ParserKeys.Type)(ParserKeys.Params))
     val signature =
-      s"$fullname$templateParams (${parameterSignature(methodType)})$returnType "
+      s"$fullname$templateParams (${parameterSignature(params)})$returnType "
```
DavidBakerEffendi, line 49: Is the whitespace at the end intentional? Might make matching on signatures buggy later.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -2,64 +2,116 @@ package io.joern.gosrc2cpg.astcreation
...
   }
 
-  private def astForMethodParameter(paramType: ParserNodeInfo): Seq[Ast] = {
-    Seq(Ast(NewMethodParameterIn()))
+  private def astForMethodParameter(params: ParserNodeInfo): Seq[Ast] = {
+    var index = 1
+    params
+      .json(ParserKeys.List)
+      .arrOpt
+      .getOrElse(ArrayBuffer())
```
DavidBakerEffendi, line 71: `ArrayBuffer`s are mutable, are they necessary? Would prefer something immutable if possible, as the memory is assumed fixed in that case and for empty collections it would be minimal.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -2,64 +2,116 @@ package io.joern.gosrc2cpg.astcreation
...
-            }) mkString (", ")
-        )
-      )
       .getOrElse(ArrayBuffer())
+      .map(x =>
+        val typeInfo                                           = createParserNodeInfo(x(ParserKeys.Type))
+        val (typeFullName, typeFullNameForcode, isVariadic, _) = processTypeInfo(typeInfo.json)
+        x(ParserKeys.Names).arrOpt
+          .getOrElse(ArrayBuffer())
+          .map(y => {
```
DavidBakerEffendi, line 109: If this is unused, then perhaps go
```scala
.map(_ => typeFullName) // We are returning same type from x object for each name in the names array.
```

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
DavidBakerEffendi: Similarly in `AstForGenDeclarationCreator`

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/Defines.scala
```diff
@@ -27,4 +27,5 @@ object Defines {
   val DynamicCallUnknownFullName = "<unknownFullName>"
 
   val LeftAngularBracket = "<"
+  val Uknown             = "<unknown>"
```
DavidBakerEffendi, line 30:
```suggestion
  val Unknown             = "<unknown>"
```

DavidBakerEffendi (commented):
Great PR, my comments are mostly minor things. Once addressed, I'll give the accept.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
pandurangpatil: That's a good catch

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
pandurangpatil: what is the other option? I could only see upickel provind `.arr` or `arrOpt`

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/Defines.scala
pandurangpatil: done

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
pandurangpatil: done

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
pandurangpatil: done

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
pandurangpatil: done

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
DavidBakerEffendi: I see there is no easy way to convert this to an immutable collection for free, I think we leave it for now

DavidBakerEffendi (approved):

Issue #3527, opened 2023-08-21
[Bug] Recent Java Type Inference Speed-Up Leads to Missing Types


DavidBakerEffendi:
**Describe the bug**
Recent changes from https://github.com/joernio/joern/commit/02a375ed132aac18e7430b55d8c70331b1639bad have lead to missing type information for certain calls. 

The example used is https://github.com/devdcores/BookStoreApp-Distributed-Application and the call in question can be queried with: `cpg.call.nameExact("getMyPaymentMethodById").lineNumber(331).l`

**To Reproduce**
1. Build Joern on commit `5d98ee8568f4f47e3f404133e0421343f6cfe8cf` (before the speedup commit) and generate CPG using the repo above
2. Run query `cpg.call.nameExact("getMyPaymentMethodById").lineNumber(331).l` and see
```
  Call(
    id = 15818L,
    argumentIndex = 2,
    argumentName = None,
    code = "getMyPaymentMethodById(o.getPaymentMethodId())",
    columnNumber = Some(value = 64),
    dispatchType = "DYNAMIC_DISPATCH",
    dynamicTypeHintFullName = ArraySeq(
      "com.devd.spring.bookstorecommons.feign.PaymentFeignClient.getMyPaymentMethodById:com.devd.spring.bookstorecommons.web.GetPaymentMethodResponse(java.lang.String)"
    ),
    lineNumber = Some(value = 331),
    methodFullName = "com.devd.spring.bookstorecommons.feign.PaymentFeignClient.getMyPaymentMethodById:com.devd.spring.bookstorecommons.web.GetPaymentMethodResponse(java.lang.String)",
    name = "getMyPaymentMethodById",
    order = 2,
    possibleTypes = ArraySeq(),
    signature = "com.devd.spring.bookstorecommons.web.GetPaymentMethodResponse(java.lang.String)",
    typeFullName = "com.devd.spring.bookstorecommons.web.GetPaymentMethodResponse"
  )
```
3. Build Joern on commit `02a375ed132aac18e7430b55d8c70331b1639bad ` (the speedup commit) and generate CPG using the repo above
4. Run query `cpg.call.nameExact("getMyPaymentMethodById").lineNumber(331).l` and see
```
  Call(
    id = 15818L,
    argumentIndex = 2,
    argumentName = None,
    code = "getMyPaymentMethodById(o.getPaymentMethodId())",
    columnNumber = Some(value = 64),
    dispatchType = "DYNAMIC_DISPATCH",
[...]

PR #3532, opened 2023-08-21
Have ANY argument types match every argument type in javasrc type inference pass

johannescoetzee:
Fixes https://github.com/joernio/joern/issues/3527

The argument <-> parameter matching added during the speed-up pr was too strict and didn't handle `ANY` types correctly. With this PR, `ANY` argument types will match every type for the purposes of finding matching methods.

The behaviour still won't be exactly the same as the pre-speedup version since that version didn't take argument types into account at all, but this version will be closer and fixes https://github.com/joernio/joern/issues/3527


DavidBakerEffendi (approved):
Confirmed the fix, thanks!

PR #3541, opened 2023-08-22
[rubysrc2cpg] Handle Conditional `next` as Conditional Return

DavidBakerEffendi:
* Added control structure type to dot visualizer
* Check if an if-statement `lhs` is a `next` command with return values, if this is the case, convert it to a conditional return instead
* Extended `Ast.subTreeCopy` to be able to replace the root node with a given node. This is used to replace the `ControlStructure(next)` with a `Return`

Fixes invalid AST due to `CONTINUE` having a non-integer child literal. Note this always assumes a return when this pattern is detected, though this is likely a reasonable assumption.


xavierpinho (approved):

DavidBakerEffendi:
Hmm, I see that the way I handle the new children of the return may dissolve the AST structure. I will add a test where a call node with arguments are in the return to validate that the AST structure is preserved.

PR #3544, opened 2023-08-24
Make the filename property in pysrc2cpg relative again.

ml86:
This the introduction of META_DATA.root, absolute path are not longer
required.

@DavidBakerEffendi can you please have a special look on the expectation i change in joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/passes/TypeRecoveryPassTests.scala
To me it looks like it has been wrong in the first place because `find_one` should be external.

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/passes/TypeRecoveryPassTests.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/passes/TypeRecoveryPassTests.scala
@@ -459,7 +459,7 @@ class TypeRecoveryPassTests extends PySrc2CpgFixture(withOssDataflow = false) {
 
     "correctly determine that, despite being unable to resolve the correct method full name, that it is an internal method" in {
       val Some(selfFindFound) = cpg.typeDecl(".*InstallationsDAO.*").ast.isCall.name("find_one").headOption: @unchecked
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/passes/TypeRecoveryPassTests.scala
```diff
@@ -459,7 +459,7 @@ class TypeRecoveryPassTests extends PySrc2CpgFixture(withOssDataflow = false) {
 
     "correctly determine that, despite being unable to resolve the correct method full name, that it is an internal method" in {
       val Some(selfFindFound) = cpg.typeDecl(".*InstallationsDAO.*").ast.isCall.name("find_one").headOption: @unchecked
-      selfFindFound.callee.isExternal.toSeq shouldBe Seq(true, false)
+      selfFindFound.callee.isExternal.toSeq shouldBe Seq(true, true)
```
DavidBakerEffendi, line 462: This contradicts the description of the test case, though I wonder what the cause of this is - perhaps import resolution is confused.

DavidBakerEffendi (approved):
LGTM, I'll just need to separately look at the external/internal thing in the type recovery.

DavidBakerEffendi:
I think go ahead with the merge and I'll make a separate PR to look into the type recovery part

PR #3547, opened 2023-08-25
use `*-latest` runners

mpollmeier:
to avoid future sunsetting of runners leaving us in the dark...


ursachec (approved):

max-leuthaeuser:
The joern admin (I think @fabsx00 should be it) needs to set the new expected checks as their names changed with this PR.

fabsx00:
Alright. We need to merge this first before the runners will appear in the UI.

mpollmeier:
grafik

Issue #3552, opened 2023-08-29
[Bug] JavaSrc2Cpg : Missing Type Information in Java after revamping Type Inference Pass


khemrajrathore:
**Describe the bug**
With the recent changes made to JavaSrc2Cpg's Type Inference Pass, we are observing some loss in the methodFullName and signature for some call nodes. 
Suspected PRs causing the delta 
- https://github.com/joernio/joern/pull/3500
- https://github.com/joernio/joern/pull/3532

One such example is being shared below

**To Reproduce**
Steps to reproduce the behaviour:
1. Using Joern version `2.0.61+`
2. Clone the repo - https://github.com/devdcores/BookStoreApp-Distributed-Application
3. Import this repo using ImportCode
4. Run the query - `cpg.call("findByUserName").lineNumber(127).l`
5. See the output with unresolved Signature
```
Call(
    id = 3544L,
    argumentIndex = 0,
    argumentName = None,
    code = "findByUserName(userName)",
    columnNumber = Some(value = 7),
    dispatchType = "DYNAMIC_DISPATCH",
    dynamicTypeHintFullName = ArraySeq(),
    lineNumber = Some(value = 127),
    methodFullName = "com.devd.spring.bookstoreaccountservice.repository.UserRepository.findByUserName:<unresolvedSignature(0)>",
    name = "findByUserName",
    order = 1,
    signature = "<unresolvedSignature(0)>",
    typeFullName = "java.util.Optional"
  ) 
```

**Expected behavior**
The Types should be resolved like it used to be till Joern version `2.0.52`
```
Call(
    id = 3544L,
    argumentIndex = 0,
    argumentName = None,
    code = "findByUserName(userName)",
    columnNumber = Some(value = 7),
    dispatchType = "DYNAMIC_DISPATCH",
    dynamicTypeHintFullName = ArraySeq(),
    lineNumber = Some(value = 127),
    methodFullName = "com.devd.spring.bookstoreaccountservice.repository.UserRepository.findByUserName:java.util.Optional(java.lang.String)",
    name = "findByUserName",
    order = 1,
    signature = "java.util.Optional(java.lang.String)",
    typeFullName = "java.util.Optional"
  )
  ```
  @johannescoetzee 
  
  cc- @DavidBakerEffendi

**Comment by johannescoetzee** [comment]
The issue here is with the `findByUserName` argument types. 

```
joern> cpg.call("findByUserName").argument.l
[...]

PR #3554, opened 2023-08-29
Fix javasrc unresolved lambda parameter types

johannescoetzee:
`<unresolvedNamespace>` should only be used in method full names. When a local/param type can't be resolved, it should be `ANY` instead.

This also fixes https://github.com/joernio/joern/issues/3552


johannescoetzee:
> Error opening https://github.com/Privado-Inc/goastgen/releases/download/v0.9.0/goastgen-macos-arm64: java.io.IOException: Server returned HTTP response code: 403 for URL: ...

Some server flakiness causing test-scripts failure

DavidBakerEffendi (approved):
Verified the fix against the issue repo on my side. Thanks!

PR #3562, opened 2023-08-31
local returns are deprecated: use scala.util.boundary instead

mpollmeier:


DavidBakerEffendi (approved):

DavidBakerEffendi:
I've recently become aware of `org.objectweb.asm.Type.getType(java.lang.String)` which does exactly what that manually implemented method does. I'll make a PR that will get rid of this method altogether.

DavidBakerEffendi:
This PR should suppercede this one https://github.com/joernio/joern/pull/3593

mpollmeier:
thank you!

PR #3568, opened 2023-09-01
fix deprecation warnings in Test/compile

mpollmeier:
n.b. I dropped the test because I didn't deem it helpful - we don't
need to test a json library that we don't maintain

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/ObjectExpressionTests.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/ObjectExpressionTests.scala
@@ -156,9 +156,9 @@ class ObjectExpressionTests extends KotlinCode2CpgFixture(withOssDataflow = fals
 
     "should contain a correctly lowered representation" in {
       val List(last: Return) =
-        cpg.method.nameExact("withFailListener").block.astChildren.l: @unchecked
+        cpg.method.nameExact("withFailListener").block.astChildren.isReturn.l
 
-      val List(c: Call) = last.astChildren.l: @unchecked
+      val List(c: Call) = last.astChildren.isCall.l
       c.methodFullName shouldBe "mypkg.PClass.addListener:void(mypkg.SomeInterface)"
       val List(objExpr: TypeDecl, l: Local, alloc: Call, init: Call, i: Identifier) =
         c.astChildren.isBlock.astChildren.l: @unchecked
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/ObjectExpressionTests.scala
```diff
@@ -156,12 +156,12 @@ class ObjectExpressionTests extends KotlinCode2CpgFixture(withOssDataflow = fals
 
     "should contain a correctly lowered representation" in {
       val List(last: Return) =
-        cpg.method.nameExact("withFailListener").block.astChildren.l
```
max-leuthaeuser, line 159: In all these cases I think its better to filter explictly.
E.g., `.astChildren.isReturn.l`. With that, the `@unchecked` annotation becomes unnecessary.

mpollmeier: good point, i was just blindly following the compiler recommendation to use the automatic rewrite. will change

mpollmeier: this one from a few lines down is better off with the `@unchecked` annotation though, right?
```
      val List(objExpr: TypeDecl, l: Local, alloc: Call, init: Call, i: Identifier) =
        c.astChildren.isBlock.astChildren.l: @unchecked
```

mpollmeier: (which actually applies to half the changed lines in this PR)

max-leuthaeuser: In other tests we explicitly filter with the `.isFoo` API and in case this would result in multiple objects with `.lineNumber` additionaly.

max-leuthaeuser: That makes sure you actually test the elements meant to be tested. 
(the test style here in this few places is not optimal in first place)

max-leuthaeuser (approved):

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/ObjectExpressionTests.scala
mpollmeier: Hmm, in the above example wouldn't that result in rather long-winded code? Not sure if that's really worth it, given that this is in test code, and at (test) runtime it'd result in a MatchError if the types don't match. 

For the single-typed lists I'm completely with you.

PR #3572, opened 2023-09-03
[gosrc2cpg] - Call node handling

pandurangpatil:
1. Updated call node methodFullName, typeFullName and signature recovered for methods defined within the same project. We first tag identified call nodes TypeFullName with `<TO_BE_FILLED>`. While processing method declarations, we cache the Method Return type full name and signature. With additional `CallTypeAndSigntureLinkerPass` pass, we look for call nodes whose TypeFullName is tagged with `<TO_BE_FILLED>` and update their type fullname and signature by doing a lookup inside the cache.
2. Listed builtin functions of go while identifying  return type and signature
3. Respective unit tests covering call node use cases. TODO: There are a few more use cases that need to be added.
4. Made changes to create `IDENTIFIER` node when a variable is only declared.
5. Identified bug while setting up the fully qualified name of the file when it's getting processed. Added unit test for use case `call to method from the same project but from another package`
6. Initial handling for argument node along with a unit test for basic dataflow across the function

TODO: 
1. Additional unit tests for built-in function call nodes.
2. Argument node handling. - Basic one is already done. Need to handle using type information to update the call method signature and add more unit tests around the same.
3. TypeDecl related handling
   - Receiver node handling
   - Constructor call handling

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/GoSrc2Cpg.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/GoSrc2Cpg.scala
@@ -4,12 +4,12 @@ import better.files.File
 import io.joern.gosrc2cpg.datastructures.GoGlobal
 import io.joern.gosrc2cpg.model.GoMod
 import io.joern.gosrc2cpg.parser.GoAstJsonParser
-import io.joern.gosrc2cpg.passes.AstCreationPass
+import io.joern.gosrc2cpg.passes.{AstCreationPass, CallTypeAndSigntureLinkerPass}
 import io.joern.gosrc2cpg.utils.AstGenRunner
 import io.joern.x2cpg.X2Cpg.withNewEmptyCpg
 import io.joern.x2cpg.X2CpgFrontend
-import io.joern.x2cpg.utils.Report
 import io.joern.x2cpg.passes.frontend.{MetaDataPass, TypeNodePass}
+import io.joern.x2cpg.utils.Report
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.codepropertygraph.generated.Languages
 import org.slf4j.LoggerFactory
@@ -30,7 +30,8 @@ class GoSrc2Cpg extends X2CpgFrontend[Config] {
           GoAstJsonParser.readModFile(Paths.get(modFile)).foreach(x => GoMod.meta = Some(x))
         )
         new AstCreationPass(cpg, astGenResult, config, report).createAndApply()
-        TypeNodePass.withRegisteredTypes(GoGlobal.typesSeen(), cpg).createAndApply()
+        new CallTypeAndSigntureLinkerPass(cpg).createAndApply()
+//        TypeNodePass.withRegisteredTypes(GoGlobal.typesSeen(), cpg).createAndApply()
         report.print()
       }
     }
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
@@ -36,9 +36,11 @@ trait AstCreatorHelper { this: AstCreator =>
         parserNodeCache.addOne(json(ParserKeys.NodeId).num.toLong, pni)
         node match
           case CallExpr =>
-            json(ParserKeys.Fun)(ParserKeys.Obj).objOpt.map(obj => {
-              createParserNodeInfo(obj(ParserKeys.Decl))
-            })
+            Try(json(ParserKeys.Fun)(ParserKeys.Obj)(ParserKeys.Decl)) match
+              case Success(obj) =>
+                // This is being called only to cache this objects
+                createParserNodeInfo(obj)
+              case _ =>
           case _ =>
         // Do nothing
         pni
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/Defines.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/Defines.scala
@@ -37,22 +37,22 @@ object Defines {
   val builtinFunctions: Map[String, (String, String, String)] =
     // Prepared referring to this - https://pkg.go.dev/builtin#pkg-functions
     Map(
-      ("append", ("append([]any, []any) []any", "append", "[]any")),
-      ("cap", ("cap(any) int", "cap", "int")),
+      ("append", ("append([]any, []any)[]any", "append", "[]any")),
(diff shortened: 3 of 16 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/GoSrc2Cpg.scala
```diff
@@ -30,7 +30,8 @@ class GoSrc2Cpg extends X2CpgFrontend[Config] {
           GoAstJsonParser.readModFile(Paths.get(modFile)).foreach(x => GoMod.meta = Some(x))
         )
         new AstCreationPass(cpg, astGenResult, config, report).createAndApply()
-        TypeNodePass.withRegisteredTypes(GoGlobal.typesSeen(), cpg).createAndApply()
+        new CallTypeAndSigntureLinkerPass(cpg).createAndApply()
+//        TypeNodePass.withRegisteredTypes(GoGlobal.typesSeen(), cpg).createAndApply()
```
DavidBakerEffendi, line 34: Shouldn't this be uncommented again?

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -36,9 +36,11 @@ trait AstCreatorHelper { this: AstCreator =>
         parserNodeCache.addOne(json(ParserKeys.NodeId).num.toLong, pni)
         node match
           case CallExpr =>
-            json(ParserKeys.Fun)(ParserKeys.Obj).objOpt.map(obj => {
-              createParserNodeInfo(obj(ParserKeys.Decl))
-            })
+            Try(json(ParserKeys.Fun)(ParserKeys.Obj)(ParserKeys.Decl)) match
+              case Success(obj) =>
+                // This is being called only to cache this objects
+                createParserNodeInfo(obj)
```
DavidBakerEffendi, line 42: Instead of a match statement with an empty default case, you could use a `.foreach` maybe?

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/Defines.scala
```diff
@@ -2,6 +2,7 @@ package io.joern.gosrc2cpg.astcreation
 
 object Defines {
   val anyTypeName: String            = "ANY"
+  val tobeFilled: String             = "<TO_BE_FILLED>"
```
DavidBakerEffendi, line 5: Why did you choose this strategy over using `ANY` instead?

DavidBakerEffendi (changes requested):

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/CallTypeAndSigntureLinkerPass.scala
```diff
@@ -0,0 +1,22 @@
...
+import io.shiftleft.semanticcpg.language.*
+import overflowdb.BatchedUpdate
+
+class CallTypeAndSigntureLinkerPass(cpg: Cpg) extends CpgPass(cpg) {
+  override def run(builder: BatchedUpdate.DiffGraphBuilder): Unit =
+    cpg.call
+      .typeFullNameExact(Defines.tobeFilled)
+      .foreach(call => {
+        val (returnTypeFullName, signature) =
+          GoGlobal.methodFullNameReturnTypeMap.getOrDefault(call.methodFullName, (Defines.anyTypeName, null))
```
DavidBakerEffendi, line 17: We prefer not to use `null` in Joern, rather use `Option.empty[String]` or something

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/GoSrc2Cpg.scala
pandurangpatil: This is not being used. I wanted to add it only when we start using it.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
pandurangpatil: Sometimes there is no `ParserKeys.Obj` key some other times there is no `ParserKeys.Decl`. This Try takes care of multiple combinations, which might throw a null pointer exception.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/Defines.scala
pandurangpatil: It's one step prior to setting `ANY` if we can't resolve the TypeFullName. I am setting it `<TO_BE_FILLED>` to search for such nodes, and in the next pass `CallTypeAndSigntureLinkerPass`, I search with `<TO_BE_FILLED>` and try to resolve TypeFullName from the method cache. This will at least resolve the TypeFullName for the method calls to the methods defined within the source code. If it fails to find, then it will set it to `ANY`. 

In principle, I am yet to ascertain I am not able to resolve this Type and then set to ANY till that time it will be set to `<TO_BE_FILLED>`. This will also be helpful to troubleshoot in future, in case if there is any node that is left with this value. Then there is some use case that is not being handled in the `CallTypeAndSigntureLinkerPass`

mpollmeier:
This introduced a compiler warning about a non-exhaustive match statement, which seems legit...

```
[warn] -- [E029] Pattern Match Exhaustivity Warning: /home/mp/Projects/shiftleft/joern/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForMethodCallExpressionCreator.scala:13:42
[warn] 13 |    val (alias, methodName) = funcDetails.node match
[warn]    |                              ^^^^^^^^^^^^^^^^
[warn]    |match may not be exhaustive.
[warn]    |
[warn]    |It would fail on pattern case: BinaryExpr, UnaryExpr, StarExpr, ParenExpr, TypeAssertExpr, CallExpr
[warn]    |(More unmatched cases are elided)
```

Can you please handle the other cases, or if you're really sure this will never happen, add something like `case other => ???` together with a comment about why this cannot happen?

DavidBakerEffendi:
@mpollmeier @pandurangpatil I'll make a follow-up PR to fix this, my bad on not catching it during the review

DavidBakerEffendi:
Follow-up: https://github.com/joernio/joern/pull/3591

pandurangpatil:
> This introduced a compiler warning about a non-exhaustive match statement, which seems legit...

I missed adding that handling, I had noticed that warning. But somehow, I missed to fix it before raising the PR. 

Anyway, thanks @DavidBakerEffendi for fixing it.

PR #3587, opened 2023-09-06
[CfgCreator] Fixed fringe erasure in try-blocks

DavidBakerEffendi:
In the case of a try-block, if an explicit return node was given then the fringe would be erased (due to the nature of the `++` operator) and the Cfg would no longer be a graph rooted in the method node. The first statement in the try block would have it's fringe erased by the Cfg of the return node and thus be dangling.

This change would detect if the return node is within a try-block and conditionally inherit the fringe.

E.g.
```
    public static int tryWithExplicitReturn(String args) {
        try {
            var x = 1;
            System.out.println("in begin");
            return x;          // <---- the culprit
         } catch (Exception e) {
            System.out.println("Something went wrong." + args);
         }
         return 1;
    }

    public static void test10(String[] args) {
       String s = "MALICIOUS";
       tryWithExplicitReturn(s);
    }
```

#### CFG Before this commit, with the return node in the try block
graphviz-2
#### CFG After this commit, with the return node in the try block
graphviz-3

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/cfgcreation/CfgCreator.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/cfgcreation/CfgCreator.scala
@@ -522,7 +522,7 @@ class CfgCreator(entryNode: Method, diffGraph: DiffGraphBuilder) {
 
     val diffGraphs = tryToCatchEdges ++ catchToFinallyEdges ++ tryToFinallyEdges
 
-    val struff = if (maybeTryBlock.isEmpty) {
+    if (maybeTryBlock.isEmpty) {
       // This case deals with the situation where the try block is empty. In this case,
       // no catch block can be executed since nothing can be thrown, but the finally block
       // will still be executed.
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi:
Damn, I suppose we shouldn't have that link from `System.out.println` -> `return 1;`

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/cfgcreation/CfgCreator.scala
```diff
@@ -245,9 +252,14 @@ class CfgCreator(entryNode: Method, diffGraph: DiffGraphBuilder) {
     * of the CFG for calculation of that expression, appended to a CFG containing only the return node, connected with a
     * single edge to the method exit node. The fringe is empty.
     */
-  protected def cfgForReturn(actualRet: Return): Cfg = {
-    cfgForChildren(actualRet) ++
-      Cfg(entryNode = Option(actualRet), edges = singleEdge(actualRet, exitNode), List())
+  protected def cfgForReturn(actualRet: Return, inheritFringe: Boolean = false): Cfg = {
```
fabsx00, line 255: Can you add a comment here that explains `inheritFringe`?

fabsx00 (approved):

DavidBakerEffendi:
Offline, confirmed with @fabsx00 that the third flow is a fine overapproximation to keep (for now)

PR #3592, opened 2023-09-07
Increase `.sbtops` Memory Alloc

DavidBakerEffendi:
I've recently noticed `jssrc2cpg` and `pysrc2cpg` tests bombing because of memory related issues once again. This bumps the memory a bit closer to the [Windows and Linux runner limits](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,2 +1,2 @@
 -J-Xms4G
--J-Xmx6G
+-J-Xmx5G
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,2 +1,2 @@
--J-Xms3G
--J-Xmx4G
+-J-Xms4G
+-J-Xmx6G
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,2 +1,2 @@
--J-Xms4G
--J-Xmx5G
+-J-Xms3584M
+-J-Xmx4608M


DavidBakerEffendi:
Perhaps our gracious overlords could sponsor large runners: https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners#understanding-billing

The less fun (and probably more sensical) thing is to optimize our memory consumption 🦀

mpollmeier (approved):

DavidBakerEffendi:
@mpollmeier Looks like the memory wasn't enough to stop Windows actions like this one from hanging indefinitely

mpollmeier:
That's just a log entry (or rather scalatest info message): https://github.com/joernio/joern/blob/master/joern-cli/src/test/scala/io/joern/joerncli/ConsoleTests.scala#L12

Some tests are deactivated on windows. 
Not sure what's going on there, or where exactly it's hanging.

mpollmeier:
now the last entry is 
```
- group uses should have the correct names for all elements in the group
```
¯\_(ツ)_/¯

DavidBakerEffendi:
Seems like Windows doesn't get enough memory for other operations and continuously fails. Closing this.

PR #3595, opened 2023-09-07
[gosrc2cpg] - support to exclude files

ankit-privado:
we can now exclude file for go frontend:

`importCode.golang.apply(inputPath = "path", projectName = "name", args = List("--exclude-regex", "regex"))`

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/utils/AstGenRunner.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/utils/AstGenRunner.scala
@@ -134,8 +134,10 @@ class AstGenRunner(config: Config) {
     }
   }
 
-  private def runAstGenNative(in: String, out: File, exclude: String): Try[Seq[String]] =
-    ExternalCommand.run(s"$astGenCommand -exclude ${exclude} -out ${out.toString()} $in", ".")
+  private def runAstGenNative(in: String, out: File, exclude: String): Try[Seq[String]] = {
+    val excludeCommand = if (exclude.isEmpty) "\"\"" else s"\"$exclude\""
+    ExternalCommand.run(s"$astGenCommand -exclude $excludeCommand -out ${out.toString()} $in", ".")
+  }
 
   def execute(out: File): AstGenRunnerResult = {
     val in = File(config.inputPath)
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi (approved):

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/utils/AstGenRunner.scala
```diff
@@ -134,13 +134,15 @@ class AstGenRunner(config: Config) {
     }
   }
 
-  private def runAstGenNative(in: String, out: File): Try[Seq[String]] =
-    ExternalCommand.run(s"$astGenCommand -out ${out.toString()} $in", ".")
+  private def runAstGenNative(in: String, out: File, exclude: String): Try[Seq[String]] = {
+    val excludeCommand = if (exclude.isEmpty) "\"\"" else s"\"$exclude\""
```
pandurangpatil, line 138: @ankit-privado Can you try following? 

```
val excludeCommand = if (exclude.isEmpty) "" else s"-exclude $exclude"
ExternalCommand.run(s"$astGenCommand $excludeCommand -out ${out.toString()} $in", ".")

```

pandurangpatil (commented):
Can you try the suggestion?

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/utils/AstGenRunner.scala
ankit-privado: initially I was also doing this, but this is command line string. if we do this then command look like `..-exclude -out..` but we need `..-exclude "" -out..`

@pandurangpatil

pandurangpatil: What I am suggesting should generate `.. -out ..`

pandurangpatil: If `goastgen` always expects a value to be passed for `-exclude`, then I suppose we need to fix that on `goastgen`

ankit-privado: okay okay.

PR #3597, opened 2023-09-07
[javasrc2cpg] Introduce new `Scope` in preparation for local/anonymous class support and fixed captures.

johannescoetzee:
# Scope changes and motivation
### Separate type and variable lookups
With the old scope, types (from imports and type parameters) and variables were grouped in the same collection and lookups were done almost identically. The only difference was that for lookups where a variable is expected, attempting to match a name to a wildcard import was not permitted. 

This generally worked well enough before, but with new captured variable handling the rules for variables and types in scope are different, so they've been separated. 

### Preparation for better capture handling

Variable lookups in scope now return a `LookupResult` which can differentiate between variables captured from outer classes and those not. This is to allow for correct capture handling for all type declarations (currently only supported for lambdas) and implementation details of the `CapturedVariable` result might change.

### Stronger restrictions on scope types

See `object Scope` for details. Restricting the types of variables that can be added to the scope means some convenience logic can be moved from the `AstCreator` to the `Scope`, helping with cleaner code in the former.

### Temporary methods

The goal with the current iteration of the scope changes is to minimise the number of changes needed in the `AstCreator`, so a couple of temporary methods were added to maintain compatibility with the old `Scope` way of doing things. All of this can be refactored in the `AstCreator` to make what's going on clearer by reducing scope magic, but those changes will follow in separate PRs. For now the placeholders are marked with `TODOs`.

# Other changes
* Add a hidden `dump-javaparser-asts` option to dump the raw javaparser ast in yaml format for dev purposes.
* Add a `takeUntil` extension method to lists that effectively returns `list.takeWhile(!predicate) ++ list.find(predicate)`, or `Nil` if no matching element exists. An implementation building the result through a mutable ListBuffer may be faster, but until the performance becomes an issue, I think this implementation is cleaner.
* Add a few other TODOs to double check logic in `AstCreator`


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/scope/Scope.scala
```diff
@@ -0,0 +1,270 @@
...
+  def pushTypeDeclScope(typeDecl: NewTypeDecl): Unit = {
+    scopeStack = new TypeDeclScope(typeDecl) :: scopeStack
+  }
+
+  def pushNamespaceScope(namespace: NewNamespaceBlock): Unit = {
+    scopeStack = new NamespaceScope(namespace) :: scopeStack
+  }
+
+  def popScope(): JavaScopeElement = {
+    scopeStack.headOption match {
```
ml86, line 54: I would just to `val scope = scopeStack.head;scopeStack = scopeStack.tail;scope`. this throws an exception if the stack is empty like you also do at the moment. The custom text message is not required.
Please apply the same to the `addXXX` methods.

ml86 (approved):

PR #3611, opened 2023-09-09
[rubysrc2cpg] Creating Locals & Detecting Non-Parenthesis Calls

DavidBakerEffendi:
This PR was intended to only address the issue where calls without parenthesis were detected as identifiers due to not having the method within immediate scope, but I found that Ruby CPGs do not generate `LOCAL` nodes which was necessary for this feature.

* Corrected `.code` property for binary operation calls
* Added `IdentifierToCallPass` to heuristically determine if an identifier is meant to be a call node and would replace it in-place.
* Modified the `scope` variable by introducing a new `RubyScope` subtype that can generate local variables of the given scope and link them to their identifiers.

This resulted in many tests to be modified if they checked on certain properties.

**Note**: `this` parameter nodes need to be tested as I saw this wasn't being generated implicitly.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/dataflow/DataFlowTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/dataflow/DataFlowTests.scala
@@ -889,7 +889,7 @@ class DataFlowTests extends RubyCode2CpgFixture(withPostProcessing = true, withD
     "find flows to the sink" in {
       val source = cpg.identifier.name("x").l
       val sink   = cpg.call.name("puts").l
-      sink.reachableByFlows(source).l.size shouldBe 2
+      sink.reachableByFlows(source).l.size shouldBe 1
     }
   }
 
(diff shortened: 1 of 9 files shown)


xavierpinho (approved):

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/dataflow/DataFlowTests.scala
```diff
@@ -889,7 +889,7 @@ class DataFlowTests extends RubyCode2CpgFixture(withPostProcessing = true, withD
     "find flows to the sink" in {
       val source = cpg.identifier.name("x").l
       val sink   = cpg.call.name("puts").l
-      sink.reachableByFlows(source).l.size shouldBe 2
+      sink.reachableByFlows(source).l.size shouldBe 1
```
khemrajrathore, line 892: @DavidBakerEffendi Shouldn't the dataflow count be `2` as there are `2 identifiers` at line number `884` and `885`?

khemrajrathore: For the statement `x = [1,2,3]` looks like we are not creating identifier, because I have observed this behaviour in other frontends as well

DavidBakerEffendi: @khemrajrathore Good catch, I'll follow up on that this morning

DavidBakerEffendi: Seems there are two identifiers being generated, but looking into the CFG and DDG now

DavidBakerEffendi: Update: The Java equivalent seems to be exactly the same, though I did find some properties incorrectly filled in for the `arrayInitializer` call node. I will test what the difference is before and after this diff.

PR #3619, opened 2023-09-12
[rubysrc2cpg] Follow-Up to Local Node & Identifier Handling 

DavidBakerEffendi:
This PR follows up from the discussion of https://github.com/joernio/joern/pull/3611. It was found that some identifiers were not linked via REF edges to their local nodes and thus causing issues with dataflow.

* Fixed instances of dangling identifier nodes from their referencing local nodes by extending `RubyScope` functionality.
* Fixed code for RHS of arrays
* Moved type/method ref assignments at the beginning of the block (after locals) so that it is before their references
* Made `signature` and `name` properties consistent with arrays and binops
* Added ScalaDoc comments to RubyScope
* Implemented constructors
* Moved to using x2cpg AstBase methods where possible
* Use `relativeFilename` instead of `filename`
* Lots of general optimizations

cc @khemrajrathore


xavierpinho (approved):

DavidBakerEffendi:
One of these tests are failing because of poor builtin/user-defined method conflict handling. This and constructors are not defined, so... I'll be fixing that here

PR #3610, opened 2023-09-09
[gosrc2cpg] - FieldAccess call node handling along with Receiver identifier handling for method call node.

pandurangpatil:
1. Receiver argument handling for method call nodes
2. Fixed the duplicate call nodes getting created for Constructor call nodes.
3. Few additional unit tests
4. FieldAccess call node handling along with field identifier.
5. Created `NotHandledType` for the ast types which we have not yet handled and logged the warning.
6. Some other code refactoring.

PR #3621, opened 2023-09-12
[gosrc2cpg]: handled method call on chain of variable

hiteshbedre:
Handled case where method call is executed on chain of variable such as:

```
1. person.address.state.getName()
2. student.contact.getEmail()
```
In existing code method call on single variable is handled.


DavidBakerEffendi (approved):

pandurangpatil (changes requested):
@hiteshbedre we need to handle field access for chained call nodes for this scenario refer to this PR - https://github.com/joernio/joern/pull/3610 for the same.

Also the current unit test covers only the external objects. Add unit tests considering the struct type of chained use case within the source same file, as well as in another file.

cc: @DavidBakerEffendi

PR #3625, opened 2023-09-13
gosrc2cpg: handledling ast creation when reference id found in node

hiteshbedre:
Handling:
If key is not found under parserNodeCache (https://github.com/joernio/joern/blob/master/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala#L47) program is breaking with `key not found` exception which is halting further processing of file.

Test Case:
Added breaking test case for which processing is skipped (where node_reference_id is exist). We need to handle it separately.
cc: @pandurangpatil @ankit-privado

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
@@ -23,7 +23,7 @@ trait AstCreatorHelper { this: AstCreator =>
   private val parserNodeCache = mutable.TreeMap[Long, ParserNodeInfo]()
 
   protected def createParserNodeInfo(json: Value): ParserNodeInfo = {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -44,8 +44,18 @@ trait AstCreatorHelper { this: AstCreator =>
...
         pni
-      case Success(nodeReferenceId) => parserNodeCache(nodeReferenceId)
-
+      case Success(nodeReferenceId) =>
+        // Get the parser node info from the cache using the node reference ID
+        val parserNodeInfo = parserNodeCache.get(nodeReferenceId)
+
+        // If the parser node info exists in the cache, return it
+        parserNodeInfo match
+          case Some(value) => value
```
DavidBakerEffendi, line 53: In general, I'd prefer more descriptive names than `value`, but not worth delaying this PR on alone

DavidBakerEffendi (approved):

PR #3627, opened 2023-09-14
[gosrc2cpg] - unit tests for type full name and few data flow unit tests

pandurangpatil:
1. Unit tests around typeFullName checks.
2. Unit tests for data flow tests around struct type.
3. Removed unwanted block node creation for while creating local nodes. Which is failing some unit tests for the loop use case. TODO: We need to get it fixed as a separate PR


DavidBakerEffendi (approved):
Insane amount of operator call tests lol

PR #3642, opened 2023-09-14
added support to handle reference object

hiteshbedre:
Handling reference node object while processing Identifier.

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
@@ -34,15 +34,7 @@ trait AstCreatorHelper { this: AstCreator =>
         val node  = nodeType(json)
         val pni   = ParserNodeInfo(node, json, c, ln, cn, lnEnd, cnEnd)
         parserNodeCache.addOne(json(ParserKeys.NodeId).num.toLong, pni)
-        node match
-          case CallExpr =>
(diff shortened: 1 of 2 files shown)


pandurangpatil (approved):
looks good to me

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -58,6 +50,22 @@ trait AstCreatorHelper { this: AstCreator =>
             nullSafeCreateParserNodeInfo(None)
   }
 
+  def cacheReferenceNode(json: Value, node: ParserNode) = {
+    // This is being called only to cache this objects
+    node match
+      case CallExpr =>
+        Try(json(ParserKeys.Fun)(ParserKeys.Obj)(ParserKeys.Decl)) match
```
DavidBakerEffendi, line 57: If the non-success branch is not handled, this can be a `map` or `foreach` instead, same as below

DavidBakerEffendi (changes requested):

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreatorHelper.scala
pandurangpatil: @DavidBakerEffendi it can throw key not found error. Which we need to swallow. We had discussed this in my earlier PR.

DavidBakerEffendi (approved):

PR #3653, opened 2023-09-15
gosrc2cpg: added channel support for make function

hiteshbedre:
Extension of: https://github.com/joernio/joern/pull/3647

added support for channel.

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/Defines.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/Defines.scala
@@ -8,6 +8,7 @@ object Defines {
   val empty                          = "<empty>"
   val dot                            = "."
   val map                            = "map"
+  val chan                           = "chan"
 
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/Defines.scala
```diff
@@ -48,7 +49,7 @@ object Defines {
       ("delete", ("delete(map[any]any, any)", "delete", voidTypeName)),
       ("imag", ("imag(ComplexType)FloatType", "imag", "FloatType")),
       ("len", ("len(any)int", "len", "int")),
-      ("make", ("make(map)map", "map", "map")),
+      ("make", ("make(map)map|make(chan)chan", "make", "map|chan")),
```
pandurangpatil, line 52: Not the best way to handle this one with type full name as `map|chan` but for now we can live with it and rightly fix it at a later stage.

pandurangpatil (approved):
looks good to me

DavidBakerEffendi (approved):

PR #3670, opened 2023-09-18
gosrc2cpg: added range loop support

hiteshbedre:
In Go we can even write a for loop in below fashion:

      for range collections {

	     isTrue = isTrue && <-shutdownChan
       }
We can able to a iterator over item in collections is not mentioned here, Go handles this implicitly.

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForStatementsCreator.scala
@@ -233,16 +233,22 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
 
   private def astForRangeStatement(rangeStmt: ParserNodeInfo): Ast = {
 
-    val keyParserNode  = createParserNodeInfo(rangeStmt.json(ParserKeys.Key))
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -233,16 +233,22 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
...
+        val forNode        = controlStructureNode(rangeStmt, ControlStructureTypes.FOR, code)
+        val declAst        = astsForStatement(declParserNode)
+        val initAst        = astForNode(rangeStmt.json(ParserKeys.X))
+        val stmtAst        = astsForStatement(rangeStmt.json(ParserKeys.Body))
+        controlStructureAst(forNode, None, initAst ++ declAst ++ stmtAst)
+      case false =>
+        val initAst = astForNode(rangeStmt.json(ParserKeys.X))
+        val code    = s"for range ${createParserNodeInfo(rangeStmt.json(ParserKeys.X)).code}"
+        val forNode = controlStructureNode(rangeStmt, ControlStructureTypes.FOR, code)
+        val stmtAst = astsForStatement(rangeStmt.json(ParserKeys.Body))
```
hiteshbedre, line 250: making this statement common for both true and false is leading to some breaking test case, so handled separately.

DavidBakerEffendi (approved):

PR #3696, opened 2023-09-27
[Kotlin2Cpg] - Continue with cpg generation even if unable to download dependency

khemrajrathore:
Currently in Kotlin2Cpg, If we are unable to download dependency for some reason, we cancel the scan, This PR proposes to continue scanning even if we are unable to download dependency

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
@@ -184,8 +184,7 @@ class Kotlin2Cpg extends X2CpgFrontend[Config] with UsesService {
         deps
       case None =>
         logger.warn(s"Could not fetch dependencies for project at path $sourceDir")
-        println("Could not fetch dependencies when explicitly asked to. Exiting.")
-        System.exit(1)
+        println("Could not fetch dependencies when explicitly asked to.")
         Seq()
     }
   }
@@ -203,8 +202,7 @@ class Kotlin2Cpg extends X2CpgFrontend[Config] with UsesService {
         coordinates.toSeq
       case None =>
         logger.warn(s"Could not fetch coordinates for project at path $sourceDir")
-        println("Could not fetch coordinates when explicitly asked to. Exiting.")
-        System.exit(1)
+        println("Could not fetch coordinates when explicitly asked to.")
         Seq()
     }
   }


ursachec:
@khemrajrathore what is the reason for this change? I explicitly made the behavior as is – if the user explicitly asks to fetch dependencies, then a fail stops the scan; if the user wants to scan without fetching dependencies, they can simply remove the flag. Continuing when there is an error might cause confusion in case the user doesn't notice the log line saying that the fetch failed

ursachec:
@khemrajrathore you haven't answered my question – if you want an approval from me on this PR please do that first.

As to your question about recovering types – it's not really helping with recovery, but rather setting FULLNAMEs explicitly for references to code using logic out of dependencies.

khemrajrathore:
@ursachec The change made in the PR is to make download dependency logic consistent with what we have with JAVA frontend. If something does go wrong while downloading dependency scan still continues.

khemrajrathore:
My understanding of downloading dependencies was that it helps in getting more types and methodFullName for call nodes 
cc - @fabsx00

ursachec:
@khemrajrathore IMO javasrc2cpg should behave like kotlin2cpg, not the other way around, so I'll leave it up to fabs or someone else to give a tie-breaker

khemrajrathore:
@ursachec 
One question - How will a user know beforehand if downloading dependency is going to result in a failure?
cc - @fabsx00

ursachec:
@khemrajrathore not sure I understand the reasoning behind that question. I think I explained my thinking behind having the behavior as is fairly clearly in my first comment – either a user goes with the default behavior in which dependencies are not fetched or they choose to use the flag to download dependencies in which case an error in that fetching stops the scan. There could be arguments for changing this, but so far I haven't heard any, that's why I'm still of the opinion that the way things currently are is the way things _should_ be, and that javasrc2cpg should follow the same convention. But again, maybe someone else has good arguments for continuing a scan after an error

tuxology:
@ursachec So I feel downloading of dependencies is intended to add extra coverage to the extent possible - which means, it's a best effort activity. In that sense, this is something external to main frontend work where we request code to be downloaded. Numerous things can happen during downloading a series of deps - so IMO intent of the frontend should be to try as best as possible - by possibly throwing an exception and carrying on to increase more coverage by proceeding to download other deps that can be downloaded. Just by 2 cents.

ursachec:
@tuxology Alright, roger that. I for one don't find the argument that downloading dependencies is a best effort activity convincing – I think it's a heavier operation which should be treated as such, especially since using this flag introduces network access to a scan.

fabsx00:
Failing entirely when one dependency can't be resolved somewhat conflicts with the idea of robustness: what if we want to analyze a project and we do not have access to some internal dependencies? This is often the case in security audits, so, I'd also vote for merging this one.

ursachec (approved):

ursachec:
fair enough, then the new behavior wins @khemrajrathore

PR #3700, opened 2023-09-28
[gosrc2cpg] - Code refactor

pandurangpatil:
1. Changed Object implementation of GoMod to create Class one as we need to use many instances and context to parse files downloaded from dependencies.

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/DownloadDependenciesPass.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/DownloadDependenciesPass.scala
@@ -1,0 +1,36 @@
+package io.joern.gosrc2cpg.passes
+
+import better.files.File
+import io.joern.gosrc2cpg.model.GoModHelper
+import io.joern.gosrc2cpg.utils.AstGenRunner.getClass
+import io.joern.x2cpg.utils.ExternalCommand
+import org.slf4j.LoggerFactory
+
+import scala.util.{Failure, Success}
+
+class DownloadDependenciesPass(parentGoMod: GoModHelper) {
+  private val logger = LoggerFactory.getLogger(getClass)
+  def process(): Unit = {
+    File.usingTemporaryDirectory("go-temp-download") { tmpDir =>
+      setupDummyProjectAndDownload(tmpDir.toString)
+    }
+  }
+
+  private def setupDummyProjectAndDownload(prjDir: String): Unit = {
+    parentGoMod
+      .getModMetaData()
+      .map(mod => {
+        ExternalCommand.run("go mod init joern.io/temp", prjDir) match
+          case Success(_) =>
+            mod.dependencies.foreach(dependency => {
+              val cmd = s"go get ${dependency.module}@${dependency.version}"
+              ExternalCommand.run(cmd, prjDir) match
+                case Success(_) => print(". ")
+                case Failure(f) =>
+                  logger.error(s"\t- command '${cmd}' failed", f)
+            })
+          case Failure(f) =>
+            logger.error("\t- command 'go mod init joern.io/temp' failed", f)
+      })
+  }
+}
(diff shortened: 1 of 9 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/DownloadDependenciesPass.scala
```diff
@@ -0,0 +1,36 @@
...
+  private def setupDummyProjectAndDownload(prjDir: String): Unit = {
+    parentGoMod
+      .getModMetaData()
+      .map(mod => {
+        ExternalCommand.run("go mod init joern.io/temp", prjDir) match
+          case Success(_) =>
+            mod.dependencies.foreach(dependency => {
+              val cmd = s"go get ${dependency.module}@${dependency.version}"
+              ExternalCommand.run(cmd, prjDir) match
+                case Success(_) => print(". ")
```
DavidBakerEffendi, line 28: We probably want to avoid calling `print` as this cannot be configured like the logger

joern-cli/frontends/gosrc2cpg/src/test/scala/io/joern/go2cpg/testfixtures/GoCodeToCpgSuite.scala
```diff
@@ -16,7 +16,9 @@ trait Go2CpgFrontend extends LanguageFrontend {
     val cpgOutFile = File.newTemporaryFile("go2cpg.bin")
     cpgOutFile.deleteOnExit()
     val go2cpg = new GoSrc2Cpg()
-    val config = Config()
+    val config = getConfig()
+      .map(_.asInstanceOf[Config])
```
DavidBakerEffendi, line 20: Does this not work? A bit safer than an unchecked cast
```suggestion
      .collectFirst { case x: Config => x }
```

DavidBakerEffendi (changes requested):

joern-cli/frontends/gosrc2cpg/src/test/scala/io/joern/go2cpg/testfixtures/GoCodeToCpgSuite.scala
pandurangpatil: I simply mimic how it's been done at other places.

DavidBakerEffendi: Ok then that's fine

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/DownloadDependenciesPass.scala
pandurangpatil: When a download is happening for a little longer time and if there is no indicator to show some sort of progress on the screen. It gives the feeling of the process of getting hanged or stuck. I guess this progress bar cannot be achieved without using `print`.

joern-cli/frontends/gosrc2cpg/src/test/scala/io/joern/go2cpg/testfixtures/GoCodeToCpgSuite.scala
pandurangpatil: I have more changes to be done for the download dependency to work completely. I will handle this one in next PR.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/DownloadDependenciesPass.scala
DavidBakerEffendi: Right, fair point. I'll accept the PR as is then.

DavidBakerEffendi (approved):

PR #3731, opened 2023-10-10
Prerequisite changes to fix #3695

### LINKED ISSUE #3695: [GO] Handle global variable and constant
Global variable and constant is declared here: https://github.com/apm-dev/go-user-rest-api/blob/main/datasources/mysql/users_db/users_db.go

global variable is used here: https://github.com/apm-dev/go-user-rest-api/blob/main/domain/users/user_dao.go
(stmt, err := users_db.Client.Prepare(queryGetAll))

End outcome to verify: MethodFullName property for below call nodes would be get resolved:

cpg.call("Prepare").l

pandurangpatil:
Prerequisite changes to fix #3695 issue.

Handling done to set the member node under given TypeDecl parent.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name                     := "joern"
 ThisBuild / organization := "io.joern"
 ThisBuild / scalaVersion := "3.3.1"
 
-val cpgVersion = "1.4.23"
+val cpgVersion = "1.4.24"
 
 lazy val joerncli          = Projects.joerncli
 lazy val querydb           = Projects.querydb
@@ -64,7 +64,7 @@ ThisBuild / compile / javacOptions ++= Seq(
 ThisBuild / scalacOptions ++= Seq(
   "-deprecation", // Emit warning and location for usages of deprecated APIs.
   "--release",
-  "11",
+  "11"
 )
 
 lazy val createDistribution = taskKey[File]("Create a complete Joern distribution")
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -64,7 +64,7 @@ ThisBuild / compile / javacOptions ++= Seq(
 ThisBuild / scalacOptions ++= Seq(
   "-deprecation", // Emit warning and location for usages of deprecated APIs.
   "--release",
-  "11",
+  "11"
```
DavidBakerEffendi, line 67: To avoid unrelated change
```suggestion
  "11",
```

pandurangpatil: This is formattee change. It has removed the unwanted last `,`

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/base/AstLinkerPass.scala
```diff
@@ -16,6 +16,9 @@ class AstLinkerPass(cpg: Cpg) extends CpgPass(cpg) with LinkingUtil {
     cpg.typeDecl.whereNot(_.astParent).foreach { typeDecl =>
       addAstParent(typeDecl, typeDecl.fullName, typeDecl.astParentType, typeDecl.astParentFullName, dstGraph)
     }
+    cpg.member.whereNot(_.astParent).foreach { member =>
+      addAstParent(member, member.typeFullName, member.astParentType, member.astParentFullName, dstGraph)
```
DavidBakerEffendi, line 20: `member.typeFullName` is not this member's full name. Members don't really have full names, though, but this is only used to log errors. Perhaps you could do something like 
```suggestion
      addAstParent(member, s"${member.astParentFullName}.${member.name}", member.astParentType, member.astParentFullName, dstGraph)
```

DavidBakerEffendi (approved):

PR #3734, opened 2023-10-11
[gosrc2cpg] - Made structural changes while creating initial AST nodes per file #3695

pandurangpatil:
This is one more partial handling for package-level Global variable handling ticket #3695 

Earlier, we were creating an AST structure per file in the following way.

NameSpaceNode(`package prefixed with filename`)=>TypeDecl(`Fake TypeDecl node representing the given file`)=>Method(`Fake Method node representing the statements (imports, gobal var and constants, types and functions) in the given file`)=>Block(`all the direct statment asts as child nodes`)

With the new structure, we wanted to create a Fake TypeDecl node representing one package (One package can have multiple go files in it). In order to achieve this, we are following the order of nodes being created for a package and then for a file.

For package, i.e. per folder

NameSpaceNode(`one namespace node representing one package`)=>TypeDecl(`Fake TypeDecl node representing the give package. File path mentioned in this case will be the folder path`)

For each file in the package. The first fake node will be added as child node of the Package TypeDecl created in the previous cache pass.

Method(`Fake Method node representing the statements (imports, gobal var and constants, types and functions) in the given file`)=>Block(`all the direct statment asts as child nodes`)

- Added a few more unit tests covering the new use cases
- Changed the existing unit tests to adjust with the updated AST structure.

TODO:

- Handle the global variable and constants to be created as member fields of the package level Fake TypeDecl we are creating. As of now, it has been created as a `LOCAL` node under the Fake file level `METHOD` node we have created.
- If the Global variable or constant is getting accessed in the same package. As of now, that is being treated as an `IDENTIFIER`. We need to change it to be converted to the FieldAccess `CALL` node.
- With the above change, we might not require the File level fake `METHOD` node for Global variables or constants. We are going to treat them as Memnber fields of level Type Decl instead of `LOCAL` nodes. So we might eliminate this file level Fake `METHOD` node.

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/MethodAndTypeCacheBuilderPass.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/MethodAndTypeCacheBuilderPass.scala
@@ -1,18 +1,19 @@
 package io.joern.gosrc2cpg.passes
 
 import io.joern.gosrc2cpg.Config
-import io.joern.gosrc2cpg.astcreation.{AstCreator, CacheBuilder}
+import io.joern.gosrc2cpg.astcreation.AstCreator
 import io.joern.gosrc2cpg.model.GoModHelper
 import io.joern.gosrc2cpg.parser.GoAstJsonParser
-import io.joern.gosrc2cpg.parser.GoAstJsonParser.ParserResult
 import io.joern.x2cpg.SourceFiles
+import io.shiftleft.codepropertygraph.Cpg
+import io.shiftleft.codepropertygraph.generated.DiffGraphBuilder
 
 import java.nio.file.Paths
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration.Duration
 import scala.concurrent.{Await, Future}
 
-class MethodAndTypeCacheBuilderPass(astFiles: List[String], config: Config, goMod: GoModHelper) {
+class MethodAndTypeCacheBuilderPass(cpgOpt: Option[Cpg], astFiles: List[String], config: Config, goMod: GoModHelper) {
   def process(): Seq[AstCreator] = {
     val futures = astFiles
       .map(file => {
@@ -20,11 +21,20 @@ class MethodAndTypeCacheBuilderPass(astFiles: List[String], config: Config, goMo
           val parserResult    = GoAstJsonParser.readFile(Paths.get(file))
           val relPathFileName = SourceFiles.toRelativePath(parserResult.fullPath, config.inputPath)
           val astCreator      = new AstCreator(relPathFileName, parserResult, goMod)(config.schemaValidation)
-          astCreator.buildCache()
-          astCreator
+          val diffGraph       = astCreator.buildCache()
+          (astCreator, diffGraph)
         }
       })
-    val allResults: Future[List[AstCreator]] = Future.sequence(futures)
-    Await.result(allResults, Duration.Inf)
+    val allResults: Future[List[(AstCreator, DiffGraphBuilder)]] = Future.sequence(futures)
+    val results                                                  = Await.result(allResults, Duration.Inf)
+    val (astCreators, diffGraphs)                                = results.unzip
+    cpgOpt.map(cpg => {
+      diffGraphs.foreach(diffGraph => {
+        overflowdb.BatchedUpdate
+          .applyDiff(cpg.graph, diffGraph, null, null)
+          .transitiveModifications()
+      })
+    })
+    astCreators
   }
 }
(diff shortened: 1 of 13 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/MethodAndTypeCacheBuilderPass.scala
```diff
@@ -1,30 +1,40 @@
...
+          val diffGraph       = astCreator.buildCache()
+          (astCreator, diffGraph)
         }
       })
-    val allResults: Future[List[AstCreator]] = Future.sequence(futures)
-    Await.result(allResults, Duration.Inf)
+    val allResults: Future[List[(AstCreator, DiffGraphBuilder)]] = Future.sequence(futures)
+    val results                                                  = Await.result(allResults, Duration.Inf)
+    val (astCreators, diffGraphs)                                = results.unzip
+    cpgOpt.map(cpg => {
```
DavidBakerEffendi, line 31: Stylistically speaking, I prefer 
```scala
cpgOpt.map { cpg =>
...
    diffGraphs.foreach { diffGraph =>
    ...
```

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/MethodAndTypeCacheBuilderPass.scala
```diff
@@ -1,30 +1,40 @@
...
           val astCreator      = new AstCreator(relPathFileName, parserResult, goMod)(config.schemaValidation)
-          astCreator.buildCache()
-          astCreator
+          val diffGraph       = astCreator.buildCache()
+          (astCreator, diffGraph)
         }
       })
-    val allResults: Future[List[AstCreator]] = Future.sequence(futures)
-    Await.result(allResults, Duration.Inf)
+    val allResults: Future[List[(AstCreator, DiffGraphBuilder)]] = Future.sequence(futures)
```
DavidBakerEffendi, line 28: Perhaps to improve efficiency, you could avoid awaiting all the tasks and applying the diffs at the end, by applying them as you go. If you use a strictly ordered collection such as `Seq` or `List` then you would apply everything in the sequential order you started them.

You may want to use an executor though so that other tasks may start while you are waiting for the result of one. See [ExectorService](https://www.baeldung.com/java-executor-service-tutorial) and consider the `.newWorkStealingPool` as it will dynamically set the parallelism based on the core count and work on tasks that unblocked as it needs.

Just keep your futures ordered and you should be able to apply them in the correct order at the end.

DavidBakerEffendi: Also, you want to additionally avoid too much concurrency if the task count gets too large, etc. so maybe fixing a thread count to something like 4 may be good for predictable memory usage.

pandurangpatil: 1. With respect to applying the diff in parallel to processing other files. I did refer to the `ConcurrentWriterCpgPass`, and the way its implemented by using `LinkedBlockingQueue`. There as well `main` thread waits for the future one by one and pushes the results into the `LinkedBlockingQueue` which will, in parallel apply those diffs to CPG. As per my analysis, I did at the time of working on the performance optimization issue in Jan 23.  The time it takes to apply the diff from the point `main` thread wait and collect the results, push it into the queue, which will then be picked up by another thread and apply them vs time it will take to apply the same in `main` thread will be faster as any way main thread has no other job than waiting for other threads to finish and then move on to next pass. Having said that I went ahead in this way to wait for all the futures to get finished and then apply the diff in a sequence in `main` thread. As I believe this should not be a very time-consuming task and should be finished faster in sequence without any concurrency. Doing some more parallelism, there might just be over-optimisation. If we have to do any optimisation, we should do it when we face any issues for the same.
2. With regards to limiting concurrency, I am using `scala.concurrent.ExecutionContext.Implicits.global`, which already does the job of creating a thread pool matching the number of cores available on the machine. So I guess that should take care of the same.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/MethodAndTypeCacheBuilderPass.scala
pandurangpatil: @DavidBakerEffendi Is there something that needs to be changed here?

DavidBakerEffendi: Yes, from a style perspective it is inconsistent with the rest of Joern code, the code is `cpgOpt.map(cpg => {` instead of `cpgOpt.map { cpg =>`

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/MethodAndTypeCacheBuilderPass.scala
DavidBakerEffendi: Okay I see, great. Then we can go ahead with it in that case.

DavidBakerEffendi (approved):

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/MethodAndTypeCacheBuilderPass.scala
DavidBakerEffendi: For time's sake, I'll merge it as is for now and this can be changed in a follow-up

pandurangpatil: Ohh I get it. I didn't notice the braces

PR #3736, opened 2023-10-16
[php2cpg] Resolve static method call name when called by self

wunused:
When a static method call is made by referring to `self`, the AST node created for it failed to resolve the proper method full name.

For example:

```
<?php
class Foo {
    static function bar($x) {
        return 0;
    }
    function baz($x) {
        self::bar($x);
    }
}
```

The method full name of the call node for `self::bar($x)` is `self::bar`, but it should be `Foo::bar`.

This PR resolves the method full name of functions called by `self`.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
@@ -973,7 +973,6 @@ class AstCreator(filename: String, phpAst: PhpFile)(implicit withSchemaValidatio
       }
 
       case Some(expr) =>
-        // composeMethodFullName(name, call.isStatic)
         s"$UnresolvedNamespace\\$codePrefix"
 
       case None if PhpBuiltins.FuncNames.contains(name) =>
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
```diff
@@ -967,10 +967,13 @@ class AstCreator(filename: String, phpAst: PhpFile)(implicit withSchemaValidatio
...
       // Static method call with a known class name
-      case Some(nameExpr: PhpNameExpr) if call.isStatic =>
-        s"${nameExpr.name}${StaticMethodDelimiter}$name"
+      case Some(nameExpr: PhpNameExpr) if call.isStatic => {
+        if (nameExpr.name == "self") composeMethodFullName(name, call.isStatic)
+        else s"${nameExpr.name}${StaticMethodDelimiter}$name"
+      }
 
       case Some(expr) =>
+        // composeMethodFullName(name, call.isStatic)
```
DavidBakerEffendi, line 976: If you want to leave in commented-out code, then please supply a note

DavidBakerEffendi (approved):
Approving since this seems sound to me

PR #3748, opened 2023-10-19
Add file content flags and support to javasrc2cpg

johannescoetzee:
This PR implements the FILE_CONTENT property changes introduced in https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1737. 

Currently a temporary `enable-file-content` option is added (disabled by default) to avoid accidental source in commercial CPGs due to release schedules, but a `disable-file-content` is also added with the intent of enabling file content by default which can be turned off by this flag (currently unused).

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/OffsetUtils.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/OffsetUtils.scala
@@ -1,0 +1,30 @@
+package io.joern.x2cpg.utils
+
+import io.joern.x2cpg.AstNodeBuilder.ZeroIndexedCoordinates
+
+object OffsetUtils {
+  def getLineOffsetIndex(fileContent: Option[String]): Array[Int] = {
+    fileContent
+      .map { content =>
+        var totalCharsCounted = 0
+        content.linesWithSeparators.map { line =>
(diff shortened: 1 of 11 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/OffsetUtils.scala
```diff
@@ -0,0 +1,30 @@
...
+          lineStartOffset
+        }.toArray
+      }
+      .getOrElse(Array.empty)
+  }
+
+  def coordinatesToOffset(lineOffsetIndex: Array[Int], coordinates: ZeroIndexedCoordinates): Option[(Int, Int)] = {
+    coordinates.match {
+      case ZeroIndexedCoordinates(startLine, startColumn, endLine, endColumn)
+          if endLine <= lineOffsetIndex.length - 1 =>
```
ml86, line 22: The length check will just hide bugs in other parts of the code. If for some reason the line numbers of certain frontends are out of range, the frontend should deal with it. But most likely it will just be a bug.

ml86: Return type is than just `(Int, Int)`

ml86 (changes requested):

ml86 (approved):

PR #3750, opened 2023-10-20
[pysrc2cpg] Model Field-like Behaviour of Module Variables

DavidBakerEffendi:
The motivation for this PR is that module-level variables in Python can be accessed by other files by some implicit global scope. This behaviour is more akin to fields rather than identifiers, and should be modelled closer to that direction.

Given the caveat that the data-flow engine sees fields a bit differently, especially if the flow goes across closures, this aims to be the least intrusive change to achieve this result.

### Main changes
* Create a module-level identifier assigned to the module's type ref to be used as the base of module-level member field accesses.
* At module-level identifier STOREs, create a block that writes to the module field, then aliases the result to the identifier, e.g., `x = 1` becomes
```python
x = {
      <module>.x = 1
      tmp = <module>.x
      tmp
    }
```
* Added helper methods, e.g., `isModuleContext` to `ContextStack`
* Separated `VariableReference` class into `Reference` trait shared between `VariableReference` and `FieldReference`
* During `createMemberLinks`, `FieldReference` calls will have a `REF` edge to their referencing member. TODO: Recover these `REF` edges during the type recovery pass. *

## Misc Changes
* Fix `pysrc2cpg/ImportResolver` bug when an `__init__.py` entity is imported from a non-sibling module.
* Added shortcut in `PythonTypeRecovery.visitStatementsInBlock` to detect these strong-update blocks and "re-sugar" them to skip propagating the type via member
* Removed Python slicer tests, I would rather be intentional in how I support Python usage slicing at a later stage


joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/dataflow/DataFlowTests.scala
```diff
@@ -20,7 +19,7 @@ class DataFlowTests extends PySrc2CpgFixture(withOssDataflow = true) {
       |""".stripMargin)
     val source = cpg.literal("42")
     val sink   = cpg.call.code("print.*").argument
-    sink.reachableByFlows(source).size shouldBe 1
+    sink.reachableByFlows(source).size should be > 0
```
fabsx00, line 22: Can we use a concrete number here to ensure that it doesn't fluctuate?

fabsx00 (changes requested):
Does this contain a test for a flow we can now find which we could not find previously? It looks to me like a few test expectations were edited, but there is no new test. Also, can you comment on why the number of flows changes? Are we ok with suddenly getting 2 flows in a place where we previously only had one?

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/passes/TypeRecoveryPassTests.scala
```diff
@@ -214,6 +215,23 @@ class TypeRecoveryPassTests extends PySrc2CpgFixture(withOssDataflow = false) {
       "bar.py"
     ).cpg
 
+    "be able to traverse from `foo.[x|y|db]` to its members" in {
```
DavidBakerEffendi, line 218: One additional (non-data-flow) test to show recovered member references that were previously not there

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/dataflow/DataFlowTests.scala
```diff
@@ -20,7 +19,7 @@ class DataFlowTests extends PySrc2CpgFixture(withOssDataflow = true) {
       |""".stripMargin)
     val source = cpg.literal("42")
     val sink   = cpg.call.code("print.*").argument
-    sink.reachableByFlows(source).size shouldBe 1
+    sink.reachableByFlows(source).size shouldBe 2
```
DavidBakerEffendi, line 22: This and the flows below are simply introduced by the extra available steps within the de-sugared block

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/dataflow/DataFlowTests.scala
```diff
@@ -589,4 +591,25 @@ class RegexDefinedFlowsDataFlowTests
...
+          |import foo
+          |
+          |print(foo.a)
+          |""".stripMargin,
+        "bar.py"
+      )
+
+    val source = cpg.literal("42").l
+    val sink   = cpg.call.code("print.*").argument.l
+    sink.reachableByFlows(source).size shouldBe 1
```
DavidBakerEffendi, line 612: Because of this PR, we can now obtain the following flow:
```
    /*
    ____________________________________________________________
    | nodeType| tracked         | lineNumber| method   | file   |
    |===========================================================|
    | Literal | <module>.a = 42 | 2         | <module> | foo.py |
    | Call    | print(foo.a)    | 4         | <module> | bar.py |
     */
```

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
```diff
@@ -174,7 +176,9 @@ class SourceToStartingPoints(src: StoredNode) extends RecursiveTask[List[CfgNode
       .or(
         _.method.nameExact(Defines.StaticInitMethodName, Defines.ConstructorMethodName, "__init__"),
         // in language such as Python, where assignments for members can be directly under a type decl
-        _.method.typeDecl
+        _.method.typeDecl,
+        // for Python, we have moved to replacing strong updates of module-level variables with their members
+        _.target.isCall.nameExact(Operators.fieldAccess).argument(1).isIdentifier.name("<module>")
```
DavidBakerEffendi, line 181: This is added to compensate for the de-sugared blocks to reason between the module-level variable identifier and its field access

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
```diff
@@ -209,6 +213,11 @@ class SourceToStartingPoints(src: StoredNode) extends RecursiveTask[List[CfgNode
 
   private def targetsToClassIdentifierPair(targets: List[AstNode]): List[(TypeDecl, AstNode)] = {
     targets.flatMap {
+      case expr: FieldIdentifier =>
+        expr.method.typeDecl.map { typeDecl => (typeDecl, expr) } ++
+          expr.inCall.fieldAccess.referencedMember.flatMap { member =>
+            member.typeDecl.map { typeDecl => (typeDecl, member) }
+          }
```
DavidBakerEffendi, line 220: This extends on the `Expression` case to access the referenced members, effectively combining the behaviour of both cases below this one

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/ExtendedCfgNode.scala
```diff
@@ -37,6 +37,18 @@ class ExtendedCfgNode(val traversal: Iterator[CfgNode]) extends AnyVal {
     reachedSources.cast[NodeType]
   }
 
+  /** Trims out path elements that represent module-level variable assignment desugaring
+    */
+  private def trimModuleVariableDesugarBlock(path: Vector[PathElement]): Vector[PathElement] = {
```
DavidBakerEffendi, line 42: @fabsx00 I've added this to remove the visibility of the desugaring, now the existing Python data-flows no longer need to be +1'd in certain places since they were duplicates in any case

fabsx00 (approved):

PR #3753, opened 2023-10-23
Delete tests depending on external download.

ml86:
Two reasons:
1. This is currently flaky and blocks other PRs.
2. Downloading in unit test is a bad idea in general:
  - It takes a long time
  - Working offline is not possible
  - External download sources can be flaky or disappear.


pandurangpatil:
@ml86 I would recommend ignoring the unit tests rather than removing them all together with note in the comment. The respective unit tests help verify the download dependency integration at least in the development stage by just changing it.

cc: @DavidBakerEffendi @fabsx00

mpollmeier:
@ml86 can you elaborate more on the flakeyness? Is it 'reproducably flakey', as in does it fail 2 out of 10 times or so? Which PRs are blocked by this?
If it turns out to be flakey and there's no good fix for it, I'm for deleting it. 

However if we can unflake it and the only issue is that it's slow and requires internet connection, I'd vote for making it an [integration test](https://www.scala-sbt.org/1.x/docs/Testing.html#Integration+Tests) that only runs on CI and on user's explicit request.

ml86:
@mpollmeier It was this one https://github.com/joernio/joern/pull/3748. Got a run in which was green in the mean time. https://github.com/joernio/joern/pull/3748

ml86:
@pandurangpatil This PR is more about raising awareness for the problems of such tests. If you think it is better to just comment out, please do so.

mpollmeier:
In case anyone else wonders: the [github action runs for the PR markus linked](https://github.com/joernio/joern/actions/workflows/pr.yml?query=branch%3Ajohannes%2Fjavasrc-method-offsets) don't contain that issue, but apparently that's only because the logs get overridden if you rerun the job :(

However, there's other PRs with the same issue:
https://github.com/joernio/joern/actions/runs/6612010703/job/17956999996#step:10:9609

In that case I take back my suggestion to make it an integration test - it would still hold up other PRs due to flakeyness. 

Maybe we can all agree on marking them as `ignored` (i.e. changing `in` to `ignore`) and adding a comment in the lines of 'ignored due to flakeyness, only to be run manually and to ensure it still compiles'?

DavidBakerEffendi:
Merged https://github.com/joernio/joern/pull/3754, shall we close this?

PR #3755, opened 2023-10-23
[rubysrc2cpg] add `--useDeprecatedFrontend` flag

xavierpinho:
* Moves the current support into a `deprecated` sub-package
* Disables the current support by placing it behind a **disabled-by-default** feature flag named `--useDeprecatedFrontend`

TODO:
* I observed that in `RubyCpgGenerator` we were passing in the default `Config`. In this patch I'm merely checking if it contains the `--useDeprecatedFrontend` flag, but it seems that this would never be the case given how it's invoked. What would the appropriate thing to do here be? Create another `SourceBasedFrontend` for the deprecated ruby variant?


DavidBakerEffendi (commented):
I generally like the change, my only other idea was using `v1`/`v2` terminology but I prefer your route since I assume we'd like to get rid of the deprecated work ASAP and have the new work to be canonical from here onwards.

> What would the appropriate thing to do here be? Create another SourceBasedFrontend for the deprecated ruby variant?

I think this is a good idea, you could even make a `RubyFrontend` trait to unify both of them if this would make life easier.

xavierpinho:
@DavidBakerEffendi Went that route. Named it `rubyDeprecated` so it should be as discoverable as when typing in `ruby`.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
```diff
@@ -28,23 +28,34 @@ class RubySrc2Cpg extends X2CpgFrontend[Config] {
...
+        newCreateCpgAction(cpg, config)
+      }
+    }
+  }
 
-        val astCreationPass = new AstCreationPass(cpg, global, parser, RubySrc2Cpg.packageTableInfo, config)
-        astCreationPass.createAndApply()
-        TypeNodePass.withRegisteredTypes(astCreationPass.allUsedTypes(), cpg).createAndApply()
+  private def newCreateCpgAction(cpg: Cpg, config: Config): Unit = {
+    // TODO
```
fabsx00, line 40: It looks like we're not yet triggering the new AST generator, or am I misreading this?

fabsx00 (approved):

fabsx00:
One thing we have to ensure here is that it is indeed possible to enable this flag. What would the code look like to do that?

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
xavierpinho: You're correct. I thought of keeping the PR digestible, so I split this import into 2 steps: this first would only bring in the flag, and the next one would bring the new AstCreator

xavierpinho:
> One thing we have to ensure here is that it is indeed possible to enable this flag. What would the code look like to do that?

* From Joern's prompt, we'd replace usages of `ruby` for `rubyDeprecated`, e.g. `importCode.rubyDeprecated("...path...")`.

* Using Ruby frontend scripts (`rubysrc2cpg.sh`, etc.), we need to pass in `--useDeprecatedFrontend` as an extra flag.

* Using Joern as if it were a library, we need to set `RubySrc2Cpg`'s configuration with an extra parameter (which is `false`, by default), like so: `Config(..., ..., useDeprecatedFrontend = true)`

PR #3787, opened 2023-11-02
c2cpg: removed org.jline dependency

max-leuthaeuser:
I don't know how this dependency slipped in there but it looks like it is not used at all.

@mpollmeier any idea? You added it with: https://github.com/joernio/joern/pull/816

All tests work fine and a locally build joern also works.


mpollmeier (approved):
Please note that this PR doesn't remove the dependency, it only changes it's version. The scala-compiler itself strangely depends on it - I tried to change that in dotty but wasn't successful. Most projects don't have a classpath dependency on the compiler, we're the exception because of the repl. 

For our frontends could actually clean up the dependency tree, since they don't really need the repl... Let me know if that's something worthwhile. In the meantime you can explore the dependency tree with e.g. `sbt c2cpg/dependencyBrowseTree`

mpollmeier:
As to why I added it in that PR you linked: no idea, seriously. Doesn't really make sense :)

max-leuthaeuser:
Ok, I'll merge it then.
Let's hope that the internal build does not explode 🙃

mpollmeier:
it's not used by c2cpg, so it'll be fine :)

PR #3794, opened 2023-11-06
readme: update and additional details for the IDEA setup

mpollmeier:


mpollmeier:
That's good to hear re scala3 with the stable build, probably makes sense to drop it then. 
For reference, I use it all the time and there's no stability issues or any other downsides I'm aware of...

DavidBakerEffendi (approved):

PR #3796, opened 2023-11-06
Kotlin2Cpg - Add handling when getting class name for unknown nodes

khemrajrathore:
Currently, when forming an Unknown node there is a change that we are unable to get the `className` effectively. This PR aims to add handling for such cases.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstNodeBuilder.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstNodeBuilder.scala
@@ -25,6 +25,7 @@ import io.shiftleft.codepropertygraph.generated.nodes.{
 }
 import org.apache.commons.lang.StringUtils
 
+import scala.util.Try
 trait AstNodeBuilder[Node, NodeProcessor] { this: NodeProcessor =>
   protected def line(node: Node): Option[Integer]
   protected def column(node: Node): Option[Integer]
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstNodeBuilder.scala
```diff
@@ -46,7 +47,7 @@ trait AstNodeBuilder[Node, NodeProcessor] { this: NodeProcessor =>
 
   protected def unknownNode(node: Node, code: String): NewUnknown = {
     NewUnknown()
-      .parserTypeName(node.getClass.getSimpleName)
+      .parserTypeName(Try(node.getClass.getSimpleName).toOption.getOrElse(Defines.Unknown))
```
DavidBakerEffendi, line 50: I see `node` is also passed to `line` and `column` etc. If `node == null` then this may be problematic (similarly for other calls too). Might be worth doing this kind of `Try` check for `line` calls too.

No need to handle nulls for the other calls for now I'd say

khemrajrathore: for `line` handling is already present

```
  override def line(element: PsiElement): Option[Integer] = {
    try {
      Some(
        element.getContainingFile.getViewProvider.getDocument
          .getLineNumber(element.getTextOffset) + 1
      )
    } catch {
      case _: Throwable => None
    }
  }
  ```

khemrajrathore: similarly present for `column` as well

DavidBakerEffendi (approved):

Issue #3745, opened 2023-10-18
reachableBy and reachableByflows not working for PHP code with Joern2.0


elirizk:
Here's a sample output. On the left is Joern2.0 and on the right is the previous version (cpg version 1.4.24).
![image](https://github.com/joernio/joern/assets/76455216/9fe7f282-dd1a-4916-96ef-65da61614f4b)

itsspike:
I can confirm that the dataflow in PHP is broken.

Given this simple PHP code
```php
<?php
function Foo() {
    $input = input();
    sink($input);
}
```
no flow from `input` to `sink` is found: 
```
joern> cpg.call("sink").reachableByFlows(cpg.call("input")).size
val res1: Int = 0
```

Whereas using JavaScript, the expected dataflow is found:
```js
function Foo() {
    var input = input();
    sink(input);
}
```
```
joern> cpg.call("sink").reachableByFlows(cpg.call("input")).size
val res1: Int = 1
```

Comparing all the graphs of both snippets, I found that the AST and CFG look pretty much the same. But the DDG is quite different.

DDG for PHP example:
<img width="761" alt="image" src="https://github.com/joernio/joern/assets/39234151/b87aeb77-d76f-438d-a63d-8762816e7683">

DDG for JavaScript example:
<img width="1387" alt="image" src="https://github.com/joernio/joern/assets/39234151/3ba4324b-23d1-4a58-83ae-2ca5b89d955b">

Shouldn't be there a connection from the assignment of `input` to the call of `sink`, as in the JS graph?

elirizk:
Thanks for solving the issue for function calls data dependencies @itsspike. However, the PDG graph is still broken for regular statements, just like the example in the screenshot on the original issue:
`<?php
$source=0;
$tmp = $source +1;
$sink=$source;
?>`

itsspike:
@elirizk This time, I'm afraid I cannot confirm 😕 

Please have a look at the statements below and give me feedback if I missed something. But it looks great so far:

```scala
joern> importCode.php.fromString("""<?php
     | $source = 0;
     | $tmp = $source;
     | $sink = $tmp + 1;
     | ?>""")
```
```scala
joern> cpg.identifier("source").l
val res1: List[io.shiftleft.codepropertygraph.generated.nodes.Identifier] = List(
  Identifier(
    id = 8L,
    argumentIndex = 1,
    argumentName = None,
    code = "$source",
    columnNumber = None,
[...]

PR #3799, opened 2023-11-06
[dataflowengineoss] Fix identifier name/code mismatch in the `sameVariable` function

itsspike:
Fixes issue https://github.com/joernio/joern/issues/3745

I've discovered that no data flow is built between function calls, whereas data flows for the same code ported to `C` or `JavaScript` are found. Tracked down the problem to the function `sameVariable` of `DdgGenerator`. 

Problem is that this function compares two Nodes and tries to figures whether they are equal or not. With `PHP`, the node's _name_ is e.g. `my_input`, but the node's _code_ would be `$my_input` (mark the trailing $-sign). So the function fails by comparing the _name_ of the left node and the _code_ of the right node.

Added unit test.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/DdgGenerator.scala
@@ -356,7 +356,7 @@ private class UsageAnalyzer(
         call.argumentOption(1).exists(x => nodeToString(use).contains(x.code))
       case call: Call =>
         nodeToString(use).contains(call.code)
-      case identifier: Identifier => nodeToString(use).contains(identifier.code)
+      case identifier: Identifier => nodeToString(use).contains(identifier.name)
       case _                      => false
     }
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi:
Great find! Hopefully this change holds across all frontends, if not, then perhaps the `.name` in the PHP AST generator.

DavidBakerEffendi (approved):

DavidBakerEffendi:
@itsspike be sure to sun `sbt scalafmt test:scalafmt` to run the code formatter

itsspike:
> @itsspike be sure to sun `sbt scalafmt test:scalafmt` to run the code formatter

Oh sorry, was not aware I need to run a code formatter. Fixed it 🙂 

BTW, this was my first contribution to this project. Is there a contribution guideline?

DavidBakerEffendi:
@itsspike No, however, that is probably long over-due. For the most part, the README has everything but lacks that. I will add one today.

DavidBakerEffendi:
Guidelines here #3801

fabsx00 (approved):

PR #3806, opened 2023-11-07
[ruby] fix: double splat issue  

tuxology:
Ruby allows using double splat arguments without the attached hash for whatever reasons 🙄 Recently saw this in production code where the parser/frontend breaks. For example this is valid function definition:

```
def foo(**)
  bar(**)
end
```

While I've just added a test for now, there is a fix needed in both the new as well as old parsers


DavidBakerEffendi:
Handled in the new parser

PR #3809, opened 2023-11-08
[rubysrc2cpg] implicit RETURN node handling for array/hash literals and method declarations

xavierpinho:
* Expands the implicit RETURN node handling for array and hash literals.
* When the last statement in a method is another method declaration, the result of invoking the outer method is the name of the inner method, e.g. invoking `f` after the following snippet returns `:g` (a symbol literal), rather than the method declaration itself:

```ruby
def f
  def g = 0
end
f # returns :g
```

In the "v1" ruby frontend we would return a METHOD_REF. In the "v2" we explicitly return the symbol name.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
@@ -27,13 +27,7 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
 
     val stmtBlockAst = ParserAst(node.body) match
       case stmtList: StatementList => astForStatementListReturningLastExpression(stmtList)
-      case _: StaticLiteral =>
-        astForStatementListReturningLastExpression(StatementList(node.body, List(node.body)))
-      case _: BinaryExpression =>
-        astForStatementListReturningLastExpression(StatementList(node.body, List(node.body)))
-      case _: SingleAssignment =>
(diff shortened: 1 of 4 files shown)


max-leuthaeuser (approved):

DavidBakerEffendi:
Curious to why we are going the route of symbol literal? Is this faithful to how Ruby works on a compiler level? I'm mostly worried about tracking invocations of this return value as it propagates through the program.

xavierpinho:
> Curious to why we are going the route of symbol literal? Is this faithful to how Ruby works on a compiler level? I'm mostly worried about tracking invocations of this return value as it propagates through the program.

That's a good question. From what I have read, this seems to be the case. And also taking a peek at the resulting bytecode seems to confirm this:

```ruby
instrs = RubyVM::InstructionSequence.compile("
  def f
    def g = 0
  end
")
puts instrs.disasm
```

which prints

```text
== disasm: #<ISeq:<compiled>@<compiled>:1 (1,0)-(4,5)> (catch: false)
0000 definemethod                           :f, f                     (   2)[Li]
0003 putobject                              :f
0005 leave

== disasm: #<ISeq:f@<compiled>:2 (2,2)-(4,5)> (catch: false)
0000 definemethod                           :g, g                     (   3)[LiCa]
0003 putobject                              :g
0005 leave                                                            (   4)[Re]

== disasm: #<ISeq:g@<compiled>:3 (3,4)-(3,13)> (catch: false)
0000 putobject_INT2FIX_0_                                             (   3)[Ca]
0001 leave                                  [Re]
```

So, a method declaration gets lowered into 2 basic instructions: `definemethod` and `putobject`. The latter is a variant of the usual "push into the stack" instruction.

DavidBakerEffendi (commented):
Chatted offline, happy with the PR to go ahead

Issue #3811, opened 2023-11-09
[x2cpg] `SourceFiles.determine` too Coupled with `X2CpgConfig`


DavidBakerEffendi:
Original report by @khemrajrathore 

**Describe the problem**

[Source Code](https://github.com/joernio/joern/blob/master/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala#L57-L77)

We have 4 overloaded variants of the `determine` function, many of which end up to one or two of the same ones, but the `determine` function at LNo 57 and LNo 77 doesn’t use the `filterFiles` function, and it doesn’t seem correct to me.

The function at LNo 57 is used at most places, which is not considering any exclusion rules provided in the field `withIgnoredFilesRegex`. This leaves the `X2CpgConfig` silently/unintentionally unutilized.

It is observed that there is a place for `determine` to be called without `X2CpgConfig`, but it is mistakenly confused by frontends that ingest source code for processing.

**Desired Solution**
Rename the function at LNo 57 to something like `determineAll` and replacing it with the function at LNo 64 where ever it is being called for processing files to be generated by the CPG. As well as inspect usages of LNo77 such that the result of the solution is that all frontends call `determine` that make use of the config.

**Alternatives**
Remove the function at LNo 57 and add another overloaded variant where the config is an optional argument so that developers are intentional about using it without `X2CpgConfig`

Points from @fabsx00:
- Perhaps the filtering should not exist in `determine` at all. This adds extra coupling and implicit functionality.
- Only a subset of the `X2CpgConfig` properties are used, and by simply passing the whole config, it is unclear what it's being used for.
- File ingestion and filtering may need a general refactor to be clearer to developers and improve maintainability.

DavidBakerEffendi:
Resolved by https://github.com/joernio/joern/pull/3813 and https://github.com/joernio/joern/pull/3822

PR #3813, opened 2023-11-09
X2CPG : Refactor implementation and usage of SourceFiles determine method

khemrajrathore:
This PR addresses the following [issue](https://github.com/joernio/joern/issues/3811)

As of today we have 5 different overloaded `determine` method for getting SourceFiles, as part of this PR
- Removed the 2 `determine` methods which used `X2CpgConfig`
- Refactored 1 `determine` method to use `ignoredDefaultRegex`, `ignoredFilesRegex` and `ignoredFilesPath`
- Removed 1 `determine` method as it became redundant after refactor done by this PR

Effectively now we have 2 `determine` methods which can be used accordingly

Note - This refactoring is not applied to `JsSrc2Cpg` frontend, as there we have our own custom handling for ignored files


fabsx00 (approved):
Much better. Build is still red, but once it's green, let's get this merged.

PR #3829, opened 2023-11-16
[jssrc2cpg] improved ASTGEN_BIN handling

max-leuthaeuser:
We handle the following cases now:
1) no astgen binary in the systems `PATH` and `ASTGEN_BIN` is not set: take astgen from joern
2) astgen binary in the systems `PATH`: take that
3) `ASTGEN_BIN` contains a valid path to a file: take that concrete file as astgen binary
4) `ASTGEN_BIN` contains a valid path to a directory: take that dir and look for an executable named `astgen` there

(higher number means higher priority)

For: https://shiftleftinc.atlassian.net/browse/SEN-1805


Ferada (approved):
makes sense in that order, thanks

PR #3832, opened 2023-11-16
Adde DYNAMIC_TYPE_HINT_FULL_NAME on literal nodes.

ml86:
This is required as type anchor for type resolution.

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/NodeBuilder.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/NodeBuilder.scala
@@ -238,6 +238,10 @@ class NodeBuilder(diffGraph: DiffGraphBuilder) {
     literalNode(string, Some(Constants.builtinStrType), lineAndColumn)
   }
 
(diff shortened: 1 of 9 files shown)


joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/NodeBuilder.scala
```diff
@@ -223,28 +223,31 @@ class NodeBuilder(diffGraph: DiffGraphBuilder) {
...
 
   def stringLiteralNode(string: String, lineAndColumn: LineAndColumn): nodes.NewLiteral = {
-    val literalNode = nodes
-      .NewLiteral()
-      .code(string)
-      .typeFullName(Constants.ANY)
-      .lineNumber(lineAndColumn.line)
-      .columnNumber(lineAndColumn.column)
-    addNodeToDiff(literalNode)
+    literalNode(string, Some(Constants.builtinStrType), lineAndColumn)
```
maltek, line 238: I think we also need to consider `bytes` literals somewhere around here. (Strings and bytes APIs have some overlap but are different.)

maltek (changes requested):

maltek (approved):

PR #3848, opened 2023-11-27
[csharp] Initial Commit

DavidBakerEffendi:
* Added new module and linked respective build files
* DotNetAstGen successfully downloaded during compilation
* Test fixtures and initial test skeletons created

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/Main.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/Main.scala
@@ -1,0 +1,35 @@
+package io.joern.csharpsrc2cpg
+
+import io.joern.csharpsrc2cpg.Frontend.{cmdLineParser, defaultConfig}
+import io.joern.x2cpg.{X2CpgConfig, X2CpgMain}
+import io.joern.x2cpg.passes.frontend.{TypeRecoveryParserConfig, XTypeRecovery}
+import io.joern.x2cpg.utils.Environment
+import scopt.OParser
+
+import java.nio.file.Paths
+
+final case class Config() extends X2CpgConfig[Config] with TypeRecoveryParserConfig[Config]
+
+object Frontend {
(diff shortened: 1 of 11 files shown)


joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/Main.scala
```diff
@@ -0,0 +1,35 @@
...
+}
+
+object Main extends X2CpgMain(cmdLineParser, new CSharpSrc2Cpg()) {
+
+  def run(config: Config, csharpsrc2cpg: CSharpSrc2Cpg): Unit = {
+    val absPath = Paths.get(config.inputPath).toAbsolutePath.toString
+    if (Environment.pathExists(absPath)) {
+      csharpsrc2cpg.run(config.withInputPath(absPath))
+    } else {
+      System.exit(1)
```
xavierpinho, line 31: Usually we don't `System.exit` at this point: maybe some remnant code?

xavierpinho (approved):

pandurangpatil:
@DavidBakerEffendi I think you need to add similar script for csharp - https://github.com/joernio/joern/blob/master/joern-cli/frontends/gosrc2cpg/gosrc2cpg.sh

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/Main.scala
DavidBakerEffendi: Yeah, straight up copy-paste lol. `jssrc2cpg` has it, but I agree that it's a gnarly side-effect to have there. I'll remove it from here but it's worth noting it's existence.

pandurangpatil: I think we need to print the log with error message saying invalid input path. Otherwise, it will be difficult to make out the reason for exit.

pandurangpatil (approved):

PR #3870, opened 2023-11-30
[gosrc2cpg] - Initial Lambda Handling

pandurangpatil:
1. Handling of basic lambda method (creation of Method and MethodRef node)
2. gitignore for dotnetastgen binaries

TODO:
1. Handling of Lambda TypeDecl along with struct type.
2. Adding more combinations of unit tests.
3. Handling of Closure use case.
4. Lambda invocation call node.

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForLambdaCreator.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForLambdaCreator.scala
@@ -1,0 +1,62 @@
+package io.joern.gosrc2cpg.astcreation
+
+import io.joern.gosrc2cpg.datastructures.GoGlobal
+import io.joern.gosrc2cpg.parser.{ParserKeys, ParserNodeInfo}
+import io.joern.x2cpg.{Ast, ValidationMode, Defines as XDefines}
+import io.shiftleft.codepropertygraph.generated.nodes.{NewMethod, NewMethodReturn, NewTypeDecl}
+import io.joern.x2cpg.datastructures.Stack.StackWrapper
+import io.joern.x2cpg.utils.NodeBuilders.newModifierNode
+import io.shiftleft.codepropertygraph.generated.{ModifierTypes, NodeTypes}
+import ujson.Value
+
+trait AstForLambdaCreator(implicit withSchemaValidation: ValidationMode) { this: AstCreator =>
+
+  def astForFuncLiteral(funcLiteral: ParserNodeInfo): Seq[Ast] = {
+    val lambdaName = nextClosureName()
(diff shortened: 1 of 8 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForLambdaCreator.scala
```diff
@@ -0,0 +1,62 @@
...
+import io.shiftleft.codepropertygraph.generated.{ModifierTypes, NodeTypes}
+import ujson.Value
+
+trait AstForLambdaCreator(implicit withSchemaValidation: ValidationMode) { this: AstCreator =>
+
+  def astForFuncLiteral(funcLiteral: ParserNodeInfo): Seq[Ast] = {
+    val lambdaName = nextClosureName()
+    // if the top of the stack function is fake file level method node (which is checked with filename) then use the fully qualified package name as base fullname
+    val baseFullName = methodAstParentStack
+      .collectFirst({ case m: NewMethod if !m.fullName.endsWith(parserResult.filename) => m.fullName })
```
DavidBakerEffendi, line 18: Maybe for future, but file-level method nodes should node start having the `MODULE` modifier, then you can check `.isModule` instead

DavidBakerEffendi (approved):

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForLambdaCreator.scala
pandurangpatil: noted

pandurangpatil: @DavidBakerEffendi I added `MODULE` modifier for file-level method nodes. However, I cannot call `.isModule` on `NewMethod` object as it's a part of `AST` node.

pandurangpatil: can you point me to any sample explaining what needs to be done. I will try to make the respective changes.

DavidBakerEffendi: There are currently [modifier accessors here](https://github.com/joernio/joern/blob/master/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/propertyaccessors/ModifierAccessors.scala#L44-L49), but it is a bit more complicated as I understand the complication of it being a `NewNode` now, so I don't think we can do anything here

PR #3871, opened 2023-12-01
fix symlink loop for AWS JDK Docker image

Ferada:
Note that all of `better-files` will have this problem, if there's a loop, you'll not be able to use the symlink-following option:

```
ln -s . jre
```

This is essentially what e.g. AWS Corretto, https://hub.docker.com/_/amazoncorretto, has, which crashes the whole process with a loop exception:

```
java.io.UncheckedIOException: java.nio.file.FileSystemLoopException: /usr/lib/jvm/java-11-amazon-corretto/jre
        at java.base/java.nio.file.FileTreeIterator.fetchNextIfNeeded(FileTreeIterator.java:87)
...
        at better.files.Dispose$.better$files$Dispose$FlatMap$Implicits$traversableFlatMap$$$_$apply$$anonfun$3(Dispose.scala:139)
...
        at io.joern.x2cpg.SourceFiles$.determine(SourceFiles.scala:88)
        at io.joern.x2cpg.SourceFiles$.determine(SourceFiles.scala:58)
        at io.joern.javasrc2cpg.typesolvers.JdkJarTypeSolver$.fromJdkPath(JdkJarTypeSolver.scala:137)
...
```

If it's necessary to be able to skip these, `java.nio.file.FileTreeWalker` would have to be used I think, but I didn't follow that particular rabbit hole much more.


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JdkJarTypeSolver.scala
```diff
@@ -134,7 +135,8 @@ object JdkJarTypeSolver {
   }
 
   def fromJdkPath(jdkPath: String): JdkJarTypeSolver = {
-    val jarPaths = SourceFiles.determine(jdkPath, Set(JarExtension, JmodExtension))
+    // not following symlinks, because some setups might have a loop, e.g. AWS's Corretto
```
mpollmeier, line 138:
```suggestion
    // not following symlinks, because some setups might have a loop, e.g. AWS's Corretto
    // see https://github.com/joernio/joern/pull/3871
```

mpollmeier (approved):

johannescoetzee (approved):

PR #3874, opened 2023-12-01
Simplify XTypeRecovery

bbrehm:
XTypeRecovery contained multiple parts that were somewhat misleading. I clarified the code, in preparation for a more substrantial refactor.

1. The earlyStop system never worked because changesWereMade was not reset in between iterations. This means that this logic should either be removed in its entirety, or fixed and debugged. Per discord, removal it is.
2. The concurrency never worked, because iterators are lazy. Calling `iter.map{u => u.fork()}.map{future => future.get()}` is single threaded. It becomes concurrent if one does `iter.map{u => u.fork()}.toList.map{future => future.get()}` or something like that. This means that the concurrency logic should either be removed in its entirety, or fixed and debugged. Per discord, removal it is.

Future PRs will introduce concurrency logic that actually has a chance of working (removing and recreating is simpler than fixing).

My intention for this PR is bug-for-bug equivalent end-to-end behavior of the code.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeRecovery.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeRecovery.scala
@@ -24,7 +24,7 @@ import scala.util.matching.Regex
 /** @param iterations
   *   the number of iterations to run.
   * @param enabledDummyTypes
-  *   whether to enable placeholder dummy values for partially resolved types.
+  *   whether to enable placeholder dummy values for partially resolved types during the final iteration.
   */
 case class XTypeRecoveryConfig(iterations: Int = 2, enabledDummyTypes: Boolean = true)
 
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XTypeRecovery.scala
```diff
@@ -159,17 +159,15 @@ trait TypeRecoveryParserConfig[R <: X2CpgConfig[R]] { this: R =>
 abstract class XTypeRecovery[CompilationUnitType <: AstNode](cpg: Cpg, state: XTypeRecoveryState) extends CpgPass(cpg) {
 
   override def run(builder: DiffGraphBuilder): Unit = {
-   compilationUnit
-      .map(unit => generateRecoveryForCompilationUnitTask(unit, builder).fork())
-      .map(_.get)
-      .reduceOption((a, b) => a || b)
-      .getOrElse(false)
+    for(cu <- compilationUnits){
```
DavidBakerEffendi, line 162: I want to measure the performance impact if I had to fix the multi-threading in a branch from this one, if it goes without a hitch (i.e., tests don't start failing for some reason), then I may be able to have a swift follow-up for a faster overall pass

bbrehm: I'll follow this up relatively quickly with a version that is potentially parallel. Could you rubber-stamp https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1748 for that?

DavidBakerEffendi (approved):
Just quickly checking around if there is going to be issues arising changing the API of `compilationUnit` -> `compilationUnits` elsewhere, then will merge

DavidBakerEffendi:
> Just quickly checking around if there is going to be issues arising changing the API of `compilationUnit` -> `compilationUnits` elsewhere, then will merge

Okay yes we will need to keep the integrity of the existing API, so we would need to add `@Deprecated` to the method.

DavidBakerEffendi:
Subsumed by https://github.com/joernio/joern/pull/3886

PR #3875, opened 2023-12-01
Refactor XTypeRecovery

bbrehm:
This includes https://github.com/joernio/joern/pull/3874

Conceptual change is that old XTypeRecoveryPass becomes XTypeRecoveryPassGenerator.

The definition of CpgPass is that we process a CPG, without committing any changes. If a later component needs changes by a previous component to be visible, then we have two passes.

The old code broke that contract.

At the initial commit, this will parallelize the type recovery. I'll disable that in a follow-up commit within this PR, using https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1748

But I'm curious whether this works out-of-the-box ;)

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/JavaTypeRecoveryPassGenerator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/JavaTypeRecoveryPassGenerator.scala
@@ -1,52 +1,41 @@
-package io.joern.kotlin2cpg.passes
+package io.joern.javasrc2cpg.passes
 
-import io.joern.kotlin2cpg.Constants
 import io.joern.x2cpg.Defines
-import io.joern.x2cpg.passes.frontend.*
+import io.joern.x2cpg.passes.frontend._
 import io.shiftleft.codepropertygraph.Cpg
 import io.shiftleft.codepropertygraph.generated.PropertyNames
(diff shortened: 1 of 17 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/JavaTypeRecoveryPassGenerator.scala
```diff
@@ -8,22 +8,22 @@ import io.shiftleft.codepropertygraph.generated.nodes._
 import io.shiftleft.semanticcpg.language._
 import overflowdb.BatchedUpdate.DiffGraphBuilder
 
-class JavaTypeRecoveryPass(cpg: Cpg, config: XTypeRecoveryConfig = XTypeRecoveryConfig())
-    extends XTypeRecoveryPass[Method](cpg, config) {
-  override protected def generateRecoveryPass(state: XTypeRecoveryState): XTypeRecovery[Method] =
-    new JavaTypeRecovery(cpg, state)
+class JavaTypeRecoveryPassGenerator(cpg: Cpg, config: XTypeRecoveryConfig = XTypeRecoveryConfig())
```
DavidBakerEffendi, line 11: We may need to do this in a slightly non-API breaking way, i.e., add @Deprecated to `JavaTypeRecoveryPass` and have it instead spawn `JavaTypeRecoveryPassGenerator`, etc. for the rest

DavidBakerEffendi:
Impressed that it all passes!

PR #3876, opened 2023-12-01
Don't follow symlinks when finding jdk jars

johannescoetzee:
Following symlinks is leading to `FileSystemLoopExceptions` in some cases when finding jdk jars or jmods for the javasrc type solver. We shouldn't need to follow these in this case, so this just disables that. This could mean that an explicit jdk path argument must be given if the frontend is invoked via a symlinked java distribution.


maltek (approved):
@Ferada already opened https://github.com/joernio/joern/pull/3871 for this. That one has more comments explaining things, so tiny preference from me that we close this PR.

johannescoetzee:
Ah, missed that. Closing in favour of https://github.com/joernio/joern/pull/3871

PR #3886, opened 2023-12-04
[XTypeRecovery] Parallelism Fix, Serial Data Structures, CPG Pass Formalisms

bbrehm:
This is the final PR in my sequence of XTypeRecovery refactors. It removes all the now-unneeded concurrent datastructures. In includes the previous change-sets https://github.com/joernio/joern/pull/3875 and https://github.com/joernio/joern/pull/3874


DavidBakerEffendi:
I should be able to take over from here - the failing compilation seems odd but likely not a large issue

DavidBakerEffendi (approved):
In summary, API migration should be:
*` XTypeRecoveryPass` -> `XTypeRecoveryPassGenerator`
*` XTypeRecoveryPassGenerator` is no longer a pass, but `.generate()` will create a sequence of passes
* `XTypeRecovery.compilationUnit` -> `XTypeRecovery.compilationUnits`
* `RecoverForXCompilationUnit.isField` -> `RecoverForXCompilationUnit.isFieldUncached`

DavidBakerEffendi:
# Informal Performance Improvement

Using the [code that was used as an example is a previous issue](https://github.com/joernio/joern/issues/3211) with the type recovery pass, I tested `master` and this branch.

## `master`

Screenshot 2023-12-06 at 13 59 23

## `bbrehm/xtype3`

Screenshot 2023-12-06 at 14 01 22

## Conclusion

Seems like real multi-threading is about 2.3 times faster than "fake" multi-threading! :D 

Thanks again @bbrehm!

PR #3888, opened 2023-12-04
[gosrc2cpg] - TypeDecl handling for inheritsFromTypeFullName with type

pandurangpatil:
1. Handling of lambda function types and mapping it with lambdas with matching signature.
2. Respective Unit tests.

TODO:

1. Handling of global package level constructor node. Along with verification of lambda defined at the package level and the respective data flow.
2. Handling of closure use case.
3. More data flow unit tests.

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
@@ -5,8 +5,7 @@ import io.joern.gosrc2cpg.parser.{ParserKeys, ParserNodeInfo}
 import io.joern.x2cpg.datastructures.Stack.*
 import io.joern.x2cpg.utils.NodeBuilders
 import io.joern.x2cpg.{Ast, ValidationMode}
-import io.shiftleft.codepropertygraph.generated.nodes.*
-import io.shiftleft.codepropertygraph.generated.{EvaluationStrategies, NodeTypes, PropertyNames}
+import io.shiftleft.codepropertygraph.generated.{EvaluationStrategies, NodeTypes}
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -127,7 +126,7 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
         val typeInfo                                           = createParserNodeInfo(x(ParserKeys.Type))
         val (typeFullName, typeFullNameForcode, isVariadic, _) = processTypeInfo(typeInfo, genericTypeMethodMap)
         x(ParserKeys.Names).arrOpt
-          .getOrElse(List())
+          .getOrElse(List(""))
```
DavidBakerEffendi, line 129: Why a blank string?

pandurangpatil: While generating the signature for a function structure

```
func test(a, b int, c string) int {
}

```
it works because if there is no situation where there will not be a parameter name exist.

When I reuse the same function for generating the signature for lambda types as below

`type Operation func(int, int) int`

Now in this case there is no parameter name exist, in order to handle this situation add this empty string, which only facilitates adding the parameter type to list.

DavidBakerEffendi: Ok cool, maybe next PR it would be good to have a comment here, since someone debugging or developing in the future might be confused/accidentally delete it.

DavidBakerEffendi (approved):

PR #3893, opened 2023-12-05
[dataflowengineoss] - Graceful Error Handling for AccessPathUsage

pandurangpatil:
1. Added ERROR log for missing node handling and continued with further processing. Added respective unit tests for the same. Note: This is not the fix, this will just unblock the further processing of the CPG.
2. Added a few comments as a result of suggestions from past PR review comments.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/AccessPathUsage.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/AccessPathUsage.scala
@@ -40,7 +40,7 @@ object AccessPathUsage {
           val path         = AccessPathHandling.memberAccessToPath(memberAccess, tail)
           (base, path)
         case _ =>
-          logger.error(s"Missing handling for node type ${node.getClass}.")
+          logger.warn(s"Missing handling for node type ${node.getClass}.")
           (TrackedUnknown, Nil)
       }
     }
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/AccessPathUsage.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/AccessPathUsage.scala
@@ -39,6 +39,9 @@ object AccessPathUsage {
           val (base, tail) = toTrackedBaseAndAccessPathInternal(argOne.get)
           val path         = AccessPathHandling.memberAccessToPath(memberAccess, tail)
           (base, path)
+        case _ =>
+          logger.error(s"Missing handling for node type ${node.getClass}.")
+          (TrackedUnknown, Nil)
       }
     }
   }
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
@@ -126,6 +126,22 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
         val typeInfo                                           = createParserNodeInfo(x(ParserKeys.Type))
         val (typeFullName, typeFullNameForcode, isVariadic, _) = processTypeInfo(typeInfo, genericTypeMethodMap)
         x(ParserKeys.Names).arrOpt
+          /*
+          While generating the signature for a function structure
+
+          func test(a, b int, c string) int {
+          }
+
+          it works because if there is no situation where there will not be a parameter name exist.
+
+          When I reuse the same function for generating the signature for lambda types as below
+
+          type Operation func(int, int) int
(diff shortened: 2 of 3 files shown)


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/AccessPathUsage.scala
```diff
@@ -39,6 +39,9 @@ object AccessPathUsage {
           val (base, tail) = toTrackedBaseAndAccessPathInternal(argOne.get)
           val path         = AccessPathHandling.memberAccessToPath(memberAccess, tail)
           (base, path)
+        case _ =>
+          logger.error(s"Missing handling for node type ${node.getClass}.")
```
DavidBakerEffendi, line 43: Since the process doesn't propagate an exception, I would choose to log a warning instead of an error
```suggestion
          logger.warn(s"Missing handling for node type ${node.getClass}.")
```

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala
```diff
@@ -695,4 +695,16 @@ class DataflowTest extends DataFlowCodeToCpgSuite {
       sink.reachableBy(identifierSource).size shouldBe 1
     }
   }
+  "Field access on TemplatedDom directly" should {
```
DavidBakerEffendi, line 698: Could you add a space between the tests, please?

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala
```diff
@@ -695,4 +695,16 @@ class DataflowTest extends DataFlowCodeToCpgSuite {
...
+    val cpg = code("""
+                     |import { useRouter } from "next/router";
+                     |
+                     |const tabComponentType = (<Tab title={"typeComponent"} />).type;
+                     |
+        |""".stripMargin)
+    "Not throw error and get handled it gracefully" in {
+      val List(x) = cpg.identifier("tabComponentType").l
```
DavidBakerEffendi, line 706: Perhaps just generate the CPG in here, then you could remove the query, since the error happens during a pass in any case.

Since you're only testing that an unhandled expression doesn't throw an error, you could do something like:

```scala
noException should be thrownBy code(""" ... """)
```

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -126,6 +126,22 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
         val typeInfo                                           = createParserNodeInfo(x(ParserKeys.Type))
         val (typeFullName, typeFullNameForcode, isVariadic, _) = processTypeInfo(typeInfo, genericTypeMethodMap)
         x(ParserKeys.Names).arrOpt
+          /*
```
DavidBakerEffendi, line 129: I don't think this should be part of this commit as it's unrelated

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala
pandurangpatil: For some reason the entire process of CPG generation doesn't happen in the `code` method invocation. You need to run some query to have CPG generated along with overlay passed getting executed.

Just now I tried what you suggested, and it doesn't throw the error and fail the test case when I try reverting the fix.

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala
pandurangpatil: done

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/AccessPathUsage.scala
pandurangpatil: done

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/dataflow/DataflowTest.scala
DavidBakerEffendi: Okay I see, must be that there is some "lazy" element implicit in the CPG test generation

DavidBakerEffendi (approved):

PR #3894, opened 2023-12-05
[c#] Basic Deserialization & File Nodes Generated + `x2cpg.astgen` API

### LINKED ISSUE #3851: [c#] `astcreation` Module & `AstCreationPass`
The foundation of any frontend is the AST creation module and pass.

This pass should accept the AST gen result comprised of all the JSON files. File names from this point on should be interpreted as relative.

### LINKED ISSUE #3860: [c#] Deserialise JSON Output From AST Gen
JSON file and its properties need to be deserialized in order to be used in Joern. This is a fairly large ticket but has other frontends to copy from so it shouldn't be too difficult other than figuring out the schema.

The schema is somewhat arbitrary for now and up to be changed, so it would be best to have some basic high-level components deserialized for testing purposes, and the rest can generate an "unknown"-style variant.

DavidBakerEffendi:
Firstly, while creating the usual AST gen components, and referring to existing work in JavaScript and Go, I've noticed a lot of useful components locked into each individual frontend. 

This PR builds the support for C#, but from a x2cpg API point onwards. This is all under an `astgen` package, but I have refrained from completely implementing these in the other frontends. Most of the commonly duplicated components have been transformed into traits and classes that can be imported in, instead of copy-pasting in an AST-Gen-like frontend.

I have also removed common dependencies that would now be brought in transitively via x2cpg.

Resolves #3860 
Resolves #3851


max-leuthaeuser:
That astgen refactoring makes a lot of sense, thanks!
Will be a nightmare to adapt https://github.com/joernio/joern/pull/3883 where I did a few thing differently (AstgenResult and in the runner itself, e.g., no version check because swiftastgen does not support that). 
Maybe we merge https://github.com/joernio/joern/pull/3883 first and adapt this PR here afterwards to get all frontends going in one run?

DavidBakerEffendi:
@max-leuthaeuser  Sounds good!

DavidBakerEffendi:
@max-leuthaeuser perhaps we could get this merged, then migrate this new `x2cpg` API accordingly in follow-ups? I think the most cohesiveness I've found so far is between `jssrc`, `gosrc`, and `csharp`. Mostly because they are based on eachother, in order.

max-leuthaeuser (approved):

PR #3899, opened 2023-12-06
[gosrc2cpg] - Package level TypeDecl constructor handling

pandurangpatil:
1. Pending Package level TypeDecl constructor handling is added to handle the initialisation of global variables with the respective RHS side. Handling this was a little tricky as global variables might get defined in different files. It is being handled by caching the package level global variable assignment statement AST mapped against the package TypeDecl fullname. We create this package-level constructor Method node in a separate pass by using this cached information.
2. Added simple data flow test as well.

TODO:
1. Need to add more unit tests around package level variable Lambda expression initialisation as well as data flow tests for the same.
2. More unit tests around the general package-level variable declaration.
3. Handling of closure use case.
4. More data flow tests around lambda expression handling.


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstForPackageConstructorCreator.scala
```diff
@@ -0,0 +1,51 @@
...
+import scala.collection.immutable.Set
+
+class AstForPackageConstructorCreator(val pacakgePath: String, statements: Set[(Ast, String)])(implicit
+  withSchemaValidation: ValidationMode
+) extends AstCreatorBase(pacakgePath) {
+
+  override def createAst(): DiffGraphBuilder = {
+    val name = StringUtils.normalizeSpace(s"$pacakgePath${XDefines.StaticInitMethodName}")
+    val fakeGlobalMethod =
+      NewMethod()
```
pandurangpatil, line 18: @DavidBakerEffendi I had to create direct nodes instead of using x2cpg APIs that's because I didn't had a corresponding Node object from which the respective line and column number information is set from the API.

DavidBakerEffendi: The `node` argument in `AstNodeBuilder` is largely null-safe so the lineNumber-style variables will just be left as their default values. You would however need to update the `code` property afterwards, but this is fine.

DavidBakerEffendi: But I suppose the intention is that they are meant to be `0`, in which case then yeah disregard the last comment.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/datastructures/GoGlobal.scala
```diff
@@ -25,6 +25,8 @@ object GoGlobal extends Global {
 
   val lambdaSignatureToLambdaTypeMap: ConcurrentHashMap[String, Set[(String, String)]] = new ConcurrentHashMap()
 
+  val pkgLevelVarAndConstantAstMap: ConcurrentHashMap[String, Set[(Ast, String)]] = new ConcurrentHashMap()
```
DavidBakerEffendi, line 28: Just a comment, but the Scala version of this is `TrieMap`

pandurangpatil: Let me try that in the next PR.

PR #3914, opened 2023-12-07
[javasrc2cpg] Add anonymous classes and captures for inner classes

johannescoetzee:
This PR adds capture logic to nested and inner classes that mimics javac logic. In short, captured variables are represented as members in the capturing class and are passed as extra constructor arguments. In addition, an `outerClass` member/parameter is created with `this` from the outer class passed in, which can then be used to access captured members and methods. For example,

```java
class Foo {
  int fooMember = 0;

  void foo(int x, int y) { ... }

  void definesClass(int capturedParameter) {
    class Bar {
      void bar() {
        foo(fooMember, capturedParameter);
      }
    }
    Bar b = new Bar();
    b.bar();
  }
```

is represented as

```java
class Foo {
  int fooMember = 0;

  void foo(int x, int y) { ... }

  void definesClass(int capturedParameter) {
    class Bar {
      int capturedParameter;
      Foo outerClass;

      public Bar(Foo outerClass, int capturedParameter) {
        this.outerClass = outerClass;
        this.capturedParameter = capturedParameter;
      }

      void bar() {
        this.outerClass.foo(this.outerClass.fooMember, this.capturedParameter);
      }
    }
    Bar b = new Bar(this, capturedParameter);
    b.bar();
  }
```

Some TODOs that would follow this PR are:
- To clean up the `BindingTableAdapters`. There are currently 3 (one for inner classes, one for regular classes and one for lambdas), but we probably only need one (a version of the inner class one). The best way to unify these would depend somewhat on the next point.
- Maybe use this representation for captures in lambdas as well. It would be nice to not have duplicate code paths, but I'm not sure how well the open source dataflow engine would deal with this, given this representation's reliance on member initializations and accesses. @DavidBakerEffendi do you have any thoughts on this?
- Clean up how javasrc deals with object creation expressions. The logic used currently is quite convoluted and it has been the single biggest issue while implementing this. It works in all the cases I've tested, but I would not be surprised if there is a case I missed.

This fixes dataflow tests in https://github.com/ShiftLeftSecurity/codescience/pull/7376

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
@@ -164,9 +164,7 @@ class AstCreator(
         .importedAs(name)
         .importedEntity(typeFullName)
 
-      if (importStmt.isStatic) {
-        // TODO: Deal with this case
-      } else {
+      if (!importStmt.isStatic) {
         scope.addTopLevelType(name, typeFullName)
       }
       importNode
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
@@ -16,7 +16,11 @@ import com.github.javaparser.ast.expr.{
 import com.github.javaparser.ast.nodeTypes.{NodeWithName, NodeWithSimpleName}
 import com.github.javaparser.ast.{CompilationUnit, Node, PackageDeclaration}
 import com.github.javaparser.resolution.UnsolvedSymbolException
-import com.github.javaparser.resolution.declarations.{ResolvedMethodLikeDeclaration, ResolvedReferenceTypeDeclaration}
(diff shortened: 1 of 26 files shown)


DavidBakerEffendi (approved):
This is a great way to get flows propagated, I wonder if this is actually a better way going forward to getting data-flows propagated using existing structures instead of what is proposed in the lambda epic.

In some way, "lowering" all this logic into anonymous but known structures may get us to where we want in terms of lambda support.

If this becomes "cannon" then we may want to document this as the official way to model lambdas in the CPG

johannescoetzee:
> If this becomes "cannon" then we may want to document this as the official way to model lambdas in the CPG

I'm not sure if this would be a good general representation for lambdas since the way Java deals with them is somewhat convoluted, with lambdas being (at least conceptually) just syntactic sugar for a sort of anonymous class declaration. For languages that support "real" anonymous functions, this intermediate/implicit object might just be an extra layer of complexity, especially since lambdas in such languages can be called directly and not via the object they define. Then again, PHP does exactly this, where lambdas implicitly create a `Closure` object and this object can then be called directly, so such a representation could probably work in general. 

Regardless of whether we choose this as a general representation, the functionality added here would be useful as a proof of concept to see how well dataflow tracks.

DavidBakerEffendi:
Yeah a good proof-of-concept, but also potentially a lot more to model (change x N frontends) than maybe a simpler change in the data-flow engine I suppose.

Issue #3901, opened 2023-12-07
[Bug][Jssrc2cpg] ReachingDefPass failed


yzn12345:
[eslint-plugin-react.zip](https://github.com/joernio/joern/files/13594005/eslint-plugin-react.zip)

```

moving cpg.bin.zip to cpg.bin because it is already a database file
Creating working copy of CPG to be safe
Loading base CPG from: C:\Users\admin\Downloads\joern-cli\joern-cli\workspace\eslint-plugin-react\cpg.bin.tmp
Code successfully imported. You can now query it using `cpg`.
For an overview of all imported code, type `workspace`.
Adding default overlays to base CPG
The graph has been modified. You may want to use the `save` command to persist changes to disk.  All changes will also be saved collectively on exit
2023-12-07 10:08:27.242 ERROR CpgPassBase: Pass io.joern.dataflowengineoss.passes.reachingdef.ReachingDefPass failed
overflowdb.SchemaViolationException: OUT edge with label REF to an adjacent METHOD is mandatory, but not defined for this METHOD_REF node with id=30852
        at io.shiftleft.codepropertygraph.generated.nodes.MethodRefDb.referencedMethod(MethodRef.scala:869) ~[io.shiftleft.codepropertygraph-domain-classes_3-1.4.25.jar:1.4.25]
        at io.shiftleft.codepropertygraph.generated.nodes.MethodRef.referencedMethod(MethodRef.scala:399) ~[io.shiftleft.codepropertygraph-domain-classes_3-1.4.25.jar:1.4.25]
        at io.shiftleft.codepropertygraph.generated.traversal.MethodRefTraversalExtGen$.referencedMethod$extension$$anonfun$1(MethodRef.scala:14) ~[io.shiftleft.codepropertygraph-domain-classes_3-1.4.25.jar:1.4.25]
        at scala.collection.Iterator$$anon$9.next(Iterator.scala:584) ~[org.scala-lang.scala-library-2.13.10.jar:?]
        at scala.collection.Iterator$$anon$10.nextCur(Iterator.scala:594) ~[org.scala-lang.scala-library-2.13.10.jar:?]
        at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:608) ~[org.scala-lang.scala-library-2.13.10.jar:?]
        at scala.collection.Iterator$$anon$6.hasNext(Iterator.scala:480) ~[org.scala-lang.scala-library-2.13.10.jar:?]
[...]

PR #3915, opened 2023-12-07
[jssrc2cpg] Fixed ObjectMethod naming and TypeNodePass performance

max-leuthaeuser:
Fixes: https://github.com/joernio/joern/issues/3901 (both the REF errors and the performance)

For whatever reason the `lambdaTypeRegex` in the TypeNodePass hangs when evaluating certain JS/TS types. So I have overridden the default there and this fixes it.


maltek (approved):
completely untested, but I think something like `raw"[^:]*\.([^.:]*):[^(]*\(.*\)".r` for `lambdaTypeRegex` would trigger a lot less backtracking while (hopefully) matching on the same inputs

But not running a regex that won't match anything for those languages will be even faster :)

PR #3954, opened 2023-12-13
[gosrc2cpg] - fix for package-level global variables from dependencies.

pandurangpatil:
1. It's been observed that the AST nodes were getting created for package-level global variables and constants from the dependencies code. As it was getting processed through the same ASTCreators code. Made changes to handle this situation.
2. Also identified that the variables, constants, types and functions from dependencies which starts with lower case letter will not be referred from the main source code. Hence, I ignored the processing for those. This will also reduce the memory footprint as we are not storing that metadata in the cache. This will also speed up the processing
3. With a lesser memory footprint, unit tests with download dependencies should work. Hence enabled those unit tests. These tests will be helpful in identifying some breaking changes. Like the other ignore test in the same file, was working before. With some recent changes, it has started breaking that flow.

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/GoSrc2Cpg.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/GoSrc2Cpg.scala
@@ -36,7 +36,9 @@ class GoSrc2Cpg extends X2CpgFrontend[Config] {
           astGenResult.parsedModFile.flatMap(modFile => GoAstJsonParser.readModFile(Paths.get(modFile)).map(x => x))
         )
         if (config.fetchDependencies) {
+          goGlobal.processingDependencies = true
           new DownloadDependenciesPass(goMod, goGlobal).process()
+          goGlobal.processingDependencies = false
         }
         val astCreators =
           new MethodAndTypeCacheBuilderPass(Some(cpg), astGenResult.parsedFiles, config, goMod, goGlobal).process()
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/CacheBuilder.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/CacheBuilder.scala
@@ -1,7 +1,6 @@
 package io.joern.gosrc2cpg.astcreation
 
-import io.joern.gosrc2cpg.datastructures.GoGlobal
-import io.joern.gosrc2cpg.parser.ParserAst.{FuncType, GenDecl, InterfaceType, StructType, ValueSpec}
+import io.joern.gosrc2cpg.parser.ParserAst.*
 import io.joern.gosrc2cpg.parser.{ParserKeys, ParserNodeInfo}
 import io.joern.gosrc2cpg.utils.UtilityConstants.fileSeparateorPattern
 import io.joern.x2cpg.{Ast, ValidationMode}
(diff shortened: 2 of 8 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/GoSrc2Cpg.scala
```diff
@@ -36,7 +36,9 @@ class GoSrc2Cpg extends X2CpgFrontend[Config] {
           astGenResult.parsedModFile.flatMap(modFile => GoAstJsonParser.readModFile(Paths.get(modFile)).map(x => x))
         )
         if (config.fetchDependencies) {
+          goGlobal.processingDependencies = true
```
DavidBakerEffendi, line 39: A suggestion in general, but if `goGlobal` was a case class, you could simply pass a deep copy as follows:
```scala
new DownloadDependenciesPass(goMod, goGlobal.copy(processingDependencies = true)).process()
```

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/CacheBuilder.scala
```diff
@@ -120,19 +119,25 @@ trait CacheBuilder(implicit withSchemaValidation: ValidationMode) { this: AstCre
...
-      case InterfaceType => Seq.empty
-      // astForStructType() function will record the member types
-      case StructType => astForStructType(typeNode, fullName)
-      // Process lambda function types to record lambda function signature mapped to TypeFullName
-      case FuncType => processFuncType(typeNode, fullName)
-      case _        => Seq.empty
-    }
-    (name, fullName, ast)
+    val name = typeSepc(ParserKeys.Name)(ParserKeys.Name).str
+    if (!goGlobal.processingDependencies || goGlobal.processingDependencies && name.headOption.exists(_.isUpper)) {
```
DavidBakerEffendi, line 123: This check is duplicated 3 times, I wonder if you could potentially make a function of this explaining the check it's looking for in one place?

DavidBakerEffendi: I see it's duplicated more below, I think it should definitely be a function

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/CacheBuilder.scala
```diff
@@ -148,22 +153,28 @@ trait CacheBuilder(implicit withSchemaValidation: ValidationMode) { this: AstCre
...
+      val (returnTypeStr, _) =
+        getReturnType(funcDeclVal(ParserKeys.Type), genericTypeMethodMap).headOption
+          .getOrElse(("", null))
+      val params = funcDeclVal(ParserKeys.Type)(ParserKeys.Params)(ParserKeys.List)
+      val signature =
+        s"$methodFullname(${parameterSignature(params, genericTypeMethodMap)})$returnTypeStr"
+      goGlobal.recordFullNameToReturnType(methodFullname, returnTypeStr, signature)
+      (name, methodFullname, signature, params, receiverInfo, genericTypeMethodMap)
+    } else
+      ("", "", "", Value("{}"), None, Map())
```
DavidBakerEffendi, line 178: When tuples start to get this long, I wonder if this shouldn't rather be a case class

DavidBakerEffendi (commented):
Some general refactoring related comments

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/GoSrc2Cpg.scala
pandurangpatil: Earlier goGlobal was a singleton object converted to a class object in order to not clear the state. If I pass the copy of the same, the cache built in this pass will not be useful in further passes.

DavidBakerEffendi: OK I see, makes sense.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/CacheBuilder.scala
pandurangpatil: done

DavidBakerEffendi (approved):

Issue #3927, opened 2023-12-11
[ruby] `Begin`, `Rescue`, `Ensure` (Try-Catch Statements)


DavidBakerEffendi:
A laymans example can be found here: https://stackoverflow.com/questions/2191632/begin-rescue-and-ensure-in-ruby

These are simply try-catch-finally statements, and there should be an attempt at that in the deprecated `AstForControlStructuresCreator`.

Java and C frontends may have better examples on this implementation. For these structures, the `index` of the blocks representing `Begin`, `Rescue`, and `Ensure` are important for CFG creation.

PR #3970, opened 2023-12-18
[ruby] Begin, Rescue, Ensure (Try-Catch Statements) for intermediate ast #3927

badly-drawn-wizards:
The related issue assumes the RubyNode intermediate AST is complete for Begin, Rescue, Ensure, but this is not the case. The scope of this issue has crept to include handling the ANTLR-to-RubyNode AST feature for these expressions, as well as the AstCreator CPG building. This appears to be the case for a few of the other Ruby issues, which will influence their ETA accordingly.

Additionally, Rescue clauses in Ruby can contain splats which isn't handled yet in parsing. For now the exception list is will be an unknown block and posted as a medium-priority issue. I am unsure if adding a RubyNode for multipleRightHandSide affects anything else.

The scope of this PR is reduced to just the intermediate AST and the AST creator changes will be in a follow up.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/RubyIntermediateAst.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/RubyIntermediateAst.scala
@@ -22,6 +22,12 @@ object RubyIntermediateAst {
     def text: String = span.text
   }
 
+  implicit class RubyNodeHelper(node: RubyNode) {
+    def asStatementList = node match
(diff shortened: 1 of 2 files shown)


badly-drawn-wizards:
An open question for myself is that it that currently the same RubyNode can result in different AST generation depending on whether it is processed by `astForExpression` or `astForStatement`. For example
```ruby
if true then end
```
results in a ControlFlowNode
and
```ruby
x = if true then end
```
result in a ternary operation

This leads to the question of why this is the case and whether rescue blocks should follow the same logic or if all control flow can be treated uniformly.

DavidBakerEffendi (approved):

PR #4009, opened 2023-12-22
Utility for General Purpose Parallelism

DavidBakerEffendi:
Following the type recovery generator parallelism fix, as well as noting that faulty parallelism exists elsewhere such as [in the slicing](https://github.com/joernio/joern/blob/38ba540a93c10e82b0550569353f4d84d131e898/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/slicing/DataFlowSlicing.scala#L23), I've created this general purpose concurrency tool as an easy means to bootstrap effective concurrency for execution order independent tasks.


DavidBakerEffendi:
Related: https://github.com/joernio/joern/pull/4010

max-leuthaeuser (approved):

max-leuthaeuser:
Please let @bbrehm double check on these two PRs before merging.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ConcurrentTaskExecutionUtil.scala
```diff
@@ -0,0 +1,54 @@
...
+      } else if (completionQueue.nonEmpty) {
+        val future = completionQueue.removeHead()
+        val res    = Try(Await.result(future, Duration.Inf))
+        results.append(res)
+      } else {
+        done = true
+      }
+    }
+
+    results.toList
```
mpollmeier, line 51:
```suggestion
    val futures = tasks.map(task => Future(task())).toSeq
    // starts them all ^

    futures.map { future =>
      Try(Await.result(future, Duration.Inf))
    }
```

mpollmeier: doesn't this do the same with less noise?

mpollmeier (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ConcurrentTaskExecutionUtil.scala
DavidBakerEffendi: The point of the queue limits how many tasks are loaded into memory - which is also helped along with the use of the iterator parameter. Not sure, on the other hand, if it makes that much of an impact using this pattern outside of CPG writing to disk.

maltek: I can offer this third way of doing things from one of my joern scripts:

```scala

def runInParallel[V](it: Iterator[() => V]): mutable.Buffer[V] = {
    StreamSupport
      .stream(Spliterators.spliteratorUnknownSize(it.asJava, Spliterator.NONNULL), /* parallel */ true)
      .map(_.apply())
      .toList
      .asScala
  }
```

I haven't benchmarked it though - I just assumed the Java streams are optimized well enough (including lazy reading from the stream/iterator)

mpollmeier: I guess both @maltek and my suggestions ignore the fact that @DavidBakerEffendi wants to run `x` tasks in parallel - our suggestions both start them all eagerly. 

So I guess it's best to just ignore us. 
Final thought: doesn't ExecutorService with a fixed threadpool do that as well?
But again, feel free to ignore, we didn't mean to hold you up, it was just a thought.

DavidBakerEffendi: Got back from vacation today, I'll see if I can use @maltek's solution can be specified with a fixed `x` size as it looks much cleaner and Java's way of doing things are likely more battle-tested.

Yeah, my idea was that I don't want to pollute the memory too much for tasks which may produce unpredictable memory demands. Let me see if I can produce a cleaner variant that addresses this, and maybe reflect the intention of the method in the name.

DavidBakerEffendi: Seems like a [ThreadPoolExecutor](https://docs.oracle.com/javase/6/docs/api/java/util/concurrent/ThreadPoolExecutor.html) is what I want, should limit the number of running tasks.

mpollmeier: yes that's what I had in mind re ExecutorService

you get bonus points if you wrap it in a scala.util.Using - that'll ensure that the executorservice is always properly closed, similar to java's try-with-resource

DavidBakerEffendi:
Given all the feedback, I've added two variants: One with a fixed size pool, and one that uses @maltek's suggestion of a spliterator.

Note that the executor service doesn't play nice with `Autocloseable` given the `shutdown` and `shutdownNow` functions make closing ambiguous, so I've used a try-finally here with a call to `shutdown`.

If this is still a meaningful utility, we can then merge it and slowly refer to it as a common source of implementing correct concurrency. This is mostly due to observing the pitfalls of my previous shortcomings in implementing concurrent execution, and avoiding this happening to others.

mpollmeier (approved):

max-leuthaeuser:
Just a small side note:
These unit tests will fail on machines where `Runtime.getRuntime.availableProcessors()` returns 1.
This can be the case for virtual environments or local/physical developer machines running on battery or in certain energy savings modes.
(maybe we can tag the test to be run only if we got > 1 available processors)

PR #4018, opened 2024-01-02
Add <init> naming for ruby constructors

badly-drawn-wizards:
- Set the name of method nodes of ruby initialize methods whose parent method ast is a TYPE_DECL to be `<init>`.
- Also set the full name AST accordingly
- Create a default `<init>` if no constructor is present
  This is done pretty jankily, as I have added a stack  of boolean flags of whether the immediate TypeDecl contains a constructor. I'd love to know if there is a better way.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
@@ -18,10 +18,14 @@ trait AstCreatorHelper { this: AstCreator =>
    * */
   protected val methodAstParentStack: Stack[NewNode] = new Stack()
 
-  /* Use to if any constructors of classes are present to know if a default constructor should be generated
+  /* Used if any constructors of classes are present to know if a default constructor should be generated
    * TODO: this seems too specific to add another stack, perhaps there is a better way in checking the class body. There are some possible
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -18,6 +18,11 @@ trait AstCreatorHelper { this: AstCreator =>
    * */
   protected val methodAstParentStack: Stack[NewNode] = new Stack()
 
+  /* Use to if any constructors of classes are present to know if a default constructor should be generated
```
fabsx00, line 21: `Use to ... to know` - two "to"s?

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -18,6 +18,11 @@ trait AstCreatorHelper { this: AstCreator =>
...
+   * TODO: this seems too specific to add another stack, perhaps there is a better way in checking the class body. There are some possible
```
fabsx00, line 22: It may be possible to use one context-stack by defining a context-class used as a type for stack elements. Not sure if this makes sense here, but that's something that has worked well in the past.

fabsx00 (approved):

PR #4022, opened 2024-01-04
[csharpsrc2cpg] Handle basic operators

karan-batavia:
This PR adds support for operators (arithmetic, logical, comparison, and bitwise) in the `csharpsrc2cpg` frontend.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ExternalCommand.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ExternalCommand.scala
@@ -10,7 +10,7 @@ object ExternalCommand {
 
   private val IS_WIN: Boolean = scala.util.Properties.isWin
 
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ExternalCommand.scala
```diff
@@ -10,7 +10,7 @@ object ExternalCommand {
 
   private val IS_WIN: Boolean = scala.util.Properties.isWin
 
-  private val shellPrefix: Seq[String] = if (IS_WIN) "cmd" :: "/c" :: Nil else "sh" :: "-c" :: Nil
```
max-leuthaeuser, line 13: Why is that? This change will affect all frontends running external programs so I just want to understand that :-)

karan-batavia: Oops! Committed this by mistake, will push a fix commit. Thanks.

max-leuthaeuser (approved):

DavidBakerEffendi (approved):

Issue #3926, opened 2023-12-11
[ruby] `when` Statements (Switch Statements)


DavidBakerEffendi:
A nice resource for understanding this is here: https://www.rubyguides.com/2015/10/ruby-case/

Note that the `default` case is called `else` in Ruby, but the `CfgCreator` requires the `name` property to be `default`.

The older modelling for this can be found in the `deprecated` frontend under `AstForControlStructuresCreator`.

PR #4029, opened 2024-01-08
[ruby] when Statements (Switch Statements) #3926

badly-drawn-wizards:
This draft translates ruby `case` expressions into `if-elif-...-else-end` chains. 

The expression matched against is assigned a temporary variable `<tmp-#>` where # is a number. I was not sure on the convention but I wanted to be sure the variable is always fresh.

A single when can contain a list of expressions to match against, which I translate into an `or-expression` if there is more than one. When matching against an expression `mExpr`, it is turned into a condition with `mExpr.=== <tmp-#>`.

This list of expressions can contain a splat at the end, which we aren't handling yet.

It also remains maybe to special case to a switch ast if all match expressions are literals.

TextSpans for the generated intermediate ast are not sane and will need to be considered more carefully.

Sanity checking, unit tests and more documentation remain to be done.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
@@ -14,6 +14,7 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
     case node: UntilExpression            => astForUntilStatement(node) :: Nil
     case node: IfExpression               => astForIfStatement(node) :: Nil
     case node: UnlessExpression           => astForUnlessStatement(node) :: Nil
+    case node: CaseExpression             => astsForCaseExpression(node)
     case node: StatementList              => astForStatementList(node) :: Nil
     case node: SimpleCallWithBlock        => astForSimpleCallWithBlock(node) :: Nil
     case node: MemberCallWithBlock        => astForMemberCallWithBlock(node) :: Nil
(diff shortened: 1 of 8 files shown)


DavidBakerEffendi (commented):
Looking good so far

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -100,6 +101,46 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
...
+        (whenClause: WhenClause, restClause: Option[RubyNode]) =>
+          // We translate multiple match expressions into an or expression.
+          // There may be a splat as the last match expression, which is currently parsed as unknown
+          // A single match expression is compared using `.===` to case target expression if it is present
+          // otherwise it is treated as a conditional.
+          val conditions = whenClause.matchExpressions.map {
+            case u: Unknown => u
+            case mExpr      => expr.map(e => MemberCall(e, ".", "===", List(mExpr))(mExpr.span)).getOrElse(mExpr)
+          }
+          // There is always at least one match expression
```
badly-drawn-wizards, line 118: I realised that there could be a single splat, I should handle that.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -100,6 +101,46 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
     controlStructureAst(ifNode, Some(notConditionAst), thenAst :: elseAsts)
   }
 
+  protected def astsForCaseExpression(node: CaseExpression): Seq[Ast] = {
+    def goCase(expr: Option[SimpleIdentifier]): List[RubyNode] = {
+      val elseThenClause: Option[RubyNode] = node.elseClause.map(_.asInstanceOf[ElseClause].thenClause)
+      val whenClauses                      = node.whenClauses.map(_.asInstanceOf[WhenClause])
+      val ifElseChain = whenClauses.foldRight[Option[RubyNode]](elseThenClause) {
```
badly-drawn-wizards, line 108: I only saw later that there is already a fold in the IfExpression handling, I will switch to that instead.

xavierpinho: Just a quick note about that fold - it was created when there was still no RubyNode and we were operating on top of ANTLR contexts. With dedicated RubyNodes for each construct, it might be more confusing than useful/elegant at times, so you are the better judge.

badly-drawn-wizards: Yeah, I later saw later I might have to do some weird special casing if I went that route for if vs elif, so this fold might be cleaner in this case. Just don't want to create a special snowflake in the the code.

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/CaseTests.scala
```diff
@@ -0,0 +1,91 @@
...
+    val List(lhs, rhs) = assign.argument.l
+
+    List(lhs).isCall.name.l shouldBe List("<tmp-0>")
+    List(rhs).isLiteral.code.l shouldBe List("0")
+
+    val headIf @ List(_)           = block.astChildren.isControlStructure.l
+    val ifStmts @ List(_, _, _, _) = headIf.repeat(_.astChildren.order(3).astChildren.isControlStructure)(_.emit).l;
+    val conds: List[List[String]] = ifStmts.condition.map { cond =>
+      val orConds = List(cond)
+        .repeat(_.isCall.where(_.name("<operator>.logicalOr")).argument)(
```
DavidBakerEffendi, line 33:
```suggestion
        .repeat(_.isCall.where(_.name(Operators.logicalOr)).argument)(
```

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/CaseTests.scala
```diff
@@ -0,0 +1,91 @@
...
+          _.emit(_.whereNot(_.isCall.name("<operator>.logicalOr")))
```
DavidBakerEffendi, line 34:
```suggestion
          _.emit(_.whereNot(_.isCall.name(Operators.logicalOr)))
```

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/CaseTests.scala
```diff
@@ -0,0 +1,91 @@
...
+      |end
+      |""".stripMargin)
+
+    val block @ List(_) = cpg.method(":program").block.astChildren.isBlock.l
+
+    val headIf @ List(_)           = block.astChildren.isControlStructure.l
+    val ifStmts @ List(_, _, _, _) = headIf.repeat(_.astChildren.order(3).astChildren.isControlStructure)(_.emit).l;
+    val conds: List[List[String]] = ifStmts.condition.map { cond =>
+      val orConds = List(cond)
+        .repeat(_.isCall.where(_.name("<operator>.logicalOr")).argument)(
```
DavidBakerEffendi, line 75:
```suggestion
        .repeat(_.isCall.where(_.name(Operators.logicalOr)).argument)(
```

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/CaseTests.scala
```diff
@@ -0,0 +1,91 @@
...
+          _.emit(_.whereNot(_.isCall.name("<operator>.logicalOr")))
```
DavidBakerEffendi, line 76:
```suggestion
          _.emit(_.whereNot(_.isCall.name(Operators.logicalOr)))
```

DavidBakerEffendi (approved):
Just a few suggestions, but otherwise we can ship it

PR #4036, opened 2024-01-10
Remove Flaky ConcurrentTaskUtilTests

DavidBakerEffendi:
These tests are too unpredictable to consistently pass, so they are to be removed.


max-leuthaeuser (approved):

max-leuthaeuser:
Benchmarking with nanotime on the JVM is not a good idea anyway: https://shipilev.net/blog/2014/nanotrusting-nanotime/

If we ever want to do somehing like this again, we better use JMH:
- https://github.com/openjdk/jmh
- https://github.com/sbt/sbt-jmh
(also usefull for periodically test for performance regressions in frontends)

DavidBakerEffendi:
Good idea for the long term - especially if a bad query ends up slowing down post-processing

PR #4045, opened 2024-01-12
Properly handle python sub module imports.

ml86:
Hierarchical imports like 'import a.b' are now represented correctly.

Do not yet merge.

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/MethodCpgTests.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/MethodCpgTests.scala
@@ -1,0 +1,23 @@
+package io.joern.pysrc2cpg.cpg
+
+import io.joern.pysrc2cpg.Py2CpgTestContext
+import io.shiftleft.semanticcpg.language._
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.should.Matchers
+
+class MethodCpgTests extends AnyFreeSpec with Matchers {
+  "A method" - {
+
+    lazy val cpg = Py2CpgTestContext.buildCpg(
+      """
+        |def method():
+        |   pass
+        |""".stripMargin, "a/b.py")
+
+    "foo" in {
+      val method = cpg.method.name("method").head
+      method.fullName shouldBe "a/b.py:<module>.method"
+    }
+  }
+
+}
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/MethodCpgTests.scala
maltek, line 1: looks like a leftover/unfinished file?

maltek (approved):

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/MethodCpgTests.scala
ml86: No, i wanted an extra test for the full name of a method. But did not give a proper name.

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/MethodCpgTests.scala
```diff
@@ -0,0 +1,25 @@
...
+
+class MethodCpgTests extends AnyFreeSpec with Matchers {
+  "A method" - {
+
+    lazy val cpg = Py2CpgTestContext.buildCpg(
+      """
+        |def method():
+        |   pass
+        |""".stripMargin,
+      "a/b.py"
```
DavidBakerEffendi, line 16: Should this not be something like `Seq("a", "b.py").mkString(File.separator)`

DavidBakerEffendi (commented):
In terms of the type recovery test expectations, could we simply not use `ignore` for now? Not sure if I want to merge this before I can make the appropriate change to keep the type recovery tests consistent.

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/MethodCpgTests.scala
ml86: It should not because the OS file handling layer will correctly handle the slash as file separator even on windows.

DavidBakerEffendi (approved):
After chatting offline, I'm happy to roll back my 2 commits and keep the tests changed. The semantics are still modelled fine.

PR #4065, opened 2024-01-18
Create CITATION.cff

DavidBakerEffendi:
Since Joern is used in research, and commonly cited, it would be good to have a single preferred way to cite the project. GitHub has a [CITATION.cff](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files) mechanism where we can provide this. The release date on the citation is consistent with the [v1.0.1 release](https://github.com/joernio/joern/releases/tag/v1.0.1).

Question: Do we add authors? Which authors qualify? e.g. some commit number threshold on the commit threshold/feature impact?

A live example can be found on the [SootUp](https://github.com/soot-oss/SootUp) repository.


mpollmeier (approved):
For my understanding, and maybe the readme: should the version and release date be updated regularly?

And shouldn't the release date correspond to the version number? 2.0.225 was released yesterday. Or is it the initial incarnation of the project?

DavidBakerEffendi:
@mpollmeier Good point on the release date, I'll have a quick look around on what existing GitHub Actions mechanisms there are to do edit these files are. Seems we can get the date as follows: https://stackoverflow.com/a/60942437 and then commit the new citation file with https://github.com/marketplace/actions/update-files-on-github

DavidBakerEffendi:
@mpollmeier The automation is pushed, but my only concern is that pushing to `master` directly will likely be rejected unless there is a rule to allow `github-actions[bot]@users.noreply.github.com` to push directly.

mpollmeier:
Hmm, `GITHUB_TOKEN` is just for authentication, the main hurdle here is the pushing to master, as you mentioned. And we don't really want to lift that rule. 
It would also pollute the git history if there's a separate commit for the citation file every day..

How important is it to always have it on the latest? If we do it on a manual basis (i.e. revert the auto-updater and just add some comments in the citation and/or readme with the context and how to update it), that should be enough for now..?

DavidBakerEffendi:
@mpollmeier Agreed, let's do that. I'll just add the current major and minor version with the latest date, and we add instructions on the message in the `CITATION.cff`

PR #4068, opened 2024-01-18
[frontends] Remove shell layer for running external commands

max-leuthaeuser:


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/utils/ExternalCommand.scala
```diff
@@ -1,30 +1,26 @@
...
-            IncludeAutoDiscovery.gccAvailable() =>
-        Success(result.toSeq)
+    val stdOutOutput  = new ConcurrentLinkedQueue[String]
+    val processLogger = ProcessLogger(stdOutOutput.add, stdOutOutput.add)
+    Try(Process(command).!(processLogger)) match {
+      case Success(0) =>
+        Success(stdOutOutput.asScala.toSeq)
+      case Success(1)
+          if IS_WIN && command != IncludeAutoDiscovery.GCC_VERSION_COMMAND && IncludeAutoDiscovery.gccAvailable() =>
+        Success(stdOutOutput.asScala.toSeq)
```
mpollmeier, line 20: I know you didn't change this, but it might be worth adding a comment if one of you knows: 
Looks like there are cases on windows where the external commands returns `1` and that's a success? My understanding so far was that also for windows, anything other than `0` is a failure..?

mpollmeier (approved):

Ferada (commented):
sgtm, only I can't test the Windows portion

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/utils/ExternalCommand.scala
max-leuthaeuser: The command to query the system header file locations within a Windows environment always returns Success(1) for whatever reason. I was not able to figure out why that is a thing.

mpollmeier, line 20:
```suggestion
        // the command to query the system header file locations within a Windows
        // environment always returns Success(1) for whatever reason...
        Success(stdOutOutput.asScala.toSeq)
```

mpollmeier: 👍🏻 thanks! For future readers we could add that as a comment (suggested below)..

mpollmeier: Also note that future readers is most likely one of us 😆

max-leuthaeuser:
I can't get this running for php2cpg. Looks like the Shell layer is required for the php parser to correctly handle newlines. Otherwise, parsing fails non-deterministically.

Who is in charge of php2cpg currently? Maybe that person should have a look later on.

gacevedo:
> I can't get this running for php2cpg. Looks like the Shell layer is required for the php parser to correctly handle newlines. Otherwise, parsing fails non-deterministically.
> 
> Who is in charge of php2cpg currently? Maybe that person should have a look later on.

@johannescoetzee  ^^

max-leuthaeuser:
@johannescoetzee The latest commit here got it running for php I think.
Using separate handlers for std out/err (like we did before) did the trick. Looks like these php parse errors did exist previously but never got caught because they landed in a different std err handler. It might be a good idea to investigate why parsing the php test code snippets sometime errors out.

PR #4069, opened 2024-01-18
Validate CITATION.cff

DavidBakerEffendi:
GitHub apparently can't parse the CITATION.cff file and gives no feedback. This GitHub action will give us more insight.


DavidBakerEffendi:
It appears that we needed an author field

mpollmeier (approved):

PR #4079, opened 2024-01-23
[swiftsrc2cpg] Added support for ForEach loops

max-leuthaeuser:
Incl. de-sugaring for ForEach loops with simple identifier, expressions, and tuple declarations with and without where clauses.

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForStmtSyntaxCreator.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForStmtSyntaxCreator.scala
@@ -5,8 +5,12 @@ import io.joern.swiftsrc2cpg.parser.SwiftNodeSyntax.*
 import io.joern.x2cpg.Ast
 import io.joern.x2cpg.ValidationMode
 import io.joern.x2cpg.datastructures.Stack.*
+import io.joern.x2cpg.Defines
 import io.shiftleft.codepropertygraph.generated.ControlStructureTypes
 import io.shiftleft.codepropertygraph.generated.nodes.NewJumpLabel
+import io.shiftleft.codepropertygraph.generated.DispatchTypes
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForStmtSyntaxCreator.scala
```diff
@@ -73,7 +77,466 @@ trait AstForStmtSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
...
+    // _iterator assignment:
+    val iteratorName      = generateUnusedVariableName(usedVariableNames, "_iterator")
+    val iteratorLocalNode = localNode(node, iteratorName, iteratorName, Defines.Any).order(0)
+    val iteratorNode      = identifierNode(node, iteratorName)
+    diffGraph.addEdge(localAstParentStack.head, iteratorLocalNode, EdgeTypes.AST)
+    scope.addVariableReference(iteratorName, iteratorNode)
+
+    val iteratorCall =
+      // TODO: add operator to schema
+      callNode(node, s"<operator>.iterator($collectionName)", "<operator>.iterator", DispatchTypes.STATIC_DISPATCH)
```
johannescoetzee, line 140: Note for the future: If we're adding support for an `iterator` operator, then I think it would make sense to use that across frontends through a common `foreach` lowering in `x2cpg`. This would be a fairly significant representation change for existing frontends, however.

johannescoetzee (approved):

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForStmtSyntaxCreator.scala
max-leuthaeuser: Yes, true. Currently, its done this way in js2cpg, jssrc2cpg, and afaik pysrc2cpg.
This should be refactored at some point.

PR #4082, opened 2024-01-24
Add content steps for METHOD and TYPE_DECL nodes.

ml86:
With these new steps the parts of FILE content belonging to a TYPE_DECL
or METHOD can be easily accessed.


maltek (approved):
There is some overlap with the existing `.dump` steps. But I think a new step makes sense since `.dump` always does fancy syntax-highlighting stuff, while this new step doesn't and shouldn't for the use-case that inspired it. (We should probably start making use of this in `.dump` when available, but I'd say that can happen later.)

PR #4084, opened 2024-01-24
[c#] Fix RHS of EqualsValueClause

### LINKED ISSUE #4078: [Bug] [c#] Simple Aliased DataFlows Not Working
**Describe the bug**
Trying out a simple dataflow test in the C# frontend. Add a dataflow test with the code below to reproduce. 

```
int i = 0;
int a = i + 1;
Console.WriteLine(a);
```

For the query, 
```
val src  = cpg.identifier.nameExact("i").l
val sink = cpg.call.nameExact("WriteLine").l
```
The expected output for `sink.reachableBy(src).size` should be `2`, but is `0` in this case.

DavidBakerEffendi:
Simple data-flows were not being propagated after some re-assignment. This was found to be caused by only a limited class of RHS nodes being handled on the EqualsValueClause. This has now been fixed and tested.

Other style-related changes added.

Resolves #4078

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/dataflow/SimpleDataflowTests.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/dataflow/SimpleDataflowTests.scala
@@ -1,0 +1,26 @@
+package io.joern.csharpsrc2cpg.querying.dataflow
+
+import io.joern.csharpsrc2cpg.testfixtures.CSharpCode2CpgFixture
+import io.shiftleft.semanticcpg.language.*
+import io.joern.dataflowengineoss.language.*
+
+class SimpleDataflowTests extends CSharpCode2CpgFixture(withDataFlow = true) {
+
+  "a source with a simple re-assignment" should {
+
+    val cpg = code(basicBoilerplate("""
+        |int i = 0;
+        |int a = i + 1;
+        |Console.WriteLine(a);
+        |""".stripMargin))
+
+    "still propagate to the sink" in {
+      cpg.method("Main").dotAst.foreach(println)
+      val src  = cpg.assignment.target.isIdentifier.nameExact("i").l
+      val sink = cpg.call.nameExact("WriteLine").l
+      sink.reachableBy(src).p.foreach(println)
+    }
+
+  }
+
+}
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/dataflow/SimpleDataflowTests.scala
```diff
@@ -0,0 +1,26 @@
...
+        |int i = 0;
+        |int a = i + 1;
+        |Console.WriteLine(a);
+        |""".stripMargin))
+
+    "still propagate to the sink" in {
+      cpg.method("Main").dotAst.foreach(println)
+      val src  = cpg.assignment.target.isIdentifier.nameExact("i").l
+      val sink = cpg.call.nameExact("WriteLine").l
+      sink.reachableBy(src).p.foreach(println)
```
karan-batavia, line 21: Is this a debug statement? We might want to remove this.

joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/dataflow/SimpleDataflowTests.scala
```diff
@@ -0,0 +1,26 @@
...
+  "a source with a simple re-assignment" should {
+
+    val cpg = code(basicBoilerplate("""
+        |int i = 0;
+        |int a = i + 1;
+        |Console.WriteLine(a);
+        |""".stripMargin))
+
+    "still propagate to the sink" in {
+      cpg.method("Main").dotAst.foreach(println)
```
karan-batavia, line 18: Same for this. Also, this test does not assert anything; is this deliberate?

karan-batavia (approved):
Some questions, LGTM otherwise.

joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/dataflow/SimpleDataflowTests.scala
DavidBakerEffendi: Ah yes good catch

DavidBakerEffendi:
@karan-batavia Fixed the test assertions, forgot to assert and remove debugging

max-leuthaeuser (approved):

PR #4092, opened 2024-01-26
Allow multiple colon-separated inference jar paths

johannescoetzee:
Allow users to specify multiple inference jar paths with a `:` separated list. I chose `:` since it's the path separator on Linux and not a valid path character in Windows.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/Main.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/Main.scala
@@ -78,9 +78,9 @@ private object Frontend {
     import builder._
     OParser.sequence(
       programName("javasrc2cpg"),
-      opt[String]("inference-jar-paths")
-        .text(s"extra jars used only for type information")
-        .action((path, c) => c.withInferenceJarPaths(c.inferenceJarPaths + path)),
+      opt[Seq[String]]("inference-jar-paths")
+        .text(s"extra jars used only for type information (comma-separated list of paths)")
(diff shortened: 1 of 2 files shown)


maltek (approved):
Windows convention for $PATH / java class path is `;`, whereas UNIX convention is `:`. I think there is some potential for confusion here.

- we could use the OS convention (`System.getProperty("path.separator")`), but then be inconsistent ourselves / force our users to do things differently depending on OS
- we could just use `:` everywhere like this PR currently does, though Windows users could possibly be confused if they don't look closely enough and are used to semi-colons
- we could just support both `:` and `;` everywhere. Downside is that users can't specify files with `;` anymore. I think I like this the best.

Not the biggest of deals either way in the end.

johannescoetzee:
I considered using the OS convention as well, but thought that would cause more problems than it solves. I like your idea of using both, however, so updated the PR with that.

max-leuthaeuser:
Why not making it `opt[Seq[String]]`?
The argument parser lib we are using there then does the splitting for you (default its `,`).

johannescoetzee:
That is certainly an option as well, but in my opinion using `:` or `;` to be consistent with OS path separators (as well as java classpath separators) is worth the extra effort of doing the splitting ourselves. With that being said, if we're currently using comma separated paths anywhere in joern, then it would be better to stay consistent with that.

maltek:
Now that I think more about it, splitting on `:` doesn't work on Windows for absolute paths - that'll split after the drive letter.

I think that makes me lean towards commas like Max suggested.

johannescoetzee:
Good point. I'll change to that then.

max-leuthaeuser (approved):

PR #4096, opened 2024-01-29
ghidra 11.0

mpollmeier:


itsacoderepo (approved):

itsacoderepo:
Ghidra specific tests are passing on my machine

mpollmeier:
From the test output:
```
[info] io.joern.scanners.c.CredentialDropTests *** ABORTED ***
[info]   java.lang.InstantiationError: org.h2.mvstore.FileStore
```

Turns out they now also ship an h2 database, which collides with the one that overflowdb ships...

Context: since ghidra doesn't publish a jar, we replicate their build and publish it ourselves...

PR #4103, opened 2024-01-31
[php2cpg] Expose frontend command line option for builtins signatures file path

wunused:
The php2cpg frontend uses the SetKnownTypes path to set types for known function signatures, like builtin functions. Currently, it reads the known type signature information from a resource file. This makes it difficult to add new type signatures at runtime, or on a per-project basis.

This PR adds a command line option to the PHP frontend to provide a path to a type signatures file at runtime, so that it can be invoked as:

```
./joern-parse --language php target/ -o target.cpg --frontend-args --known-types-file=known_function_signatures.txt
```


wunused:
@DavidBakerEffendi this is related to the PHP type inference passes from earlier, does this look good to merge?

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/PhpSetKnownTypes.scala
```diff
@@ -23,17 +26,20 @@ case class KnownFunction(
   pTypes: Seq[Seq[String]] = Seq.empty
 )
 
+case class PhpSetKnownTypesConfig(knownTypesFilePath: Option[String] = None)
```
DavidBakerEffendi, line 29: Feel free to extend this on top of `TypeRecoveryPass` as I am currently doing something similar in `csharpsrc2cpg`

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/PhpSetKnownTypes.scala
```diff
@@ -74,3 +80,27 @@ class PhpSetKnownTypesPass(cpg: Cpg, knownTypesFile: Option[JFile] = None)
     if (types.size == 1) builder.setNodeProperty(n, PropertyNames.TYPE_FULL_NAME, types.head)
     else builder.setNodeProperty(n, PropertyNames.DYNAMIC_TYPE_HINT_FULL_NAME, types)
 }
+
+trait PhpSetKnownTypesParserConfig[R <: X2CpgConfig[R]] { this: R =>
```
DavidBakerEffendi, line 84: Similarly here, it can just be on top of `TypeRecoveryParserConfig`

DavidBakerEffendi (changes requested):

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/PhpSetKnownTypes.scala
wunused: Is there an example in the C# frontend that I should reference and match? I don't see it in `master`. 

I use both the `TypeRecoveryParserConfig` and the `PhpSetKnownTypesParserConfig` traits for the php2cpg.Config class in Main.scala; do you prefer that `PhpSetKnownTypesParserConfig` extends directly from `TypeRecoveryParserConfig`, rather than being used along side it as a trait?

DavidBakerEffendi: Nothing in C# yet, but I'm deserializing the object into `TypeMap` which you can find in C#. Right now it's pre-defined, but I'd like to extend it to user-defined paths like this too.

I mean that you can add the `knownTypesFilePath` property to `TypeRecoveryParserConfig` and get rid of `PhpSetKnownTypesParserConfig`.

DavidBakerEffendi: Actually, better yet, keep it and rename it to `TypeStubParserConfig` and then it can be re-used across frontends. It can be placed in the `io.joern.x2cpg` package next to `TypeRecoveryParserConfig` somewhere in `frontends`

wunused: Got it, thanks! That makes sense.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/PhpSetKnownTypes.scala
```diff
@@ -74,3 +79,17 @@ class PhpSetKnownTypesPass(cpg: Cpg, knownTypesFile: Option[JFile] = None)
     if (types.size == 1) builder.setNodeProperty(n, PropertyNames.TYPE_FULL_NAME, types.head)
     else builder.setNodeProperty(n, PropertyNames.DYNAMIC_TYPE_HINT_FULL_NAME, types)
 }
+
+object PhpSetKnownTypes {
+
+  def parserOptions[R <: X2CpgConfig[R] with TypeStubsParserConfig[R]]: OParser[_, R] = {
```
DavidBakerEffendi, line 85: Could we move this to `object TypeStubsParserConfig`? It seems language agnostic enough

DavidBakerEffendi (changes requested):

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/PhpSetKnownTypes.scala
wunused: Yes, makes sense. I've moved it, and changed the names of the PHP specific pass to match the TypeStubs convention for consistency.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/PhpTypeStubsParser.scala
```diff
@@ -27,14 +31,15 @@ case class KnownFunction(
   *
   * TODO: Need to handle variadic arguments.
   */
-class PhpSetKnownTypesPass(cpg: Cpg, knownTypesFile: Option[JFile] = None)
+class PhpTypeStubsParserPassGenerator(cpg: Cpg, config: XTypeStubsParserConfig = XTypeStubsParserConfig())
```
DavidBakerEffendi, line 34: The reason the `XTypeRecoveryPassGenerator` is named a generator is because it generates passes. This pass doesn't so it should stay `PhpSetKnownTypesPass`

DavidBakerEffendi (changes requested):
Last comment, then it should be approved

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/PhpTypeStubsParser.scala
wunused: Makes sense, fixed.

DavidBakerEffendi (approved):

PR #4105, opened 2024-02-01
[C#] Ast creation for jump statements

karan-batavia:
This PR includes,
1. AST Creation for `break`, `continue` and `goto` statements.
2. Corresponding tests.

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForStatementsCreator.scala
@@ -198,16 +198,16 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
       case BreakStatement    => Seq(Ast(controlStructureNode(jumpStmt, ControlStructureTypes.BREAK, jumpStmt.code)))
       case ContinueStatement => Seq(Ast(controlStructureNode(jumpStmt, ControlStructureTypes.CONTINUE, jumpStmt.code)))
       case GotoStatement     => astForGotoStatement(jumpStmt)
-      case _                 => Seq(Ast())
+      case _                 => Seq.empty
   }
 
   private def astForGotoStatement(gotoStmt: DotNetNodeInfo): Seq[Ast] = {
-    val idenitifierAst = Option(gotoStmt.json(ParserKeys.Expression)) match
+    val identifierAst = Option(gotoStmt.json(ParserKeys.Expression)) match
       case Some(value: ujson.Obj) => astForNode(createDotNetNodeInfo(value))
-      case _                      => Seq(Ast())
+      case _                      => Seq.empty
 
     val gotoAst = Ast(controlStructureNode(gotoStmt, ControlStructureTypes.GOTO, gotoStmt.code))
 
-    idenitifierAst :+ gotoAst
+    identifierAst :+ gotoAst
   }
 }
--- a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForStatementsCreator.scala
@@ -4,15 +4,19 @@ import io.circe.Json
 import io.joern.csharpsrc2cpg.parser.DotNetJsonAst.{
   BinaryExpr,
   Block,
+  BreakStatement,
   CasePatternSwitchLabel,
   CaseSwitchLabel,
+  ContinueStatement,
   DefaultSwitchLabel,
   DoStatement,
   ExpressionStatement,
   ForEachStatement,
   ForStatement,
   GlobalStatement,
+  GotoStatement,
   IfStatement,
+  JumpStatement,
   LiteralExpr,
   SwitchStatement,
   ThrowStatement,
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -188,4 +193,21 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
     astForNode(globalStatement.json(ParserKeys.Statement))
   }
 
+  private def astForJumpStatement(jumpStmt: DotNetNodeInfo): Seq[Ast] = {
+    jumpStmt.node match
+      case BreakStatement    => Seq(Ast(controlStructureNode(jumpStmt, ControlStructureTypes.BREAK, jumpStmt.code)))
+      case ContinueStatement => Seq(Ast(controlStructureNode(jumpStmt, ControlStructureTypes.CONTINUE, jumpStmt.code)))
+      case GotoStatement     => astForGotoStatement(jumpStmt)
+      case _                 => Seq(Ast())
```
DavidBakerEffendi, line 201: This can be just `Seq.empty`, same on line 207

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -188,4 +193,21 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
...
+  }
+
+  private def astForGotoStatement(gotoStmt: DotNetNodeInfo): Seq[Ast] = {
+    val idenitifierAst = Option(gotoStmt.json(ParserKeys.Expression)) match
```
DavidBakerEffendi, line 205: Typo on `idenitifierAst` -> `identifierAst`

joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/JumpStatementsTests.scala
```diff
@@ -0,0 +1,95 @@
...
+  "AST creation for jump statements" should {
+    "be correct for break" in {
+      val cpg = code(basicBoilerplate("""
+          |for (int i = 0; i < 10; i++) {
+          | break;
+          |}
+          |""".stripMargin))
+
+      inside(cpg.method("Main").controlStructure.controlStructureTypeExact(ControlStructureTypes.BREAK).l) {
+        case breakStatement :: Nil =>
```
DavidBakerEffendi, line 17: There should be a default case (`case _ =>`) with a fail message

DavidBakerEffendi: Similarly on the other tests

DavidBakerEffendi (changes requested):

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForStatementsCreator.scala
karan-batavia: Ah, good catch.

DavidBakerEffendi (approved):

joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/JumpStatementsTests.scala
```diff
@@ -0,0 +1,98 @@
...
+class JumpStatementsTests extends CSharpCode2CpgFixture {
+  "AST creation for jump statements" should {
+    "be correct for break" in {
+      val cpg = code(basicBoilerplate("""
+          |for (int i = 0; i < 10; i++) {
+          | break;
+          |}
+          |""".stripMargin))
+
+      inside(cpg.method("Main").controlStructure.controlStructureTypeExact(ControlStructureTypes.BREAK).l) {
```
pandurangpatil, line 16: @karan-batavia 
I would recommend not to use `inside` as you cannot debug from the IDE code inside the `inside` block. Whereas if you take the return list and extract the element out of it like in the below example that I tried to rewrite part of this test case.

 In fact, it makes the code more readable.

e.g.

```
val List(breakStatement) = cpg.method("Main").controlStructure.controlStructureTypeExact(ControlStructureTypes.BREAK).l
breakStatement.controlStructureType shouldBe ControlStructureTypes.BREAK
breakStatement.code shouldBe "break;"
breakStatement.lineNumber shouldBe Some(10)
breakStatement.columnNumber shouldBe Some(1)
```

cc: @DavidBakerEffendi @tuxology

karan-batavia: @pandurangpatil we have been following this conventions for all the tests right from AST Creation.

DavidBakerEffendi: I've had success with the IDE, and the reason I want to use this is since it encourages using `failure` with helpful error messages which is readable from CLI failures.

pandurangpatil: I noticed that but debugging the inside block is difficult. In go as well I was following in a similar fashion, in fact there would be some test cases following this style. However, I once got badly stuck while debugging the dataflow test. I spent almost half a day trying to understand why I was not able to monitor the variables inside the inside block. 

Then I removed the `inside` block and changed the code in the way shown in the example. I could debug it easily.

Going ahead you can follow the recommended structure.

tuxology: I did face some IDE issues. My middle ground for this is that for a breaking query (not just a failing test), I replicate the query above inside block and debug. Once its done, and test is fixed, I remove it and place it back in inside 😬

Issue #4057, opened 2024-01-16
Move IdentifierToCall node pass logic inside AST Creation flow itself.


pandurangpatil:
`IdentifierToCall` determines, after the AstCreation, if an identifier behaves like a parenthesis-less call, and if so, replaces that node with a similar call node. This is not ideal, as base passes may treat a node differently based on the type.

This issue suggests using the `PackageContext.packageTable.methodTableMap` given to the `AstCreator` to determine if an identifier shares a name with a known method, and immediately creates a call instead.

pandurangpatil:
@DavidBakerEffendi add more details in this ticket

DavidBakerEffendi:
Development at [dave/ruby/identifier-call-replacement](https://github.com/joernio/joern/tree/dave/ruby/identifier-call-replacement)

khemrajrathore:
Tried verifying against -  https://github.com/pay-rails/pay/blob/master/app/controllers/pay/webhooks/braintree_controller.rb#L9
we have `verified_event`  identifier, which is basically a call to the method `verified_event`
cc - @DavidBakerEffendi @pandurangpatil

PR #4111, opened 2024-02-01
[ruby] IdentifierToCall Pass Replacement

DavidBakerEffendi:
* Making use of package table & removed identifier to call pass
* Moved parsing outside of ast creator to help enable a pre-parse
* If an identifier is encountered, and it matches the name of a pre-parsed method name in the package table, then a call node with the corresponding method full name is created
* Fixed behaviour of handling internal methods/modules
* Categorized IDE/OS gitignores
* Cleared package table after each run

Resolves [#4057](https://github.com/joernio/joern/issues/4057)


xavierpinho (approved):

DavidBakerEffendi:
Note @khemrajrathore @pandurangpatil, the reason this pipeline failed after the previous commit is because the package table was a global mutable object in the `RubySrc2Cpg` companion object. 

I've now forced it to clear after every `createCpg` call, so this may or may not affect downstream usages of the package table if it's used (i.e, it should be re-initialized before every CPG run since it's now used for storing application specific method/module data).

PR #4116, opened 2024-02-02
[C#] support record types

### LINKED ISSUE #3997: [c#] Record Types
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record

tuxology:
- AST creation for two variants of `record` types
 - Tests
 
 Resolves #3997

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForDeclarationsCreator.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForDeclarationsCreator.scala
@@ -10,6 +10,7 @@ import io.shiftleft.codepropertygraph.generated.nodes.*
 import io.shiftleft.proto.cpg.Cpg.EvaluationStrategies
 
 import scala.util.Try
+import scala.collection.mutable.ArrayBuffer
 import io.joern.csharpsrc2cpg.CSharpOperators as CSharpOperators
 import io.joern.x2cpg.Defines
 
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForDeclarationsCreator.scala
```diff
@@ -45,6 +46,47 @@ trait AstForDeclarationsCreator(implicit withSchemaValidation: ValidationMode) {
...
+    val modifiers         = astForModifiers(recordDecl)
+    val paramList         = recordDecl.json(ParserKeys.ParameterList)
+    var membersFromParams = Seq[Ast]()
+
+    // Covers the case where record type can be declared as `record Person(string Name);`
+    // Here, Person should be a TypeDecl and Name should be a member instead of a parameter
+    if (!paramList.isNull) {
+      membersFromParams = paramList(ParserKeys.Parameters).arr
+        .map(p => {
+          val metaData     = p(ParserKeys.MetaData)
```
DavidBakerEffendi, line 63: This should use `createDotNetNodeInfo` as it extracts all of the meta data information and sets the type for you.
You should the be able to do something like
```scala
paramList(ParserKeys.Parameters).arr
  .map(createDotNetNodeInfo)
  .map { paramNode => 
      val name = nameFromNode(paramNode) // the helper function for extracting identifier info
      val typeFullName = nodeTypeFullName(paramNode) // helper function for obtaining type info (uses type map)
      memberNode(paramNode, name, ... etc)
  }.toSeq
```

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForDeclarationsCreator.scala
```diff
@@ -45,6 +46,47 @@ trait AstForDeclarationsCreator(implicit withSchemaValidation: ValidationMode) {
...
+    val fullName = astFullName(recordDecl)
+    val typeDecl = typeDeclNode(recordDecl, name, fullName, relativeFileName, code(recordDecl))
+    scope.pushNewScope(TypeScope(fullName))
+    val modifiers         = astForModifiers(recordDecl)
+    val paramList         = recordDecl.json(ParserKeys.ParameterList)
+    var membersFromParams = Seq[Ast]()
+
+    // Covers the case where record type can be declared as `record Person(string Name);`
+    // Here, Person should be a TypeDecl and Name should be a member instead of a parameter
+    if (!paramList.isNull) {
```
DavidBakerEffendi, line 60: Could we perhaps instead do something like
```scala
val membersFromParams = Try { 
  recordDecl.json(ParserKeys.ParameterList)(ParserKeys.Parameters)
    .arr
    .map(createDotNetNodeInfo)
    .toSeq
  }
 .toOption
 .getOrElse(Seq.empty)
 .map { ... }
 ```
You can also try simply getting a return from the `if` statement instead of prompting an exception, but I would avoid using `var`

DavidBakerEffendi (changes requested):

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForDeclarationsCreator.scala
tuxology: Good point

DavidBakerEffendi (approved):

PR #4124, opened 2024-02-06
[c#] Create <clinit> Method for Fields Assigned to Defaults

### LINKED ISSUE #3982: [c#] Create `<clinit>` Method for Fields Assigned to Defaults
To model this similarly to Java, fields with static initializers or default values should have this assignment done in a special `<clinit>` method.

AndreiDreyer:
* Create constructor type for normal and static constructors
* Move field initialization into respective constructor blocks

TODO: Add constructor and static constructor blocks if it is not explicitly defined in the code being parsed.

Resolves #3982


DavidBakerEffendi (approved):
Looks good! Could you add the TODOs to issues here under the TODO column? https://github.com/orgs/joernio/projects/13

PR #4126, opened 2024-02-06
[swiftsrc2cpg] Added basic dataflow tests

max-leuthaeuser:


johannescoetzee (approved):
I didn't look at all the test expectations in detail, but it looks good overall

max-leuthaeuser:
They are a shameless (unchanged) ripoff from c2cpg (I just translated the test code snippets to swift) so that should be fine :-)

PR #4131, opened 2024-02-06
[x2cpg] Program Summary & Typed Scope

DavidBakerEffendi:
A few frontends each face the problems of:
* What types are available at this point in the program? How do I resolve a type alias to its full name?
* How do I deal with tracking possible method calls?
* How do I deal with conflicts if more than one type/method share the same name?
* Do I need to download my dependencies to provide the type meta-data?

This PR standardises the various structures found across frontends that try to deal with this by having a program summary and modified scope object.

The `ProgramSummary` gives enough bare-bones structure to describe the meta-data of high level interprocedurally scoped objects. The idea is that this is populated via a pre-parse before the `AstCreator` runs, and can be populated with pre-computed type stubs. The `TypedScope` trait allows one to keep track of which types are currently available at the current point of the program to help manage type references and aliases.

See the test for an illustration of this scoped type look-up behaviour.

The `TypeMap` and `CSharpScope` structures were pilot studies into this strategy and have proven effective. This structure tries to be fairly broad and flexible, and will be back ported into CSharpSrc and new implemented in the new Ruby frontend (useful for parenthesisless calls of methods imported into the scope).


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -0,0 +1,253 @@
+package io.joern.x2cpg.datastructures
+
+import scala.collection.mutable
+
+/** A heirarchical data-structure that stores the result of types and their respective members. These types can be
```
mpollmeier, line 5:
```suggestion
/** A hierarchical data-structure that stores the result of types and their respective members. These types can be
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -0,0 +1,253 @@
...
+
+}
+
+/** An implementation of combining the typed scoping structures to manage the available type information at namespace
+  * levels.
+  *
+  * @param summary
+  *   the program summary.
+  */
+class DefaultTypedScope[I, V, S <: TypedScopeElement, M <: MethodLike, F <: FieldLike, T <: TypeLike[M, F]](
```
mpollmeier, line 141: This is very heavy in type parameters - can you please give them more meaningful names? Like with regular parameters, single-char names are ok in very concise functions, i.e. if the scope is very small. That's not the case here, so I'd be helpful to the reader if they have proper names. 

To be honest, I don't want people to see this and blame it all on "Scala is sooo complicated", throw it all away and rewrite it in go. I've seen it happen before, trust me :)

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -0,0 +1,253 @@
...
+  * The utility of this object is in assisting resolving shorthand types during AST creation.
+  *
+  * @tparam M
+  *   the method/function meta data class.
+  * @tparam F
+  *   the field/object property meta data class.
+  * @tparam T
+  *   the type/class meta data class.
+  */
+trait ProgramSummary[M <: MethodLike, F <: FieldLike, T <: TypeLike[M, F]] {
```
mpollmeier, line 17: It doesn't look like `M` and `F` are really needed here, are they? This might help to simplify thinkgs as well. Can you please check the same for other type parameters as well?

```suggestion
trait ProgramSummary[T <: TypeLike[_, _]] {
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -0,0 +1,253 @@
...
+  */
+class DefaultTypedScope[I, V, S <: TypedScopeElement, M <: MethodLike, F <: FieldLike, T <: TypeLike[M, F]](
+  summary: ProgramSummary[M, F, T]
+) extends Scope[I, V, S]
+    with TypedScope[M, F, T, S](summary) {
+
+  /** Pops the scope, adding types from the scope if necessary.
+    */
+  override def pushNewScope(scopeNode: S): Unit = {
+    scopeNode match
```
mpollmeier, line 149: Minor note: I'd prefer to keep our current style to use braces around match statements - we didn't jump onto the bandwagon of significant indentation, and IMO that's a good thing.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
DavidBakerEffendi: Good advice, I'll see what I can do to reduce this and make it a lot friendlier, even with maybe some default scope elements

joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/datastructures/ProgramSummaryTests.scala
```diff
@@ -0,0 +1,90 @@
...
+
+    "successfully resolve a type once the namespace scope is pushed" in {
+      val mockScope = DefaultTypedScope(summary = mockSummary)
+      mockScope.pushNewScope(NamespaceTSEImpl("io.joern"))
+      mockScope.tryResolveTypeReference("Foo") match
+        case None      => fail("Unable to resolve type!")
+        case Some(typ) => typ.name shouldBe "io.joern.Foo"
+    }
+
+    "unable to resolve a type once the namespace scope is popped and pushed off again" in {
```
AndreiDreyer, line 67: Minor wording change: `unable to resolve a type once the namespace scope is pushed and popped off again`.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
mpollmeier: Another option that could help simplify things in some situations might be to use type members instead of type parameters:
```
class A1[B]
// vs
class A2 { type B }
```

DavidBakerEffendi: I've pushed some new changes to address reducing the type arguments, including modelling module-like scope behaviour. 

I still figure `DefaultTypedScope` should be seen as more of a reference implementation, instead of a canonical way to use these tools, but I've chosen to design it in a way that now it should suite most usages I've observed.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -0,0 +1,318 @@
...
+
+  /** Pops the scope, removing types from the scope if necessary.
+    */
+  override def popScope(): Option[TypedScopeElement] = {
+    super.popScope() match {
+      case Some(n: NamespaceLikeScope) =>
+        summary.typesUnderNamespace(n.fullName).foreach(typesInScope.remove)
+        Some(n)
+      case x => x
+    }
```
mpollmeier, line 217:
```suggestion
    super.popScope().map {
      case n: NamespaceLikeScope =>
        summary.typesUnderNamespace(n.fullName).foreach(typesInScope.remove)
        n
      case x => x
    }
```

mpollmeier (approved):

PR #4133, opened 2024-02-07
[php2cpg] Fix orphan identifier issue

johannescoetzee:
There's a pattern in `php2cpg` where ASTs are sometimes created, but not added to the graph (doing it this way seemed sensible at the time, but now I'm doubting that). Regardless, this was causing an issue with `astForNameExpression` which added ref edges from the newly created node to the declaring node directly to the graph, even if that AST was not added, resulting in orphan identifier nodes. These orphan nodes, in turn, caused issues with some backend passes.

This PR cleans up a few of the cases where the AST is created but not added, but also changed `astForNameExpr` to return the ref edges in the identifier AST instead of adding them to the graph directly, which fixes the orphan issue even if created ASTs are discarded.

This should also fix https://shiftleftinc.atlassian.net/browse/SEN-2172

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
@@ -663,29 +663,32 @@ class AstCreator(filename: String, phpAst: PhpFile)(implicit withSchemaValidatio
 
   private def astsForStaticStmt(stmt: PhpStaticStmt): List[Ast] = {
     stmt.vars.flatMap { staticVarDecl =>
-      val variableAst   = astForVariableExpr(staticVarDecl.variable)
-      val maybeValueAst = staticVarDecl.defaultValue.map(astForExpr)
+      staticVarDecl.variable match {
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
```diff
@@ -1027,11 +1029,9 @@ class AstCreator(filename: String, phpAst: PhpFile)(implicit withSchemaValidatio
   private def astForNameExpr(expr: PhpNameExpr): Ast = {
     val identifier = identifierNode(expr, expr.name, expr.name, TypeConstants.Any)
 
-    scope.lookupVariable(identifier.name).foreach { declaringNode =>
-      diffGraph.addEdge(identifier, declaringNode, EdgeTypes.REF)
-    }
+    val declaringNode = scope.lookupVariable(identifier.name)
 
-    Ast(identifier)
+    Ast(identifier).withRefEdges(identifier, declaringNode.toList)
```
johannescoetzee, line 1034: This part is the proper fix

ml86 (approved):

PR #4135, opened 2024-02-07
Fix php switch and match cases

johannescoetzee:
The last PR to fix the unused AST pattern mentioned in https://github.com/joernio/joern/pull/4133. This adds switch/match case conditions to the AST directly after the corresponding jump target.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
@@ -46,15 +46,20 @@ class ControlStructureTests extends PhpCode2CpgFixture {
         inside(switchStmt.whenTrue.astChildren.l) {
           case List(
                 case0: JumpTarget,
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
```diff
@@ -887,23 +899,35 @@ class ControlStructureTests extends PhpCode2CpgFixture {
...
 
+            bCond.code shouldBe "$b"
+            bCond.lineNumber shouldBe Some(4)
+
             cTarget.code shouldBe "case $c"
             cTarget.lineNumber shouldBe Some(4)
 
+            cCond.code shouldBe "$c"
+            cCond.lineNumber shouldBe Some(4)
+
```
johannescoetzee, line 930: For a case like 
```
match (...) {
  $b, $c => ...
}
```
I'm not sure if we want separate jump targets for the 2 variables or one shared. Having the separate targets creates a clearer link between jump target and condition, in my opinion.

PR #4143, opened 2024-02-08
dynamically pass through available terminal width for help table rendering

mpollmeier:
depends on https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1754

preview for different table sizes:

image

image


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/Path.scala
```diff
@@ -25,7 +26,7 @@ object Path {
   // TODO replace with dynamic rendering based on the terminal's width, e.g. in scala-repl-pp
   lazy val maxTrackedWidth = sys.env.get("JOERN_DATAFLOW_TRACKED_WIDTH").map(_.toInt).getOrElse(DefaultMaxTrackedWidth)
```
maltek, line 27: the comment above says we should get rid of this in a PR like this one here.

maltek (approved):

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/Path.scala
mpollmeier: True that. This is used to shorten the tracked statement (see a few lines below). Given that our tables now support cells with arbitrary length (which is being wrapped into multiple lines), we can probably just drop this altogether... 

Would you agree? 
And would you mind giving this a quick poweruser sanity test, once https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1754 is merged? I'll let you know.

mpollmeier:
`.p` output looks suboptimal - this isn't new in any way and just a result on how the repl renders the strings, but I think I'll just add a `\n` somewhere to make it look nicer

image

mpollmeier:
now it's better
image

mpollmeier:
should get fixed by https://github.com/ShiftLeftSecurity/overflowdb/pull/417

maltek:
it looks like the new library just makes all columns the same width? that's not great for flows

mpollmeier:
yes that's true. will experiment with column spans, maybe in a follow-up PR though, given that this is already better than the status quo

PR #4144, opened 2024-02-08
[swiftsrc2cpg] Added Support for Tuple Pattern Variable Declarations

max-leuthaeuser:


max-leuthaeuser:
@DavidBakerEffendi Any idea how this PR can fail some ruby tests?

max-leuthaeuser:
Thanks!
Yes, it needs to be fixed then as it delays all future PRs.

DavidBakerEffendi (approved):

PR #4146, opened 2024-02-08
[ruby] Handle Slurping of Multi-Assignments

DavidBakerEffendi:
* Modified grammar to expect splat operator in more positions on the multi-lhs
* Tested the splat operator slurping RHS variables for three tests
* Modified MandatoryParameter to contain its identifier
* Handle implicit slurping

Continuation of [#4136](https://github.com/joernio/joern/pull/4136)

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
@@ -126,8 +126,8 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
           } ++ (whenClause.matchSplatExpression.iterator.flatMap {
             case u: Unknown => List(u)
             case e =>
-              logger.warn("Splatting not implemented for `when` in ruby `case`")
-              List(Unknown()(e.span))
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -126,8 +126,8 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
           } ++ (whenClause.matchSplatExpression.iterator.flatMap {
             case u: Unknown => List(u)
             case e =>
-              logger.warn("Splatting not implemented for `when` in ruby `case`")
-              List(Unknown()(e.span))
+              logger.warn(s"Splatting not implemented for `when` in ruby `case`")
```
max-leuthaeuser, line 129: We don't need the interpolated string here right?

max-leuthaeuser (approved):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
DavidBakerEffendi: Ah some debugging code I rolled back

PR #4147, opened 2024-02-09
[swiftsrc2cpg] Several crash fixes

max-leuthaeuser:
Had a few crashes on https://github.com/apple/swift due to broken encoding (they use it as parser/compiler robustness checks - nothing we need, but we should never crash).


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForDeclSyntaxCreator.scala
```diff
@@ -34,7 +36,58 @@ trait AstForDeclSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
...
+
+    val name                     = typeNameForDeclSyntax(node)
+    val (typeName, typeFullName) = calcTypeNameAndFullName(name)
+    registerType(typeFullName)
+
+    val existingTypeDecl = global.seenTypeDecls.keys().asScala.find(_.name == typeName)
+    if (existingTypeDecl.isEmpty) {
+
+      val astParentType     = methodAstParentStack.head.label
+      val astParentFullName = methodAstParentStack.head.properties("FULL_NAME").toString
```
DavidBakerEffendi, line 54: `import io.shiftleft.codepropertygraph.generated.PropertyNames`
```suggestion
      val astParentFullName = methodAstParentStack.head.properties(PropertyNames.FULL_NAME).toString
```

DavidBakerEffendi (approved):

PR #4148, opened 2024-02-09
[ruby] Handle Multi-Assignment Splitting

### LINKED ISSUE #3931: [ruby] Destructured Assignments & Splatting
Assignments (from [this example](https://dev.to/ridhwana/destructuring-in-ruby-53j0)) can be destructured in many forms. E.g. 

```ruby
a, b, c = 1, 2, 3

a => 1 
b => 2 
c => 3 
```

The above can simply be `desugared` into 3 assignments.

---

```ruby
a, b, c, d = 1, 2, 3

a => 1 
b => 2 
c => 3 
d =>  undefined local variable or method `d' for main:Object
```
In this case, `d` would not get an assignment, but would get a corresponding `LOCAL` node.

---

```ruby
a, b, c = 1, 2, 3, 4
```
The `4` would remain unassigned, and so I would likely say it's best to leave it out, since the CPG schema always expects an assignment to have 2 AST children.

---

The splat operator (`*`) is tricky, e.g.

```ruby
a, b, *c = 1, 2, 3, 4

a => 1 
b => 2 
c => [3, 4]
```

However, it has well-defined rules on which we could group the outcomes together into the assignments as necessary.

```ruby
a, *b, c = 1, 2, 3, 4

a => 1 
b => [2, 3] 
c => 4
```

More complicated example:

```ruby
array =  [2, 3, 4, 5]

a, *b, c = 1, *array
```

(More rules and examples on the link above)
---

We may need to simulate how these evaluate in order to desugar well and make sure that the lists and the single values are correctly assumed.

DavidBakerEffendi:
* Handles the splat operator on an array on the RHS of a multi-assignment
* Rolls back change where we could expect two splats on LHS, but had to work around some other grammar restrictions for the splat operator being on other points in the multi-assignment, e.g., `*a, b, c = ...` and `a, *b, c = ...`

In the following example
```ruby
list = [2, 3]
a, b, c = 1, *list
```
We would get
```ruby
a => 1
b => 2
c => 3
```
But we may not always be able to see `list`, instead, this is approximated with an `<operator>.splat` call node to look like
```ruby
a => 1
b => *list
c => *list
```

Finally Resolves #3931

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
@@ -331,13 +331,9 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
       *   how many more duplicates to create.
       */
     def slurp(nodes: List[RubyNode], expandSize: Int): List[RubyNode] = nodes match {
-      case (head: SplattingRubyNode) :: tail if expandSize > 0 && tail.exists(_.isInstanceOf[SplattingRubyNode]) =>
-        val newTail = slurp(tail, expandSize - 1)
-        head :: slurp(head :: newTail, expandSize - 2)
-      case (head: SplattingRubyNode) :: tail if expandSize > 0 =>
-        head :: slurp(head :: tail, expandSize - 1)
-      case head :: tail => head :: slurp(tail, expandSize)
-      case Nil          => List.empty
+      case (head: SplattingRubyNode) :: tail if expandSize > 0 => head :: slurp(head :: tail, expandSize - 1)
+      case head :: tail                                        => head :: slurp(tail, expandSize)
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
```diff
@@ -358,52 +340,107 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
       case Nil          => List.empty
```
badly-drawn-wizards, line 340: The top case is no longer applicable (assuming it is impossible in the grammar), so it should be dead code. Not affecting functionality, so up to you what to do with it.

DavidBakerEffendi: Good point, I should remove it as it is probably the least eloquent code and may be adding complexity

PR #4151, opened 2024-02-12
[ruby] Program Summary & Typed Scope

### LINKED ISSUE #4128: [ruby] Pre-Parse For High Level Structures
To enable efficient type resolution during AST creation, as well as determine whether an identifier is a call or parenthesis-less invocation, we need some level of interprocedural context before AST creation time.

Adopting CSharpSrc's TypeMap, with some tweaks to consider the duck-typing nature of Ruby, we should be able to accomplish this.

DavidBakerEffendi:
This is a big one, but here we go

* Created basic implementation classes for `ProgramSummary` and `TypedScopeElement` in Ruby
* Added program summary hooks with implementation in `AstSummaryVisitor` that re-uses some `AstCreator` processes.
* Implemented pre-parse of high-level structures 
* Used this to fix implicit constructor bugs (and other bugs that were lurking around due to having 3 stacks)
* Differentiated modules/types to make sure implicit constructor wasn't created for modules
* Removed other stacks, only using `scope` now

Now we should be able to figure out if an `identifier` is a call or not with the lookahead that `RubyScope` and `RubyProgramSummary` provides us with.

Resolves #4128

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
@@ -8,10 +8,6 @@ import io.shiftleft.codepropertygraph.generated.nodes.DeclarationNew
   *
   * The utility of this object is in assisting resolving shorthand types during AST creation.
   *
-  * @tparam M
-  *   the method/function meta data class.
-  * @tparam F
-  *   the field/object property meta data class.
   * @tparam T
   *   the type/class meta data class.
   */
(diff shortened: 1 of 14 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -40,6 +38,25 @@ trait ProgramSummary[T <: TypeLike[_, _]] {
...
 
+object ProgramSummary {
+
+  /** Combines two namespace-to-type maps.
+    */
+  def combine[T <: TypeLike[_, _]](a: Map[String, Set[T]], b: Map[String, Set[T]]): Map[String, Set[T]] = {
+    val accumulator = mutable.HashMap.from(a)
+    val allKeys     = accumulator.keySet ++ b.keySet
+
+    allKeys.foreach(k =>
```
xavierpinho, line 49: Nit: since `accumulator` is already taken from `a`, we could probably just traverse `b`'s keys, right?

xavierpinho (approved):
I like this!

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
DavidBakerEffendi: Good point, will implement the suggestion!

PR #4159, opened 2024-02-13
Andrei/c#/implicit member access

AndreiDreyer:
This PR includes: 
 * Implicit member access


joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -39,6 +39,20 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
     Seq(Ast(literalNode(_literalNode, code(_literalNode), nodeTypeFullName(_literalNode))))
   }
 
+  protected def astForExpressionNode(expressionNodeInfo: DotNetNodeInfo): Seq[Ast] = {
```
DavidBakerEffendi, line 42: I wonder if we should rename this to `astForOperand` if that's the only use case for this special case method

DavidBakerEffendi: Also, there exists `astForExpression`, so that's why I'm pretty keen for the name to be a bit more specific to reduce ambiquity

joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/MemberTests.scala
```diff
@@ -178,13 +178,16 @@ class MemberTests extends CSharpCode2CpgFixture {
...
         |  static int nonInitMaxSpeed;  // field
+        |  int b;
         |
         |  static Car() { // static constructor
         |     this.nonInitMaxSpeed = 2000;
         |  }
         |
         |  public Car() {
         |     this.color = "red";
+        |     b = 1000;
```
DavidBakerEffendi, line 189: Could you add a test case where there is a local variable shadowing the field, so that it correctly gets determined as a non-field access? e.g.
```java
public class Foo {
   int a;
   public Foo() {
      var a = 2;
      this.a = a;
   }
}
```

DavidBakerEffendi (changes requested):

PR #4161, opened 2024-02-13
Enabled `CodeDumper` to use `File.content`

DavidBakerEffendi:
* Enabled `CodeDumper` to use `File.content` if it's not empty
* Added `CSHARPSRC` to list of supported languages
* Tested this using `JAVASRC`

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
@@ -28,6 +28,7 @@ class JavaSrc2Cpg extends X2CpgFrontend[Config] {
   override def createCpg(config: Config): Try[Cpg] = {
     withNewEmptyCpg(config.outputPath, config: Config) { (cpg, config) =>
       new MetaDataPass(cpg, language, config.inputPath).createAndApply()
+      println(s"File content disabled! ${config.disableFileContent}")
       val astCreationPass = new AstCreationPass(config, cpg)
       astCreationPass.createAndApply()
       new ConfigFileCreationPass(cpg).createAndApply()
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
@@ -28,7 +28,6 @@ class JavaSrc2Cpg extends X2CpgFrontend[Config] {
   override def createCpg(config: Config): Try[Cpg] = {
     withNewEmptyCpg(config.outputPath, config: Config) { (cpg, config) =>
       new MetaDataPass(cpg, language, config.inputPath).createAndApply()
-      println(s"File content disabled! ${config.disableFileContent}")
       val astCreationPass = new AstCreationPass(config, cpg)
       astCreationPass.createAndApply()
       new ConfigFileCreationPass(cpg).createAndApply()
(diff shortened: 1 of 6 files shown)


max-leuthaeuser (approved):

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/CodeDumperTests.scala
```diff
@@ -0,0 +1,41 @@
...
+      inside(cpg.method.nameExact("test").dumpRaw.l) {
+        case content :: Nil =>
+          content.linesIterator.map(_.strip).l shouldBe List(
+            "public void test() { /* <=== Foo.test:void() */",
+            "var a = 1;",
+            "var b = 2;",
+            "var c = a + b;",
+            "}"
+          )
+        case _ => fail("Expected exactly 1 content dump")
```
mpollmeier, line 35:
```suggestion
        case content => fail(s"Expected exactly 1 content dump, but got: $content")
```

mpollmeier (approved):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/codedumper/CodeDumper.scala
```diff
@@ -74,6 +76,14 @@ object CodeDumper {
                       line
                   }
                   .mkString("\n")
+              } else if (m.file.exists(_.content != PropertyDefaults.Content)) {
+                sliceCode(
```
maltek, line 80: there's a `m.content` step already that should help here

maltek: ah no, this code isn't dumping the method but (potentially) parts of it. Won't help much then.

DavidBakerEffendi: Thanks! That simplifies things. Making the related change now

DavidBakerEffendi: > ah no, this code isn't dumping the method but (potentially) parts of it. Won't help much then.

Ok, let's see if the tests pass, but I've used `m.content` and then split the lines so that the comment arrow can still work

DavidBakerEffendi:
> javasrc2cpg uses platform dependent file reading:
> 
> https://github.com/joernio/joern/blob/5fc3938207b7e7b57dfc7ef9510c963d8281d85e/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/SourceParser.scala#L44
> 
> Maybe it helps to change that to the utils provided by x2cpg.

Thanks, let me give it a try

DavidBakerEffendi:
Update: If I disable `CodeDumperTests` then the `OffsetTests` appear to pass. Strange, I wonder what side-effect is causing this...

DavidBakerEffendi:
Ah, this was keeping configs from old tests within the same test class
```
trait LanguageFrontend {
  private var config: Option[X2CpgConfig[_]] = None
```

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
```diff
@@ -28,6 +28,7 @@ class JavaSrc2Cpg extends X2CpgFrontend[Config] {
   override def createCpg(config: Config): Try[Cpg] = {
     withNewEmptyCpg(config.outputPath, config: Config) { (cpg, config) =>
       new MetaDataPass(cpg, language, config.inputPath).createAndApply()
+      println(s"File content disabled! ${config.disableFileContent}")
```
max-leuthaeuser: Leftover log message?

PR #4165, opened 2024-02-14
[C#] Populate `inheritsFromTypeFullName`, and some more type resolution

### LINKED ISSUE #3992: [c#] Handle Inheritance on Class Declarations
https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/object-oriented/inheritance

karan-batavia:
This PR includes.
1. Populating the `inheritsFromTypeFullName` property for `typeDecl` nodes.
2. Resolving types if short-hand types are present on the LHS of variable declaration.
3. Tests.

Resolves #3992


joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/TypeMap.scala
```diff
@@ -100,8 +100,8 @@ class TypeMap(astGenResult: AstGenRunnerResult, initialMappings: List[NamespaceT
       .map(AstCreatorHelper.createDotNetNodeInfo(_))
       .filter { x =>
         x.node match
-          case NamespaceDeclaration | FileScopedNamespaceDeclaration | ClassDeclaration => true
-          case _                                                                        => false
+          case NamespaceDeclaration | FileScopedNamespaceDeclaration | ClassDeclaration | InterfaceDeclaration => true
```
DavidBakerEffendi, line 103: Perhaps it would make sense to bind these together under a single trait, something like `NamespaceOrTypeDeclaration`. Then we could add `Enum` and `Record` too

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpScope.scala
```diff
@@ -47,7 +48,16 @@ class CSharpScope(typeMap: TypeMap) extends Scope[String, DeclarationNew, ScopeT
     .exists(x => x.scopeNode.isInstanceOf[MethodScope] || x.scopeNode.isInstanceOf[TypeLikeScope])
 
   def tryResolveTypeReference(typeName: String): Option[String] = {
-    typesInScope.find(_.name.endsWith(typeName)).flatMap(typeMap.namespaceFor).map(n => s"$n.$typeName")
+    typesInScope
+      .find(_.name.endsWith(typeName))
+      .flatMap(typeMap.namespaceFor)
+      .map(n => {
+        boundary:
```
DavidBakerEffendi, line 55: I'm not a fan of `boundary-break`, this is a side-effect which we would like to avoid. Why not simply continue mapping?

DavidBakerEffendi: Also, you can use `==` in Scala, strings are equated by value

joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/InheritanceFullNameTests.scala
```diff
@@ -0,0 +1,103 @@
...
+        |   Bar b = new Bar();
+        | }
+        |}
+        |""".stripMargin)
+
+    "resolve full names for classes with inheritance" in {
+      inside(cpg.typeDecl.nameExact("Bar").l) {
+        case bar :: Nil =>
+          bar.fullName shouldBe "Bar"
+          bar.inheritsFromTypeFullName shouldBe Seq("Foo")
```
DavidBakerEffendi, line 26: `Foo`'s full name is equal to its short name. Could you wrap this in some namespace to make sure it's a full name, please?

DavidBakerEffendi: Similarly on other tests below

joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/InheritanceFullNameTests.scala
```diff
@@ -0,0 +1,103 @@
...
+        case _ => fail("No class named `Bar` found")
+      }
+    }
+
+    "resolve identifier full names instantiated from the class" in {
+      inside(cpg.identifier.nameExact("b").l) {
+        case b :: Nil =>
+          b.typeFullName shouldBe "Bar"
```
DavidBakerEffendi, line 34: Similarly it's difficult to tell if the type is resolved here or not

DavidBakerEffendi (changes requested):

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpScope.scala
karan-batavia: The IDE threw a warning, thought we should avoid using if this is discontinued. Will revert it back if you think it's not a good idea.

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpScope.scala
```diff
@@ -47,7 +48,17 @@ class CSharpScope(typeMap: TypeMap) extends Scope[String, DeclarationNew, ScopeT
...
 
   def tryResolveTypeReference(typeName: String): Option[String] = {
-    typesInScope.find(_.name.endsWith(typeName)).flatMap(typeMap.namespaceFor).map(n => s"$n.$typeName")
+    typesInScope
+      .find(_.name.endsWith(typeName))
+      .flatMap(typeMap.namespaceFor)
+      .map(n => {
+        // To avoid recursive type prefixing on assignment calls.
+        if (typeName.startsWith(n)) {
+          return Some(typeName)
```
DavidBakerEffendi, line 57: It's generally better to use `Option` instead of `Some` directly, but I wont block the PR for that now

DavidBakerEffendi: ^ Reason being, `Some(null) != Option(null)`

DavidBakerEffendi (approved):

Issue #4168, opened 2024-02-14
[ruby] Handle Constructor Invocation


DavidBakerEffendi:
The constructor invocation is simply a `MemberCall` and the `new` method name is not interpreted as a constructor. This needs to be handled.

DavidBakerEffendi:
One should also note the comments under the preceding issue about checking the base of `.new` https://github.com/joernio/joern/pull/4171

PR #4171, opened 2024-02-14
[ruby] Anonymous & Singleton Class Handling 

### LINKED ISSUE #3934: [ruby] Anonymous & Singleton Classes
Some more Ruby-foo to handle, see [this example](https://dev.to/factorial/a-trick-with-ruby-anonymous-classes-11pp)

```ruby
Class.new do
  def hello
    puts "Hello world!"
  end
end
```

The JavaScript frontend has a mechanism for handling such things, which can be used as a guideline. The corresponding structure in JavaScript looks something like:

```javascript
{ 
   "foo": "bar"
}
```

This object is something like creating the type and invoking it at the same time. I would be hesitant to simply call it a dictionary or map, since there is already such a structure for that. This is a lot more around the dynamic handing of properties.

DavidBakerEffendi:
* Added anonymous class name generator with an incrementer variable. This will be unique on a per-file level, which is fine given the fully qualified name.
* Handling anonymous classes as a form of class declaration.
* Added TODO for constructor invocation handling
* Unified singleton classes and anonymous classes under the trait `AnonymousTypeDeclaration`
* Singleton classes trigger the scope to add the singleton type to the possible types of the variables that may be instances of the base type

Resolves #3934 
Next #4168 

### Thought process behind the singleton class implementation

The singleton class is similar in behaviour to `implicit class` in Scala, but a lot more throw-away. This is treated as a type of anonymous class that inherits the behaviours of the base class it extends, but also now extends the functionality of nodes of the base type.

In terms of the CPGs ability to resolve the calls of this new behaviour, the `RubyScope` object now appends this anonymous class to the `possibleTypes` of the variable. The type recovery should then be able to do the rest i.t.o call resolution for open-source Joern users. The only reason the test is ignored is that the constructors aren't handled and thus any points-to won't succeed since there is no `<operator>.new` or call referencing a constructor, this means the variables in scope don't have this type associated as a possible type.


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
```diff
@@ -548,7 +558,10 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
       )(ctx.toTextSpan)
     }
 
-    if (hasBlock) {
+    if (hasBlock && ctx.methodName().getText == "new") {
```
xavierpinho, line 561: I'm wondering if we shouldn't check if it's `Class.new` exactly, as there might be other methods named `new`

xavierpinho (approved):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
DavidBakerEffendi: Yeah good point

DavidBakerEffendi:
Hmm I think I've made an incorrect assumption about this being similar to Scala's `implicit class`. This only affects the variable that is being extended, which is different to changing the meta class too.

PR #4187, opened 2024-02-19
[kotlin] Parser Performance Improvement & Additional Null Safety

pandurangpatil:
1. Changed input directory list parameter to a single main repo path. We were earlier passing a list of all directories in the repository. Which was resulting into `KotlinToJVMBytecodeCompiler.INSTANCE.analyze` method taking too much of time along with consuming too much of memory in the process. Tried with the single path and it reduced the time as well as
memory footprint with the same results.
2. Made changes into warning logs to add the file path to get the details of the file where we are facing the given problem.
3. Removed two assert statements and handled the situation gracefully.


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForStatementsCreator.scala
```diff
@@ -341,15 +341,21 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) {
...
   ): Ast = {
-    assert(expr.getSubjectExpression != null)
-
-    val astForSubject = astsForExpression(expr.getSubjectExpression, Some(1)).headOption.getOrElse(Ast())
-    val finalAstForSubject = expr.getSubjectExpression match {
-      case p: KtProperty =>
-        val block = blockNode(p, "", "").argumentIndex(1)
-        blockAst(block, List(astForSubject))
-      case _ => astForSubject
+    val (astForSubject, finalAstForSubject) = Some(expr.getSubjectExpression) match {
```
DavidBakerEffendi, line 344: `Some(null)` will return `Some(null)` and not `None`
```suggestion
    val (astForSubject, finalAstForSubject) = Option(expr.getSubjectExpression) match {
```

DavidBakerEffendi (changes requested):

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForStatementsCreator.scala
pandurangpatil: @DavidBakerEffendi thnx for catching this one. I will make changes and update.

pandurangpatil: Done

DavidBakerEffendi (approved):

Issue #4117, opened 2024-02-03
[c#] Clean Up InvocationExpression Handling 


DavidBakerEffendi:
[This PR](https://github.com/joernio/joern/pull/4179) was largely to unblock downstream users but introduced some ugly code with unclear heuristics. This tickets stands as a need to clean this up and replace it with a more robust & clear solution.

PR #4190, opened 2024-02-19
[C#] Enhanced type resolution for internal calls, and handle some missing Ast Structures

### LINKED ISSUE #4191: [C#] Resolve types where Generic Names are used
For the scope of this issue, only resolve types where it is used, and not where it is declared.

### LINKED ISSUE #4192: [C#] Add types from upstream namespaces to scope.
An example of an upstream namespace is, 
`namespace Foo.Bar `-> `Foo` is an upstream namespace for `Foo.Bar`, and types from Foo can be used inside `Foo.Bar` without a `using` statement.

### LINKED ISSUE #4193: [C#] Add all the types which are inherited to scope
All the types of the inherited class are accessible in the child classes, without a corresponding `using` statement.

karan-batavia:
This PR includes, 
1.  Handling for the GenericName syntax property.
2. Resolve types if shorthand types are present on the LHS.
3. Add upstream namespaces types to scope.
4. Add all the types from the inherited classes to scope.

Resolves #4191 
Resolves #4192 
Resolves #4193


joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -238,8 +246,14 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
 
   def astForObjectCreationExpression(objectCreation: DotNetNodeInfo): Seq[Ast] = {
     val dispatchType = DispatchTypes.STATIC_DISPATCH
-    val typeFullName = nodeTypeFullName(objectCreation)
-    val arguments    = astForArgumentList(createDotNetNodeInfo(objectCreation.json(ParserKeys.ArgumentList)))
+    val typeFullName = Try(createDotNetNodeInfo(objectCreation.json(ParserKeys.Type))).toOption match {
```
DavidBakerEffendi, line 249: I think it's best if we simply use `nodeTypeFullName(typenode)` for all of this? e.g.
`Try(createDotNetNodeInfo(objectCreation.json(ParserKeys.Type))).toOption.map(nodeTypeFullName).getOrElse(Defines.Any)`

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForPrimitivesCreator.scala
```diff
@@ -1,23 +1,26 @@
...
     if identifierName != "_" then {
-      val variableOption = scope.lookupVariable(identifierName)
-      variableOption match
-        case Some(variable) =>
+      List(scope.lookupVariable(identifierName), scope.findFieldInScope(identifierName)) match {
+        case List(Some(variable: DeclarationNew), None) =>
           val node = identifierFromDecl(variable, Option(ident))
           Ast(node).withRefEdge(node, variable)
+        case List(None, Some(field: FieldDecl)) =>
+          Ast(identifierNode(ident, field.name, ident.code, field.typeFullName))
```
DavidBakerEffendi, line 20: If this is a field, shouldn't it rather be a field access + field identifier? @AndreiDreyer has been working on field accesses and may be able to advise.

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -149,6 +151,12 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
         } else {
           (Option(baseIdentifier), Option(_typeFullName))
         }
+      case IdentifierName =>
+        val typeFullName = scope.findTypeFromMethodName(nameFromNode(expression)) match {
```
DavidBakerEffendi, line 155: What is an example of this? Shouldn't we consider rather popping of the stack to find the first instance of a type that has a method that shares this name? That way we can ensure we aren't in trouble from shadowing.

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpScope.scala
```diff
@@ -119,9 +127,20 @@ class CSharpScope(typeMap: TypeMap) extends Scope[String, DeclarationNew, ScopeT
     */
   override def pushNewScope(scopeNode: ScopeType): Unit = {
     scopeNode match
-      case NamespaceScope(fullName) => typesInScope.addAll(typeMap.classesIn(fullName))
-      case TypeScope(name, _)       => typesInScope.addAll(typeMap.findType(name))
-      case _                        =>
+      case NamespaceScope(fullName) => {
```
DavidBakerEffendi, line 130: Could we rather move this logic under `astForNamespaceDeclaration` or something similar? Also, a potentially cleaner solution that avoids blank strings (that could be placed under `astForNamespaceDeclaration`) could be something like
```scala
def recurseNamespace(parts: List[String], prefix: List[String] = List.empty): Unit = {
  parts match {
    case head :: tail => 
      val currentFullName = prefix :+ head
      scope.pushNewScope(NamespaceScope(currentFullName.mkString(".")))
      recurseNamespace(tail, currentFullName)
    case Nil => // nothing
  }
}
val namespaceParts = fullName.split("[.]").toList
recurseNamespace(namespaceParts)
// Do ast stuff
namespaceParts.foreach(_ => scope.popScope())
```
Note the popping at the end is also important.

DavidBakerEffendi (changes requested):

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForExpressionsCreator.scala
karan-batavia: https://github.com/Privado-Inc/joern/blob/46674de7d30a8e6a9f08f30af8ad00177e6fe27e/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/MethodTests.scala#L80

Refer this test.

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForPrimitivesCreator.scala
karan-batavia: I've seen cases where a `member` is accessed just by it's name, do we want to create a `fieldAccess` node here as well?

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForExpressionsCreator.scala
karan-batavia: You're right, but this would fail in case of multiple inheritance where we'll push multiple types on the stack and we might not get the required type when we pop the stack.

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForPrimitivesCreator.scala
DavidBakerEffendi: When a local variable is overshadowing a field then the local is used, but if it's referring to the field, then a field access should be used. See this test case for an example: https://github.com/joernio/joern/blob/03e0ba64e3800760d59db9c72057a9c1242b99e5/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/MemberTests.scala#L273

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForExpressionsCreator.scala
DavidBakerEffendi: I'm re-structuring this as part of [#4117](https://github.com/joernio/joern/issues/4117) so I wouldn't block you on this, the tests will fail if I break something here.

DavidBakerEffendi:
Another note, please refrain from doing so many issues in one PR, it makes it difficult to review. Additionally, it blocks the multiple tickets at a time, especially if you haven't assigned yourself on the Kanban appropriately.

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForPrimitivesCreator.scala
```diff
@@ -1,23 +1,26 @@
...
 trait AstForPrimitivesCreator(implicit withSchemaValidation: ValidationMode) { this: AstCreator =>
 
   protected def astForIdentifier(ident: DotNetNodeInfo, typeFullName: String = Defines.Any): Ast = {
     val identifierName = nameFromNode(ident)
     if identifierName != "_" then {
-      val variableOption = scope.lookupVariable(identifierName)
-      variableOption match
-        case Some(variable) =>
+      List(scope.lookupVariable(identifierName), scope.findFieldInScope(identifierName)) match {
+        case List(Some(variable: DeclarationNew), None) =>
```
DavidBakerEffendi, line 21: These first two cases seem redundant, and these two seem to be invariant, could be something more concise like
```scala
      scope.lookupVariable(identifierName) match {
        case Some(variable: DeclarationNew) =>
          val node = identifierFromDecl(variable, Option(ident))
          Ast(node).withRefEdge(node, variable)
        case None =>
          scope.findFieldInScope(identifierName) match {
            case Some(field) if field.node.node != DotNetJsonAst.VariableDeclarator =>
              val fieldAccess =
                newOperatorCallNode(
                  Operators.fieldAccess,
                  field.node.code,
                  Some(field.typeFullName).orElse(Option(typeFullName)),
                  field.node.lineNumber,
                  field.node.columnNumber
                )
              val identifierAst = Ast(newIdentifierNode(identifierName, typeFullName))
              val fieldIdentifier = Ast(
                NewFieldIdentifier()
                  .code(field.name)
                  .canonicalName(field.name)
                  .lineNumber(field.node.lineNumber)
                  .columnNumber(field.node.columnNumber)
              )
              callAst(fieldAccess, Seq(identifierAst, fieldIdentifier))
            case _ => Ast(identifierNode(ident, identifierName, ident.code, typeFullName))
          }
      }
```

DavidBakerEffendi (changes requested):

DavidBakerEffendi (approved):

PR #4198, opened 2024-02-19
[Jimple2cpg] catch error caused by soot and asm

Liyw979:
```
  protected def astsForAnnotations(annotationTag: AnnotationTag, host: AbstractHost): Ast = {
    val typeFullName = registerType(annotationTag.getType.parseAsJavaType)
    val name         = typeFullName.split('.').last
    val elementNodes = annotationTag.getElems.asScala.map(astForAnnotationElement(_, host)).toSeq
    val code = s"@$name(${elementNodes.flatMap(_.root).flatMap(_.properties.get(PropertyNames.CODE)).mkString(", ")})"

    val annotation = annotationNode(host, code, name, typeFullName)
    annotationAst(annotation, elementNodes)
  }
```
Some times the `annotationTag.getType` returns a `MethodRef` and  `Type.getType(s)`  will throw excpetions.
Just catch this for the moment.

### DIFF
--- a/joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/astcreation/AstCreator.scala
@@ -287,7 +287,7 @@ implicit class JvmStringOpts(s: String) {
     * <code>I</code> to <code>int</code>.
     * @return
     */
-  def parseAsJavaType: String = 
-    Try{ Type.getType(s).getClassName.replaceAll("/", ".") }.getOrElse("unknownType")
+  def parseAsJavaType: String =
+    Try { Type.getType(s).getClassName.replaceAll("/", ".") }.getOrElse("unknownType")
 
 }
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/astcreation/AstCreator.scala
```diff
@@ -287,6 +287,7 @@ implicit class JvmStringOpts(s: String) {
     * <code>I</code> to <code>int</code>.
     * @return
     */
-  def parseAsJavaType: String = Type.getType(s).getClassName.replaceAll("/", ".")
+  def parseAsJavaType: String =
+    Try { Type.getType(s).getClassName.replaceAll("/", ".") }.getOrElse("unknownType")
```
DavidBakerEffendi, line 291: Perhaps use `io.joern.x2cpg.Defines` here
```suggestion
    Try { Type.getType(s).getClassName.replaceAll("/", ".") }.getOrElse(Defines.Any)
```

DavidBakerEffendi (changes requested):

DavidBakerEffendi (approved):

PR #4209, opened 2024-02-20
SimpleCache: wipe old files

mpollmeier:
In order for the cache not to grow up too much in size, we want to clear old entries.
This cache is so simple though that we have no idea when a file was last recently used, so we bite the bullet
and simply delete all files older than 30 days.
(In theory one could change the api to allow for that, but that'd make the SimpleCache api less 'simple')

### DIFF
--- a/project/SimpleCache.scala
+++ b/project/SimpleCache.scala
@@ -1,7 +1,11 @@
-import sbt.{IO, URL}
+import sbt.IO
+
 import java.io.File
-import java.net.URI
-import java.net.URLEncoder
+import java.net.{URI, URLEncoder}
+import java.nio.file.attribute.BasicFileAttributes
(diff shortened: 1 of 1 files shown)


maltek (approved):
It's a bit aggressive, in that every 30 days even files still in use will be deleted. Works for me still. Maybe we can refresh the date on existing files in `downloadMaybe` to avoid that?

DavidBakerEffendi:
> It's a bit aggressive, in that every 30 days even files still in use will be deleted. Works for me still. Maybe we can refresh the date on existing files in `downloadMaybe` to avoid that?

+1

mpollmeier:
I came up with a different idea that better suits our needs IMO: https://github.com/joernio/joern/pull/4217

PR #4210, opened 2024-02-21
[jssrc2cpg][swiftsrc2cpg] Improved logging of files being ignored by default

max-leuthaeuser:
astgen and swiftastgen do not log ignored files.
That may result in an empty CPG without the user even noticing. With this PR, we info log any file that was ignored in that particular case.


johannescoetzee:
As far as I can tell, this change would log each individual ignored file at the info level. For large ignored directories (test directories, for example), this could get very noisy. Perhaps we should keep the individual logs at debug level, but the number of files parsed at info level or warn if it's 0 (or just info/warn on 0)?

max-leuthaeuser:
Yeah, good point. I changed it.

maltek:
this duplicates the list of ignore patterns in two places, right? Once here, once in the astgen implementations? That seems brittle - it's only a matter of time until those lists (or their interpretation) diverge.

Would be better to either have the astgens return the list of files they've decided to ignore as part of their output, or get their ignore patterns as input.

max-leuthaeuser:
@maltek 
Yes, thats right and may become a problem in the future once they diverge.
I am working on a solution for that: all the astgen tools will produce a single .json file describing parsed files, un-parsable files (with the cause for that), and ignored files.
With this approach, no frontend will have to parse the stdout/stderr output anymore and we have standard for handing over information from astgen -> frontend.
This PR is just a intermediate solution to show at least some information why your CPG is empty.

maltek:
OK - no objection from me if this is just meant as a temporary solution :+1:

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/AstGenRunner.scala
```diff
@@ -130,12 +142,26 @@ class AstGenRunner(config: Config) {
   private def runAstGenNative(in: File, out: File): Try[Seq[String]] =
     ExternalCommand.run(s"$astGenCommand -o $out", in.toString())
 
+  private def checkParsedFiles(files: List[String], in: File): List[String] = {
+    val numOfParsedFiles = files.size
+    logger.info(s"Parsed $numOfParsedFiles files.")
+    if (numOfParsedFiles == 0) {
+      logger.info("You may want to check the DEBUG logs for a list of files that are ignored by default.")
```
johannescoetzee, line 149: I think this should be a warning, but looks good otherwise.

johannescoetzee (approved):

PR #4214, opened 2024-02-21
[jimple2cpg] - Enable Decompilation with CFR

### LINKED ISSUE #3498: [jimple2cpg] CFR Decompilation on `File.content`
The `.dumps` feature gives us Jimple, but this isn't always the easiest read. It would be great to have Java.

* The CPG is Jimple, so we also can't ignore it, so perhaps we should display both, Java-first then Jimple-second. Alternatively, we give Java if we have class files available (see below) and Jimple if no class files are found.
* The [CFR-Decompiler](https://github.com/leibnitz27/cfr) is what Fabs has been using so far, and we could do this on-the-fly.
* The new `File.content` and `Method.offset*` properties are useful here. [JavaSrc2Cpg](https://github.com/joernio/joern/tree/master/joern-cli/frontends/javasrc2cpg) has a good reference implementation of setting the CLI argument for file content and setting up an offset table.
* Main challenges in decompilation are:
  - Line numbers are only available if the bytecode was compiled in debug mode.
  - Some symbols may be missing and will produce invalid Java. This will make it hard to use a traditional parser for gathering offsets if line numbers are not available.
  - The points above mean that, for now, we may need to get creative with offset calculation, or explore additional CFR decompilation configurations. For the scope of this ticket, the test suite enables debug mode and simply use line numbers for offset calculations, this however may not fully correspond with the line numbers in the decompiled code.
* The idea is that in the [CodeDumper](http://projects.whirlylabs.com/io.shiftleft.semanticcpg.codedumper) would then give us the method on `.dump` calls. See this PR: https://github.com/joernio/joern/pull/4161

AndreiDreyer:
This PR Includes:
 * Decompile `.class` to `.java` and store it in the `/tmp` dir next to the `.class` file using the CFR library
 * Create `File` and populate `.content` with decompiled Java.

Partially resolves #3498


joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/astcreation/AstCreator.scala
```diff
@@ -49,9 +49,11 @@ class AstCreator(
     val astRoot = astForCompilationUnit(cls)
     storeInDiffGraph(astRoot, diffGraph)
 
-    val fileNode = NewFile().name(filename.replace(".class", ".java")).order(0)
-    fileContent.foreach(fileNode.content(_))
-    diffGraph.addNode(fileNode)
+    if (fileContent.isDefined) {
+      val fileNode = NewFile().name(filename.replace(".class", ".java")).order(0)
```
DavidBakerEffendi, line 53: Given that this is a bytecode frontend, the source file will always be the `.class` file
```suggestion
      val fileNode = NewFile().name(filename)).order(0)
```

DavidBakerEffendi: This also ensures that other nodes that set the `filename` property are linked to this node

DavidBakerEffendi (changes requested):

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/astcreation/AstCreator.scala
AndreiDreyer: So the name on this fileNode should be the `.class` instead of the `.java`?

DavidBakerEffendi: To make sure this works, enable file content on the configs for File and Method tests and see if they still pass

DavidBakerEffendi: > So the name on this fileNode should be the `.class` instead of the `.java`?

Yes

AndreiDreyer: Change made, confirmed still working for File and Method tests with file content enabled.

image

image

DavidBakerEffendi (approved):

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/astcreation/AstCreator.scala
DavidBakerEffendi: Great!

Issue #4172, opened 2024-02-14
[Bug] Joern gets number of arguments to call in PHP wrong


neochristou:
Here is a minimal example: 

```
<?php
$x = file("fasdfa");
define('SOME_CONST', 'somedir/');
$file = SOME_CONST . 'config.php';
include_once($file);
?>
``` 
In the last last, I would expect the call to `include_once` to only have one argument (the `$file` identifier), but I actually get the following:

```
joern> cpg.call.id(21).l(0).argument.l
val res5: List[io.shiftleft.codepropertygraph.generated.nodes.Expression] = List(
  Identifier(
    id = 22L,
    argumentIndex = 1,
    argumentName = None,
    code = "$file",
    columnNumber = None,
    dynamicTypeHintFullName = ArraySeq(),
    lineNumber = Some(value = 7),
    name = "file",
    order = 1,
    possibleTypes = ArraySeq(),
    typeFullName = "file"
  ),
  MethodRef(
    id = 87L,
    argumentIndex = 2,
    argumentName = None,
    code = "file",
    columnNumber = None,
    dynamicTypeHintFullName = ArraySeq(),
    lineNumber = Some(value = 7),
    methodFullName = "file",
    order = -1,
    possibleTypes = ArraySeq(),
    typeFullName = "<empty>"
  )
)
```

This only happens when both the call to `file` in the second line and the constant in the definition of the `$file` variable are there.

To reproduce:
1. Copy the example above in a file (e.g., `test.php`).
2. Generate its graph (`joern-parse --language php test.php --output graph`).
3. Start an interactive shell and check the call's arguments.

The expected behavior would be to only see the first argument (`Identifier`), but instead there is an additional `MethodRef` argument to the `file` function.

This was tested on Debian v12, Joern version 2.0.258, and Java version 17.0.9.

PR #4216, opened 2024-02-21
[XTypeRecovery] Fix type recovery function pointer bug

johannescoetzee:
This PR partially fixes an issue that came up in https://github.com/joernio/joern/issues/4172.

In this case, the `file` identifier matches the `file` method name, so in `XTypeRecovery` the `handlePotentialFunctionPointer` confuses them and creates a `METHOD_REF` in any case. The `handlePontentialFunctionPointer` function comment 
```
  /** In the case this field access is a function pointer, we would want to make sure this has a method ref.
    */
```
suggests that change in this PR is the intended behaviour, but I'm not sure about that and this isn't a proper fix for php either. Even if the `file` identifier were referring to a field, we'd still not want to enable this behaviour. A better fix could be to move `handlePotentialFunctionPointer` into the `XTypeRecovery` class to make it possible to adjust this behaviour per language.

I haven't looked at the type recovery code in quite a while though, so please let me know if I'm missing something.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
@@ -8,6 +8,18 @@ import io.shiftleft.codepropertygraph.generated.nodes.{Call, Identifier}
 import io.shiftleft.semanticcpg.language._
 
 class CallTests extends PhpCode2CpgFixture {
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi (approved):
Given that all the existing tests still pass, I think this fix is closer to the intended behaviour. 

This was originally motivated by frameworks in Python that would encourage a design of passing method references as arguments when defining rest endpoints, and there was some difficulty in matching the import as a method ref.

The current work on the `ProgramSummary` pre-parse should render this downstream check as redundant.

PR #4217, opened 2024-02-21
replace SimpleCache with DownloadHelper

mpollmeier:
While trying to improve SimpleCache (https://github.com/joernio/joern/pull/4209) I realised
that we no longer have the use case that it was originally introduced for. Plus, we added
some file-fiddling in the interim that would delete some of the files on
`sbt clean`, which isn't really the 'sbt way' :tm:

This PR suggests a different way to handle our current needs:
1) we do not need to delete old cache entries and re-download them every x days
2) we can still upgrade versions and it'll automatically replace the file
3) it's still fairly simple - if we take the build.sbt code into account it's actually far simpler now


max-leuthaeuser (approved):
Very nice!

All -internal repos would need an update then: e.g., https://github.com/ShiftLeftSecurity/jssrc2cpg-internal/blob/master/build.sbt#L28

DavidBakerEffendi (approved):

PR #4225, opened 2024-02-23
WIP fg transition

mpollmeier:
Instructions for local development:

publish [flatgraph](https://github.com/ShiftLeftSecurity/overflowdbv2/) locally:
```bash
git checkout master
git pull
sbt publishLocal
...
[info]  published core_3 to /home/mp/.ivy2/local/io.joern.flatgraph/core_3/0.0.13+1-bb8754d1/jars/core_3.jar
...
```

publish codepropertygraph branch locally:
```bash
git checkout michael/flatgraph
git pull
sbt publishLocal
...
[info]  published codepropertygraph_3 to /home/mp/.ivy2/local/io.shiftleft/codepropertygraph_3/1.6.5+12-385c92a1/jars/codepropertygraph_3.jar
...
```

stage joern locally:
```bash
git checkout michael/odb2-transition/dropin-replacement-wip
git pull
sbt stage
```

create a small and a medium cpg and import into joern:
```bash
./c2cpg.sh -o linux4-drivers-net-ethernet-broadcom.bin ~/tmp/cpgtesting/linux-kernel/linux-4.1.16/drivers/net/ethernet/broadcom
./c2cpg.sh -o linux4-drivers-net-ethernet.bin ~/tmp/cpgtesting/linux-kernel/linux-4.1.16/drivers/net/ethernet

./joern
importCpg("linux4-drivers-net-ethernet-broadcom.bin")
importCpg("linux4-drivers-net-ethernet.bin")
```

Current status: in comparison to joern master (which uses overflowdb), the importCpg is much slower while we apply the default layers. The smaller cpg (ethernet-broadcom) imports in 53s on my machine (compared to 17s in joern master). The medium cpg (ethernet) takes 3 mins on master and didn't finish after 15mins with flatgraph. 

Early profiling points to `DiffGraphApplier.addEdges` (line 541), but that line looks fast and I haven't figured out how to drill in further yet...

1-diffgraph-addEdges-slow


bbrehm:
Well spotted! The stack trace and problem description are both excellent enough that I don't even need to reproduce to see the issue.

The immediate culprit of the performance problem is quite clear: ConcurrentWriterCpgPass.

The issue is that ConcurrentWriterCpgPass is not properly batched: Instead of collecting all the updates into one large batched update, it creates lots of mini-batches that are applied concurrently to the creation of new updates.

This approach was chosen because of grumbling about peak memory load during these passes, and because odbv1 doesn't make use of the batching anyway. This way, it was thought, odb could overflow parts of the graph during this process, and would never need to keep the entire update in memory at the same time.

In flatgraph, this will expectedly explode into our face.

The correct fix to that is to make ConcurrentWriterCpgPass a type-alias to the standard cpg pass in the ported codepropertygraph version. Back when I wrote ConcurrentWriterCpgPass, I anticipated this situation -- the APIs should be exactly the same, i.e. no code adjustment in any passes should be necessary. (I am not sure whether joern will need to be recompiled or whether this can be fixed by dropping a modified codepropertygraph into the classpath, but we don't need binary compat here anyways)

mpollmeier:
Indeed: simply `s/ConcurrentWriterCpgPass/ForkJoinParallelCpgPass/g` fixes the bottleneck. Thank you!

PR #4227, opened 2024-02-24
[x2cpg] Performance improvements in Overlay Passes

pandurangpatil:
1. Introduced parallelism inside `LinkingUtil.linkToSingle` as well as inside `LinkingUtil.linkToMultiple` which is being called from multiple overlay passes. This has improved the performance of those passes.
2. Similarly introduced parallelism with `StaticCallLinkerPass` along with some refactoring of the code and importantly getting rid of the `methodFullNameToNode` map being created. Which was unnecessary we can query the required data from CPG itself. There were two factors affecting the high memory utilisation in this pass
   a. `methodFullNameToNode` map was taking up too much of memory with the given large repo.
   b. When CPG is getting modified with the edges being created the `DiffGraph` modification recorded was consuming huge memory. So I initially tried dividing the processing-based batching of the CALL nodes to be processed in a number of batches matching the number of processors available on that machine. However, even with these changes it was consuming a huge amount of memory. Then I tried breaking those batches into smaller chunks of 100 records. I could get it working.

TODO: There are still places where we can improve the performance as well as memory footprint in the overall process including finding data flow.


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/StaticCallLinker.scala
```diff
@@ -1,64 +1,40 @@
...
+import io.shiftleft.semanticcpg.language.*
 import org.slf4j.{Logger, LoggerFactory}
 
-import scala.collection.mutable
+class StaticCallLinker(cpg: Cpg) extends ConcurrentWriterCpgPass[Seq[Call]](cpg) {
 
-class StaticCallLinker(cpg: Cpg) extends CpgPass(cpg) {
-
-  import StaticCallLinker._
-  private val methodFullNameToNode = mutable.Map.empty[String, List[Method]]
```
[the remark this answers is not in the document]
pandurangpatil: @DavidBakerEffendi how does this cache help? will the lookup be faster than the querying the DB?

fabsx00:
@bbrehm could you check if this usage of `ConcurrentWriterPass` potentially introduces race conditions or whether you see any other issues here that may introduce flakiness?

pandurangpatil:
> I think the `java.util.ConcurrentModificationException: null` error in `testScripts` references some thread-safety issues this PR introduces

I the failure is pointing at different pass - https://github.com/joernio/joern/actions/runs/8044750384/job/21968839681

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallsToFieldAccessTests.scala
```diff
@@ -40,5 +40,71 @@ class CallsToFieldAccessTests extends KotlinCode2CpgFixture(withOssDataflow = fa
...
+        |class BClass(var acls: AClass){
+        |}
+        |
+        |fun main() {
+        |    val a = AClass("A_MESSAGE")
+        |    val b = BClass(a)
+        |    val m = b.acls.x
+        |}
+        |""".stripMargin)
+    "Create only two fieldAccess nodes" in {
```
DavidBakerEffendi, line 98:
```suggestion
    "Create only four fieldAccess nodes" in {
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/StaticCallLinker.scala
```diff
@@ -1,64 +1,40 @@
...
+    calls.foreach { call =>
       try {
-        linkCall(call, dstGraph)
+        call.dispatchType match {
+          case DispatchTypes.STATIC_DISPATCH | DispatchTypes.INLINED =>
+            val resolvedMethodOption = cpg.method.fullNameExact(call.methodFullName)
+            resolvedMethodOption.foreach(dst => builder.addEdge(call, dst, EdgeTypes.CALL))
+            val size = cpg.method.fullNameExact(call.methodFullName).size
+            // Add the debug logs with number of METHOD nodes found for given method full name
+            if size > 5 then logger.debug(s"Total ${size} METHOD nodes found for -> ${call.methodFullName}")
```
ml86, line 29: Why log at 5? To me it would only make sense to log at >1 or not at all.

pandurangpatil: I agree with you @ml86 let me make this modification

bbrehm:
I like the idea of multi-threading this kind of thing. Unfortunately this doesn't vibe with the API:

1. We don't really want to start multi-threading within a function call; the intended approach is for multi-threading to go through the CpgPass interface. This would not be a blocker, though.
2. There is a problem with race conditions: The typical  `dstNotExistsHandler` mutates the hashmap that backs  `dstNodeMap` (it often also mutates the backing builder!).
3. If you naively add locks / synchronized, then performance will suck.

I don't think it is possible to properly fix this race condition while keeping the API.

But I think it is very possible to rip out linkToSingle and linkToMultiple -- this is a very bad API. Then we can think about a new API that passes can use for this kind of operation that can be properly parallelized.

In cases where such mutations occur, we'd ideally use an immutable dict. Volatile read on read access (no atomics or locks needed), and either atomic rmw-loop or locks for writes.

pandurangpatil:
> 2\. There is a problem with race conditions: The typical  `dstNotExistsHandler` mutates the hashmap that backs  `dstNodeMap` (it often also mutates the backing builder!).

I didn't connect this point, could you please point to the code location to which I can refer to connect the dots?

pandurangpatil:
> 1. We don't really want to start multi-threading within a function call; the intended approach is for multi-threading to go through the CpgPass interface. This would not be a blocker, though.

Agree with what you are pointing at, we would have to change all the passes that are consuming these `linkToSingle` and `linkToMutliple`. Post having a quick discussion with @DavidBakerEffendi we decided to follow this approach.  Let me know if I go ahead and change the passes to make use of ParallelPasses.

pandurangpatil:
> > 2. There is a problem with race conditions: The typical  `dstNotExistsHandler` mutates the hashmap that backs  `dstNodeMap` (it often also mutates the backing builder!).
> 
> I didn't connect this point, could you please point to the code location to which I can refer to connect the dots?

I guess I got the code that you are pointing at.

bbrehm:
>I didn't connect this point, could you please point to the code location to which I can refer to connect the dots?

Your new linkToSingle spawns multiple tasks that run concurrently. These tasks concurrently call the provided  `dstNotExistsHandler` and `dstNodeMap`.

Therefore your construction has only a chance of working if these two functions are threadsafe, for all users of `linkToSingle`. A typical example of `dstNotExistsHandler` looks like `FileCreationPass`:
```
    def createFileIfDoesNotExist(srcNode: StoredNode, destFullName: String): Unit = {
      if (destFullName != File.PropertyDefaults.Name) {
        val dstFullName = if (destFullName == "") { FileTraversal.UNKNOWN }
        else { destFullName }
        val newFile = newFileNameToNode.getOrElseUpdate(
          dstFullName, {
            val file = NewFile().name(dstFullName).order(0)
            dstGraph.addNode(file)
            file
          }
        )
        dstGraph.addEdge(srcNode, newFile, EdgeTypes.SOURCE_FILE)
      }
    }
```
This is eminently not threadsafe. Adding locks around `createFileIfDoesNotExist` does not solve the problem, because of the concurrent read accesses to `newFileNameToNode`. If you add locks to each read access, then performance is shot.

>ConcurrentWriterPass

Don't use ConcurrentWriterPass unless you run out of memory on real machines (>24g heap size) on real projects and have validated that this specific use of ConcurrentWriterPass fixes that issue (and doesn't merely delay it!).

ConcurrentWriterPass will be removed with flatGraph / odbv2 anyways.

pandurangpatil:
> This is eminently not threadsafe. Adding locks around `createFileIfDoesNotExist` does not solve the problem, because of the concurrent read accesses to `newFileNameToNode`. If you add locks to each read access, then performance is shot.

I agree and makes sense. This will introduce the race conditions.

pandurangpatil:
> Then the diffgraph applier goes through the diffgraph -- which only contains edge additions! -- in order, and adds the NewFile at the first moment it is referenced. This is where the linearization / ordering of updates happens: It is racy which thread gets to allocate the NewFile and which gets to reuse it, but we don't care about this, the order of edges in the final diffgraphbuilder is independent of execution order.

Get it perfectly. If I use `ForkJoinParallelCpgPass` processing still can be parallelised.

bbrehm:
>If I use ForkJoinParallelCpgPass processing still can be parallelised.

Yep. I think I would write a baseclass `ParallelLinkerPass extends ForkJoinParallelCpgPass` and use that. This means a major refactor/rewrite of all users of linkToSingle and linkToMultiple.

Generally, a lot of thought has gone into the ForkJoinParallelCpgPass + DiffGraphApplier + NewNode, in order to ensure that output is independent of execution order and to ensure a minimum of synchronization overhead.

Feel free to ask questions about that!

bbrehm:
>The second point of discussion was about using ConcurrentWriterCpgPass for the StaticCallLinker pass. Do you see any race conditions there? 

I see no races.

But don't use ConcurrentWriterCpgPass, this class is the moral equivalent equivalent to sun.misc.Unsafe. Use it if you must, but not without an essay-length comment that explains why it is unavoidable. Instead use ForkJoinParallelCpgPass (it has exactly the same API).

Furthermore, you don't need to batch with ForkJoinParallelCpgPass: Java stream collect is capable of dealing with small workunits.

pandurangpatil:
> But don't use ConcurrentWriterCpgPass, this class is the moral equivalent equivalent to sun.misc.Unsafe. Use it if you must, but not without an essay-length comment that explains why it is unavoidable. Instead, use ForkJoinParallelCpgPass (it has exactly the same API).

Can you elaborate more on why it is unsafe to `ConcurrentWriterCpgPass` if the modifications are independent? Is it from the point of view of developers who use it unknowingly where there could be race conditions? more than that identifying such scenarios is very complicated and painful. Or is there some other reason? 

I am asking this question, as I think we can even go beyond `ConcurrentWriterCpgPass` and write every small modification directly to the graph through a writer queue instead of recording the changes in batches of DiffGraph.

pandurangpatil:
> In short: The thread-safety of overflowdb is fundamentally questionable. It is questionable how safe it is for two threads to concurrently write. It is even questionable how safe it is for one thread to write while another reads, even if it is not the same data (e.g. one thread adds an `astOut` edge while another thread reads a `cfgIn` edge of the same node).

With writes queued and done through a single thread takes care of writing data to the DB without any race conditions. This is by far the best solution for having concurrent processing making use of CPU resources to the fullest and make sure writes are consistent. Even with the odb successor we can follow similar approach of writing to it through a single thread behind the queue and continue using `ConcurrentWriterCpgPass`. 

I guess concerning reads being not consistent is expected if writes and reads are done in parallel. If there is a use case that requires one to make sure the previous writes are done then we need a separate pass which will run after the first pass. 

I just hope you are not getting offended by what I am saying below :). 

I am still not convinced why we shouldn't be using `ConcurrentWriterCpgPass` and use only `ForkJoinParallelCpgPass`. Of course unless there is a race condition situation that you have explained in earlier examples where we should use `ForkJoinParallelCpgPass`.

bbrehm:
>I am still not convinced why we shouldn't be using ConcurrentWriterCpgPass

Why? What do you believe we could gain by using ConcurrentWriterCpgPass?

Memory consumed by the buffer? The line `val sourceNodes = cpg.graph.nodes(srcLabels: _*).toList` already buffers roughly as much memory as the total diffgraph.

>Even with the odb successor we can follow similar approach of writing to it through a single thread behind the queue and continue using ConcurrentWriterCpgPass.

Imagine we had a simple datastructure: Just a collection of items. The only supported update is `append`. A simple implementation could be something like an ArrayList. Easy to append items.

Ah, but our writes are all buffered. Then we can afford to store items in an `Array` that we copy on each update.

This has the negative effect that no database write is ever O(1). This doesn't matter if we are guaranteed to only do < 100 updates over the lifetime of the structure! 

This is, roughly, what we're doing for the successor: Optimize the case of very large buffered updates, fileIO, sequential reads and random reads; give up all other database properties. We can afford that because our program logic is buffered into `CpgPass`es.

mpollmeier:
This might be a silly question but I'd like understand it either way: can we teach `ForkJoinParallelCpgPass` to perform chunking, i.e. having a maximum buffer size for the batched changes, and apply them once it's full? That way we could handle much larger graphs, and we could make it configurable and be able to turn the dial between speed and memory.

As an example, when I run joern-flatgraph 115g heap I can _just about_ import the linux4 kernel, and after all the regular passes have been applied the heap drops to 20g. Linux5 is larger and runs out of heap with the same setup. As discussed above we could switch the very large passes back to `ConcurrentWriterCpgPass`, but if the `ForkJoinParallelCpgPass` supported chunking that would be much better and flexible IMO. Also we wouldn't need to make this judgement for every (current and new) pass.

bbrehm:
Changing ForkJoinPass to do that automatically: I don't think this would be a good idea.

Making a flatgraph-adapted version of ConcurrentWriterPass: This may be necessary. I think I can save some more memory by changing around the internals of the diffgraph applier. Adapting probably means that ConcurrentWriterPass decides on batch-size (i.e. it buffers diffgraphs until they are beyond some threshold and then applies one merged diffgraph). 

Regardless, the passes discussed in this PR do not qualify for ConcurrentWriter. 

First, they generate only very small diffgraphs. No reason to use ConcurrentWriter.

Second, they read from the graph. This is super dangerous wrt races. Ast generation (which is the prime candidate) does not read from the graph.

mpollmeier:
That makes sense, and my comment wasn't really coined to this PR - sorry for derailing the discussion, the previous discussions here simply fit this though very well.

bbrehm:
So, I was asked why ConcurrentWriterPass is so dangerous.

The answer is that some parts of the updates are written while other parts of the pass still run.

Suppose that part1 updates `lineNumber` of node X, and that part2 tries to read the `lineNumber`.

Under ConcurrentWriterPass this is a data race.

If we did heroics, we could guarantee the following: part2 reads either the old or the new lineNumber.

We don't do heroics, and it is possible that part2 gets a NullPointerException: The access checks whether lineNumber is null; it isn't, so it is put into an `Option`, but in the meantime it was set to `null`, and we get `Some(null)`.

To make this safe we would need to make the field `volatile`.

Even worse: If you look into overflowdb, there are all the various accesses to the arrays backing `adjacentNodes`. These array accesses would need various types of `volatile` semantics. Volatile access to array elements only really works well since java9 VarHandle, but odb is java8 (this is a new jvm feature, not a question of the java language). Sure we could go sun.misc.unsafe pointer-slinging.

And even worse: Out code is littered with stuff like `if(node.lineNumber.isDefined) node.lineNumber.get else -1`. This would need to become `val tmp = node.lineNumber; if(tmp.isDefined) tmp.get else -1`. It is utterly impossible to audit our code to a quality that avoids all such pitfalls.

So the general rule is: Only use ConcurrentWriterPass if:
1. It is verified that the specific pass generates too much output to buffer
2. The pass never reads anything from the graph
3. Alternatively to (2), if you know all graph accesses and all overflowdb internals and the JMM. This is very unlikely: We have 0-2 people qualified to do that (both me and @mpollmeier are possible candidates, but both of us are uncomfortable with this. And @mpollmeier wrote odb!).

pandurangpatil:
@DavidBakerEffendi @bbrehm Pushed this PR https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1759 to intercept and show some time, CPU and Memory usage by passes. By default performance matrix collection is disabled, it will only log debug logs when a pass is invoked and ended. It will be helpful for doing some benchmarking of performance tests.

pandurangpatil:
@bbrehm PR is ready with final changes. 

cc: @DavidBakerEffendi @fabsx00

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/base/TypeUsagePassTwo.scala
```diff
@@ -0,0 +1,40 @@
...
+  def generateParts(): Array[List[Node]] = cpg.graph.nodes(srcLabels: _*).toList.grouped(BATCH_SIZE).toArray
+  def runOnPart(builder: DiffGraphBuilder, part: List[overflowdb.Node]): Unit = {
+    newLinkToSingle(
+      cpg = cpg,
+      srcNodes = part,
+      srcLabels = srcLabels,
+      dstNodeLabel = NodeTypes.TYPE,
+      edgeType = EdgeTypes.EVAL_TYPE,
+      dstNodeMap = typeFullNameToNode(cpg, _),
+      dstFullNameKey = "TYPE_FULL_NAME",
```
DavidBakerEffendi, line 35:
```suggestion
      dstFullNameKey = PropertyNames.TYPE_FULL_NAME,
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/base/TypeUsagePassTwo.scala
```diff
@@ -0,0 +1,40 @@
+package io.joern.x2cpg.passes.base
+
+import io.joern.x2cpg.utils.LinkingUtil
+import io.shiftleft.codepropertygraph.Cpg
+import io.shiftleft.codepropertygraph.generated.{EdgeTypes, NodeTypes}
+import io.shiftleft.passes.ForkJoinParallelCpgPass
+import overflowdb.Node
+import overflowdb.traversal.*
+
+class TypeUsagePassTwo(cpg: Cpg) extends ForkJoinParallelCpgPass[List[Node]](cpg) with LinkingUtil {
```
DavidBakerEffendi, line 10:
```suggestion
class TypeEvalPass(cpg: Cpg) extends ForkJoinParallelCpgPass[List[Node]](cpg) with LinkingUtil {
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/base/TypeUsagePassOne.scala
```diff
@@ -0,0 +1,27 @@
+package io.joern.x2cpg.passes.base
+
+import io.joern.x2cpg.utils.LinkingUtil
+import io.shiftleft.codepropertygraph.Cpg
+import io.shiftleft.codepropertygraph.generated.{EdgeTypes, NodeTypes, PropertyNames}
+import io.shiftleft.passes.ForkJoinParallelCpgPass
+import overflowdb.Node
+import overflowdb.traversal.*
+
+class TypeUsagePassOne(cpg: Cpg) extends ForkJoinParallelCpgPass[List[Node]](cpg) with LinkingUtil {
```
DavidBakerEffendi, line 10:
```suggestion
class TypeRefPass(cpg: Cpg) extends ForkJoinParallelCpgPass[List[Node]](cpg) with LinkingUtil {
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/StaticCallLinker.scala
```diff
@@ -1,64 +1,39 @@
...
+class StaticCallLinker(cpg: Cpg) extends ForkJoinParallelCpgPass[Seq[Call]](cpg) {
 
-class StaticCallLinker(cpg: Cpg) extends CpgPass(cpg) {
-
-  import StaticCallLinker._
-  private val methodFullNameToNode = mutable.Map.empty[String, List[Method]]
-
-  override def run(dstGraph: DiffGraphBuilder): Unit = {
+  private val logger: Logger = LoggerFactory.getLogger(classOf[StaticCallLinker])
+  private val BATCH_SIZE     = 100
```
DavidBakerEffendi, line 13: This is quite an arbitrary batch size. Could this perhaps be determined more dynamically based on the number of processors? This value might not be optimized for all CPUs

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/StaticCallLinker.scala
```diff
@@ -1,64 +1,39 @@
...
+            val size = cpg.method.fullNameExact(call.methodFullName).size
+            // Add the debug logs with number of METHOD nodes found for given method full name
+            if size > 1 then logger.debug(s"Total ${size} METHOD nodes found for -> ${call.methodFullName}")
+          case DispatchTypes.DYNAMIC_DISPATCH =>
+          // Do nothing
+          case _ => logger.warn(s"Unknown dispatch type on dynamic CALL ${call.code}")
+        }
       } catch {
         case exception: Exception =>
           throw new RuntimeException(exception)
```
DavidBakerEffendi, line 35: We should probably not throw an exception here. I know this is existing code, but could you replace this with an error log?

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/StaticCallLinker.scala
```diff
@@ -1,64 +1,39 @@
...
+  }
 
-    cpg.call.foreach { call =>
+  override def runOnPart(builder: DiffGraphBuilder, calls: Seq[Call]): Unit = {
+    calls.foreach { call =>
       try {
-        linkCall(call, dstGraph)
+        call.dispatchType match {
+          case DispatchTypes.STATIC_DISPATCH | DispatchTypes.INLINED =>
+            val resolvedMethodOption = cpg.method.fullNameExact(call.methodFullName)
```
DavidBakerEffendi, line 24: This is of type `Iterator` not `Option`. Rather place this into memory with `.l` and then you can use this same variable later when you check the size. Right now you're making this query twice when it could be only done once.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/LinkingUtil.scala
```diff
@@ -1,28 +1,23 @@
...
+import java.util.concurrent.Callable
 import scala.collection.mutable
-import scala.jdk.CollectionConverters._
+import scala.jdk.CollectionConverters.*
+import scala.util.{Failure, Success}
 
 trait LinkingUtil {
 
   import overflowdb.BatchedUpdate.DiffGraphBuilder
+  val BATCH_SIZE = 100
```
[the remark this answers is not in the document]
DavidBakerEffendi: Another arbitrary batch size which could be determined differently. Why have two batch sizes? If this is to be inherited then you can use the `protected` modifier.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/LinkingUtil.scala
```diff
@@ -41,11 +36,9 @@ trait LinkingUtil {
...
     cpg.graph.indexManager.lookup(PropertyNames.FULL_NAME, x).asScala
 
-  /** For all nodes `n` with a label in `srcLabels`, determine the value of `n.\$dstFullNameKey`, use that to lookup the
-    * destination node in `dstNodeMap`, and create an edge of type `edgeType` between `n` and the destination node.
-    */
-  def linkToSingle(
+  def newLinkToSingle(
     cpg: Cpg,
+    srcNodes: List[Node],
     srcLabels: List[String],
```
DavidBakerEffendi, line 42: This parameter is no longer used, it can be removed

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/LinkingUtil.scala
```diff
@@ -41,11 +36,9 @@ trait LinkingUtil {
   def nodesWithFullName(cpg: Cpg, x: String): mutable.Seq[NodeRef[_ <: NodeDb]] =
     cpg.graph.indexManager.lookup(PropertyNames.FULL_NAME, x).asScala
 
-  /** For all nodes `n` with a label in `srcLabels`, determine the value of `n.\$dstFullNameKey`, use that to lookup the
-    * destination node in `dstNodeMap`, and create an edge of type `edgeType` between `n` and the destination node.
-    */
-  def linkToSingle(
+  def newLinkToSingle(
```
DavidBakerEffendi, line 39: Could this not be overridden? `newLinkToSingle` is not a descriptive name. 
```suggestion
  def linkToSingle(
```

DavidBakerEffendi (changes requested):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/LinkingUtil.scala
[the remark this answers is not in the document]
pandurangpatil: I tried creating batches based on the number of available cores on the machine with few combinations. It was translating into batches becoming too large. 

Resulting in two unwanted situations.
1. DiffGraph builder buffer becoming large consuming more memory than required for that period.
2. Towards the end one of the batches would take more time and in turn not provide the parallelism for that period.

I chose a random number 100 considering it to be not too small and not too large. Which will make batches in chunks. I could see that reduced the memory requirement as well as the time it took to finish the processing.

I will remove one from StaticCallLinker and use it from LinktingUtil itself.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/LinkingUtil.scala
pandurangpatil: It's getting used at line no 52.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/LinkingUtil.scala
pandurangpatil: I didn't change all the passes, so kept the original method as is. I am not sure whether it gets used inside private repos or not so chose to keep the original method as is. 

I will change other passes from where this method is being called.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/LinkingUtil.scala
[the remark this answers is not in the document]
DavidBakerEffendi: > I chose a random number 100 considering it to be not too small and not too large

This is still relative to the machine you're testing on though. I'm just not really convinced that this is a universal best. e.g. if there are 200 tasks on a 4 core machine, two cores will be idle.

Perhaps we can compromise on something like `Math.min(tasks / numProcessors, 100)`

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/StaticCallLinker.scala
```diff
@@ -1,64 +1,39 @@
...
+import io.shiftleft.codepropertygraph.generated.nodes.*
 import io.shiftleft.codepropertygraph.generated.{DispatchTypes, EdgeTypes}
-import io.shiftleft.passes.CpgPass
-import io.shiftleft.semanticcpg.language._
+import io.shiftleft.passes.{ConcurrentWriterCpgPass, ForkJoinParallelCpgPass}
+import io.shiftleft.semanticcpg.language.*
 import org.slf4j.{Logger, LoggerFactory}
 
-import scala.collection.mutable
+class StaticCallLinker(cpg: Cpg) extends ForkJoinParallelCpgPass[Seq[Call]](cpg) {
```
DavidBakerEffendi, line 10: Also, why not use `Method` as the part? If method is the part you can query `cpg.call.methodFullNameExact(method.fullName)` exactly once for each method, whereas you may query your existing `cpg.method.fullNameExact(call.methodFullName)` potentially multiple times. This would avoid the need for a batch size at all.

DavidBakerEffendi: While a batch size would be more consistent for each thread, does it outweigh the cost of the additional graph reads? I suppose it's debatable

DavidBakerEffendi: I would say so yes, maybe do the check on `finish`, example of a check you could add to detect multiple methods for one static call.
```
override def finish(): Unit = {
   val invalidCallRelations = cpg.call.count(x => x.dispatchType == STATIC_DISPATCH && x.callee.size > 1)
   if (invalidCallRelations > 0) {
      logger.warn(s"Invalid CPG structure detected, as $invalidCallRelations static calls have multiple callees")
   }
}
```

mpollmeier:
Sorry again for taking over part of the discussion here - for future reference I spawned a new discussion ticket for this here: https://github.com/joernio/joern/issues/4256

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/LinkingUtil.scala
```diff
@@ -41,11 +34,16 @@ trait LinkingUtil {
   def nodesWithFullName(cpg: Cpg, x: String): mutable.Seq[NodeRef[_ <: NodeDb]] =
     cpg.graph.indexManager.lookup(PropertyNames.FULL_NAME, x).asScala
 
+  protected def getBatchSize(totalItems: Int): Int =
+    Math.max(totalItems / ConcurrentTaskUtil.MAX_POOL_SIZE, MAX_BATCH_SIZE)
```
DavidBakerEffendi, line 38: Why `max`? This would choose `totalItems / ConcurrentTaskUtil.MAX_POOL_SIZE` instead of `MAX_BATCH_SIZE` if `totalItems / ConcurrentTaskUtil.MAX_POOL_SIZE > MAX_BATCH_SIZE`

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/StaticCallLinker.scala
```diff
@@ -1,64 +1,51 @@
...
+import io.shiftleft.codepropertygraph.generated.nodes.*
 import io.shiftleft.codepropertygraph.generated.{DispatchTypes, EdgeTypes}
-import io.shiftleft.passes.CpgPass
-import io.shiftleft.semanticcpg.language._
+import io.shiftleft.passes.ForkJoinParallelCpgPass
+import io.shiftleft.semanticcpg.language.*
 import org.slf4j.{Logger, LoggerFactory}
 
-import scala.collection.mutable
+implicit val resolver: ICallResolver = NoResolve
```
DavidBakerEffendi, line 11: Since this only gets used on line 45, instead of introducing a global val, you could do `x.callee(NoResolve).size`

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/StaticCallLinker.scala
```diff
@@ -1,64 +1,51 @@
...
+class StaticCallLinker(cpg: Cpg) extends ForkJoinParallelCpgPass[Seq[Method]](cpg) with LinkingUtil {
```
DavidBakerEffendi, line 12: I suppose in my initial comment, I meant `ForkJoinParallelCpgPass[Method]` in an unbatched way

DavidBakerEffendi: But I guess this works too at reducing query count

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/callgraph/StaticCallLinker.scala
```diff
@@ -1,64 +1,51 @@
...
+    val allMethods = cpg.method.l
+    allMethods.grouped(getBatchSize(allMethods.size)).toArray
+  }
 
-    cpg.call.foreach { call =>
+  override def runOnPart(builder: DiffGraphBuilder, methods: Seq[Method]): Unit = {
+    methods.foreach { method =>
       try {
-        linkCall(call, dstGraph)
+        cpg.call
```
DavidBakerEffendi, line 24: In hindsight, we can already optimize this by filtering in the dispatch types directly.
```scala
cpg.call
   .dispatchType(DispatchTypes.STATIC_DISPATCH, DispatchTypes.INLINED)
   .methodFullNameExact
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/LinkingUtil.scala
pandurangpatil: Typing mistake

DavidBakerEffendi:
@pandurangpatil it appears we don't need to batch when using `ForkJoin` pass https://github.com/joernio/joern/issues/4256

pandurangpatil:
> @pandurangpatil it appears we don't need to batch when using `ForkJoin` pass #4256

@DavidBakerEffendi batching will help reduce the time process has to invest rescheduling smaller chunks of tasks again and again. It is as good as saying max batch size = 1.

DavidBakerEffendi:
@mpollmeier @pandurangpatil I'm observing `_evalOut` returning `null` in `codescience` tests and when I revert this commit then this appears to be fixed. I'm considering reverting this, as it is a silent failure until one does something like `call.receiver.typ` and finds nothing.

PR #4229, opened 2024-02-26
[ruby] Bug Fix - Generate Implicit Return for IndexAccess

### LINKED ISSUE #4230: [ruby][Bug] Implicit Return node for IndexAccess not generated
Sample code does not generate return node: 
```ruby
def foo
  l = [1,2,3]
  l[1]
end
```

AndreiDreyer:
This PR includes:
 * generate implicit return for index access

Resolves #4230

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodReturnTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodReturnTests.scala
@@ -106,6 +106,24 @@ class MethodReturnTests extends RubyCode2CpgFixture {
     c.methodFullName shouldBe Operators.arrayInitializer
   }
 
+  "implicit RETURN node for index access exists" in {
+    val cpg = code("""
+                     |def f
+                     | l = [1, 2, 3]
+                     | l[1]
+                     |end
+                     |""".stripMargin)
+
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodReturnTests.scala
```diff
@@ -106,6 +106,24 @@ class MethodReturnTests extends RubyCode2CpgFixture {
...
+                     |""".stripMargin)
+
+    val List(f)         = cpg.method.name("f").l
+    val List(r: Return) = f.methodReturn.cfgIn.l: @unchecked
+
+    r.code shouldBe "l[1]"
+    r.lineNumber shouldBe Some(4)
+
+    val List(c: Call) = r.astChildren.isCall.l
+    c.methodFullName shouldBe Operators.indexAccess
```
DavidBakerEffendi, line 124: Just one thing I want you to double check as well, could you make sure that the arguments of `Operators.indexAccess` are `1` and `2`? I think there is a bug where it will be `0` and `1` because the `l` for example is set to the receiver. `Operators.*` don't have receivers generally, and a common mistake in frontend dev is where `field` and `index` operators use a receiver. I think the mistake is made here and may be logged somewhere.

DavidBakerEffendi (approved):

PR #4237, opened 2024-02-27
[c#] Add more built-in types

tuxology:
We have used a way to generate built-in types using DLLs and PDB symbols for some key packages. This PR adds support to populate types from multiple JSON files. These files have been generated using the new feature added in `DotNetAstGen` via https://github.com/joernio/DotNetAstGen/pull/20


joern-cli/frontends/csharpsrc2cpg/src/main/resources/builtin_types/Microsoft.AspNetCore.Mvc/builtin_types.json
```diff
@@ -0,0 +1,69 @@
...
+              "System.Action`1<Microsoft.AspNetCore.Mvc.MvcOptions>"
+            ]
+          ],
+          "isStatic": true
+        }
+      ],
+      "fields": []
+    },
+    {
+      "name": "Microsoft.Extensions.DependencyInjection.MvcServiceCollectionExtensions/FrameworkAssemblyPart",
```
DavidBakerEffendi, line 36: When I look for the types suffixed with `/` I don't seem to find them in the documentation, but I do find the parent. https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.mvcservicecollectionextensions?view=aspnetcore-7.0

Perhaps we can shorten these collections by removing those with `/` components?

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpProgramSummary.scala
```diff
@@ -51,6 +55,30 @@ object CSharpProgramSummary {
...
+      InputStream.nullInputStream()
+    } else {
+      val mergedJsonObjects = ListBuffer[ujson.Obj]()
+      for (resourcePath <- resourcePaths) {
+        val inputStream = classLoader.getResourceAsStream(resourcePath)
+        val jsonString  = Source.fromInputStream(inputStream).mkString
+        val jsonObject  = ujson.read(jsonString).obj
+        mergedJsonObjects += jsonObject
+      }
+      val mergedJson = write(mergedJsonObjects.map(_.value).reduce(_ ++ _))
```
DavidBakerEffendi, line 77: We may want to go for a `reduceOption` in case `mergedJsonObjects` remains empty

DavidBakerEffendi (commented):
Have we checked if all these types/methods are public? I think we can filter out private/internal methods.

joern-cli/frontends/csharpsrc2cpg/src/main/resources/builtin_types/Microsoft.AspNetCore.Mvc/builtin_types.json
tuxology: Added this filter in https://github.com/joernio/DotNetAstGen/pull/20. Will update this shortly. You can merge  https://github.com/joernio/DotNetAstGen/pull/20 🙏

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpScope.scala
```diff
@@ -29,7 +30,11 @@ class CSharpScope(summary: CSharpProgramSummary)
   }
 
   override def isOverloadedBy(method: CSharpMethod, argTypes: List[String]): Boolean = {
-    method.parameterTypes.filterNot(_._1 == "this").size == argTypes.size
+    method.parameterTypes.filterNot(_._1 == "this").size == argTypes.size && method.parameterTypes
+      .filterNot(_._1 == "this")
+      .map(_._2)
+      .zip(argTypes)
+      .count({ case (x, y) => x != y }) == 0
```
DavidBakerEffendi, line 37: We can't make this guarantee because of polymorphism. I suggest we comment this out until the scope object can understand whether an argument is a valid subtype of a parameter

joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/ControlStructureTests.scala
```diff
@@ -25,7 +25,7 @@ class ControlStructureTests extends CSharpCode2CpgFixture {
             case Some(exp: Call) =>
               exp.code shouldBe "new Exception(\"Error!\")"
               exp.name shouldBe Defines.ConstructorMethodName
-              exp.typeFullName shouldBe "Exception" // TODO: Should really be System.Exception
+              exp.typeFullName shouldBe "System.InsufficientMemoryException"
```
DavidBakerEffendi, line 28: This doesn't seem valid?

DavidBakerEffendi (changes requested):

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpScope.scala
karan-batavia: If we comment this out, other methods (ex - Console.WriteLine) which have multiple overloads will pick up the first available type it gets from the TypeMap.

joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/ControlStructureTests.scala
karan-batavia: https://github.com/Privado-Inc/joern/blob/35698e1b283485fb65cb0f68afa5244f64a85111/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala#L94

This line picks up all the types in scope, which end with the typeName parameter. What is the best way to handle this?

karan-batavia: Maybe, we can match on the last part ( part meaning types split on "." ) ?

joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpScope.scala
DavidBakerEffendi: Then I suggest we have some kind of fall-back, otherwise we're going to lose coverage. e.g.
1. Try match on exact types
2. If no result, then try match only on param count

joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/ControlStructureTests.scala
DavidBakerEffendi: @karan-batavia yes let's do that, match the last elements in each respective split

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -91,6 +91,7 @@ trait TypedScope[M <: MethodLike, F <: FieldLike, T <: TypeLike[M, F]](summary:
   def tryResolveTypeReference(typeName: String): Option[T] = {
     typesInScope
       .collectFirst {
+        case typ if typ.name.split("[.]").lastOption.getOrElse("").equals(typeName)       => typ
         case typ if typ.name.endsWith(typeName)                                           => typ
```
karan-batavia, line 95: This check was still needed incase we encounter `fullyQualifiedName`.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -91,6 +91,7 @@ trait TypedScope[M <: MethodLike, F <: FieldLike, T <: TypeLike[M, F]](summary:
   def tryResolveTypeReference(typeName: String): Option[T] = {
     typesInScope
       .collectFirst {
+        case typ if typ.name.split("[.]").lastOption.getOrElse("").equals(typeName)       => typ
```
DavidBakerEffendi, line 94: `typ.name.split("[.]").lastOption.contains(typeName)` would be more concise (for next time)

DavidBakerEffendi (approved):

joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/ControlStructureTests.scala
DavidBakerEffendi, line 28: I think this should change to `System.Exception` right?

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
DavidBakerEffendi: Ah wait, if you always assume `typeName` is not qualified then yes, but if you make the first check:
```scala
typ.name.split("[.]").lastOption == typeName.split("[.]").lastOption
```
then it should work, but we would want to also eventually want some way to match something like
`Foo.Bar.Baz` with `Bar.Baz` (partially qualified if `using Bar;` was qualified at the top. So for that we would want to look for some intersection of the suffixing parts, but let's leave a TODO for now

DavidBakerEffendi (approved):

PR #4240, opened 2024-02-27
add some comments on common pitfalls

bbrehm:
This adds some fixme notes on fundamental architecture/API/performance issues I spotted.

I also started to write down some of the general pitfalls. How do we want these general pitfalls documented @fabsx00 ?

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
@@ -43,7 +43,7 @@ object ProgramSummary {
   /** Combines two namespace-to-type maps.
     */
   def combine[T <: TypeLike[_, _]](a: Map[String, Set[T]], b: Map[String, Set[T]]): Map[String, Set[T]] = {
-    //fixme: Use mutable datatypes, otherwise folds are quadratc
+    // fixme: Use mutable datatypes, otherwise folds are quadratc
     val accumulator = mutable.HashMap.from(a)
 
     b.keySet.foreach(k =>
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
@@ -43,6 +43,7 @@ object ProgramSummary {
   /** Combines two namespace-to-type maps.
     */
   def combine[T <: TypeLike[_, _]](a: Map[String, Set[T]], b: Map[String, Set[T]]): Map[String, Set[T]] = {
+    //fixme: Use mutable datatypes, otherwise folds are quadratc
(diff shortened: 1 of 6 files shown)


DavidBakerEffendi (approved):
Lol, I'm always happy to provide the example of code for people to learn what **not** to do 😉 

Perhaps we can make an entry in the Joern documentation? I suppose a general "Developer Guidelines" section so that it doesn't get hidden and forgotten.

In fact, seems like there is no developer guidelines page on the [doc website](https://github.com/joernio/website/tree/master/docs.joern.io/content), only some on the [Joern README](https://github.com/joernio/joern/blob/62e31dd05e7f631a35277cbe58af4d0f4f8bed18/README.md?plain=1#L83)

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -43,6 +43,7 @@ object ProgramSummary {
   /** Combines two namespace-to-type maps.
     */
   def combine[T <: TypeLike[_, _]](a: Map[String, Set[T]], b: Map[String, Set[T]]): Map[String, Set[T]] = {
+    //fixme: Use mutable datatypes, otherwise folds are quadratc
```
DavidBakerEffendi, line 46: Is this more suggesting that we'd want something more like `DiffGraphBuilder.absorb`?

maltek:
Maybe some of the discussion on time complexity of collection classes can be replaced with a link to the Scala docs on the topic? (I found these very useful when I finally decided I should actually know where each of these types is appropriate.)

https://docs.scala-lang.org/overviews/collections-2.13/concrete-immutable-collection-classes.html
https://docs.scala-lang.org/overviews/collections-2.13/performance-characteristics.html

(These links go to the Scala 2.13 docs, because the Scala 3 book [still links there, too.](https://docs.scala-lang.org/scala3/book/collections-classes.html#more-details))

DavidBakerEffendi:
Before this branch gets ignored into oblivion, I've made a candidate entry on the Joern docs https://github.com/joernio/website/pull/109

bbrehm:
>Lol, I'm always happy to provide the example of code for people to learn what not to do 😉

Don't feel bad about it -- concurrency is hard, lazy iterators are really scary API, and I am also giving examples from the java standard library like `Collectors.toList()`, so you're in very good company ;)

DavidBakerEffendi:
@pandurangpatil as per https://github.com/joernio/joern/issues/4256, it seems like don't need to batch if `ForkJoin` pass is used.

DavidBakerEffendi:
> Yeah, let's put it in the docs. The chances of it being read there are higher.

@bbrehm then are you happy with https://github.com/joernio/website/pull/109? I'll give it another comb through first then add you as a reviewer once I feel it's ready

DavidBakerEffendi:
It's been on my backlog for a while, but will be implementing the mutable program summary maps and concurrent util stream update today

DavidBakerEffendi:
Closed by https://github.com/joernio/joern/pull/4620

PR #4242, opened 2024-02-27
Fix issue that resulted in error messages being used as namespace names

johannescoetzee:
This PR fixes an issue that was introduced in https://github.com/joernio/joern/pull/2831, where an error message in the unsolved symbol exception name would sometimes be appended to a wildcard.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/CallTests.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/CallTests.scala
@@ -12,6 +12,22 @@ import overflowdb.traversal.toNodeTraversal
 
 class NewCallTests extends JavaSrcCode2CpgFixture {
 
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/CallTests.scala
```diff
@@ -12,6 +12,22 @@ import overflowdb.traversal.toNodeTraversal
...
+        |class Test {
+        |
+        |  void test() {
+        |    foo().bar();
+        |  }
+        |}
+        |""".stripMargin)
+
+    cpg.call.name("foo").typeFullName.l shouldBe List("ANY")
+    cpg.call.name("foo").methodFullName.l shouldBe List("Test.foo:<unresolvedSignature>(0)")
```
johannescoetzee, line 28: In the fullName here, we probably shouldn't assume `foo` is a method in `Test`, but this is working as "intended" and should be fixed separately

max-leuthaeuser (approved):

PR #4244, opened 2024-02-28
[ruby] handling of self Singleton methods

AndreiDreyer:
This PR includes:
 * Handling of `self` Singleton methods for both cases:
```ruby
class C
  class << self
    def foo(x)
      # do something
    end
  end
end
```
```ruby
class C
  def self.foo(x)
    # do something
  end
end
```

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
@@ -3,8 +3,9 @@ package io.joern.rubysrc2cpg.astcreation
 import io.joern.rubysrc2cpg.astcreation.RubyIntermediateAst.*
 import io.joern.rubysrc2cpg.datastructures.{ConstructorScope, MethodScope}
 import io.joern.rubysrc2cpg.passes.Defines
+import io.joern.x2cpg.utils.NodeBuilders.newThisParameterNode
 import io.joern.x2cpg.{Ast, ValidationMode, Defines as XDefines}
-import io.shiftleft.codepropertygraph.generated.{EvaluationStrategies, NodeTypes, Operators}
+import io.shiftleft.codepropertygraph.generated.{EvaluationStrategies, NodeTypes}
 import io.shiftleft.codepropertygraph.generated.nodes.{NewLocal, NewMethodParameterIn, NewTypeDecl}
 
 trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { this: AstCreator =>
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -219,4 +193,42 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
...
+    )(TextSpan(None, None, None, None, ""))
+  }
+
+  private def astForMethodBody(body: RubyNode, optionalStatementList: StatementList): Ast = {
+    if (this.parseLevel == AstParseLevel.SIGNATURES) {
+      Ast()
+    } else {
+      body match
+        case stmtList: StatementList =>
+          astForStatementListReturningLastExpression(
```
DavidBakerEffendi, line 218: https://github.com/joernio/joern/pull/4241 made some changes in this behaviour, is this code duplicated? May need to reflect the similar level of implicit return behaviour in editted in #4241

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
```diff
@@ -152,8 +152,74 @@ class MethodTests extends RubyCode2CpgFixture {
...
+                     |""".stripMargin)
+
+    inside(cpg.typeDecl.name("C").l) {
+      case classC :: Nil =>
+        inside(classC.method.name("f").l) {
+          case funcF :: Nil =>
+            inside(funcF.parameter.l) {
+              case thisParam :: xParam :: Nil =>
+                thisParam.code shouldBe "this"
+                thisParam.typeFullName shouldBe "Test0.rb:<global>::program.C:f"
```
DavidBakerEffendi, line 176: The `this` parameter type should be the class, so only `Test0.rb:<global>::program.C`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
```diff
@@ -656,10 +656,42 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
...
+            singletonClassDeclaration.baseClass match {
+              case Some(selfIdentifier: SelfIdentifier) =>
+                singletonClassDeclaration.body match {
+                  case singletonClassStmtList: StatementList =>
+                    singletonClassStmtList.statements.map {
+                      case method: MethodDeclaration =>
+                        SingletonMethodDeclaration(selfIdentifier, method.methodName, method.parameters, method.body)(
+                          method.span
+                        )
+                      case nonMethodStatement => nonMethodStatement
```
DavidBakerEffendi, line 683: Have you investigated if something like 
```ruby
 class << self
   x = 1
 end
```
Needs to be lowered to `self.x = 1`?

DavidBakerEffendi: If this is the case, we can leave a TODO and create a separate issue to handle it

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
DavidBakerEffendi: Ignore :) my changes are in `astsForImplicitReturnStatement`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
AndreiDreyer: Created separate issue as instance and class members are not handled yet. #4246

DavidBakerEffendi (approved):

PR #4249, opened 2024-02-28
[C#] Fix not found exception on jar path for builtin_types

karan-batavia:
A `FileNotFoundException` was thrown for the `builtin_types` directory in the C# frontend, when the frontend was invoked from the release/local build. Turns out it is a age old bug in Java itself. This is a small workaround to solve this problem. 

Bug: https://bugs.java.com/bugdatabase/view_bug?bug_id=4523159


DavidBakerEffendi (approved):
Ah yes I should've caught this, I've experienced it before

karan-batavia:
> Ah yes I should've caught this, I've experienced it before
@DavidBakerEffendi Is this the right way of solving this?

DavidBakerEffendi:
@karan-batavia yes,loading via class loader as streams should ensure resources in the JAR are available to you

Issue #3939, opened 2023-12-11
[ruby] Import Nodes & Require Support


DavidBakerEffendi:
The require statements need to be handled as `Import` nodes. Note that this issue does not relate to `RubyImportResolverPass`

PR #4250, opened 2024-02-29
[ruby] Import Nodes & Require Support #3939

badly-drawn-wizards:
Add the import nodes in a new pass. The X2Cpg generic pass requires an assignment node for the importAs, but this doesn't fit the Ruby import mechanics which behave more like c `#includes` with dedup.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/ImportsPass.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/ImportsPass.scala
@@ -7,8 +7,6 @@ import io.shiftleft.passes.ConcurrentWriterCpgPass
 import io.joern.x2cpg.Imports.createImportNodeAndLink
 import io.joern.x2cpg.X2Cpg.stripQuotes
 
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/ImportsPass.scala
```diff
@@ -0,0 +1,26 @@
...
+  override def generateParts(): Array[Call] = cpg
+    .call(importCallName)
+    .toArray
+
+  override def runOnPart(diffGraph: DiffGraphBuilder, call: Call): Unit = {
+    val importedEntity = stripQuotes(call.argument.isLiteral.code.l match {
+      case s :: _ => s
+      case _      => ""
+    })
+    val importedAs = ""
```
DavidBakerEffendi, line 23: `importedAs` == `importedEntity`. When that is not the case, then it is usually imported under some alias, e.g.

`import foo as bar` or, as it's used in the Java world `import bar.Foo` can be accessed as `Foo`.

DavidBakerEffendi (approved):

PR #4282, opened 2024-03-06
[pysrc2cpg] Added auto-discovery for venv directories

max-leuthaeuser:
- directories with a `pyvenv.cfg` file are ignored if `ignoreVenvDir == true`
- nothing is ignored if `ignoreVenvDir == false`
- if `ignoreVenvDir == true` and there is at least one folder set via old `--venvDir` or new `--venvDirs`: only these are ignored; no auto-discovery then

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/Py2CpgOnFileSystem.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/Py2CpgOnFileSystem.scala
@@ -75,7 +75,7 @@ class Py2CpgOnFileSystem extends X2CpgFrontend[Py2CpgOnFileSystemConfig] {
           ignoredFilesPath = Option(config.ignoredFiles)
         )
         .map(x => Path.of(x))
-        .filter { file => filterEnvDir(config, file) }
+        .filter { file => filterEnvDir(config, file, inputPath) }
         .filter { file => filterIgnoreDirNames(file, inputPath, ignoreDirNamesSet) }
         .filter { file => !absoluteIgnorePaths.exists(ignorePath => file.startsWith(ignorePath)) }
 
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/Py2CpgOnFileSystem.scala
```diff
@@ -93,6 +97,22 @@ class Py2CpgOnFileSystem extends X2CpgFrontend[Py2CpgOnFileSystemConfig] {
     }
   }
 
+  private def filterEnvDir(config: Py2CpgOnFileSystemConfig, file: Path, inputPath: Path): Boolean = {
```
maltek, line 100: just nit-picking, but I find this function a bit hard to follow: I had to look at the place where this is called to figure out what the return values actually mean. I think it might be clearer if we flip the returned bool around and rename the function to `isAutoDectedVenv`? (And adjust the call-site to use `filterNot`.)

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/config/VenvExcludeTests.scala
```diff
@@ -0,0 +1,124 @@
...
+        Set(
+          ".venv/a1.py",
+          ".venv/sub/s1.py",
+          ".venv2/a2.py",
+          ".venv2/sub/s2.py",
+          "folder/b.py",
+          "folder/c.py",
+          "main.py"
+        )
+      ),
```
maltek, line 75: doesn't really matter, but we could write these tests as

```scala
"exclude nothing if ignoreVenvDir is false" should {
   assert(testWithArguments(false, Seq.empty, Seq(...)))
  }
}
```

should be a little shorter in the end, and result in somewhat nicer errors when the tests fail

maltek (approved):

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/Py2CpgOnFileSystem.scala
max-leuthaeuser: I wanted to keep it consistent to `filterIgnoreDirNames` down below (also their callsites).
But you are right, I guess both could benefit from a small refactoring there for readibility.

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/config/VenvExcludeTests.scala
maltek: OK, leave it as-is then.

I've had some problems figuring out which table entry was failing for some scalatest tests in the past, so my knee-jerk reaction is to avoid them. But this might very well not apply here.

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/Py2CpgOnFileSystem.scala
max-leuthaeuser: I implemented the refactoring you suggested.

maltek (approved):

Issue #4188, opened 2024-02-19
[c#] Download Dependency Option


DavidBakerEffendi:
Following #4180, we can use the information to query the [NuGet API](https://www.nuget.org/packages/ICSharpCode.Decompiler) and obtain library DLLs. The API documentation is here: https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource

This would need to be enabled by a false-by-default CLI argument as is present in some other frontends.

PR #4293, opened 2024-03-07
[c#] Download & Summarise Dependencies

DavidBakerEffendi:
* Added download dependencies CLI argument
* If downloading of dependencies is enabled, then these packages are downloaded and unzipped from NuGet
* Calling DotNetAstGen to summarize results and merges these with internal summaries

TODO: Choose the version of DLL to summarize more intelligently. If a package publishes binaries with more than one dot net version, the first is taken.
TODO: Extend `DotNetAstGen` to parse DLLs even when PDB files are not present. Right now dependencies without PDB files are skipped.

Resolves [#4188](https://github.com/joernio/joern/issues/4188)

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/utils/DependencyDownloader.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/utils/DependencyDownloader.scala
@@ -1,29 +1,31 @@
 package io.joern.csharpsrc2cpg.utils
 
 import better.files.File
+import io.joern.csharpsrc2cpg.Config
 import io.joern.csharpsrc2cpg.datastructures.CSharpProgramSummary
-import io.joern.semanticcpg.utils.SecureXmlParsing
+import io.joern.x2cpg.astgen.AstGenRunner.DefaultAstGenRunnerResult
 import io.shiftleft.codepropertygraph.generated.Cpg
 import io.shiftleft.codepropertygraph.generated.nodes.Dependency
 import io.shiftleft.semanticcpg.language.*
 import org.slf4j.LoggerFactory
 
-import java.io.{BufferedReader, FileOutputStream, InputStreamReader}
-import java.net.{HttpURLConnection, URI, URL, URLConnection}
+import java.io.FileOutputStream
+import java.net.{HttpURLConnection, URI, URLConnection}
 import java.util.zip.{GZIPInputStream, InflaterInputStream, ZipEntry}
 import scala.util.{Failure, Success, Try, Using}
-import scala.xml.Elem
 
-/** Queries NuGet for the artifacts of a programs' dependencies.
+/** Queries NuGet for the artifacts of a programs' dependencies, according to the V2 API.
+  *
+  * TODO: Note that NuGet has API throttling, so beware of that if this is to be processed concurrently.
   *
   * @see
   *   <a href="https://learn.microsoft.com/en-us/nuget/api/overview">NuGet API</a>
   */
 class DependencyDownloader(
   cpg: Cpg,
+  config: Config,
   internalProgramSummary: CSharpProgramSummary,
-  nugetRepositoryBaseUrl: String = "www.nuget.org",
-  nugetApiVersion: String = "v2"
+  nugetRepositoryApi: String = "www.nuget.org/api/v2"
 ) {
 
   private val logger = LoggerFactory.getLogger(getClass)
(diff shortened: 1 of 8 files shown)


joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/utils/DependencyDownloader.scala
```diff
@@ -0,0 +1,186 @@
...
+    *   the directory to download to.
+    * @param dependency
+    *   the dependency to download.
+    * @return
+    *   a disposable version of the directory where the dependencies live.
+    */
+  private def downloadDependency(targetDir: File, dependency: Dependency): Unit = {
+    Seq("package", "symbolpackage")
+      .map(packageType =>
+        URI(s"https://$nugetRepositoryApi/$packageType/${dependency.name}/${dependency.version}").toURL
```
tuxology, line 70: In relation to https://github.com/joernio/joern/pull/4295 we might get no versions - this is actually a common case I am seeing (eg. https://github.com/dotnet-architecture/eShopOnWeb/blob/main/src/ApplicationCore/ApplicationCore.csproj)  When this happens, the `package` gets downloaded fine because NuGet API redirects to latest package. However we get a 400 since `symbolpackage` expects exact `$dependency.version`. Here is an example: 
 
400 : https://www.nuget.org/api/v2/symbolpackage/System.IdentityModel.Tokens.Jwt/
200: https://www.nuget.org/api/v2/symbolpackage/System.IdentityModel.Tokens.Jwt/7.4.0

One workaround could be to first download the nupkg, get it's version (which is latest) and then append to the `symbolpackage` URL and re-download.

A better approach could be to do something like this: https://api.nuget.org/v3-flatcontainer/System.IdentityModel.Tokens.Jwt/index.json and fetch package versions and then pick last one (Ref: https://learn.microsoft.com/en-us/nuget/api/package-base-address-resource#enumerate-package-versions) Note, this is using v3 API and to me seems a more robust way.

tuxology (changes requested):
A few changes are needed to handle the empty version issues which I found quite common in code I was trying. 😿 Feel free to bring minor changes in #4295 in this PR and close that one. The test I was trying was this and we can add this in this PR as well: 

```
 diff --git a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/passes/DependencyPass.scala b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/passes/DependencyPass.scala
index 250b8ea4c..3458281d9 100644
--- a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/passes/DependencyPass.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/passes/DependencyPass.scala
@@ -25,7 +25,7 @@ class DependencyPass(cpg: Cpg, buildFiles: List[String]) extends ForkJoinParalle
             case packageReference if packageReference.label == "PackageReference" =>
               Try {
                 val packageName    = packageReference.attribute("Include").map(_.toString()).get
-                val packageVersion = packageReference.attribute("Version").map(_.toString()).get
+                val packageVersion = packageReference.attribute("Version").map(_.toString()).getOrElse("")
                 val dependencyNode = NewDependency()
                   .name(packageName)
                   .version(packageVersion)
diff --git a/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/io/DependencyDownloadTests.scala b/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/io/DependencyDownloadTests.scala
index 3769a5690..baf526433 100644
--- a/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/io/DependencyDownloadTests.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/io/DependencyDownloadTests.scala
@@ -16,6 +16,7 @@ class DependencyDownloadTests extends CSharpCode2CpgFixture {
         |    <ItemGroup>
         |      <PackageReference Include="CommandLineParser" Version="2.9.1" />
         |      <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.0.0" />
+        |      <PackageReference Include="System.IdentityModel.Tokens.Jwt" />
         |    </ItemGroup>
         |
         |    <ItemGroup>
@@ -37,6 +38,10 @@ class DependencyDownloadTests extends CSharpCode2CpgFixture {
       summary.typesUnderNamespace("CommandLine.Text") should not be empty
     }

+    "summarize Jwt which has no version but publishes with PDB files" in {
+      summary.typesUnderNamespace("System.IdentityModel.Tokens.Jwt") should not be empty
+    }
+
     "summarize CommandLineParser's transient dependencies" in {
       summary.typesUnderNamespace("RailwaySharp.ErrorHandling") should not be empty
       summary.typesUnderNamespace("CSharpx") should not be empty
```

DavidBakerEffendi:
@tuxology I'm working on cleaning this issue up today

tuxology (approved):

PR #4295, opened 2024-03-07
[c#] minor dependency resolution fix

tuxology:
Allows deps with no version number to still populate


DavidBakerEffendi (approved):
Does the dependency downloader still generate valid URLs/ default to latest version? If so, looks good

tuxology:
> Does the dependency downloader still generate valid URLs/ default to latest version? If so, looks good but maybe adding a versionless dependency to the existing test would be good too

Good point. Let me add a test. I ran this against a repo and almost all repos are getting 400. This could be the issue. Let me investigate more and improve this. Will also add a test

PR #4296, opened 2024-03-08
3939 ruby import nodes require support

badly-drawn-wizards:
- Adds new RequireCall and ImportCall nodes
- Adds pathToTypes to RubyProgramSummary that map require paths
- Handles require_relative, require and include with `addImported*` methods.

What has yet to be done
- tryResolveTypeReference should be changed to use the new information. Unit tests written now mostly pass trivially unless some ambiguous resolution needs to be made. They will mostly act as regression.
- Nested modules are not handled (either in the summary or both the summary and the main pass, not sure)  
- Unit tests require multi-file cpgs to processed together so that program summary has all information, which appears not be the the case
  - as a workaround, I can directly invoke the frontend on a test folder in a unit test for now, but not ideal 

TODO accidentally committed ImportsPass, or not rebased properly, will fix


badly-drawn-wizards:
Accidentally used old branch

Issue #3940, opened 2023-12-11
[ruby] Import Resolution


DavidBakerEffendi:
This issue relates to `RubyImportResolverPass` and depends on [#3939](https://github.com/joernio/joern/issues/3939).

If there is something like string concatentation, it is fine to leave that out for now. This is mostly about resolving paths to internal code (i.e., something we have an internal CPG node for). See `PythonImportResolverPass` for a good example on this.

PR #4297, opened 2024-03-08
[ruby] Import Resolution #3940

badly-drawn-wizards:
- Adds new RequireCall and ImportCall nodes
- Adds pathToTypes to RubyProgramSummary that map require paths
- Handles require_relative, require and include with `addImported*` methods.

What has yet to be done
- tryResolveTypeReference should be changed to use the new information. Unit tests written now mostly pass trivially unless some ambiguous resolution needs to be made. They will mostly act as regression.
- Nested modules are not handled (either in the summary or both the summary and the main pass, not sure)  
- Unit tests require multi-file cpgs to processed together so that program summary has all information, which appears not be the the case
  - as a workaround, I can directly invoke the frontend on a test folder in a unit test for now, but not ideal

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/RubyIntermediateAst.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/RubyIntermediateAst.scala
@@ -208,7 +208,7 @@ object RubyIntermediateAst {
   final case class StaticLiteral(typeFullName: String)(span: TextSpan) extends RubyNode(span) with LiteralExpr {
     def isSymbol: Boolean = text.startsWith(":")
 
-    def isString: Boolean = text.startsWith("\"")
+    def isString: Boolean = text.startsWith("\"") || text.startsWith("'")
 
     def innerText: String = text match
       case s":'$content'" => content
@@ -256,6 +256,20 @@ object RubyIntermediateAst {
   final case class SimpleCall(target: RubyNode, arguments: List[RubyNode])(span: TextSpan)
       extends RubyNode(span)
       with RubyCall
+
+  final case class RequireCall(target: RubyNode, argument: RubyNode, isRelative: Boolean)(span: TextSpan)
+      extends RubyNode(span)
+      with RubyCall {
+    def arguments: List[RubyNode] = List(argument)
+    def asSimpleCall: SimpleCall = SimpleCall(target, arguments)(span)
+  }
+
+  final case class IncludeCall(target: RubyNode, argument: RubyNode)(span: TextSpan)
+      extends RubyNode(span)
+      with RubyCall {
+    def arguments: List[RubyNode] = List(argument)
+    def asSimpleCall: SimpleCall = SimpleCall(target, arguments)(span)
+  }
 
   /** Represents standalone `proc { ... }` or `lambda { ... }` expressions
     */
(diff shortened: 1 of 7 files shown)


DavidBakerEffendi:
Your last point seems odd, the files should be processed together, what observation makes you think otherwise?

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/datastructures/RubyProgramSummary.scala
```diff
@@ -4,18 +4,26 @@ import io.joern.x2cpg.datastructures.{FieldLike, MethodLike, ProgramSummary, Typ
...
+    RubyProgramSummary(
+      ProgramSummary.combine(this.namespaceToType, other.namespaceToType),
+      ProgramSummary.combine(this.pathToType, other.pathToType)
+    )
   }
 
 }
 
-case class RubyMethod(name: String, parameterTypes: List[(String, String)], returnType: String) extends MethodLike
+case class RubyMethod(name: String, parameterTypes: List[(String, String)], returnType: String) extends MethodLike {
```
DavidBakerEffendi, line 25: These brackets seem redundant

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/RubyIntermediateAst.scala
```diff
@@ -208,7 +208,7 @@ object RubyIntermediateAst {
   final case class StaticLiteral(typeFullName: String)(span: TextSpan) extends RubyNode(span) with LiteralExpr {
     def isSymbol: Boolean = text.startsWith(":")
 
-    def isString: Boolean = text.startsWith("\"")
+    def isString: Boolean = text.startsWith("\"") || text.startsWith("'")
```
DavidBakerEffendi, line 211: Perhaps something more robust
```scala
"^['\\\"]\\w*['\\\"]$".r.findFirstMatchIn(text).isDefined
```

DavidBakerEffendi (commented):
Good solution so far, I like it. Now to figure out those tests

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/datastructures/RubyProgramSummary.scala
badly-drawn-wizards: Vestigial from previous experimentation, will remove

badly-drawn-wizards:
> Your last point seems odd, the files should be processed together, what observation makes you think otherwise?

I was confused with the way TestCpg was implemented with by constructing the graph lazily. I think I made the same assumption a few months back as well and totally forgot I made the same mistake earlier.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/RubyIntermediateAst.scala
badly-drawn-wizards: I agree this smells and is dodgy, but since the parsing of the static literal has already happened in ANTLR, there's no need to do it again. Unless there's a static literal in ruby other than single-quoted strings that begins with `'`.

badly-drawn-wizards:
Added unit tests that test for resolving ambigious types/methods for requires and includes. Will give better descriptive names later.

DavidBakerEffendi (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -108,15 +111,13 @@ trait TypedScope[M <: MethodLike, F <: FieldLike, T <: TypeLike[M, F]](summary:
     * @return
     *   the method meta data if found.
     */
-  def tryResolveMethodInvocation(
-    callName: String,
-    argTypes: List[String],
-    typeFullName: Option[String] = None
+  def tryResolveMethodInvocation(callName: String, argTypes: List[String], typeFullName: Option[String] = None)(implicit
+    tag: ClassTag[M]
```
DavidBakerEffendi, line 115: Is this tag used? My IDE says no

PR #4299, opened 2024-03-08
[C#] Ast creation for `CastExpression`

### LINKED ISSUE #4300: [C#] CastExpression in C#
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast

karan-batavia:
Includes Ast Creation and in turn type resolution for Cast Expression. Type coverage should increase as a result of this PR.

Resolves #4300


joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstCreator.scala
```diff
@@ -101,7 +101,6 @@ class AstCreator(
       case CatchDeclaration                                      => astForCatchDeclaration(nodeInfo)
       case PropertyDeclaration                                   => astForPropertyDeclaration(nodeInfo)
       case ExpressionElement                                     => astForExpressionElement(nodeInfo)
-      case _: LiteralExpr                                        => astForLiteralExpression(nodeInfo)
```
karan-batavia, line 104: Removed this, as the statement below would anyway pass control to the same method.

DavidBakerEffendi (approved):

PR #4315, opened 2024-03-12
[c#] builtins volume 2

tuxology:
* downloaded DLLs from nuget
 * generated PDBs and symbols using ilspycmd and DotNetAstGen


DavidBakerEffendi:
I see the total size is about 21mb, and maybe we should start zipping these files and unzipping with Java ZIP streams?

DavidBakerEffendi (approved):

PR #4320, opened 2024-03-12
Add String/Regexp operator =~

badly-drawn-wizards:
- Treat `=~` as a member call.
- I'm a bit uncomfortable with not having a constant for "=~", but not sure where to put it.
- Unit tests are a bit bare, but not sure what else needs to be checked that is not done for member calls in general.


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -73,7 +73,8 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
       "|"   -> Operators.or,
       "^"   -> Operators.xor,
       "<<"  -> Operators.shiftLeft,
-      ">>"  -> Operators.logicalShiftRight
+      ">>"  -> Operators.logicalShiftRight,
+      "=~"  -> "=~"
```
DavidBakerEffendi, line 77: You can place a constant for this under `io.joern.rubysrc2cpg.passes.Defines`

DavidBakerEffendi (approved):

PR #4321, opened 2024-03-12
[ruby] Object Instantiation Type Hint

DavidBakerEffendi:
Add a type hint to the LHS identifier of an object instantiation assignment to be hinted that the type would be that of the created object.

Question @max-leuthaeuser, is this a good place to use `dynamicTypeHintFullName`? Or should I stick to `possibleTypes`?


max-leuthaeuser:
Looks like its an actual type fullname. And not the kind of garbage we sometimes generate for TS/JS so I guess its fine.

Issue #4318, opened 2024-03-12
[ruby] Splatting not implemented for `when` in `case` Statements


DavidBakerEffendi:
The following test emits `Unknown` nodes on the splatting arguments.
```ruby
case 0
  when 0 
    0
  when 1,2 then 1
  when 3, *[4,5] then 2
  when *[6] then 3
  else 4
end
```
One can approximate this as all being children to `arrayInitializer` operators most likely, as our `case` variable is undergoing a membership check which is similar to `__builtin.Array.contains` in some way

PR #4328, opened 2024-03-13
[ruby] implement splatting for when in case statements (#4318)

badly-drawn-wizards:
For cases not matching against an expression  use `.any?` and with it we use `.include? expr`. This is not entirely faithful to ruby semantics, as matching is done with `.===` but doing so would be uglier with not much upside.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
@@ -140,7 +140,7 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
                   List(MemberCall(exprList, ".", "any?", List())(splat.span))
                 }
             case e =>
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -119,15 +119,28 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
...
+              expr
+                .map { e =>
+                  List(MemberCall(exprList, ".", "include?", List(e))(exprList.span))
+                }
+                .getOrElse {
+                  List(MemberCall(exprList, ".", "any?", List())(exprList.span))
+                }
             case e =>
-              logger.warn("Splatting not implemented for `when` in ruby `case`")
+              logger.warn("Unrecognised RubyNode in case match splat expression")
```
DavidBakerEffendi, line 143:
```suggestion
              logger.warn(s"Unrecognised RubyNode (${e.getClass}) in case match splat expression")
```

PR #4340, opened 2024-03-15
[ruby] Initial handling of Instance Variables (Dynamic Fields)

AndreiDreyer:
This PR handles:
 * Instance variables (`@var`).
 * The general structure for instance variables:
```ruby
class Foo
  def foo
    @a = 10
  end
end
```
->
```ruby
class Foo
  @a

  def <init>
    @a = nil
  end

  def foo
    @a = 10
  end
end
```
TODO:
 * Handle the case where instance variables are assigned in method bodies where the method body is not a `StatementList`
 * Filter out instance variables that are already assigned a default value in the constructor of the class. The current implementation does not account for this, so you end up with a double assignment in the constructor:
```ruby
class Foo
  def initialize
    @a = 10
  end
end
```
->
```ruby
class Foo
  @a
  def <init>
    @a = nil
    @a = 10
  end
```


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
```diff
@@ -693,11 +693,76 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
...
+      }
+      .collect { case x: InstanceFieldIdentifier =>
+        x
+      }
+  }
+
   override def visitClassDefinition(ctx: RubyParser.ClassDefinitionContext): RubyNode = {
+    val loweredClassDecls = lowerSingletonClassDeclarations(ctx.bodyStatement())
+
+    val (stmts, fields) = loweredClassDecls match {
```
DavidBakerEffendi, line 712: Could we refactor this new block into a private field?

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -28,23 +26,52 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
+
+    node match {
+      case instanceField: InstanceFieldIdentifier =>
+        scope.findFieldInScope(name) match {
+          case None =>
+            val fullName = s"${scope.surroundingTypeFullName.getOrElse(Defines.UnresolvedNamespace)}.$name"
+            scope.pushField(FieldDecl(name, Defines.Any, false, false, node))
+            astForFieldAccess(
+              MemberAccess(
+                DummyNode(identifierNode(instanceField, "this", "this", Defines.Any))(
```
DavidBakerEffendi, line 38: Could we refactor `"this"` into a constant? Perhaps under `io.joern.rubysrc2cpg.passes.Defines`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -77,9 +78,20 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this:
         methodDecl ++ bodyAsts
       case bodyAsts => bodyAsts
     }
+
+    val fieldMemberNodes = node match {
+      case classDecl: ClassDeclaration =>
+        classDecl.fields.map { x =>
+          val name     = code(x)
+          val fullName = s"${scope.surroundingTypeFullName.getOrElse(XDefines.UnresolvedNamespace)}.$name"
```
DavidBakerEffendi, line 86: Is `fullName` used? From the GH review panel it doesn't seem so

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ClassTests.scala
```diff
@@ -389,4 +390,75 @@ class ClassTests extends RubyCode2CpgFixture {
...
+        |
+        | def barfoo
+        |   puts @a
+        |   puts @c
+        |   @o = "a"
+        | end
+        |end
+        |""".stripMargin)
+
+    "be correct" in {
```
DavidBakerEffendi, line 416:
```suggestion
    "create respective member nodes" in {
```

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ClassTests.scala
```diff
@@ -389,4 +390,75 @@ class ClassTests extends RubyCode2CpgFixture {
...
+              // Test that all members in class are present
+              aMember.code shouldBe "@a"
+              bMember.code shouldBe "@b"
+              cMember.code shouldBe "@c"
+              dMember.code shouldBe "@d"
+              oMember.code shouldBe "@o"
+            case _ => fail("Expected 5 members")
+          }
+
+          inside(fooType.method.name(Defines.ConstructorMethodName).l) {
```
DavidBakerEffendi, line 430: The block below could be under a different heading like `"create nil assignments under the class initializer" in {`

DavidBakerEffendi (changes requested):

DavidBakerEffendi (approved):

PR #4344, opened 2024-03-15
[javasrc2cpg] Add PackageRootFinder and use that to determine delombok input paths

johannescoetzee:
Delombok expects a package root directory or filename as input otherwise it just outputs the unmodified source code, so the main purpose of this PR is to fix that. To do so, I've added the `PackageRootFinder` which finds all package root directories for input files we care about (so not filtered out by our excludes) by comparing package statements to directory structure. For projects that break the package name === project structure convention, it just returns filenames for the non-conforming files (which may end up slow for large such projects, but it should be rare).

I've also extracted the test code writing functionality previously in `TestCpg` into a separate trait to allow re-use by tests that need the test code, but not the cpg (in this case tests for the `PackageRootFinder`)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
```diff
@@ -30,6 +30,7 @@ class JavaSrc2Cpg extends X2CpgFrontend[Config] {
       new MetaDataPass(cpg, language, config.inputPath).createAndApply()
       val astCreationPass = new AstCreationPass(config, cpg)
       astCreationPass.createAndApply()
+      astCreationPass.sourceParser.cleanupDelombokOutput()
```
johannescoetzee, line 33: I tried `delombokTempDir.deleteOnExit` first, but for some reason that didn't work. Probably some unclosed `File/Reader/Writer`, but I couldn't find it.

max-leuthaeuser (approved):

PR #4352, opened 2024-03-18
[Gosrc2cpg] download dependencies and caching improvements 

pandurangpatil:
1. Removed some unwanted brackets
2. Parallelised downloading of the dependencies and processing them.
3. Unit tests for the optimisation changes to improve the processing time as well as reduce the memory footprint.
4. Handling for used dependencies
       - Handling for the downloading of dependencies only if those are getting used in the main code.
       - While doing optimisation, came across a bug where if more than one package with the same name was created in the code. Then it was creating package level `TypeDecl` and `NamspaceBlock` only once. Introduced a few unit tests which cover these use cases as well as made the respective handling for the same.
5. Changes to not cache unwanted imports
       - Made changes to not cache unwanted imports from dependency source code.
       - Made changes to cache only used imports in source code with all the non-aliased imports to global cache and aliased ones in the context of file `AstCreator`.
       - Caching only those packages whose package name is different from the enclosing folder name inside the global cache.
6. Changes to not cache non-used lambdaTypeInformation.
7. Optimisations to store method meta data along with struct type metadata
       - Changed the storage structure to minimize the amount of data being stored for method meta data cache and struct type members type information.
       - Made respective changes to fix all the breaking unit tests.


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/DownloadDependenciesPass.scala
```diff
@@ -25,21 +28,25 @@ class DownloadDependenciesPass(parentGoMod: GoModHelper, goGlobal: GoGlobal, con
...
-                  case Success(_) =>
-                    print(". ")
-                    processDependency(dependencyStr)
-                  case Failure(f) =>
-                    logger.error(s"\t- command '${cmd}' failed", f)
+            val futures = mod.dependencies
+              .filter(dep => !dep.beingUsed)
+              .map(dependency => {
+                Future {
+                  val dependencyStr = s"${dependency.module}@${dependency.version}"
```
DavidBakerEffendi, line 38: I suppose `go get` will handle this, but one issue that comes to mind is rate limiting of the dependency APIs

DavidBakerEffendi: Would also be good to have a version of this where the dependencies are downloaded directly from the APIs like C# and Ruby so that we don't require `go` as a dependency

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/GoSrc2Cpg.scala
```diff
@@ -22,31 +22,36 @@ import io.shiftleft.codepropertygraph.generated.Languages
...
-class GoSrc2Cpg extends X2CpgFrontend[Config] {
+class GoSrc2Cpg(goGlobalOption: Option[GoGlobal] = Some(GoGlobal())) extends X2CpgFrontend[Config] {
   private val report: Report = new Report()
 
+  private var goMod: Option[GoModHelper] = None
   def createCpg(config: Config): Try[Cpg] = {
     withNewEmptyCpg(config.outputPath, config) { (cpg, config) =>
       File.usingTemporaryDirectory("gosrc2cpgOut") { tmpDir =>
-        val goGlobal = GoGlobal()
+        val goGlobal = goGlobalOption.getOrElse(GoGlobal())
```
DavidBakerEffendi, line 32: This might be cleaner
```suggestion
        goGlobalOption.getOrElse(GoGlobal()).foreach { goGlobal =>
```

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/DownloadDependenciesPass.scala
pandurangpatil: > I suppose go get will handle this, but one issue that comes to mind is rate limiting of the dependency APIs


`go get` doesn't parallelise it, I did a test to check whether it improves the time it takes to process. I do see the improvements.

pandurangpatil: > Would also be good to have a version of this where the dependencies are downloaded directly from the APIs like C# and Ruby so that we don't require go as a dependency

I think that will take time, I will handle that through a separate ticket.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/GoSrc2Cpg.scala
pandurangpatil: @DavidBakerEffendi following will work `goGlobalOption.foreach { goGlobal => ... }`  but that will not help. I can't call `.foreach` after  call `getOrElse(GoGlobal())`

DavidBakerEffendi (approved):

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/GoSrc2Cpg.scala
pandurangpatil: done

Issue #3928, opened 2023-12-11
[ruby] `do-end` Blocks (with implicit `&Proc` and `yield` handling)


DavidBakerEffendi:
Example of some Ruby gymnastics
```
irb(main):051:1* def test
irb(main):052:2*   def lol
irb(main):053:2*     yield
irb(main):054:1*   end
irb(main):055:1*   lol do return 1 end
irb(main):056:1*   nil
irb(main):057:0> end
=> :test
irb(main):058:0> test
=> 1
```
In the above code, we define a function `test`, with a child closure `lol` that contains a `yield`.

`yield`, in this context, would expect `lol` to implicitly pass a `do`-block for it to handle. The `do`-block in this case has a `return` which then returns within the `test`-context rendering the `nil` as "dead" code. 

We can assume that `do` arguments of this kind may not be given as a method reference, so it will always be defined within this scope. With this information, we can always first evaluate the argument, and then determine if we need to consider that `lol` will cause a `return` node to be prepended in the control-flow or not.

However, `lol` is a function which may be defined elsewhere in the program, so we may not know that it has this `yield` operation that may lead to this weird behaviour.

Since Joern generally evaluates control-structures in a path-insensitive way, we may as well model both cases. We can then convert

```
lol do return 1 end
nil
```

To something like

```
<stmt> -> return lol do return 1 end
   |----> return nil
```

This way, we can track both cases in the data-flow.

@badly-drawn-wizards and I have chatted offline about this, and the above is more-or-less what we decided would be a sound way to represent this.

PR #4359, opened 2024-03-18
[Ruby] do-end Blocks (with implicit &Proc and yield handling) #3928

badly-drawn-wizards:
- Adds yield RubyNode
- Handles yield in RubyNodeCreator
- Generates a ProcParameter with a unique identifier if a yield is used in a method with no block parameter
- Generates a call to the block parameter for yields


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -365,6 +369,19 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
     astForSimpleCall(node.asSimpleCall)
   }
 
+  protected def astForYield(node: YieldExpr): Ast = {
+    scope.useProcParam match {
+      case Some(param) =>
+        astForMemberCall(
+          MemberCall(SimpleIdentifier()(node.span.spanStart(param)), ".", "call", node.arguments)(node.span)
```
DavidBakerEffendi, line 376: Could we lower this differently? My concern is that this is closer to reflection in how this is executed, and not sure that our call graph could resolve this.

Could this instead be a `SimpleCall`?

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -76,12 +79,18 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
       astForMethodBody(node.body, optionalStatementList)
     }
 
+    val anonProcParam = scope.anonProcParam.map { param =>
+      val paramNode = ProcParameter(param)(node.span.spanStart(s"&$param"))
+      val index = parameterAsts.lastOption.flatMap(_.root).map { case m: NewMethodParameterIn => m.index }.getOrElse(0)
+      astForParameter(paramNode, index)
```
DavidBakerEffendi, line 85: Should this not add one?
```suggestion
      astForParameter(paramNode, index + 1)
```

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/datastructures/RubyScope.scala
```diff
@@ -117,6 +118,42 @@ class RubyScope(summary: RubyProgramSummary, projectRoot: Option[String])
...
+    case scope @ ScopeElement(MethodScope(fullName, param, _), variables) =>
+      (ScopeElement(MethodScope(fullName, param, true), variables), param.fold(x => x, x => x))
+  }
+
+  def anonProcParam: Option[String] = stack.collectFirst { case ScopeElement(MethodScope(_, Left(param), true), _) =>
+    param
+  }
+
+  def setProcParam(param: String): Unit = updateSurrounding {
+    case scope @ ScopeElement(MethodScope(fullName, _, _), variables) =>
```
badly-drawn-wizards, line 153: TODO remove unused `scope @`s

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -268,8 +290,20 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
 
         val stmtBlockAst = astForMethodBody(node.body, optionalStatementList)
 
+        val anonProcParam = scope.anonProcParam.map { param =>
+          val paramNode = ProcParameter(param)(node.span.spanStart(s"&$param"))
+          val index =
+            parameterAsts.lastOption.flatMap(_.root).map { case m: NewMethodParameterIn => m.index }.getOrElse(1)
```
DavidBakerEffendi, line 296: This `getOrElse` is at `1`, while the above is at `0`, is this intentional? I would think, on the condition that you should append `1` to index at `astForParameter`, that the truth should be `0`

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ProcParameterAndYieldTests.scala
```diff
@@ -0,0 +1,53 @@
...
+          forAll(cpgs)(_.call.code("yield").receiver.isCall.argument.code.l shouldBe List("<proc-param-0>", "call"))
+        }
+
+        "add a block argument" in {
+          forAll(cpgs)(_.method("foo").parameter.code("&.*").name.l shouldBe List("<proc-param-0>"))
+        }
+      }
+    }
+    "that don't have a yield nor a proc parameter" should {
+      val cpg1 = code("""| def foo() end
```
DavidBakerEffendi, line 41: These code blocks are single line, perhaps we could make it more concise with a normal string? e.g. `code("def foo() end")`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
DavidBakerEffendi: Also, concerning flows, the yield needs to map to `MethodReturn`, how is this done in `pysrc2cpg`/should we add a `return` here?

DavidBakerEffendi (changes requested):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
badly-drawn-wizards: I should add that to the test.

badly-drawn-wizards: But the off by one error is not.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
badly-drawn-wizards:
```scala
      case node: ast.Yield          => unhandled(node)
      case node: ast.YieldFrom      => unhandled(node)
```
it isn't handled there, but they would need to be handled vastly differently anyway.

DavidBakerEffendi (approved):
Looks good, let's follow your suggestion (from our chat offline) to add some kind of custom conditional to handle the case where yields can return or generate a collection, etc.

PR #4369, opened 2024-03-19
[C#] Handle chained calls and fields with conditional access and null forgiving operators

### LINKED ISSUE #4370: [C#] Handle chained calls and field accesses when null forgiving and conditional accesses are used.
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-

Calls and fields can be chained.

karan-batavia:
Includes changes to handle chained calls and field accesses when a conditional access or a null  forgiving operator is used.
Resolves #4370

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/MemberAccessTests.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/MemberAccessTests.scala
@@ -153,4 +153,125 @@ class MemberAccessTests extends CSharpCode2CpgFixture {
       }
     }
   }
+
+  "conditional method access expressions for chained calls" should {
+    val cpg = code("""
+        |namespace Foo {
+        | public class Baz {
+        |   public int Qux() {}
+        |   public Baz Fred(int a) {}
+        | }
+        | public class Bar {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/MemberAccessTests.scala
```diff
@@ -153,4 +153,125 @@ class MemberAccessTests extends CSharpCode2CpgFixture {
...
+        |   }
+        | }
+        |}
+        |""".stripMargin)
+
+    "have correct types and attributes both on the LHS and RHS" in {
+      inside(cpg.assignment.l.sortBy(_.lineNumber).drop(1).l) {
+        case a :: Nil =>
+          inside(a.argument.l) {
+            case (lhs: Identifier) :: (rhs: Call) :: Nil =>
```
DavidBakerEffendi, line 177: None of these tests validate the calls chained on the RHS. Please add these, then I'm happy to merge

DavidBakerEffendi (changes requested):

DavidBakerEffendi (approved):

PR #4371, opened 2024-03-20
[c#] Avoid Downloading an Existing Summary & Improve Summary Merging

### LINKED ISSUE #4368: [C#] Do not summarize dependency if already summarized
Currently we override any `builtins` summarization if we encounter the same package in a `.csproj` file through download dependencies. We want the builtins ( or already present dependencies ) to take precedence.

DavidBakerEffendi:
* Adds a check to existing summaries if existing namespacing starting with the dependency name exists.
* Added `TypeLike` addition handler to handle member collisions.
* Upgraded DotNetAstGen version

Resolves #4368

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
@@ -2,6 +2,7 @@ package io.joern.x2cpg.datastructures
 
 import io.shiftleft.codepropertygraph.generated.nodes.DeclarationNew
 
+import scala.annotation.targetName
 import scala.collection.immutable.Map
 import scala.collection.mutable
 import scala.reflect.ClassTag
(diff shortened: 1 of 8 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -299,6 +308,38 @@ trait TypeLike[M <: MethodLike, F <: FieldLike] {
...
+  @targetName("add")
+  def +(o: TypeLike[M, F]): TypeLike[M, F]
+
+  /** Helper method for creating the sum of two type-like's methods, while preferring this types' methods on collisions.
+    * @param o
+    *   the other type-like.
+    * @return
+    *   the combination of the two type-like's methods.
+    */
+  protected def mergeMethods(o: TypeLike[M, _]): List[M] = {
```
karan-batavia, line 326: Can we have a test where a `builtin` type is used, but the same dependency is also present in the `.csproj` file?

karan-batavia: Example would be the `ModelBuilder` class in `Microsoft.EntityFrameworkCore`.

karan-batavia (approved):

AndreiDreyer (approved):

Issue #4351, opened 2024-03-18
[ruby] Method Alias Handling


DavidBakerEffendi:
[This line on `sendgrid`](https://github.com/sendgrid/sendgrid-ruby/blob/main/lib/sendgrid/helpers/mail/category.rb#L16) is giving a syntax error.
```ruby
    alias category= name=
```

Seems that the `alias` command is aliasing a method name. This style of method name seems to be a popular choice for field setters. e.g.
```ruby
class Foo
    @x = 0
    def x=(z)
        @x = z
    end
    
    def x
        @x
    end
end

foo = Foo.new

foo.x= 1

puts foo.x # => 1
```

Given that identifiers can't have equals, we should perhaps make a parser rule for this `alias` keyword. Something like `AliasExpression` with it's own parser rule. More information: https://medium.com/rubycademy/alias-in-ruby-bf89be245f69.

In terms of the CPG structure, we can simply duplicate the signature of the aliased method and make the alias method simply pass on the parameters to the aliased method.

PR #4373, opened 2024-03-20
[ruby] Method Aliases

DavidBakerEffendi:
* Fixed `=` syntax error in method names
* Made sure `FieldIdentifier`s can be implicitly returned
* Created method aliases which are method declarations themselves which forward parameters to the aliasing call.

Resolves [#4351](https://github.com/joernio/joern/issues/4351)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
```diff
@@ -807,8 +805,57 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
...
+      methodParamMap.get(alias.oldName) match {
+        case Some(aliasingMethodParams) =>
+          MethodDeclaration(
+            alias.newName,
+            aliasingMethodParams,
+            StatementList(
+              SimpleCall(
+                SimpleIdentifier(None)(alias.span.spanStart(alias.oldName)),
+                aliasingMethodParams.map { x => SimpleIdentifier(None)(alias.span.spanStart(x.span.text)) }
+              )(alias.span) :: Nil
```
DavidBakerEffendi, line 835: TODO: I Should fix these span texts

AndreiDreyer (approved):

DavidBakerEffendi:
It appears calls that look like `x=1,2,3` and multi-assignments are being (rightly) confused, so this needs further massaging on the parser end

PR #4374, opened 2024-03-20
[ruby] non-`self` singleton methods

### LINKED ISSUE #3935: [ruby] Non-`self` Singleton Methods
[This article](https://www.geeksforgeeks.org/ruby-singleton-methods/) gives a decent introduction to singleton methods.

This may need quite a nuanced approach to be soundly and concisely represented, and I think we can look to Java proxying as an example for a solution:

* We generate an anonymous type that is a subtype of the object being overridden with a singleton method. (Perhaps type name + some suffix)
* The object's type should then be set to this type.

With the above solution, we can have a subtype with only the singleton method, and it would inherit the other methods from the supertype.

The main challenge is that we need to know the type of the object - may need some type propagation. We can always make some anon type with `inheritsFrom` as the name of the object and handle this is some post-processing pass after type propagation.

AndreiDreyer:
This PR handles
 * Module methods
eg:
```ruby
module F
  def F.bar(x)
    x
  end
end

F::bar(p)
```

resolves #3935


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
```diff
@@ -256,4 +256,55 @@ class MethodTests extends RubyCode2CpgFixture {
...
+              thisParam.name shouldBe "F"
+              thisParam.typeFullName shouldBe "Test0.rb:<global>::program.F"
+
+              xParam.name shouldBe "x"
+            case xs => fail(s"Expected two parameters, got ${xs.name.mkString(", ")}")
+          }
+
+          inside(baz.parameter.l) {
+            case thisParam :: xParam :: Nil =>
+              thisParam.name shouldBe "F"
```
DavidBakerEffendi, line 288:
```suggestion
              thisParam.name shouldBe "this"
```
Then the `.code` can be `F`

DavidBakerEffendi (changes requested):

DavidBakerEffendi (approved):

PR #4400, opened 2024-03-26
[ruby] Handling nodes directly under typedecl

### LINKED ISSUE #4395: [ruby] Field Declarations Directly Under TypeDecl
While work was done to declare field members and their default initializers, the original declarations are still being made under the type decl instead of in an init method.

AndreiDreyer:
This PR Handles:
 * Added new trait `AllowedTypeDeclerationChild` for all nodes that are allowed to be directly under `TypeDecl`. 
 * Moved all nodes at start of class definition not allowed under `TypeDecl` into the `<init>` method.
 * `astForConstructorMethodBody` parses method body for constructor without the implicit return.

Resolves #4395

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
@@ -774,6 +774,11 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
           x
         }
 
+        val (scopeDecls, stmtListWithoutScopeDecls) = rest.partition {
+          case x: SimpleCall if x.target.span.text == "scope" => true
+          case _                                              => false
+        }
+
         val fieldsInMethodDecls = findFieldsInMethodDecls(methodDecls)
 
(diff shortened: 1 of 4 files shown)


AndreiDreyer:
Lots of tests failing, investigating.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
```diff
@@ -774,6 +774,11 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
           x
         }
 
+        val (allowedTypeDeclChildren, nonAllowedTypeDeclChildren) = rest.partition {
```
DavidBakerEffendi, line 777:
```suggestion
        val (allowedTypeDeclChildren, nonAllowedTypeDeclChildren) = rest.partition(_.isInstanceOf[AllowedTypeDeclarationChild])
```

DavidBakerEffendi: Given where this is being filtered, it may be filtering out `AliasStatements` before they are lowered. Perhaps this partitioning should happen right before the `TypeDecl` node is repacked

DavidBakerEffendi (commented):
LGTM, just going to test on `codescience`

DavidBakerEffendi (approved):
All good

PR #4404, opened 2024-03-26
[swiftsrc2cpg] Implemented Actor/Async tests

max-leuthaeuser:
Also put in proper method signatures and fullnames.

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/AsyncSyntaxTests.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/AsyncSyntaxTests.scala
@@ -3,28 +3,48 @@
 package io.joern.swiftsrc2cpg.passes.ast
 
 import io.joern.swiftsrc2cpg.testfixtures.AstSwiftSrc2CpgSuite
-
-import io.joern.swiftsrc2cpg.testfixtures.AstSwiftSrc2CpgSuite
+import io.shiftleft.codepropertygraph.generated.*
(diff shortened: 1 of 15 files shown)


joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/AsyncSyntaxTests.scala
```diff
@@ -3,28 +3,48 @@
...
       val cpg = code("""
         |typealias AsyncFunc1 = () async -> ()
         |typealias AsyncFunc2 = () async throws -> ()
         |typealias AsyncFunc3 = (_ a: Bool, _ b: Bool) async throws -> ()
         |""".stripMargin)
-      ???
+      val List(f1, f2, f3) = cpg.typeDecl.where(_.aliasTypeFullName).l
+      f1.name shouldBe "AsyncFunc1"
+      f1.fullName shouldBe "Test0.swift:<global>:AsyncFunc1"
+      f1.aliasTypeFullName shouldBe Option("Test0.swift:<global>:_anon_cdecl_0")
```
maltek, line 36: seems this isn't new in this PR, but it would be good if these generated names had some invalid characters in them, to guarantee that they can't clash with names from actual code. (Like with `<operator>` and `<lambda>`.)

maltek (approved):

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/AsyncSyntaxTests.scala
max-leuthaeuser: Will do.

PR #4406, opened 2024-03-27
[console] Fixed handling of single files when importing JS code

max-leuthaeuser:
This came up already multiple times here as issue and also via Discord.

This PR adds support for importing single JS files by copying them into a temporary folder.

### DIFF
--- a/console/src/main/scala/io/joern/console/cpgcreation/ImportCode.scala
+++ b/console/src/main/scala/io/joern/console/cpgcreation/ImportCode.scala
@@ -57,9 +57,9 @@ class ImportCode[T <: Project](console: io.joern.console.Console[T])(implicit
     new SourceBasedFrontend("python", Languages.PYTHONSRC, "Python Source Frontend", "py")
   def golang: SourceBasedFrontend = new SourceBasedFrontend("golang", Languages.GOLANG, "Golang Source Frontend", "go")
   def javascript: SourceBasedFrontend =
-    new SourceBasedFrontend("javascript", Languages.JAVASCRIPT, "Javascript Source Frontend", "js")
+    new JsFrontend("javascript", Languages.JAVASCRIPT, "Javascript Source Frontend", "js")
   def jssrc: SourceBasedFrontend =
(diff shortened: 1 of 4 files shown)


console/src/main/scala/io/joern/console/cpgcreation/ImportCode.scala
```diff
@@ -103,6 +124,28 @@ class ImportCode[T <: Project](console: io.joern.console.Console[T])(implicit
     def isAvailable: Boolean =
       cpgGeneratorForLanguage(language, config.frontend, config.install.rootPath.path, args = Nil).exists(_.isAvailable)
 
+    def fromDirectory(inputPath: String, projectName: String = "", args: List[String] = List()): Cpg = {
```
mpollmeier, line 127: given that it's so related and already in the same file, would it make sense to have the two `fromDirectory` and `fromFile` directly next to each other?

console/src/main/scala/io/joern/console/cpgcreation/ImportCode.scala
```diff
@@ -150,6 +193,17 @@ class ImportCode[T <: Project](console: io.joern.console.Console[T])(implicit
     }
   }
 
+  class JsFrontend(name: String, language: String, description: String, extension: String)
+      extends SourceBasedFrontend(name, language, description, extension) {
+    override def apply(inputPath: String, projectName: String, args: List[String]): Cpg = {
+      if (!File(inputPath).isDirectory) {
```
mpollmeier, line 199: please add a comment (and may error message?) that JS doesn't support this, just to clarify

mpollmeier (approved):

console/src/main/scala/io/joern/console/cpgcreation/ImportCode.scala
max-leuthaeuser: Well, we need them at `ImportCode` and `Frontend` so that e.g., both `importCode.fromFile` and `importCode.jssrc.fromFile` work.

maltek:
Do we really want/need a new public API for this? Can't this different handling just be an implementation detail for the JS frontends?

max-leuthaeuser:
@maltek Good idea, I guess thats even better.

PR #4407, opened 2024-03-27
[gosrc2cpg] Memory cleanup 

pandurangpatil:
As we are processing the main source code in two phases. We create the initial list of `AstCreator` first pass and we hold the data like ParserResult, Line no to line code mapping, parserNodeInfo cache etc for each file. Till the time we don't run through the second phase. 

1. Made changes to release some of the data points post first phase and write some caching information to the file. NOTE: couldn't write the `parserNodeInfo` cache as upiclke map is not serializable. Rebuild the released information back in the second phase when it starts processing.
2. Also made changes in download dependency pass parallelism. In download dependency pass few days back we made few optimisation to start processing the dependencies in parallel. However as we were also creating multiple futures for each file of the dependency. It's been observed multiple `Future` threads (which are created per dependency) were getting blocked waiting for the results. In turn lesser number of `Future` threads being used or available while processing the dependency files. Made changes and removed the Futures while downloading the dependencies, anyway we had to synchronise the section which executes the command. Processing of individual dependency code is queued in the queue which will be picked up by another writer thread. This will make sure processing is triggered in parallel, however it will also make sure only one dependency code is being processed at time.


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreator.scala
```diff
@@ -36,23 +41,30 @@ class AstCreator(
...
+  protected val tempAliasToNameSpaceMappingFilePath: Option[Path] =
+    tmpDir.map(dir => Paths.get(dir.pathAsString, s"alias-cache${UUID.randomUUID().toString}"))
   protected val methodAstParentStack: Stack[NewNode]                 = new Stack()
   protected val scope: Scope[String, (NewNode, String), NewNode]     = new Scope()
-  protected val aliasToNameSpaceMapping: mutable.Map[String, String] = mutable.Map.empty
-  protected val lineNumberMapping: Map[Int, String]                  = positionLookupTables(parserResult.fileContent)
-  protected val declaredPackageName = parserResult.json(ParserKeys.Name)(ParserKeys.Name).str
-  protected val fullyQualifiedPackage =
-    goMod.getNameSpace(parserResult.fullPath, declaredPackageName)
+  protected var aliasToNameSpaceMapping: mutable.Map[String, String] = mutable.Map.empty
```
DavidBakerEffendi, line 49: Line 49-56 is going to become technical debt and hard to maintain. Can we try to find a solution that avoids so much global mutable state in this class?

DavidBakerEffendi: Reading the code below, perhaps you want to look at what happens in `AstSummaryVisitor` in `RubySrc2Cpg` and `CSharpSrc2Cpg` for how to do a pre-parse without all this mutable state?

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreator.scala
```diff
@@ -103,4 +115,70 @@ class AstCreator(
...
+  /** Deserialize the ByteArray back to Object.
+    *
+    * @param bytes
+    *   \- Array[Byte] to be deserialized
+    * @return
+    *   \- Deserialized object
+    */
+  private def deserialise(bytes: Array[Byte]): Any = {
+    val ois   = new ObjectInputStream(new ByteArrayInputStream(bytes))
+    val value = ois.readObject
```
DavidBakerEffendi, line 180: `readObject` is an RCE waiting to happen, it is an extremely dangerous function.

We have a whole talk at BSides about exploiting Java deserialization https://www.youtube.com/watch?v=8qo0ZGK0tt0

https://msgpack.org is on the classpath from ODB, rather serialize data using that format.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreator.scala
```diff
@@ -103,4 +115,70 @@ class AstCreator(
...
+  }
+
+  def loadCacheToProcess(): Unit = {
+    tempAliasToNameSpaceMappingFilePath.map(file => {
+      val deserialised = deserialise(Files.readAllBytes(file))
+      aliasToNameSpaceMapping = deserialised.asInstanceOf[mutable.Map[String, String]]
+    })
+  }
+
+  def cleanup(): Unit = {
```
DavidBakerEffendi, line 144: I would rather suggest using `close` and having `AstCreator` extend `AutoCloseable`, this can enforce `Using.resource(astCreator) => ` which guarantees this global mutable state is at least closed otherwise we may sit with an implicit protocol which may not be followed and result in a memory leak.

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreator.scala
DavidBakerEffendi: At the very least, we need to move these variables to another trait, as it's becoming very difficult to follow this file.

DavidBakerEffendi (changes requested):

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/astcreation/AstCreator.scala
pandurangpatil: I guess serialising the alias to namespace mapping is anyway shouldn't make much of the difference. I need to try removing this.

pandurangpatil:
This is no more valid

PR #4408, opened 2024-03-27
[javasrc2cpg] - introduce JavaSummaryTypeSolver

xavierpinho:
Introduces an analog to `CSharpProgramSummary` for Java. The intended goal is to be able to provide a filesystem path containing type descriptions (in JSON) of external types to aid type resolution.

This is achieved by building a custom `TypeSolver` out of those JSON files. However, I wasn't able this time to elegantly build the TypeSolver: the quick solution I found was to translate (on-the-fly) those JSON files into Java declarations and in turn parse those, as if they were provided with the source-code being scanned. @johannescoetzee - do you know of a better way?

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/Main.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/Main.scala
@@ -21,7 +21,8 @@ final case class Config(
   showEnv: Boolean = false,
   skipTypeInfPass: Boolean = false,
   dumpJavaparserAsts: Boolean = false,
-  cacheJdkTypeSolver: Boolean = false
+  cacheJdkTypeSolver: Boolean = false,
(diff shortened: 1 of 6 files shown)


DavidBakerEffendi (commented):
What I ended up doing (which you'll find in C# and Ruby and may already be aware of) is that I have set a `SIGNATURES` flag to describe that the AstCreator should not generate method bodies, and then generate mini in-memory CPGs that I can parse for this type info for the external types. 

The ideal scenario is to also be able to do this with dependency downloading, where you'd need to parse JVM bytecode, and then generate those JSON files on-the-fly (and re-use them later). This is something we've done in C# and RubySrc now too. Although the existing usage of Gradle and Maven has major limitations and would recommend parsing those files and downloading the dependencies manually.

This is mostly to promote re-use of the existing logic in AstCreator to parse the JavaParser entities.

xavierpinho:
Abandoning/Closing this, as it doesn't fit very well into the existing architecture.

PR #4419, opened 2024-04-02
[C#] Fix `dotnetastgen` runner to handle space containing paths

karan-batavia:
This solves an issue where for a joern scan, a directory containing spaces in the name would not get scanned by DotNetAstGen


tuxology (approved):
Tested and looks good

PR #4425, opened 2024-04-05
[ruby] Handle chain calls

### LINKED ISSUE #4396: [ruby] Chained Calls
The grammar seems to not handle chains calls in certain expressions (or at all). E.g.
```ruby
class User < ApplicationRecord

  def self.authenticate(email, password)
    auth = nil
    user = find_by_email(email)
    raise "#{email} doesn't exist!" if !(user)
    if user.password == Digest::MD5.hexdigest(password)
      auth = user
    else
      raise "Incorrect Password!"
    end
    return auth
  end

end
```

The code `Digest::MD5.hexdigest(password)` is only picked up as a `Digest.MD5` member access, and `hexdigest` onwards is discarded within the context of the equality expression.

AndreiDreyer:
This PR handles:
 * Chained calls - separated the `commandOrPrimaryValue` for classes and otherwise. This allows us to move the `primaryValue` up for other cases, without it breaking on class definitions with inheritance (`class Foo < Bar::Baz`)

resolves #4396


DavidBakerEffendi (approved):

xavierpinho (approved):
🎉

PR #4430, opened 2024-04-08
[php] Implemented TypeHintCallLinker & PathSep Supports String

### LINKED ISSUE #4402: [PHP] Types not propagated for certain cases
**Describe the bug**
We should technically see types propagated in cases were a new object is created, but we do not see them sometimes. Conside the following code:

```php
use Curler\Client;

$client = new Client();
$response = $client->get('https://example2.com/data', $userId);
echo $response->getBody();
```

In this case, when we try to get the `methodFullName` for `get()` call, we don't get the fully qualified name:

```
Client() :  """Curler\Client->__construct""",
get() :     """<unresolvedNamespace>\$client->get""",
getBody() : """<unresolvedNamespace>\$response->getBody"""
```

**To Reproduce**
1. Use the above PHP code and generate CPG using PHP frontend
2. cpg.call.methodFullName.l
3. Observe the above missing method fullnames (except the constructor)

**Expected behavior**
We should get correct type and method fullname: `Curler\Client->__construct->\$client->get`

DavidBakerEffendi:
* Implemented `PhpTypeHintCallLinker`
* Allowed for `pathSep` to be a string to accommodate PHP's `->` syntax

Resolves #4402

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/PhpTypeHintCallLinker.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/PhpTypeHintCallLinker.scala
@@ -15,8 +15,6 @@ class PhpTypeHintCallLinker(cpg: Cpg) extends XTypeHintCallLinker(cpg) {
   override protected def calls: Iterator[Call] = {
     cpg.call
       .nameNot("<operator>.*", "<operators>.*")
-      .filter(c =>
-        calleeNames(c).nonEmpty && c.callee.fullNameNot(Pattern.quote(Defines.UnresolvedNamespace) + ".*").isEmpty
-      )
+      .filter(c => calleeNames(c).nonEmpty && c.callee.forall(_.fullName.startsWith(Defines.UnresolvedNamespace)))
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/passes/PhpTypeHintCallLinker.scala
```diff
@@ -0,0 +1,22 @@
...
+
+class PhpTypeHintCallLinker(cpg: Cpg) extends XTypeHintCallLinker(cpg) {
+
+  override protected val pathSep = "->"
+
+  override protected def calls: Iterator[Call] = {
+    cpg.call
+      .nameNot("<operator>.*", "<operators>.*")
+      .filter(c =>
+        calleeNames(c).nonEmpty && c.callee.fullNameNot(Pattern.quote(Defines.UnresolvedNamespace) + ".*").isEmpty
```
fabsx00, line 19: `.startsWith`?

tuxology (approved):

AndreiDreyer (approved):

PR #4439, opened 2024-04-09
[jssrc2cpg] Fix for dynamicTypeHintFullName on this identifier

max-leuthaeuser:
Follow-up for: https://github.com/joernio/joern/pull/4426

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/TypeHelper.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/TypeHelper.scala
@@ -137,10 +137,8 @@ trait TypeHelper { this: AstCreator =>
       case Some(tpe) => Seq(tpe)
       case None if node.isDefined =>
         typeFor(node.get) match {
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/TypeHelper.scala
```diff
@@ -134,8 +137,10 @@ trait TypeHelper { this: AstCreator =>
       case Some(tpe) => Seq(tpe)
       case None if node.isDefined =>
         typeFor(node.get) match {
-          case t if t != Defines.Any && t != "this" => Seq(t)
-          case _                                    => rootTypeDecl.map(_.fullName).toSeq
+          case tpe if tpe != Defines.Any && tpe != "this" =>
+            val typeFullName = if (Defines.isBuiltinType(tpe)) tpe else Defines.Any
```
maltek, line 141: since this is used for a dynamic type hint, we shouldn't put `ANY` there. Dynamic type hints kill all further type resolution, so if we're unsure about the type (or can't express it), we shouldn't put *something* there but leave it empty instead.

maltek (approved):

PR #4441, opened 2024-04-09
[php2cpg] Don't use target code in php method full names

johannescoetzee:
We're sometimes using the receiver/base code as part of the method full name in php2cpg, which led to the weird-looking expectations changes in https://github.com/ShiftLeftSecurity/codescience/pull/7595. This was a deliberate decision at the time since we effectively never have type full names for php cpgs anyways and at least that way we get some information, but in case that's not desired behaviour, this is an alternative.


johannescoetzee:
For some reason scalafmt is failing on my laptop ("failing to download /path/to/joern/.scalafmt.conf" ???), so I'll fix formatting when I'm home.

maltek (approved):

johannescoetzee:
Since https://github.com/ShiftLeftSecurity/codescience/pull/7595 was approved, I take that as the weird names are ok and  I'm closing this in favour of that.

PR #4442, opened 2024-04-10
Add environment variable to javasrc to fetch dependencies

johannescoetzee:
Tested this manually since I'm not aware of a way to easily do this in our unit test setup.

```
❯ export JAVASRC_FETCH_DEPENDENCIES=t
❯ ./joern-cli/frontends/javasrc2cpg/javasrc2cpg ~/code/java/HelloShiftLeft
...
[INFO ] Enabling dependency fetching: Environment variable JAVASRC_FETCH_DEPENDENCIES is set
[INFO ] got 63 Maven dependencies
```

The idea is to set this by default in SL, but then customers can still unset it if they want to (which should be a rare case).

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
@@ -66,7 +66,8 @@ class AstCreationPass(config: Config, cpg: Cpg, sourcesOverride: Option[List[Str
   }
 
   private def getDependencyList(inputPath: String): List[String] = {
-    val shouldFetch = if (System.getenv(JavaSrcEnvVar.FetchDependencies.name).nonEmpty) {
+    val envVarValue = Option(System.getenv(JavaSrcEnvVar.FetchDependencies.name))
+    val shouldFetch = if (envVarValue.exists(_.nonEmpty)) {
       logger.info(s"Enabling dependency fetching: Environment variable ${JavaSrcEnvVar.FetchDependencies.name} is set")
       true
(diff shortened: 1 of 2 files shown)


max-leuthaeuser (approved):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
```diff
@@ -66,17 +66,24 @@ class AstCreationPass(config: Config, cpg: Cpg, sourcesOverride: Option[List[Str
...
-      List()
+      logger.info("dependency resolving not enabled")
+      false
+    }
+
+    Option.when(shouldFetch)(DependencyResolver.getDependencies(Paths.get(inputPath))).flatten match {
+      case Some(deps) => deps.toList
+      case None =>
+        logger.warn(s"Could not fetch dependencies for project at path $inputPath")
+        List()
```
maltek, line 85: if I'm reading this right, this would log "could not fetch" when fetching has merely not been requested. So I'd rewrite this to sth like (untested):

```suggestion
    if (shouldFetch) {
      DependencyResolver.getDependencies(Paths.get(inputPath)).getOrElse {
        logger.warn(s"Could not fetch dependencies for project at path $inputPath")
        Nil
      }
    } else  {
      Nil
```

maltek (approved):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
johannescoetzee: I was just about to commit almost exactly this fix

PR #4452, opened 2024-04-15
[php2cpg] Annotation Support

### LINKED ISSUE #4453: [php2cpg] Annotations Support
Adding support for annotations, referred to as attributes in php. 
Refer: https://www.php.net/manual/en/language.attributes.overview.php

karan-batavia:
This PR includes creating annotation nodes in the `php2cpg` frontend, and linking them to their corresponding parent AST nodes (methods, classes, parameters). 

Resolves #4453

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/AnnotationTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/AnnotationTests.scala
@@ -1,0 +1,39 @@
+package io.joern.php2cpg.querying
+
+import io.joern.php2cpg.testfixtures.PhpCode2CpgFixture
+import io.shiftleft.semanticcpg.language.*
+
+class AnnotationTests extends PhpCode2CpgFixture {
+  "annotations for nodes" should {
+    "be populated" in {
+      val cpg = code("""
+          |<?php
+          | #[Route("/api")]
+          | class Foo {
+          |   #[Route("/edit", name: "hello")]
+          |   public function bar(#[SomeAttr] $pBar){}
+          | }
+          |>
+          |""".stripMargin)
+
+      inside(cpg.typeDecl("Foo").annotation.l) { case route :: Nil =>
+        route.name shouldBe "Route"
+        inside(route.astChildren.l) { case arg1 :: Nil =>
+          arg1.code shouldBe "\"/api\""
+        }
+      }
+
+      inside(cpg.method("bar").annotation.l) { case route :: Nil =>
+        route.name shouldBe "Route"
+        inside(route.astChildren.l) { case arg1 :: arg2 :: Nil =>
+          arg1.code shouldBe "\"/edit\""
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/AnnotationTests.scala
```diff
@@ -0,0 +1,39 @@
...
+          |<?php
+          | #[Route("/api")]
+          | class Foo {
+          |   #[Route("/edit", name: "hello")]
+          |   public function bar(#[SomeAttr] $pBar){}
+          | }
+          |>
+          |""".stripMargin)
+
+      inside(cpg.typeDecl("Foo").annotation.l) { case route :: Nil =>
```
DavidBakerEffendi, line 19: Could you include the default case with a `fail` call?

DavidBakerEffendi: Same with all tests in this file

karan-batavia: Done.

DavidBakerEffendi (approved):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/AnnotationTests.scala
```diff
@@ -0,0 +1,50 @@
...
+          |>
+          |""".stripMargin)
+
+      inside(cpg.typeDecl("Foo").annotation.l) {
+        case route :: Nil =>
+          route.name shouldBe "Route"
+          inside(route.astChildren.l) {
+            case arg1 :: Nil =>
+              arg1.code shouldBe "\"/api\""
+            case _ => fail("Expected exactly 1 argument for the annotation")
```
johannescoetzee, line 25: `inside` works on partial functions, so it's not necessary to provide the default case here. With how these tests are written, a simple `match` would work:

```
route.astChildren.l match {
  case arg1 :: Nil => ...
  case _ => ...
}
```

as would 

```
inside(route.astChildren.l) { case List(arg1) => ... }
```

The second option is more consistent with how we write tests elsewhere in joern and is easier to maintain when changing tests.

karan-batavia: This was @DavidBakerEffendi's suggestion, and how we wrote tests in `csharpsrc2cpg`.  
Refer the above review comment.

DavidBakerEffendi: @johannescoetzee My motivation is that it's easier to debug a failure from nested `inside` statements, but I suppose one can go to the line numbers. If the norm is not to have a default case, then we can go forward as such

PR #4455, opened 2024-04-15
[x2cpg] SemVer Utility Class & Parsing

DavidBakerEffendi:
While handling build files and parsing dependencies, I often find myself wanting something that can help parse the actual versions while preserving their natural ordering. To address this, I've gone ahead with building this myself.

* Added parsing for SemVer according to https://semver.org/
* Added fuzzy semver handling, where it's simply converted to a normalized semver
* Added JSON deserialization and serialization with `ujson`


max-leuthaeuser:
I want to throw https://github.com/sh4869/semver-parser-scala into the ring.

DavidBakerEffendi:
@max-leuthaeuser Then I think it's probably better to use one of those dependencies, I should've probably done my homework lol.

Issue #4279, opened 2024-03-06
[Bug] SchemaViolationException frequently


cyw3:
**Describe the bug**

Hello, I still encounter this occasional exception frequently. Every time it occurs, the execution result of the tool will lack many Results that should originally exist, and the number of the log `[INFO ] Enhancement io.joern.console.scan.ScanPass completed in 4450 ms. 130  + 650 changes committed from 25 parts.` changed.

**To Reproduce**
Here is a project that can be reproduced frequently (an average of 1 time in 4 executions), which can be tested based on this project.
https://github.com/Haptic-Apps/Slide
1. git clone
```bash
git clone https://github.com/Haptic-Apps/Slide
```
2. exec cmd
```bash
/data/test/joern-cli/joern-scan Slide/ --names=too-many-params --language=java --overwrite
```

**Expected behavior**
It is expected that no exception will be reported when running, and the number of execution results will be stable.

**Screenshots**

```
Caused by: overflowdb.SchemaViolationException: OUT edge with label AST to an adjacent METHOD_RETURN is mandatory, but not defined for this METHOD node with id=134
        at io.shiftleft.codepropertygraph.generated.nodes.MethodDb.methodReturn(Method.scala:660)
        at io.shiftleft.codepropertygraph.generated.nodes.Method.methodReturn(Method.scala:286)
        at io.joern.x2cpg.passes.controlflow.cfgcreation.CfgCreator.<init>(CfgCreator.scala:55)
        at io.joern.x2cpg.passes.controlflow.CfgCreationPass.runOnPart(CfgCreationPass.scala:23)
        at io.joern.x2cpg.passes.controlflow.CfgCreationPass.runOnPart(CfgCreationPass.scala:17)
        at io.shiftleft.passes.ConcurrentWriterCpgPass.$anonfun$createApplySerializeAndStore$2(ParallelCpgPass.scala:252)
        at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:678)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:467)
        at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1426)
        at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
        at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
[...]

PR #4463, opened 2024-04-17
[slicing] Dataflow Slicing Concurrency Fix

### LINKED ISSUE #4459: Can not slice
When I try to use `joern-slice data-flow --sink-filter exec\([\S\s]+\) cpg.bin` for slicing, I encounter an error `WARN DataFlowSlicing$: Exception encountered during slicing task
java.util.concurrent.ExecutionException: java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.length()" because "this.text" is null`

Here is the source code:

```
var exec = require('child_process').exec
  , path = require('path')
  ;


/**
 * open a file or uri using the default application for the file type.
 *
 * @return {ChildProcess} - the child process object.
 * @param {string} target - the file/uri to open.
 * @param {string} appName - (optional) the application to be used to open the
 *      file (for example, "chrome", "firefox")
 * @param {function(Error)} callback - called with null on success, or
 *      an error object that contains a property 'code' with the exit
 *      code of the process.
 */

module.exports = open;

function open(target, appName, callback) {
  var opener;

  if (typeof(appName) === 'function') {
    callback = appName;
    appName = null;
  }

  switch (process.platform) {
  case 'darwin':
    if (appName) {
      opener = 'open -a "' + escape(appName) + '"';
    } else {
      opener = 'open';
    }
    break;
  case 'win32':
    // if the first parameter to start is quoted, it uses that as the title
    // so we pass a blank title so we can quote the file we are opening
    if (appName) {
      opener = 'start "" "' + escape(appName) + '"';
    } else {
      opener = 'start ""';
    }
    break;
  default:
    if (appName) {
      opener = escape(appName);
    } else {
      // use Portlands xdg-open everywhere else
      opener = path.join(__dirname, './xdg-open');
    }
    break;
  }

  if (process.env.SUDO_USER) {
    opener = 'sudo -u ' + process.env.SUDO_USER + ' ' + opener;
  }
  return exec(opener + ' "' + escape(target) + '"', callback);
}

function escape(s) {
  return s.replace(/"/g, '\\\"');
}
```

I want get the slice of `exec(opener + ' "' + escape(target) + '"', callback)`.

Someone can help me?

Thank you very much!

DavidBakerEffendi:
It appears that running `.code` in a multi-threaded environment may lead to the database returning `null` to the regex matcher. It is a bit bizarre, but this PR moves the `code` filter in a serial position. See #4459 for more details.

Resolves #4459

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/slicing/DataFlowSlicing.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/slicing/DataFlowSlicing.scala
@@ -10,7 +10,7 @@ import org.slf4j.LoggerFactory
 
 import java.util.concurrent.Callable
 import scala.concurrent.ExecutionContext
-import scala.util.{Failure, Success}
+import scala.util.{Failure, Success, Try}
 
 object DataFlowSlicing {
 
@@ -23,7 +23,7 @@ object DataFlowSlicing {
     val tasks = (config.fileFilter match {
       case Some(fileName) => cpg.file.nameExact(fileName).method.call
       case None           => cpg.call
-    }).method.withMethodNameFilter.withMethodParameterFilter.withMethodAnnotationFilter.call.withExternalCalleeFilter
+    }).method.withMethodNameFilter.withMethodParameterFilter.withMethodAnnotationFilter.call.withExternalCalleeFilter.withSinkFilter
       .map(c => () => new TrackDataFlowTask(config, c).call())
       .iterator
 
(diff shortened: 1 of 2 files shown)


max-leuthaeuser:
Maybe this also helps for: https://github.com/joernio/joern/issues/4279?
(i.e., a similar fix for joern-scan?)

AndreiDreyer (approved):

DavidBakerEffendi:
@max-leuthaeuser That seems different, as I often encounter that bug when memory is limited. However, it's work looking at but I'm not familiar with how the scanner works in detail

DavidBakerEffendi:
@max-leuthaeuser  Having a look now

Issue #4462, opened 2024-04-16
[Bug] joern-parse produces incorrect graph on preprocessed C files


will-leeson:
**Describe the bug**

I am attempting to use Joern to generate graphs from files in the [SV-Comp benchmarks](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks). Many of the files have been preprocessed using the CPP. The outputs of which are the .i files in the repository. When I run 

```
joern-parse [/path/to/file.i]
```

on one of the preprocessed files, I receive an error stating it can't identify the language. I then then specified that it is a C File using

```
joern-parse --language "c" [/path/to/file.i]
```

and the command finishes with no errors. When I attempt to export the graph using the following command

```
joern-export --repr=all --format=dot
```

I receive no errors, but I get a graph significantly smaller than expected and incorrect. As a sanity test, I ran joern-parse and joern-export on a file with the following content:

```C
int main(){
    int x = 2;

    int y = 3;

    int z = y - x;

    return z; 
}
```

I received the expected output. I then generated the preprocessed file using `cpp [file.c] > [file.i]` and ran the exact same workflow. This produced the same graph as the SV-Comp program, which differs significantly.

**To Reproduce**
1. Preprocess C file and save output
2. Run `joern-parse --language "c" [/path/to/file.i]` on preprocessed file
3. Run `joern-export --repr=all --format=dot`

**Expected behavior**
I expected joern to be able to handle preprocessed files as they are valid C. 

**Desktop (please complete the following information):**
 - OS - Ubuntu 22.04
 - Joern Version - 2.0.226
 - Java version - JDK 19.0.2

**Additional context**
If you don't support this use case, that is fine. I'm just bringing this to your attention as it seems it should work fine. Preprocessed C files are valid C.

max-leuthaeuser:
Only files with C or CPP extensions (see: https://github.com/joernio/joern/blob/master/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/parser/FileDefaults.scala) are parsed.
*.i files won't work. If your file contains valid C oder CPP code anyway you may simply rename it.

PR #4466, opened 2024-04-17
[c2cpg] Handle preprocessed files

max-leuthaeuser:
cpp/gcc/g++ offer the -E switch. This switch forces the compiler to stop after the preprocessing phase and spit out the result where all preprocessor directives are expanded or processed, all macros expanded, and all header files included. Typically, this result is stored as *.i file.

With this PR we process such files and exclude all the others with the exact same name.

Fixes: https://github.com/joernio/joern/issues/4462


max-leuthaeuser:
Do not merge yet - will put that behind a command line flag.

ml86 (approved):

PR #4472, opened 2024-04-18
Dependency Housekeeping: Removed `circe`, `jackson` & Organised Versions

DavidBakerEffendi:
Joern has **a lot** of dependencies, and I've found `meta-build` out of sync with some `build.sbt` files specifying the same dependencies.

Also, `Versions.scala` is used haphazardly and is inconsistent. This change congregates all versions there, so we can see all versions of dependencies in Joern.

This all started by wondering why we have `jackson-databind`, `ujson`, and `circe` to deserialize JSON, so I removed `circe` and `jackson-databind` while keeping `ujson`. In any-case, this organizes dependencies and paves the way for removing unused ones.

Note, some code in `jssrc2cpg` (@max-leuthaeuser) and `gosrc2cpg` (@pandurangpatil) were modified for the migration to deserialize with `ujson` instead of the existing libraries.


pandurangpatil:
@DavidBakerEffendi as long as all failing unit tests are working changes look good from my side. 

cc: @tuxology

max-leuthaeuser:
LGTM once the failing go tests are fixed. There something wrong with the new json handling there.

mpollmeier (approved):
thank you, very much appreciated!

PR #4473, opened 2024-04-19
[ruby] Download builtin package dependencies

AndreiDreyer:
This PR handles:
* Scrape [rubydoc](https://ruby-doc.org/3.3.0/) to generate a `NamespaceTypeMap` for builtin ruby packages
* Generate `.zip` folder with `MessagePack` files for each of the gems to be re-used
* Write to JSON for sanity checking (to be removed)

TODO for next branch:
* Return `NamespaceTypeMap` from `BuiltinPackageDownloader` if run
* Implement reading from a `ZipStream` if the `BuiltinPackageDownloader` is not run and a `.zip` folder is present to build the `NamespaceTypeMap`


max-leuthaeuser:
I am against putting .zip files (or any binary in general) directly into this repo.
Even with Git LFS we had trouble in the past (running into the GitHub traffic limitation, any further downloads / actions disabled).

DavidBakerEffendi:
@max-leuthaeuser Another solution I've thought of is to treat this like the AstGen binaries, where some repo generates these and during release-time they're compressed and packaged in a minimal way. Would that be an acceptable next step? Right now I believe these are ~100mb uncompressed

max-leuthaeuser:
> Would that be an acceptable next step? 

Guess thats much better, yes.

DavidBakerEffendi:
This will be replaced by https://github.com/joernio/joern/pull/4501

DavidBakerEffendi:
Superceded by https://github.com/joernio/joern/pull/4501

DavidBakerEffendi:
Superceded by https://github.com/joernio/joern/pull/4501

PR #4477, opened 2024-04-22
[ruby] `RubyProgramSummary` types derive ReadWriter

AndreiDreyer:
This PR:
 * adds `derives ReadWriter` to `RubyField`, `RubyType` in `RubyProgramSummary`
 * adds custom serialization/deserialization logic for `RubyMethod`

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/datastructures/RubyProgramSummary.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/datastructures/RubyProgramSummary.scala
@@ -34,14 +34,14 @@ case class RubyMethod(
 object RubyMethod {
   implicit val rubyMethodRwJson: ReadWriter[RubyMethod] = readwriter[ujson.Value].bimap[RubyMethod](
     x => ujson.Obj("name" -> x.name),
-    json => RubyMethod(json("name").str, List.empty, "", Option.empty)
+    json => RubyMethod(json("name").str, List.empty, XDefines.Any, Option.empty)
   )
 }
 
 case class RubyField(name: String, typeName: String) extends FieldLike derives ReadWriter
 
 case class RubyType(name: String, methods: List[RubyMethod], fields: List[RubyField])
-    extends TypeLike[RubyMethod, RubyField] derives ReadWriter {
+    extends TypeLike[RubyMethod, RubyField] {
 
   @targetName("add")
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi:
Here is a thread on what the error code means on this runner: https://stackoverflow.com/questions/77219743/what-is-exit-code-3221225781-on-windows

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/datastructures/RubyProgramSummary.scala
```diff
@@ -29,11 +30,12 @@ case class RubyMethod(
   returnType: String,
   baseTypeFullName: Option[String]
 ) extends MethodLike
+    derives ReadWriter
```
DavidBakerEffendi, line 33: Something like `baseTypeFullName` just represents the surrounding `RubyType`, so perhaps we could create a more defined serialization/serialization for `RubyMethod`? Similarly, `returnType` and `parameterTypes` can be excluded.

DavidBakerEffendi (changes requested):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/datastructures/RubyProgramSummary.scala
```diff
@@ -30,10 +31,17 @@ case class RubyMethod(
   baseTypeFullName: Option[String]
 ) extends MethodLike
 
-case class RubyField(name: String, typeName: String) extends FieldLike
+object RubyMethod {
+  implicit val rubyMethodRwJson: ReadWriter[RubyMethod] = readwriter[ujson.Value].bimap[RubyMethod](
+    x => ujson.Obj("name" -> x.name),
+    json => RubyMethod(json("name").str, List.empty, "", Option.empty)
```
DavidBakerEffendi, line 37: Note `Option.empty` should be the parent `RubyType`. Thinking about this now, we should probably write a `ReadWriter` for `RubyType` that can re-map the `RubyTypes.methods` field with `baseTypeFullName` to `RubyType.name`
```suggestion
    json => RubyMethod(json("name").str, List.empty, Defines.Any, Option.empty)
```

DavidBakerEffendi (changes requested):

DavidBakerEffendi (approved):

PR #4478, opened 2024-04-22
[php2cpg] Handle utf-8 offsets correctly

johannescoetzee:
This PR fixes offset calculations when processing utf-8 files with the file content feature enabled. The 2 main issues with this PR are that it assumes a utf-8 encoding (which will work for ascii as well, but not for more niche encodings) and a possible performance hit from the offset calculation. A potential optimization would be to cache known offsets and to calculate the new offset from the closest one, but since this is not a common operation, it likely won't be too much of a slow-down and we can revisit this decision if performance does turn out to be an issue.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
@@ -17,6 +17,8 @@ import io.shiftleft.semanticcpg.language.types.structure.NamespaceTraversal
 import org.slf4j.LoggerFactory
 import overflowdb.BatchedUpdate
 
+import java.nio.charset.StandardCharsets
+
 class AstCreator(filename: String, phpAst: PhpFile, fileContent: Option[String], disableFileContent: Boolean)(implicit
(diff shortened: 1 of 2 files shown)


max-leuthaeuser (approved):

johannescoetzee:
Checks failed with
```
Getting installed Swift version swift-5.9.2-RELEASE
C:\hostedtoolcache\windows\swift-5.9.2-RELEASE-windows10\5.9.2\x86_64\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swift.exe --version
Error: The process 'C:\hostedtoolcache\windows\swift-5.9.2-RELEASE-windows10\5.9.2\x86_64\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swift.exe' failed with exit code 3221225781
```

Seems unrelated to this, so I'm just going to rerun the jobs

max-leuthaeuser:
GH might has been updating the Windows runners in the background and something broke w.r.t. dlls missing now. I am not sure how to fix that.
Maybe we need to update the Swift install action.

max-leuthaeuser:
I had a look and something in the runner images broke. I guess we have to disable the swiftsrc2cpg tests temporarily to unblock you guys.

PR #4484, opened 2024-04-23
[ruby] Implicit Require Handling

DavidBakerEffendi:
[Autoloading in Ruby](https://www.rubyguides.com/2019/08/autoloading-in-ruby/) is fairly common, however it's important for us to represent these import nodes explicitly.

This pass adds a check for modules with no high-level require calls, to check call receivers and constructor allocations for types from other files, and add explicit require calls for them.


DavidBakerEffendi:
Some weird build errors related to the citation validation happening. We don't really need this step anymore so if this build passes without it then I'll merge

DavidBakerEffendi:
Hmm seems `sbt` is now missing? Weird

max-leuthaeuser:
This PR needs to be merged with master once https://github.com/joernio/joern/pull/4490 is in.

PR #4489, opened 2024-04-24
[javasrc2cpg] Clear JavaParser caches after AST creation

xavierpinho:
We reach out-of-memory exceptions when running consecutive scans (within the same process) and providing the so-called «inference jar paths» (`--inference-jar-paths`).

This can be observed if we run all javasrc2cpg unit-tests with some additional inference jar. In the following experiment, I've used the same 100MB .jar throughout unit-tests and the 8GB limit is quickly reached, precluding them from finishing.

Screenshot 2024-04-24 at 16 11 49

Poking around, it seems clear that each time an inference jar is loaded its TypeSolver remains in memory. 

By clearing JavaParser caches the situation improves considerably:

Screenshot 2024-04-24 at 16 16 14


DavidBakerEffendi (approved):
Similar to the issue we encountered in Ruby with the ANTLR cache. Interesting find!

max-leuthaeuser (approved):
Maybe we should make `AstCreationPass` here a `Releasable` and use `Using` (https://www.scala-lang.org/api/3.4.1/scala/util/Using$.html) to clean up stuff (cleanupDelombokOutput, clearJavaParserCaches)

Also this PR needs to be merges with master once https://github.com/joernio/joern/pull/4490 is in.

PR #4490, opened 2024-04-25
Add sbt and php to MacOS runner

max-leuthaeuser:
Looks like they are no longer pre-installed.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -18,6 +18,9 @@ jobs:
       - name: Install sbt
         if: matrix.os == 'macos-latest'
         run: brew install sbt
+      - name: Install php
+        if: matrix.os == 'macos-latest'
+        run: brew install php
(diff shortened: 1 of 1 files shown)


mpollmeier:
yes that would make sense. `macos 12` does list sbt, from 13 onwards it's gone:
https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md
https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md
https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md

php is listed as preinstalled even on macos 14 though... (PHP 8.3.6)

max-leuthaeuser:
Which is very strange as we are using macos-latest for quite a while now and it worked until yesterday...
php2cpg failed in a previous run complaining that no valid php could be found. Thats why I added that also.

mpollmeier (approved):

PR #4492, opened 2024-04-25
[C#] Fix: Built-in Type Stubs not Loaded When Packaged

### LINKED ISSUE #4474: [C#]:The option " --download-dependencies " doesn't take effect
Test project url: [https://github.com/chengkenyong/web-Application](url)

Describe the bug:
When I use the "importCode("E:\\Csharp_output\\web-Application", args=List(--download-dependencies))" command to import C# code and download additional dependencies, I found that using the --download-dependencies option did not help in restoring the namespace and signature in the methodFullName attribute of the call nodes. 
微信截图_20240419175346

Upon debugging, I discovered that the generated cpg.graph does not contain nodes labeled as DEPENDENCY, which prevents the download() method from obtaining the names of the dependencies to be downloaded.
微信截图_20240419175748

 It seems that the X2cpg/MetaDataPass class lacks an addDependency method to add nodes labeled as DEPENDENCY to the graph, but I'm not very clear about the specific implementation. I hope to seek your help.
 
微信截图_20240419175957

Desktop and Release:

- Win11
- Joern-v2.0.329

chengkenyong:
The 'csharpsrc2cpg/src/main/resources/builtin_types' is not loaded into the classpath after the stage. 

Resolves #4474

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpProgramSummary.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpProgramSummary.scala
@@ -70,7 +70,9 @@ object CSharpProgramSummary {
       case Some(url) if url.getProtocol == "jar" =>
         val connection = url.openConnection.asInstanceOf[JarURLConnection]
         Using.resource(connection.getJarFile) { jarFile =>
-          jarFile.entries().asScala
+          jarFile
+            .entries()
+            .asScala
             .toList
             .map(_.getName)
             .filter(_.startsWith(builtinDirectory))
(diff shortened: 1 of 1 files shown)


DavidBakerEffendi (commented):
Interesting catch, is there a way to write a test case for this, or is this specifically an issue when Joern is staged as a JAR?

DavidBakerEffendi:
cc: @AndreiDreyer @badly-drawn-wizards Note the way the resources are read here, may be important for Ruby

DavidBakerEffendi:
@chengkenyong For formatting, run `sbt scalafmt`

AndreiDreyer:
Think this issue is specifically to the packaged JAR, since I've been struggling with the same issue after pulling in the ruby type-stubs from the `joern-type-stubs` and getting them read during a `joern-parse` run. This solution works on the Ruby side too

PR #4496, opened 2024-04-29
[csharp] Filter unused type stubs

badly-drawn-wizards:
- Collect imports in program summary
- Filter type stubs based on import
- Add ignored tests for some method resolution problems found along the way.


DavidBakerEffendi (approved):
What are the next steps to get something like `"".ToLower` working?

PR #4501, opened 2024-04-29
[ruby] Builtin type-stubs for `Ruby`

AndreiDreyer:
This PR handles:
 * Downloads compressed type-stubs from `joern-type-stubs` release during `sbt stage` and sha512 checksums the downloaded files with the sha512 on `joern-type-stubs` release page
 * Populates the initial `namespaceToType` map from the builtin type-stubs

Todo:
* Filter out types from the `namespaceToType` map that are not imported
*


DavidBakerEffendi (approved):
Very nice!

DavidBakerEffendi:
@AndreiDreyer Could you also test that this works in the Joern shell before you merge?

PR #4509, opened 2024-04-30
[ruby] Enable `ImplicitRequirePass` if `zeitwerk` is Detected

DavidBakerEffendi:
Uses `Gemsfile.lock` which is parsed (if present) in the `DependencyPass` to determine if implicit imports are to be enabled.


DavidBakerEffendi:
Seems there are additional semantics I can follow to help hone this down: https://github.com/fxn/zeitwerk?tab=readme-ov-file#file-structure

AndreiDreyer (approved):

PR #4515, opened 2024-05-02
[kotlin2cpg] Test case for CompilerAPI.makeEnvironment with single source dir

max-leuthaeuser:
This test case demonstrates the issues with `CompilerAPI.makeEnvironment` on a single path (the `sourceDir`) compared to `dirsForSourcesToCompile`. Just change `Kotlin2Cpg.scala` line 139 back to `Seq(sourceDir)` and watch it fail again.


pandurangpatil:
@max-leuthaeuser Thank, this will be helpful to fix the issue properly.

pandurangpatil (approved):

PR #4524, opened 2024-05-02
[ruby] `loop` Do Block & Break Expressions

### LINKED ISSUE #4508: [ruby] `BreakStatement` cannot represent expression in loop
```ruby
x = 0
num  = -1
loop do
  num = x + 1
  x = x + 1
  if x > 10
    break
  end
end
puts num
```

Image

DavidBakerEffendi:
* Handles `break` statements on an expression level
* Lower `loop do` statements as DoWhile expressions

Resolves #4508


DavidBakerEffendi:
Looking into failing associated data-flow test

AndreiDreyer (approved):

PR #4528, opened 2024-05-02
[jimple] Remove jsp add html and json

itsacoderepo:
I would not consider jsp to be config files but code. Also added json and HTML files. The latter usually also does not contain configurations but i would expect it there (currently).


DavidBakerEffendi (commented):
For Jimple this choice is fine I think, but remember to add `.html` to the filters in `io.joern.x2cpg.passes.frontend.JavaConfigFileCreationPass` otherwise this won't be picked up. Alternatively, one can add a config test similar to what is in Java.

PR #4540, opened 2024-05-07
[ruby] Improved type resolution

### LINKED ISSUE #4531: Type Stub Schema For Overloadable & Dynamic Languages
## Problem

C# and Ruby currently make use of the experimental program summaries and `[Method|Type|Field]Like` traits. However, C# is an overloadable language that includes parameter types in the method full name, whereas dynamic languages are generally not overloadable and thus don't need to concern parameter types. There is wasted effort, processing, and memory usage on languages that don't care about overloading while this info is processed by default.

## Suggested Solution

Two problems come to mind here

### Method Resolution Behaviour

`io.joern.x2cpg.datastructures.TypedScope::tryResolveMethodInvocation` and `io.joern.x2cpg.datastructures.TypedScope::isOverloadedBy` have overloading detection behaviour which isn't language agnostic as discussed above.

I suggest adding a new trait, such as

```scala
package io.joern.x2cpg.datastructures

/* Other scope and program summary stuff above */

trait OverloadableScope { this: TypedScope[?, ?, ?] => 

   override protected def tryResolveMethodInvocation /* override method resolution with overloading enabled */

   protected def isOverloadedBy /* this should only occur in OverloadableScopeScope */

}
```

C#'s TypedScope is an example of a scope which should use this.

### Serialization

For languages that don't overload, we don't need to worry about parameters at all. Only the method name and return type. So similarly we can add traits that extend `MethodLike` with parameters, e.g.

```scala

trait MethodLike extends MemberLike {

  def name: String

  def returnType: String

}

trait OverloadableMethod { 

  def parameters: List[(String, String)]
  
 }

// e.g.
case class CSharpMethod(name: String, parameters: List[(String, String)], returnType: String) extends MethodLike with OverloadableMethod
```

Thus, serialized stubs for Ruby don't have unnecessary `parameters` fields and C# will.

### Type Stubs

Similarly, it would be useful for the resolution to know if the types in the `ProgramSummary` came from the application class, or from a type stub. This may help with prioritising conflicting types in scopes (e.g., built-in types which are always in scope) and user-defined methods.

As above with how we extended `MethodLike` with an `OverloadableMethod` trait, we can extend `TypeLike` with `StubbedType` to signal that this should not be prioritised over other, user-defined types and this comes from a pre-defined file.

AndreiDreyer:
This PR handles:
 * Improved type resolution for method calls to differentiate between internal and stubbed types
 * Removed overloading behaviour on `tryResolveMethodInvocation` for `ProgramSummary`
 * Introduces `OverloadableScope` for languages that allow overloading with the old `tryResolveMethodInvocation` implementation that checks for method overloading (currently C#)
 * Introduces `StubbedType` and `StubbedRubyType` which is used for stubbed types (builtin-type stubs, external dependencies)

Resolves #4531


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstSummaryVisitor.scala
```diff
@@ -17,7 +17,7 @@ import scala.util.Using
 
 trait AstSummaryVisitor(implicit withSchemaValidation: ValidationMode) { this: AstCreator =>
 
-  def summarize(): RubyProgramSummary = {
+  def summarize(dependencyDownloadPass: Boolean = false): RubyProgramSummary = {
```
DavidBakerEffendi, line 20: Might make for clearer consistent language as the CPG has the whole `isExternal` property
```suggestion
  def summarize(asExternal: Boolean = false): RubyProgramSummary = {
```

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/datastructures/RubyProgramSummary.scala
```diff
@@ -121,6 +122,40 @@ object RubyMethod {
...
+    json =>
+      RubyStubbedType(
+        name = json("name").str,
+        methods = json.obj.get("methods") match {
+          case Some(jsonMethods) =>
+            val methodsList = read[List[RubyMethod]](jsonMethods)
+
+            methodsList.map { func =>
+              val splitName        = func.name.split("\\.")
+              val baseTypeFullName = splitName.dropRight(1).mkString(".")
```
DavidBakerEffendi, line 148: Is this not the same as `json("name").str`?

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -107,6 +107,10 @@ trait TypedScope[M <: MethodLike, F <: FieldLike, T <: TypeLike[M, F]](summary:
       }
   }
 
+  def matchingM(callName: String)(implicit tag: ClassTag[M]): PartialFunction[MemberLike, M] = {
```
DavidBakerEffendi, line 110:
```suggestion
  private def matchingM(callName: String)(implicit tag: ClassTag[M]): PartialFunction[MemberLike, M] = {
```

DavidBakerEffendi: Or `protected`

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -122,22 +126,13 @@ trait TypedScope[M <: MethodLike, F <: FieldLike, T <: TypeLike[M, F]](summary:
...
-      // This function uses the `implicit tag` (IntelliJ incorrectly marks it as unused)
-      def matchingM: PartialFunction[MemberLike, M] = { case m: M if m.name == callName => m }
       // TODO: The typesInScope part is to imprecisely solve the unimplemented polymorphism limitation
-      membersInScope.collectFirst(matchingM).orElse { typesInScope.flatMap(_.methods).collectFirst(matchingM) }
-    case Some(tfn) =>
-      val methodsWithEqualArgs = tryResolveTypeReference(tfn).flatMap { t =>
-        Option(
-          t.methods.filter(m => m.name == callName && m.parameterTypes.filterNot(_._1 == "this").size == argTypes.size)
-        )
+      membersInScope.collectFirst(matchingM(callName)).orElse {
```
DavidBakerEffendi, line 130: Could we give `argTypes` a default so that nonoverloaded languages can ignore this? E.g. `argTypes: List[String] = Nil`

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -223,6 +215,51 @@ trait TypedScope[M <: MethodLike, F <: FieldLike, T <: TypeLike[M, F]](summary:
...
+    case Some(tfn) =>
+      val methodsWithEqualArgs = tryResolveTypeReference(tfn).flatMap { t =>
+        Option(
+          t.methods.filter(m => m.name == callName && m.parameterTypes.filterNot(_._1 == "this").size == argTypes.size)
+        )
+      }
+
+      methodsWithEqualArgs
+        .getOrElse(List.empty[M])
+        .find(isOverloadedBy(_, argTypes)) match {
```
DavidBakerEffendi, line 239:
```suggestion
        .find(isOverloadedBy(_, argTypes))
        .orElse(methodsWithEqualArgs.getOrElse(List.empty[M]).headOption)
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -223,6 +215,51 @@ trait TypedScope[M <: MethodLike, F <: FieldLike, T <: TypeLike[M, F]](summary:
...
+  )(implicit tag: ClassTag[M]): Option[M] = typeFullName match {
+    case None =>
+      // TODO: The typesInScope part is to imprecisely solve the unimplemented polymorphism limitation
+      membersInScope.collectFirst(matchingM(callName)).orElse {
+        typesInScope.flatMap(_.methods).collectFirst(matchingM(callName))
+      }
+    case Some(tfn) =>
+      val methodsWithEqualArgs = tryResolveTypeReference(tfn).flatMap { t =>
+        Option(
+          t.methods.filter(m => m.name == callName && m.parameterTypes.filterNot(_._1 == "this").size == argTypes.size)
```
DavidBakerEffendi, line 233: I feel we should use `isOverloadedBy` here, I think this solution mitigates `this` since `argTypes` doesn't include the receiver when used in `CSharp2Cpg`'s AstCreator. If possible, it would be a nice fix here but might be out of scope if too large
```suggestion
          t.methods.filter(m => m.name == callName && isOverloadedBy(m, argTypes))
```

DavidBakerEffendi (changes requested):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
AndreiDreyer: This causes some C# tests to fall over, added TODO comment for now.

DavidBakerEffendi (approved):

PR #4543, opened 2024-05-07
[javasrc2cpg] Preserve path in ConfigFile name

xavierpinho:
When going from 2.0.352 to 2.0.353, we noticed that `ConfigFile`'s `.name` property no longer contained their path, but rather just their file name. 

An alternative fix would be to explicitly provide `rootDir = Option(config.inputPath)` in https://github.com/joernio/joern/blob/ebd93bf07ab3261c25fb27756d772768fec02979/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala#L34

but since XConfigFileCreationPass is expected to gather that from `metaData.root` when omitted, figured it would be preferable to patch it instead.


johannescoetzee (approved):

DavidBakerEffendi (approved):

johannescoetzee:
The change looks fine, other than windows separator issues in the unit test

PR #4545, opened 2024-05-07
[javasrc2cpg] fix ext method call full name with same name of a scope var

xavierpinho:
When dealing with:
* external method calls whose receiver type we can't resolve, and
* if there is a variable in scope with that same name, then

 we are resolving to that variable's type, as in:

```java
import org.Client;
import org.Builder;
...
Client foo = new Builder().foo().build(); // build's receiver is org.Client
Client bar = new Builder().somethingElse().build(); // build's receiver is <unresolvedNamespace>
```

This gets worse: if in the 2nd statement above we had used `foo()` again instead of `somethingElse()`, it would resolve again to `org.Client`.

This patch stinks, though: I'm not a fan of that edge condition. Hopefully we can find something nicer.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
@@ -303,7 +303,7 @@ class AstCreator(
       scope.lookupType(annotation.getNameAsString)
     case namedExpr: NodeWithName[_] =>
       scope.lookupVariableOrType(namedExpr.getNameAsString)
-    case namedExpr: NodeWithSimpleName[_] =>
+    case namedExpr: NodeWithSimpleName[_] if !expr.isMethodCallExpr =>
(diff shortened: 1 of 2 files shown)


xavierpinho:
> Ideally, our namespace would be `org.Builder` right? Is it a large effort (or unsound in some cases) to achieve this result based on imports?

Not always, would depend on the return type of `foo()` and `somethingElse()`. Some libraries would return different types so that the order of operations/methods we can call is coherent.

johannescoetzee (approved):
This is a good catch! I think this patch is fine to merge to fix it for now. I wonder about matching the `NodeWithSimpleName` there in general. I originally did it that way to keep the code simple, but I didn't consider all the cases. It might make more sense to only match the specific node types we want to look up (so possibly just identifiers and types, but we'd have to consider all options)

johannescoetzee:
> Ideally, our namespace would be `org.Builder` right? Is it a large effort (or unsound in some cases) to achieve this result based on imports?

The namespace should already be `org.Builder` for the `foo` and `somethingElse` calls based on the imports. We don't know anything about the return types of those methods though, so can't say anything sensible about the namespace for the `build` calls.

PR #4550, opened 2024-05-10
 [jssrc2cpg] Fix for TypeDecl creation from TypeAliasNode for function params

max-leuthaeuser:
Assignment calls from TypeAliasNode members created for object parameters to functions are not stored directly under the TypeDecl any longer. We only attach members now.

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/types/TSTypesTests.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/types/TSTypesTests.scala
@@ -152,11 +152,10 @@ class TSTypesTests extends AstJsSrc2CpgSuite {
      |""".stripMargin,
       "Test0.ts"
     ).withConfig(Config().withTsTypes(true))
-    inside(cpg.identifier.l) { case List(idX) =>
-      idX.name shouldBe "x"
-      idX.code shouldBe "x"
-      idX.typeFullName shouldBe Defines.String // we can actually follow type intrinsics
-    }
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/types/TSTypesTests.scala
```diff
@@ -152,11 +152,10 @@ class TSTypesTests extends AstJsSrc2CpgSuite {
...
     ).withConfig(Config().withTsTypes(true))
-    inside(cpg.identifier.l) { case List(idX) =>
-      idX.name shouldBe "x"
-      idX.code shouldBe "x"
-      idX.typeFullName shouldBe Defines.String // we can actually follow type intrinsics
-    }
+    val List(idX) = cpg.identifier.nameExact("x").l
+    idX.name shouldBe "x"
+    idX.code shouldBe "x"
+    idX.typeFullName shouldBe Defines.String // we can actually follow type intrinsics
```
maltek, line 158: unlike the `inside` from the original code, this non-exhaustive match will cause a compiler warning.

maltek (approved):

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/types/TSTypesTests.scala
max-leuthaeuser: Did not see that in Intellij. Will revert this change.

PR #4553, opened 2024-05-13
[python] Decorator Call Names

DavidBakerEffendi:
* Added names to decorator calls

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
@@ -259,13 +259,8 @@ class PythonAstVisitor(
   def wrapMethodRefWithDecorators(methodRefNode: nodes.NewNode, decoratorList: Iterable[ast.iexpr]): nodes.NewNode = {
     decoratorList.foldRight(methodRefNode)((decorator: ast.iexpr, wrappedMethodRef: nodes.NewNode) =>
       val (decoratorNode, decoratorName) = convert(decorator) match {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
```diff
@@ -252,13 +252,22 @@ class PythonAstVisitor(
...
    * f1(arg)(f2(func))
    */
   def wrapMethodRefWithDecorators(methodRefNode: nodes.NewNode, decoratorList: Iterable[ast.iexpr]): nodes.NewNode = {
     decoratorList.foldRight(methodRefNode)((decorator: ast.iexpr, wrappedMethodRef: nodes.NewNode) =>
-      createCall(convert(decorator), "", lineAndColOf(decorator), wrappedMethodRef :: Nil, Nil)
+      val (decoratorNode, decoratorName) = convert(decorator) match {
+        case decoratorNode: NewIdentifier =>
+          decoratorNode -> decoratorNode.name
+        case decoratorNode: NewCall =>
+          decoratorNode -> decoratorNode.name
```
ml86, line 265: Only for identifiers giving a name makes somewhat sense. In the `NewCall` case we have no clue about the name of the returned decorator function.

ml86 (changes requested):

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
ml86: So in that case please leave the name empty.

ml86 (approved):

PR #4559, opened 2024-05-15
[dataflowengineoss] globalFromLit only when it's a standalone assignment

### LINKED ISSUE #4549: [python] incoherent flow when using a literal vs identifier argument
# Setup

## Query

```scala
def source = cpg.literal("20")
def sink = cpg.call("sink").argument
sink.reachableByFlows(source)
```

## Sample 1

```python
x = foo(20)
sink(x)
```

## Sample 2

```python
y = 20
x = foo(y)
sink(x)
```

# Results

For Sample 1, we obtain the following 2 flows:

1. LITERAL 20
2. CALL foo(20)
3. IDENTIFIER x (the assignment)
4. IDENTIFIER x (the argument)

and

1. LITERAL 20
2. IDENTIFIER x (the assignment)
3. IDENTIFIER x (the argument)

For Sample 2, we obtain a single flow:

1. LITERAL 20
2. IDENTIFIER y (the assignment)
3. IDENTIFIER y (the argument)
4. CALL foo(y)
5. IDENTIFIER x (the assignment)
6. IDENTIFIER x (the argument)

# Discussion

Sample 1 should not contain its second flow -- the one that jumps over the `foo` call directly to the assigned variable.

Interestingly:
* in Sample 2 `foo` has methodFullName `__builtin.int.foo`
* in Sample 1 `foo` has methodFullName `<unknownFullName>`

xavierpinho:
In #4549, we noticed an extra flow in the following sample:

```python
x = foo(20) # with source being the LITERAL
sink(x)
```

This was traced to `globalFromLiteral`, in which the statement `x = foo(20)` would consider `x`, adding it to the starting points of the query.

In this patch we only trigger if the literal is the RHS of the assignment. 

I may be missing some relevant context here that I didn't gather from the failing unit-tests. Before going through the trouble of fixing them all, is this patch coherent?

EDIT: Resolves #4549

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/dataflow/SingleAssignmentTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/dataflow/SingleAssignmentTests.scala
@@ -16,14 +16,12 @@ class SingleAssignmentTests extends RubyCode2CpgFixture(withPostProcessing = tru
     val source = cpg.literal.l
     val sink   = cpg.method.name("puts").callIn.argument.l
     val flows  = sink.reachableByFlows(source).map(flowToResultPairs).distinct.sortBy(_.length).l
-    flows.size shouldBe 6
-    val List(flow1, flow2, flow3, flow4, flow5, flow6) = flows
+    val List(flow1, flow2, flow3, flow4, flow5) = flows
     flow1 shouldBe List(("y = 1", 2), ("puts y", 3))
-    flow2 shouldBe List(("y = 1", 2), ("x = y = 1", 2), ("puts x", 4))
-    flow3 shouldBe List(("y = 1", 2), ("z = x = y = 1", 2), ("puts z", 5))
-    flow4 shouldBe List(("y = 1", 2), ("puts y", 3), ("puts x", 4))
+    flow2 shouldBe List(("y = 1", 2), ("puts y", 3), ("puts x", 4))
+    flow3 shouldBe List(("y = 1", 2), ("x = y = 1", 2), ("puts x", 4))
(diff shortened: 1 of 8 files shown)


DavidBakerEffendi:
This patch looks coherent, I see the `inAssignment` was unintentional in its recursive upward behaviour. Looking at the tests now.

DavidBakerEffendi (approved):
I agree with this way of expressing the query, and the failing tests lose 1 flow but they are never 0, so I don't think this is an issue and rather a side effect of overtainting in the original query

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/dataflow/SingleAssignmentTests.scala
```diff
@@ -16,14 +16,12 @@ class SingleAssignmentTests extends RubyCode2CpgFixture(withPostProcessing = tru
     val source = cpg.literal.l
     val sink   = cpg.method.name("puts").callIn.argument.l
     val flows  = sink.reachableByFlows(source).map(flowToResultPairs).distinct.sortBy(_.length).l
-    flows.size shouldBe 6
-    val List(flow1, flow2, flow3, flow4, flow5, flow6) = flows
+    val List(flow1, flow2, flow3, flow4, flow5) = flows
```
xavierpinho, line 19: The previously called `flow3` is removed, but I reckon the new `flow4` covers it better.

xavierpinho:
Was actually expecting more tests to fail, as I don't have the required setup locally to test all frontends. Turns out there was only 7 of them and all on ruby. Took the opportunity to make those flow tests more explicit, instead of just looking at their sizes. Please kindly see if the changes committed after your approval still make sense, @DavidBakerEffendi

DavidBakerEffendi (approved):
LGTM

PR #4562, opened 2024-05-16
Add MethodRefeferenceExpr to javasrc asts

johannescoetzee:
Currently these are just `Unknown` nodes in the AST


max-leuthaeuser (approved):

maltek (approved):
how does currently deal with expressions like `someObject::method`? (I don't to block this PR on complete handling of the capturing semantics. Just want to mention that's a thing that's not covered by the tests, and hence I'm unsure what happens with those.)

johannescoetzee:
> how does currently deal with expressions like `someObject::method`? (I don't to block this PR on complete handling of the capturing semantics. Just want to mention that's a thing that's not covered by the tests, and hence I'm unsure what happens with those.)

I didn't consider this case and it turns out it wasn't handled as well as it could be. I've updated the PR with type lookups more consistent with the rest of javasrc and some tests

maltek (approved):

PR #4565, opened 2024-05-16
[ruby] `require*` and `load` import nodes

DavidBakerEffendi:
Creating import nodes for `require_all`, `require_relative`, and `load` calls. `require_all` is interpreted as a wildcard import as it imports a whole directory.

Additionally handling the `require_relative` and `require_all` correctly as per their semantics with accompanying tests.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ImportTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ImportTests.scala
@@ -230,4 +230,61 @@ class ImportTests extends RubyCode2CpgFixture with Inspectors {
       }
     }
   }
+
+  "Modifying `$LOADER` with an additional entry" should {
+    val cpg = code(
+      """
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ImportTests.scala
```diff
@@ -178,4 +178,56 @@ class ImportTests extends RubyCode2CpgFixture with Inspectors {
       }
     }
   }
+
+  "`require_all` on a directory" should {
```
AndreiDreyer, line 182: Will resolution for `require_relative` be implemented at a later stage?

DavidBakerEffendi: So while I interpret `require_all` and `require_relative` as both starting paths from root, (where only `require_all` is explicitly tested), I should actually also keep an eye out for `$LOADER << ` calls which mean that I can't only consider root alone. Let me add handling for that

AndreiDreyer (approved):

PR #4586, opened 2024-05-22
[ruby] add missing token used to decide regex start

xavierpinho:
The lexer predicate that decides when a `/.../` regex literal is about to start looks at the previous token to decide... and it was missing the `=>` token 🤦 

Credit for finding: @DavidBakerEffendi


DavidBakerEffendi (approved):
Ah, that's one place I didn't think to look for a fix!

PR #4588, opened 2024-05-23
[ruby] Grouped `logger` bug fixes

AndreiDreyer:
The [logger](https://github.com/ruby/logger) repository causes this frontend to throw a lot of warnings. This PR fixes these warnings

* Handles regexs used in conditions of IF statements
* Added `<<` as a valid `definedMethodName`
* Added handling of `rescue` as a modifier statement seen in the logger repo with `@dev.close rescue nil`

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RegexLiteralHandling.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RegexLiteralHandling.scala
@@ -69,10 +69,17 @@ trait RegexLiteralHandling { this: RubyLexerBase =>
     * may not start with a WS. E.g. `puts /x/` is valid, but `puts / x/` is not.
     */
   private def isInCommandArgumentPosition: Boolean = {
-    val previousNonWsIsIdentifier = previousNonWsTokenTypeOrEOF() == LOCAL_VARIABLE_IDENTIFIER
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RegexLiteralHandling.scala
```diff
@@ -69,10 +69,19 @@ trait RegexLiteralHandling { this: RubyLexerBase =>
...
+    val previousNonWsIsIdentifier =
+      previousNonWsTokenTypeOrEOF() == LOCAL_VARIABLE_IDENTIFIER || isControlStructureStart
+    val previousIsWs = previousTokenTypeOrEOF() == WS
+    val nextCharIsWs = _input.LA(1) == ' '
     previousNonWsIsIdentifier && previousIsWs && !nextCharIsWs
   }
 
+  private def isControlStructureStart: Boolean = {
+    previousNonWsTokenTypeOrEOF() == IF
+    // TODO: Determine which of the below control structures are also valid for regex use
```
DavidBakerEffendi, line 81: I think we should address this condition in this PR. Maybe do a quick check in an interpreter, then we avoid having to revisit this

DavidBakerEffendi (approved):
Also note this PR touching regexes: https://github.com/joernio/joern/pull/4586
I'll fix the conflict there too in the meantime

PR #4591, opened 2024-05-23
[pysrc2cpg] Use ControlStructureTypes for Try-Catch-Clauses

max-leuthaeuser:
Uses `ControlStructureTypes.CATCH`, `ControlStructureTypes.FINALLY`, and `ControlStructureTypes.ELSE` now instead of relying on explicit order values.

Will do the same for all the other Joern frontends in individual PRs.

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/TryCpgTests.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/TryCpgTests.scala
@@ -1,0 +1,40 @@
+package io.joern.pysrc2cpg.cpg
+
+import io.joern.pysrc2cpg.Py2CpgTestContext
+import io.shiftleft.semanticcpg.language.*
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.should.Matchers
+
+class TryCpgTests extends AnyFreeSpec with Matchers {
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/TryCpgTests.scala
```diff
@@ -0,0 +1,40 @@
...
+import io.shiftleft.semanticcpg.language.*
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.should.Matchers
+
+class TryCpgTests extends AnyFreeSpec with Matchers {
+  private val cpg = Py2CpgTestContext.buildCpg("""
+      |def divide(x, y):
+      |  try:
+      |    result = x / y
+      |  except ZeroDivisionError:
```
johannescoetzee, line 13: This could already be an issue with try/catch, but it looks like we currently don't include what has actually been caught in the CPG. I can't remember if that's a thing with python, but it could be an issue with java where `catch (Exception e)` can actually declare a variable.

It's not necessarily something to block this PR on, but I think it is something worth thinking about before we settle on a long-term representation

max-leuthaeuser: Right, that's a thing in some languages. For Swift I simply put the declaration as first element into the repective catch block.

ml86 (approved):

PR #4597, opened 2024-05-24
[jssrc2cpg] Use ControlStructureTypes for Try-Catch-Clauses

max-leuthaeuser:
Uses ControlStructureTypes.CATCH and ControlStructureTypes.FINALLY now instead of relying on explicit order values. 
Also add catch clause parameters (declarations) to their bodies content (which was missing entirely).

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForStatementsCreator.scala
@@ -104,8 +104,6 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
   protected def astForTryStatement(tryStmt: BabelNodeInfo): Ast = {
     val tryNode = createControlStructureNode(tryStmt, ControlStructureTypes.TRY)
     val bodyAst = astForNodeWithFunctionReference(tryStmt.json("block"))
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -85,28 +86,46 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
...
+    setArgumentIndices(blockStatementAsts)
+    localAstParentStack.pop()
+    scope.popScope()
+    blockAst(blockNode, blockStatementAsts)
+  }
 
   protected def astForTryStatement(tryStmt: BabelNodeInfo): Ast = {
     val tryNode = createControlStructureNode(tryStmt, ControlStructureTypes.TRY)
     val bodyAst = astForNodeWithFunctionReference(tryStmt.json("block"))
+    setOrderExplicitly(bodyAst, 1)
```
ml86, line 107: Why do we still need to set t he order explicitly here and for `catchAst` and `finalAst`?

max-leuthaeuser: Right, not required anymore. But I need to set the index. Will fix.

ml86 (approved):

PR #4603, opened 2024-05-27
[ruby] Method Bindings & Constructor Modifier

### LINKED ISSUE #4593: [ruby] Method Bindings
For methods exported by Ruby scripts, we require a corresponding type declaration as they can be assigned to variables, etc., and the identifiers' type hint will be this method. 

In order to then link back to the method, we need a binding between the type decl and the method.

One should consider JavaScript and Python frontends for examples.

DavidBakerEffendi:
Methods require binding/ref edges with their corresponding type declaration in dynamic languages to correctly model their object-like behaviour. This PR implements these bindings following `pysrc2cpg`'s `createBinding` function.

Additionally, adds the `CONSTRUCTOR` modifier to constructors.

Resolves #4593

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
@@ -3,28 +3,40 @@ package io.joern.rubysrc2cpg.querying
 import io.joern.rubysrc2cpg.testfixtures.RubyCode2CpgFixture
 import io.joern.x2cpg.Defines
 import io.joern.rubysrc2cpg.passes.Defines as RDefines
-import io.shiftleft.codepropertygraph.generated.{ControlStructureTypes, Operators}
+import io.shiftleft.codepropertygraph.generated.{ControlStructureTypes, NodeTypes, Operators}
 import io.shiftleft.codepropertygraph.generated.nodes.{Call, Identifier, Literal, MethodRef, Return, TypeRef}
 import io.shiftleft.semanticcpg.language.*
 
 class MethodTests extends RubyCode2CpgFixture {
 
-  "`def f(x) = 1` is represented by a METHOD node" in {
+  "`def f(x) = 1`" should {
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
```diff
@@ -3,28 +3,40 @@ package io.joern.rubysrc2cpg.querying
...
-    x.index shouldBe 1
-    x.isVariadic shouldBe false
-    x.lineNumber shouldBe Some(2)
+      val List(x) = f.parameter.name("x").l
+      x.index shouldBe 1
+      x.isVariadic shouldBe false
+      x.lineNumber shouldBe Some(2)
+    }
+
+    "have a corresponding bound type" in {
```
xavierpinho, line 31: This (having a typeDecl for the method as well) caught me by surprise, I'll admit. 

Is the intent to capture bound variables? Or is it due to some Ruby-specific behaviour I'm not familiar with, maybe?

ml86 (approved):

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
DavidBakerEffendi: @xavierpinho You'll see a similar thing in Python/JavaScript etc. 

Imagine 
```ruby
class Foo
  def bar
  end
end

foo = Foo.new
a = foo.method(:bar)
```

`a.typeFullName` would point to `Foo.bar`, but since the property is `typeFullName`, the linker will look for a `TypeDecl`, not a `Method`. The binding would then point to a method if this is a method pointer instead of a traditional type.

DavidBakerEffendi:
(On hold, checking `codescience` tests)

DavidBakerEffendi:
Hmm I see a data-flow test failing on `codescience` Ruby tests, but not from this PR so I'll merge

PR #4614, opened 2024-05-30
[python] prevent overtainting dictLiterals

### LINKED ISSUE #4605: [python] over-tainting dictionary literal arguments
# Brief

Dictionary literals in argument position are over-tainted even when providing `FlowSemantic`s specifying otherwise. Compare Tests 1 and 2 with 3 below.

# Setup

Joern version: 2.0.386

## Test 1 ✅ 

This works as expected. `print` is an external call so it's sound to say that `1` taints `x`. 

```scala
"Scenario 1" in {
  val cpg = code(
    """
      |x = {'x': 10}
      |print(1, x)
      |""".stripMargin)
  def source = cpg.literal
  def sink = cpg.call("print").argument.argumentIndex(2)
  val result = sink.reachableByFlows(source).l
  result.map(_.elements.head.code) shouldBe List("1", "10")
}
```

## Test 2 ✅ 

This works as expected. If we can say that `print` does not cross-taint its arguments, then the only literal tainting `x` (the argument) is `10`.

```scala
"Scenario 2" in {
  val cpg = code(
    """
      |x = {'x': 10}
      |print(1, x)
      |""".stripMargin)
    .withExtraFlows(List(FlowSemantic(".*print", List(FlowMapping(1, 1), FlowMapping(2, 2)), true)))
  def source = cpg.literal
  def sink = cpg.call("print").argument.argumentIndex(2)
  val result = sink.reachableByFlows(source).l
  result.map(_.elements.head.code) shouldBe List("10")
}
```

## Test 3 ❌ 

This sample is equivalent to that of Test 2. However, we get the same results of Test 1, i.e. `1` is also tainting the second argument. Confirmed that the methodFullName of `print` didn't change.

```scala
"Scenario 3" in {
  val cpg = code(
    """
      |print(1, {'x': 10})
      |""".stripMargin)
    .withExtraFlows(List(FlowSemantic(".*print", List(FlowMapping(1, 1), FlowMapping(2, 2)), true)))
  def source = cpg.literal
  def sink = cpg.call("print").argument.argumentIndex(2)
  val result = sink.reachableByFlows(source).l
  result.map(_.elements.head.code) shouldBe List("10")
}
```

# Notes

Related to #4549 in the sense that the original behaviour of `globalFromLiteral` would have Test 3 passing.

xavierpinho:
In #4583, I didn't account for how Python dictionary literals are lowered as 3 address code. Specifically, the problem is at `inCall.assignment` here:

https://github.com/joernio/joern/blob/ff9c08f62539eba4194eee5c0a96f5d28d586ab7/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/package.scala#L19

In an expression such as `print(1, {'x': 10})`, `1.inCall.assignment` wouldn't be empty - it would contain the assignments that build `{'x': 10}`. The intent behind `inCall.assignment` was to filter the surrounding call by assignment. So, introduced a new CallTraversal `isAssignment` to do just that and used it here instead.

I'm a bit concerned about introducing a new API for CallTraversal. Any thoughts about that?

Resolves #4605

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/package.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/package.scala
@@ -16,7 +16,7 @@ package object dataflowengineoss {
     */
   def globalFromLiteral(lit: Literal, recursive: Boolean = true): Iterator[Expression] = lit.start
     .where(_.method.isModule)
-    .flatMap(t => if (recursive) t.inAssignment else t.inCall.assignment)
+    .flatMap(t => if (recursive) t.inAssignment else t.inCall.isAssignment)
     .target
 
   def identifierToFirstUsages(node: Identifier): List[Identifier] = node.refsTo.flatMap(identifiersFromCapturedScopes).l
--- a/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/dataflow/DataFlowTests.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/dataflow/DataFlowTests.scala
@@ -1,7 +1,7 @@
 package io.joern.pysrc2cpg.dataflow
 
 import io.joern.dataflowengineoss.language.toExtendedCfgNode
-import io.joern.dataflowengineoss.semanticsloader.FlowSemantic
+import io.joern.dataflowengineoss.semanticsloader.{FlowMapping, FlowSemantic, PassThroughMapping}
 import io.joern.pysrc2cpg.PySrc2CpgFixture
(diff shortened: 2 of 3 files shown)


maltek:
> I'm a bit concerned about introducing a new API for CallTraversal. Any thoughts about that?

👍 from me on that. I'd use that API. (Unlike the existing `inAssignment`/`assignment`, for the same reasons why you had to move away from them here.) Only comment from me is whether it shouldn't filter by all the fullnames from https://github.com/joernio/joern/blob/15e241d3174ecba9e977a399793c9c6a1249d819/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/operatorextension/package.scala#L24

DavidBakerEffendi (approved):
Tested on my side too, looks good!

PR #4615, opened 2024-05-30
[javasrc2cpg] Remove generic types from unresolved fullnames

johannescoetzee:
At the last level of our type fallback mechanisms, we use `javaParserNode.getTypeAsString` to get the type name, but this still includes generic types which is inconsistent with what we do elsewhere. This PR strips the generic types from these fallback type names.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/GenericsTests.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/GenericsTests.scala
@@ -4,55 +4,6 @@ import io.joern.javasrc2cpg.testfixtures.JavaSrcCode2CpgFixture
 import io.shiftleft.semanticcpg.language._
 
 class GenericsTests extends JavaSrcCode2CpgFixture {
-
-  "resolved generic type declarations" should {
-    val cpg = code("""package box;
-                     |
-                     |public class Box<T> {
-                     |  public Box(T value) {}
-                     |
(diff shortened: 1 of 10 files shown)


max-leuthaeuser (approved):

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/GenericsTests.scala
```diff
@@ -4,195 +4,296 @@ import io.joern.javasrc2cpg.testfixtures.JavaSrcCode2CpgFixture
...
+                     |class Box<T> {
+                     |  public <S> Box<S> into() { return null; }
+                     |
+                     |  public T get() { return null; }
+                     |
+                     |  static void test(Box<String> stringBox) {
+                     |    stringBox.<Integer>into().get();
+                     |  }
+                     |}
+                     |""".stripMargin)
```
ml86, line 66: Is this just a smoke test? If so please add a comment stating this.

ml86 (approved):

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/GenericsTests.scala
johannescoetzee: Ah, I think this is a test I was halfway through writing before moving over to something else. I'll add the rest of the test now. It's not strictly speaking relevant to this PR, but is a gap in our generic testing

PR #4617, opened 2024-05-30
[javasrc2cpg] Don't add unknown nodes to cpg for unhandled captured variables

johannescoetzee:
Due to an interaction between regular/anonymous class capture logic and lambda capture logic, the assumption that a `this` parameter will always be available in the direct enclosing method when handling captures is sometimes false. In this case, an `Unknown` node was added to the cpg, causing dynamic call linker failures.

This pr fixes the lambda issue by seaching the entire scope for a `this` param, which correctly (-ish, see note below) creates the field access with the correct type, but also replaces the `Unknown` fallback with an identifier which would still result in dynamic linking failing (probably), but won't crash.

Note on captured params in lambdas: the current representation just treats them as `this.capturedField`, but with the type of `this` set to the class in which it is declared. I would be very surprised if this works, so a rework is required (creating an `outerClass` local in the lambda method with a closure binding pointing at the captured this parameter, for example), but that's out of the scope of this PR

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
@@ -743,28 +743,71 @@ class LambdaTests extends JavaSrcCode2CpgFixture {
     }
   }
 
-  "calls on captured variables in lambdas" should {
+  "calls on captured variables in lambdas contained in anonymous classes" should {
     val cpg = code("""
-                     |
-                     |public class Foo {
-                     |
-                     |    public static void sink(String s) {};
-                     |
-                     |    public static Object test(Foo captured) {
-                     |      Visitor v = new Visitor() {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala
```diff
@@ -742,4 +742,72 @@ class LambdaTests extends JavaSrcCode2CpgFixture {
...
+        |      Visitor v = new Visitor() {
+        |        public void visit(Visited visited) {
+        |          visited.getList().forEach(lambdaParam -> captured.remove(lambdaParam));
+        |        }
+        |      };
+        |    }
+        |}
+        |""".stripMargin)
+
+    // This behaviour isn't exactly correct, but is on par with how we currently handle field captures in lambdas.
```
ml86, line 763: Please add `TODO` here and on the test cases below.

ml86 (approved):

PR #4633, opened 2024-06-04
use generated Cpg class

mpollmeier:
also to minify the diff for the flatgraph port


bbrehm (approved):
Couldn't we instead have a `type Cpg = io.shiftleft.codepropertygraph.generated.Cpg` in the `io.shiftleft.codepropertygraph` package?

mpollmeier:
We could, after we get rid of the `object io.shiftleft.codepropertygraph.Cpg`. 
But: we could do the same for every other generated type, but why would we?

PR #4634, opened 2024-06-04
[javascript] Object Property Call Linker

DavidBakerEffendi:
A common pattern in JavaScript is something like

```javascript
var foo = {};
foo.bar = {};
foo.bar.someFunc = function someFunc() {};

foo.bar.someFunc();
```

This PR adds a post-processing pass to find instances where the definition and the call live in the same file and to link them. This "common file" limitation aims to reduce false linking.

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/ObjectPropertyCallLinker.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/ObjectPropertyCallLinker.scala
@@ -1,0 +1,42 @@
+package io.joern.jssrc2cpg.passes
+
+import io.shiftleft.codepropertygraph.generated.nodes.{Call, MethodRef}
+import io.shiftleft.codepropertygraph.generated.{Cpg, PropertyNames}
+import io.shiftleft.passes.CpgPass
+import overflowdb.BatchedUpdate
+import io.shiftleft.semanticcpg.language.*
+
+/** Perform a simple analysis to find a common pattern in JavaScript where objects are dynamically assigned function
+  * pointers. To keep this precise, this will only match objects defined within the scope of the same file.
+  *
+  * This relies on JavaScriptTypeHintCallLinker.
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/ObjectPropertyCallLinker.scala
```diff
@@ -0,0 +1,42 @@
...
+/** Perform a simple analysis to find a common pattern in JavaScript where objects are dynamically assigned function
+  * pointers. To keep this precise, this will only match objects defined within the scope of the same file.
+  *
+  * This relies on JavaScriptTypeHintCallLinker.
+  */
+class ObjectPropertyCallLinker(cpg: Cpg) extends CpgPass(cpg) {
+
+  override def run(builder: BatchedUpdate.DiffGraphBuilder): Unit = {
+    val propertyCallRegex = "^(?:\\{.*\\}|.*<returnValue>):<member>\\((.*)\\):.*$".r
+    val objectCalls       = cpg.call.methodFullName("^(?:\\{.*\\}|.*<returnValue>):<member>\\(.*\\):.*$").l
```
max-leuthaeuser, line 18: Isnt that `propertyCallRegex`?

max-leuthaeuser (approved):

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/passes/ObjectPropertyCallLinker.scala
DavidBakerEffendi: `\\(.*\\)` versus `\\((.*)\\)`

DavidBakerEffendi: Might be useful to combine this into a regex generator function to reduce duplication

PR #4645, opened 2024-06-06
[ruby] `Kernel` Methods are Static Dispatch

DavidBakerEffendi:
To simplify the call graph, methods resolved to the `Kernel` class are dispatched as static.

Additionally, replaced `__builtin` with `Kernel` to map more closely to Ruby's internals and implemented `MemberAccessCommandContext`.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
@@ -684,7 +684,7 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) { 
       }
     val argumentAst = node.arguments.map(astForMethodCallArgument)
     val dispatchType =
-      if methodName.startsWith(GlobalTypes.builtinPrefix) then DispatchTypes.STATIC_DISPATCH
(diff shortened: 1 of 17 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -367,11 +371,15 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
   protected def astForSimpleIdentifier(node: RubyNode & RubyIdentifier): Ast = {
     val name = code(node)
 
-    scope.lookupVariable(name) match {
-      case Some(_) => handleVariableOccurrence(node)
-      case None if scope.tryResolveMethodInvocation(node.text).isDefined =>
-        astForSimpleCall(SimpleCall(node, List())(node.span))
-      case None => handleVariableOccurrence(node)
+    if (name == GlobalTypes.Kernel) {
```
ml86, line 374: Not only the name `Kernel` needs to be matched. According to this list https://ruby-doc.org/docs/ruby-doc-bundle/Manual/man-1.4/classes.html also those:
Comparable
Enumerable
Errno
FileTest
GC
Kernel
Marshal
Math
ObjectSpace
Precision
Process

ml86 (approved):

PR #4651, opened 2024-06-09
[javasrc2cpg] add AstCreator `code` implementation removing comments 

xavierpinho:
The `.code` field for arguments presently includes comments, e.g. in

```java
Foo.create(
  "username", // hehe, silly comment
  "password"
);
```

the `.code` for `argument(2)` is ` // hehe, silly comment\n  "password"`.

This patch:
1. provides a default implementation for AstCreator's `code(Node): String` method that prints the node's text as if it didn't contain any comments attached, and
2. uses it (instead of `.toString`) throughout AstCreator.

Credit for finding: @sean-gahagan

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
@@ -129,7 +129,7 @@ class AstCreator(
   protected def column(node: Node): Option[Integer]    = node.getBegin.map(x => Integer.valueOf(x.column)).toScala
   protected def lineEnd(node: Node): Option[Integer]   = node.getEnd.map(x => Integer.valueOf(x.line)).toScala
   protected def columnEnd(node: Node): Option[Integer] = node.getEnd.map(x => Integer.valueOf(x.column)).toScala
-  protected def code(node: Node): String               = "" // TODO: javasrc2cpg uses custom code strings everywhere
+  protected def code(node: Node): String               = node.clone.removeComment().toString
 
   private val lineOffsetTable = OffsetUtils.getLineOffsetTable(fileContent)
 
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
@@ -15,6 +15,8 @@ import com.github.javaparser.ast.expr.{
 }
 import com.github.javaparser.ast.nodeTypes.{NodeWithName, NodeWithSimpleName}
 import com.github.javaparser.ast.{CompilationUnit, ImportDeclaration, Node, PackageDeclaration}
+import com.github.javaparser.printer.configuration.DefaultPrinterConfiguration.ConfigOption
+import com.github.javaparser.printer.configuration.{DefaultConfigurationOption, DefaultPrinterConfiguration}
 import com.github.javaparser.resolution.UnsolvedSymbolException
 import com.github.javaparser.resolution.declarations.{
   ResolvedMethodDeclaration,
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
```diff
@@ -129,7 +129,7 @@ class AstCreator(
   protected def column(node: Node): Option[Integer]    = node.getBegin.map(x => Integer.valueOf(x.column)).toScala
   protected def lineEnd(node: Node): Option[Integer]   = node.getEnd.map(x => Integer.valueOf(x.line)).toScala
   protected def columnEnd(node: Node): Option[Integer] = node.getEnd.map(x => Integer.valueOf(x.column)).toScala
-  protected def code(node: Node): String               = "" // TODO: javasrc2cpg uses custom code strings everywhere
+  protected def code(node: Node): String               = node.clone.removeComment().toString
```
max-leuthaeuser, line 132: Just out of curiosity: is the clone call actually necessary?

xavierpinho: Indeed, I don't like it either. Fortunately, it's not "recursive" and we are only using it in this patch for basic nodes - javasrc2cpg crafts the `code` itself for more complex nodes.

`removeComment` destructively updates the node, and I didn't want to bring any side-effects into this, hence the clone. Other than that, I first tried to do `getTokenRange`, filter out comment nodes (cf. `getCategory`), but that has the disadvantage of missing whitespace - e.g. `1 + 2` would be written as `1+2`, since whitespace tokens are not in the token range. Design-wise, comments are attached to nodes, but I couldn't find a reasonable way to discard them other than with `removeComment` -- maybe @johannescoetzee knows, as he's more familiar with javaparser than I am?

xavierpinho: Alternatively, this removal of comment nodes is mostly relevant when we're dealing with literals, since their code is their value. I'm happy to restrict this patch to those only until we find a better solution, as well.

johannescoetzee: `removeComment` modifies the original node, so it is necessary except if we specifically handle comments before code (although javasrc2cpg doesn't currently handle comments at all, if I recall correctly)

johannescoetzee (approved):
This is a sensible default code implementation in my opinion. The `code` fields for many node types reflect the lowered representation, so it's not one we should adopt in general (although a case-by-case review could be useful at some point), but in simple `toString` cases like these it works well

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
johannescoetzee: > I'm happy to restrict this patch to those only until we find a better solution

In my opinion, handling that case specifically adds more code complexity which outweighs the performance gain. With that being said, it does look like `clone` does do a deep clone, so an uglier, but more performant option that doesn't rely on handling any specific case could be to remove the comment, get the string representation and then add the comment again

johannescoetzee:
It looks like there could be a better way with printer configuration options. I'll post an example here in a minute

johannescoetzee:
The PrettyPrinter can be configured to ignore comments with the below
```
val config = new DefaultPrinterConfiguration();
config.removeOption(new DefaultConfigurationOption(ConfigOption.PRINT_COMMENTS))
node.toString(config)
```

xavierpinho:
> The PrettyPrinter can be configured to ignore comments with the below

Thanks so much! Much cleaner indeed.

> It might make sense to disable javadoc printing too, although I think it's unlikely we'll use this method for method code fields

Added as well for coherence's sake.

johannescoetzee (approved):

PR #4653, opened 2024-06-10
[scan] Taint Analysis Queries & Re-use

DavidBakerEffendi:
The goal of this is to re-use as much of the query DB mechanisms as possible, while only having `reachableByFlows` run once for this subset.

Right now, queries define some subset of sources and sinks, but any source flowing to any sink is a problem. This changes allows queries tagged with `taint` + `[source|sink|sanitizer]` to be included in the taint analysis scan, where flows and results are gathered dynamically.

Additionally, only the language's semantics are included in this scan.

Note: It appears that `joern-scan` options regarding query filtering are not being propagated.


DavidBakerEffendi:
Will be following up with some rules for JavaScript and Python based on some benchmarks I'm running

DavidBakerEffendi:
I feel this might be best as a standalone tool somewhere else, closing this for now

PR #4656, opened 2024-06-11
[TypeRecovery] Makes sure the receiver's index matches the dispatch type

xavierpinho:
The starting point here was a discrepancy wrt `methodFullName`s in Python for the following two samples:

```python
a = 10
foo(a) # methodFullName = __builtin.int.foo
```
vs
```python
foo(10) # methodFullName = <unknownFullName>
```

It was noticed that in `setTypeInformationForRecCall` the argument `i` isn't always set according to the convention that the receiver should have argument index 0 when it's a DYNAMIC_DISPATCH call and 1 when it's a STATIC_DISPATCH. This patch requires the convention to be satisfied. In turn, had to ignore 2 (deprecated) Ruby tests that were not verifying this convention.


xavierpinho:
Taking a look at the failing Kotlin test.

DavidBakerEffendi (approved):

PR #4666, opened 2024-06-13
Wrap all javaparser getType calls in Try

johannescoetzee:
Related to https://github.com/joernio/joern/pull/4665

The AstCreator is crashing on a customer project due to an `UnsolvedSymbolException`. As mentioned in the PR, there isn't enough information in the logs to locate the exact issue, but my theory is that it is due to a `getType` call somewhere. For this PR, I've wrapped all `getType` calls in a `Try` (unless they were already wrapped) since the JavaParser symbol solver and the interactions between the various kinds of solvers are too complex to easily tell when these calls should be safe and when not.

Revision 2: I've changed `Try` to `tryWithSafeStackOverflow`, which also logs stack traces for errors now. This isn't called during regular type resolution, so we won't log every failed type (since those are often to be expected), but we should keep an eye on javasrc log sizes and more selectively log these if it turns out to be too much

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
@@ -244,7 +244,17 @@ class AstCreator(
 
   private[astcreation] def tryWithSafeStackOverflow[T](expr: => T): Try[T] = {
     try {
-      Try(expr)
+
+      /** JavaParser throws UnsolvedSymbolExceptions if a type cannot be solved, which is usually an expected occurrence
+        * that does not warrant specific failure logging. In some cases, such as node.getType, this can be unexpected,
+        * so log such instances (regular resolution is handled in the TypeInfoCalculator and does not use this method).
+        */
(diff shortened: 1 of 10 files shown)


ml86 (approved):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala
```diff
@@ -244,7 +246,35 @@ class AstCreator(
...
+        * that does not warrant specific failure logging. Since it's impossible to tell whether these are legitimately
+        * unresolved types or a bug, don't log them.
+        */
+      Try(expr) match {
+        case success: Success[_]                         => success
+        case Failure(exception: UnsolvedSymbolException) => Failure(exception)
+        case failure: Failure[_] =>
+          val exceptionType = failure.exception.getClass
+
+          val shouldLog = loggedExceptionCounts.get(exceptionType) match {
```
ml86, line 260: You will need to use an atomic read + update operation on the concurrent map. Like this the code is not thread safe.

ml86 (changes requested):

ml86 (approved):

PR #4668, opened 2024-06-14
[gosrc2cpg] Partial gosrc2cpg frontend perforamnce optimisations

pandurangpatil:
1. Changed parsing of AST json inside AST Creator itself.
2. Made changes in download dependency processing to generate AST only for the used packages and subsequently only processing used packages. There was a change required inside `goastgen` utility to support the `-include` input option.

TODO:
At this moment `AstCreator` gets instantiated while first-level processing for building the cache and it stays in memory till `AstCreationPass` is done. In subsequent changes, we will split the processing in such a way that we create `AstCreator` for that pass and destroy it once it's used. Instantiate it again for `AstCreationPass`.

The above change is done as part of the larger change.

### DIFF
--- a/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/DownloadDependenciesPass.scala
+++ b/joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/DownloadDependenciesPass.scala
@@ -18,9 +18,9 @@ import scala.util.{Failure, Success, Try}
 class DownloadDependenciesPass(parentGoMod: GoModHelper, goGlobal: GoGlobal, config: Config) {
   private val logger = LoggerFactory.getLogger(getClass)
   def process(): Unit = {
-    val writer       = new Writer()
-    val writerThread = new Thread(writer)
-    writerThread.start()
+    val processor       = new DependencyProcessorQueue()
+    val processorThread = new Thread(processor)
+    processorThread.start()
     File.usingTemporaryDirectory("go-temp-download") { tmpDir =>
       val projDir = tmpDir.pathAsString
       parentGoMod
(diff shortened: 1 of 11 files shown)


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/DownloadDependenciesPass.scala
```diff
@@ -12,65 +12,85 @@ import org.slf4j.LoggerFactory
...
-        .withIgnoredFiles(config.ignoredFiles.toList)
-      // TODO: Need to implement mechanism to filter and process only used namespaces(folders) of the dependency.
-      // In order to achieve this filtering, we need to add support for inclusive rule with goastgen utility first.
-      val astGenResult = new AstGenRunner(depConfig).execute(astLocation).asInstanceOf[GoAstGenRunnerResult]
-      val goMod = new GoModHelper(
-        Some(depConfig),
-        astGenResult.parsedModFile.flatMap(modFile => GoAstJsonParser.readModFile(Paths.get(modFile)).map(x => x))
-      )
-      new MethodAndTypeCacheBuilderPass(None, astGenResult.parsedFiles, depConfig, goMod, goGlobal).process()
+  private class DependencyProcessorQueue extends Runnable {
```
pandurangpatil, line 53: @DavidBakerEffendi This dependency processor queue I created to start the processing of downloaded dependency in parallel to downloading other dependencies. Earlier, it was downloading one dependency process it and then pickup another dependency. 

NOTE: We cannot parallelise downloading dependency as we use `go get <depdency>` command, which can be executed one at a time.

DavidBakerEffendi: Right, we would have to download directly from the repository otherwise, but this is fine, we don't want to put too much strain on the dependency server either as some have DDoS protection

DavidBakerEffendi (approved):

PR #4671, opened 2024-06-17
fix jssrc astgen invocation on linux/aarch64

maltek:
I don't like the amount of code we have there that pretends as if "ARM" is a single architecture... so I've moved things in a slightly better direction renaming some constants to ARMv8. (But I'm not going to fix that all completely in one day.)

The downloaded binaries actually already exist, just the integration here was missing.

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/build.sbt
+++ b/joern-cli/frontends/jssrc2cpg/build.sbt
@@ -18,8 +18,8 @@ lazy val astGenVersion = settingKey[String]("astgen version")
 astGenVersion := appProperties.value.getString("jssrc2cpg.astgen_version")
 
 libraryDependencies ++= Seq(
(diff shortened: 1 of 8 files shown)


joern-cli/frontends/jssrc2cpg/build.sbt
```diff
@@ -50,20 +51,15 @@ astGenBinaryNames := {
   if (hasCompatibleAstGenVersion(astGenVersion.value)) {
     Seq.empty
   } else if (sys.props.get("ALL_PLATFORMS").contains("TRUE")) {
-    Seq(AstgenWin, AstgenLinux, AstgenMac, AstgenMacArm)
+    Seq(AstgenWin, AstgenLinux, AstgenLinuxArm, AstgenMac, AstgenMacArm)
```
mpollmeier, line 54: `AstgenLinuxArmV8`? since you added v8 elsewhere..?

Issue #4625, opened 2024-05-31
[Bug] Joern tab completion can freeze joern and print "exception caught when loading module class module-info$: java.lang.AssertionError: assertion failed: attempt to parse java.lang.Object from classfile"


Trolldemorted:
**Describe the bug**
Printing `cfg.<tab>` in a linux kernel project makes joern unresponsive and prints the error message `exception caught when loading module class module-info$: java.lang.AssertionError: assertion failed: attempt to parse java.lang.Object from classfile`.

**To Reproduce**
1. Download Ubuntu24's kernel source (linux 6.8.0)
2. Start a joern container with 400GB memory, give 200 of that to joern's xmx
3. Have it parse the linux kernel source code, ignore the occasional `java.lang.OutOfMemoryError: java.lang.OutOfMemoryError: Java heap space`
4. Wait until the interactive shell is back, observe that joern still occupies 200GB memory and 1-10 CPU cores (usage going up and down)
5. Type `cfg.<tab>`  in the shell (probably not the brightest idea I had today)
6. Joern shell is completely unresponsive, joern decides that it now wants 10 cores all the time, let it run some more
7. Press `ctrl+c` a few times
8. See error ("exception caught when loading module class module-info$: java.lang.AssertionError: assertion failed: attempt to parse java.lang.Object from classfile")
9. Wait even more for the shell to be come responsive again

**Expected behavior**
Joern does not become unresponsive through the press of tab, and does not print assertion errors - is this a bad thing? It seems to be scala-specific, all google results I find are related to scala: https://github.com/scala/scala3/issues/20421

**Desktop (please complete the following information):**
 - OS `Ubuntu`
 - Joern/Java Version `ghcr.io/joernio/joern:nightly`

DavidBakerEffendi:
I've experienced this too, does not need that much to reproduce either, usually if one doesn't give an initial character and presses tab this will occur, e.g. `cpg.<TAB>` versus `cpg.c<TAB>`
@mpollmeier  any ideas?

mpollmeier:
Was able to minify it, I'm working on a fix/workaround, there's a few option, bear with me.

mpollmeier:
The root cause is is a scala compiler bug which get's triggered if there's a module-info.class in the classpath - which is a magic JPMS thing generated from the also magic module-info.java that doesn't behave like a regular classfile. 

[...]

PR #4672, opened 2024-06-17
reduce classpath dependencies from frontends: javasrc

mpollmeier:
Please review with care - this is the archetype setup for all other frontends. 

As discussed in e.g. https://github.com/joernio/joern/issues/4625#issuecomment-2166427270
we want to avoid having classpath dependencies on the frontends and instead invoke them
frontends as external processes (i.e. execute their start script). Otherwise we'll end in
jar hell with various incompatible versions of many different dependencies, and complex
issues with things like OSGI and JPMS.

There's no perfect way to do so, this is the one I currently like the most...

Fixes tab completion bug https://github.com/joernio/joern/issues/4625

### DIFF
--- a/console/build.sbt
+++ b/console/build.sbt
@@ -5,7 +5,6 @@ enablePlugins(JavaAppPackaging)
 dependsOn(
   Projects.semanticcpg,
   Projects.macros,
-  Projects.javasrc2cpg,
   Projects.jssrc2cpg,
   Projects.php2cpg,
   Projects.pysrc2cpg,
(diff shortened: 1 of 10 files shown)


DavidBakerEffendi:
Will take a look at this today, especially with testing against `standalone-ext` etc.

DavidBakerEffendi (approved):
I like this strategy too, and `javasrc2cpg` works nicely in `standalone-ext` when building this locally so I'm happy.

Issue #4664, opened 2024-06-13
[dataflowengineoss] Over-tainting inlined literals


xavierpinho:
# Setup

Tested on Joern 2.0.405

Compare Scenario 1 with Scenario 3 and Scenario 2 with Scenario 4 below. Even though these pairs of samples are equivalent, their results are different: in the failing samples, the literal `20` is tainting `print`, and the sole difference is that we inline the literal. 

If we hadn't specified FlowSemantics for `foo`, the results would coincide, but we can also note that they work as expected when the literal is not inlined, i.e. FlowSemantics are respected when the literal is not inlined.

## Scenario 1 ✅ 

```scala
"scenario 1" in {
  val cpg = code(
    """
      |from helpers import foo
      |a = 20
      |print(foo(a))
      |""".stripMargin)
    .withExtraFlows(
      List(FlowSemantic("helpers.py:<module>.foo", List(FlowMapping(0,0))))
  )
  def source = cpg.literal("20")
  def sink = cpg.call("print").argument(1)
  def flows = sink.reachableByFlows(source)
  flows shouldBe empty
}
```

## Scenario 2 ✅ 

```scala
"scenario 2" in {
  val cpg = code(
    """
      |def foo(x):
      |   return x
      |
      |a = 20
      |print(foo(a))
      |""".stripMargin)
    .withExtraFlows(List(
      FlowSemantic("Test0.py:<module>.foo", List(FlowMapping(0,0)))
    ))

  def source = cpg.literal("20")
  def sink = cpg.call("print").argument(1)
  def flows = sink.reachableByFlows(source)
  flows shouldBe empty
}
```

## Scenario 3 ❌ 

```scala
"scenario 3" in {
  val cpg = code(
    """
      |from helpers import foo
      |print(foo(20))
      |""".stripMargin)
    .withExtraFlows(
      List(FlowSemantic("helpers.py:<module>.foo", List(FlowMapping(0,0))))
    )

  def source = cpg.literal("20")
  def sink = cpg.call("print").argument(1)
  def flows = sink.reachableByFlows(source)
  flows shouldBe empty
}
```

## Scenario 4 ❌ 

```scala
"scenario 4" in {
  val cpg = code(
    """
      |def foo(x):
      |   return x
      |
      |print(foo(20))
      |""".stripMargin)
    .withExtraFlows(List(
      FlowSemantic("Test0.py:<module>.foo", List(FlowMapping(0,0)))
    ))

  def source = cpg.literal("20")
[...]

PR #4673, opened 2024-06-17
[dataflowengineoss] Refactor isCallRetVal

xavierpinho:
Apropos #4664, adds a couple of ignored failing tests with the intention of validating the intended behaviour. I was able to resolve a few of them (not included in this patch), but wanted to make sure this is right expectation before causing too much damage.


DavidBakerEffendi (approved):
I agree with the observation found in #4664 to be inconsistent. I recall a related issue with deeply nested arguments to calls not being detected, so perhaps there was some workaround that forced data-flow from the bottom to the top argument, which is why we're seeing this overtainting behaviour when inlined

PR #4680, opened 2024-06-19
[dataflowengineoss] refine return value semantics honouring

xavierpinho:
In expressions such as `foo(x)`, for which there are semantics for `foo` that don't taint its return value (cf. `isCallRetval`), we prevent creating an edge from `x` to `foo(x)`.

Otherwise, an expression such as `print(foo(x))` would be tainted by `x` when it shouldn't (since there are semantics preventing from tainting `foo`'s return value.)

### DIFF
--- a/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/dataflow/DataFlowTests.scala
+++ b/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/dataflow/DataFlowTests.scala
@@ -49,7 +49,7 @@ class DataFlowTests extends DataFlowCodeToCpgSuite {
 
     "find flows to `free`" in {
       val source = cpg.identifier
-      val sink   = cpg.call.name("free")
+      val sink   = cpg.call.name("free").argument(1)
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/dataflow/DataFlowTests.scala
```diff
@@ -49,7 +49,7 @@ class DataFlowTests extends DataFlowCodeToCpgSuite {
 
     "find flows to `free`" in {
       val source = cpg.identifier
-      val sink   = cpg.call.name("free")
+      val sink   = cpg.call.name("free").argument(1)
```
xavierpinho, line 52: @max-leuthaeuser I believe the intention here wasn't to check that the return value of `free` is tainted by `p`. The semantics for `free` currently specify that `1 -> 1`. What do you think?

DavidBakerEffendi (approved):
I agree with this, nice edge case that has been missed for a long time.

I've seen many get into the habit of setting a sink to `cpg.call.name("foo")` when they mean for the sink to be an argument of `foo`.

PR #4682, opened 2024-06-19
reduce classpath dependencies from frontends: jssrc2cpg

mpollmeier:
similar to https://github.com/joernio/joern/pull/4672, with one additional
refactor: reuse the scopt parser rather than manually fiddling with the
commandline args

### DIFF
--- a/console/build.sbt
+++ b/console/build.sbt
@@ -5,7 +5,6 @@ enablePlugins(JavaAppPackaging)
 dependsOn(
   Projects.semanticcpg,
   Projects.macros,
-  Projects.jssrc2cpg,
   Projects.php2cpg,
   Projects.pysrc2cpg,
   Projects.rubysrc2cpg,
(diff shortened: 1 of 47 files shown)


DavidBakerEffendi (commented):
Lots of API changes here, I suppose the "correct" thing would be to leave some `@Deprecated` notices here and there. In any case, I'll notify stakeholders on Discord

DavidBakerEffendi (approved):

mpollmeier:
I'd say 'couple' rather than 'lots' :)
I just added back a deprecated `XTypeRecovery.parserOptions`

DavidBakerEffendi:
> I'd say 'couple' rather than 'lots' :) I just added back a deprecated `XTypeRecovery.parserOptions`

Yes, I suppose I was thinking about the quantity of panicked DM's I might get from version upgrades breaking things versus the reality 😂

PR #4684, opened 2024-06-20
reduce classpath dependencies from frontends: remaining frontends

mpollmeier:
* similar to #4672 and #4682
* leaving out rubysrc2cpg for now, since a) it only adds one tiny 
  dependency and b) it's quite complex and also contains a second deprecated frontend


DavidBakerEffendi (approved):

DavidBakerEffendi:
Seems the test script for PHP isn't finding this file: https://github.com/joernio/joern/blob/master/joern-cli/frontends/php2cpg/src/main/resources/known_function_signatures.txt

Could possibly be because it isn't using the connection as a JAR connection perhaps

PR #4690, opened 2024-06-24
[c2cpg] Improve full names and signatures for C++.

ml86:
The C++ method full names did not include signatures which caused a lot
of methods to have the same full name which is not allowed. The
situation is not yet fully resolved because method/function declarations
independent from the implementation still created method stubs with full
name collisions. But it is a step in the right direction and the stubs
are easy to filter out via `isStub` step.

The call node generation has been reworked to accomodate this change.
Most noteable changes there:
- Full names where required contain signatures
- Signatures do not contain method names anymore
- Dispatch type is properly set if known.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
@@ -3,10 +3,29 @@ package io.joern.c2cpg.astcreation
 import io.shiftleft.codepropertygraph.generated.nodes.{NewCall, NewIdentifier, NewMethodRef}
 import io.shiftleft.codepropertygraph.generated.{DispatchTypes, Operators}
 import io.joern.x2cpg.{Ast, ValidationMode}
+import io.joern.x2cpg.Defines as X2CpgDefines
+import org.eclipse.cdt.core.dom.ast
 import org.eclipse.cdt.core.dom.ast.*
 import org.eclipse.cdt.core.dom.ast.cpp.*
 import org.eclipse.cdt.core.dom.ast.gnu.IGNUASTCompoundStatementExpression
-import org.eclipse.cdt.internal.core.dom.parser.cpp.CPPASTQualifiedName
+import org.eclipse.cdt.core.model.IMethod
+import org.eclipse.cdt.internal.core.dom.parser.c.{
+  CASTFieldReference,
+  CASTFunctionCallExpression,
+  CASTIdExpression,
+  CBasicType,
+  CFunctionType,
+  CPointerType
+}
+import org.eclipse.cdt.internal.core.dom.parser.cpp.semantics.{EvalBinding, EvalFunctionCall}
+import org.eclipse.cdt.internal.core.dom.parser.cpp.{
+  CPPASTIdExpression,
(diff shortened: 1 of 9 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -63,66 +82,265 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
     }
+  }
+
+  private def createCFunctionCallAst(
+    call: CASTFunctionCallExpression,
+    idExpr: CASTIdExpression,
+    callTypeFullName: String
+  ): Ast = {
+    val name      = idExpr.getName.getLastName.toString
+    val signature = ""
```
DavidBakerEffendi, line 304: What is the reason for blank string instead of `None`? Is it to simply be explicit in some way that the call signature is "known" to be empty, versus simply empty but might possibly be non-empty?

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -63,66 +82,265 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+
+    val callCpgNode = callNode(call, code(call), name, name, dispatchType, Some(signature), Some(callTypeFullName))
+    val args        = call.getArguments.toList.map(a => astForNode(a))
+
+    createCallAst(callCpgNode, args)
+  }
+
+  private def createPointerCallAst(call: IASTFunctionCallExpression, callTypeFullName: String): Ast = {
+    val functionNameExpr = call.getFunctionNameExpression
+    val name             = "<operator>.pointerCall"
```
DavidBakerEffendi, line 316: Probably useful to have as some constant somewhere

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/types/NamespaceTypeTests.scala
```diff
@@ -50,14 +50,25 @@ class NamespaceTypeTests extends C2CpgSuite(fileSuffix = FileDefaults.CPP_EXT) {
...
+          f.fullName shouldBe "Q.V.f:int()"
           m.name shouldBe "m"
-          m.fullName shouldBe "Q.V.C.m"
+          m.fullName shouldBe "Q.V.C.m:int()"
         }
       }
 
     }
 
+    "foo" in {
```
DavidBakerEffendi, line 61: Some debug testing to be removed

DavidBakerEffendi (commented):
I'm generally in favour of having these kinds of signatures for overloaded methods.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
ml86: Yes, the signature is conceptionally not `None` but the `""`. It is just that `None` ends up to be represented the same way.

DavidBakerEffendi (approved):

PR #4630, opened 2024-06-03
flatgraph

mpollmeier:
depends on https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1769

PR #4693, opened 2024-06-25
use scala primitives (e.g. Integer -> Int) and some cleanup

mpollmeier:
motivation: minify diff for flatgraph migration (https://github.com/joernio/joern/pull/4630)

depends on https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1771 and https://github.com/ShiftLeftSecurity/overflowdb-codegen/pull/271

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name                     := "joern"
 ThisBuild / organization := "io.joern"
 ThisBuild / scalaVersion := "3.4.2"
 
-val cpgVersion = "1.6.14+3-edeffe48"
+val cpgVersion = "1.6.15"
 
 lazy val joerncli          = Projects.joerncli
 lazy val querydb           = Projects.querydb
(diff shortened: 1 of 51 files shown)


build.sbt
```diff
@@ -2,7 +2,7 @@ name                     := "joern"
 ThisBuild / organization := "io.joern"
 ThisBuild / scalaVersion := "3.4.2"
 
-val cpgVersion = "1.6.13"
+val cpgVersion = "1.6.14+1-df813683"
```
DavidBakerEffendi, line 5: Seems this is causing some failures

mpollmeier: Yup this relies on the other PRs linked above, so this only compiles if you have those published locally. I was hoping that you can review it nonetheless.

DavidBakerEffendi: Right, duh lol.

DavidBakerEffendi (approved):
Looks like it **should** be okay. I see that the codegen will refer to Scala `Int` so this will minimize any room for issues

bbrehm (approved):

PR #4696, opened 2024-06-27
minify flatgraph migration: promote neighbor accessors (remove _ prefix)

mpollmeier:


mpollmeier:
discussed via slack, will leave as is since these are technical accessors which are different from the named steps..

PR #4697, opened 2024-06-27
ignore case while matching excluded files

Ferada:
Otherwise specifying one variant might not match the same file on e.g. Windows systems, when in fact it should match. This PR replaces the path comparisons with case-insensitive ones.

Test cases are happy, but I don't have a Windows one to test this. Ideally the comparisons should be purely on the paths, but instead there's an `isDirectory` call in there - it might be better (faster) to require callers to specify the ending slash `/` for directories instead and get rid of that case.

c.f. https://shiftleft.zendesk.com/agent/tickets/102700 and https://shiftleftsecurity.slack.com/archives/CE822HDC1/p1719386396027259

https://shiftleftinc.atlassian.net/browse/SEN-2828


max-leuthaeuser (approved):

max-leuthaeuser (changes requested):
We will have to do this differently.
Path insensitive comparison should only happen for Windows.

ml86 (approved):

max-leuthaeuser (approved):

max-leuthaeuser:
Side info: MacOS also has a case insensitive filesystem by default.

PR #4713, opened 2024-07-02
[x2cpg] Rethrow any Throwable from createCpg

max-leuthaeuser:


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/X2Cpg.scala
```diff
@@ -173,6 +173,7 @@ trait X2CpgFrontend[T <: X2CpgConfig[?]] {
   /** Create CPG according to given configuration, printing errors to the console if they occur. The CPG is closed and
     * not returned.
     */
+  @throws[RuntimeException]("if createCpg throws any Throwable")
```
Ferada, line 176: Well, either `RuntimeException`, or `Throwable`, but not both, right?

max-leuthaeuser: Yeah, right. Will change that.

mpollmeier (approved):

PR #4715, opened 2024-07-02
[speculative] minor work on reachingDef

bbrehm:
@mpollmeier can you test whether this solves / improves the memory issue?


mpollmeier:
Testing with the joern-flatgraph branch shows that this change reduces the minimum required Xmx for importing linux4 from 45g down to 30g (maybe even less) 🎉 🎉 🎉

PR #4719, opened 2024-07-03
[c2cpg] Safe .getOverload.getType access / safer AstCreator.createAst

max-leuthaeuser:
May crash with a NullPointerException.

Also: safer AstCreator.createAst so we do not crash c2cpg completely if createAst throws. (same as all other frontends now)

For: https://shiftleftinc.atlassian.net/browse/SEN-2777


mpollmeier (approved):
For my understanding: what part are you referring to with "May crash with a NullPointerException."?

max-leuthaeuser:
The .getOverload.getType call. Can be seen in the logs in https://shiftleftinc.atlassian.net/browse/SEN-2777

PR #4722, opened 2024-07-03
[c2cpg] Register call typefullnames correctly

max-leuthaeuser:
For: https://shiftleftinc.atlassian.net/browse/SEN-2840

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/TypeNodePass.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/TypeNodePass.scala
@@ -74,6 +74,6 @@ object TypeNodePass {
   }
 
   def fullToShortName(typeName: String): String = {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/TypeNodePass.scala
```diff
@@ -74,6 +74,11 @@ object TypeNodePass {
   }
 
   def fullToShortName(typeName: String): String = {
-    typeName.takeWhile(_ != ':').split('.').lastOption.getOrElse(typeName)
+    if (typeName.endsWith(">")) {
```
max-leuthaeuser, line 77: Example:
Without this change: `"foo.bar<package.someGeneric>" -> "someGeneric>"`
With this change: `"foo.bar<package.someGeneric>" -> "bar"`

DavidBakerEffendi (approved):

PR #4724, opened 2024-07-03
[gosrc2cpg] Multi module support

pandurangpatil:
Earlier, if we pass the directory path which contains multiple go modules. Processing was done with all the `.go` files mapped to the single `go.mod` file.

With this change, we have segregated the processing by first isolating all the files mapped to respective `go.mod`. This will make sure to clean up the memory footprint after every module is being processed. However, this will increase the processing time when used with a download dependency as it will process all the `go.mod` files by identifying and
processing used dependencies.

tests covering multiple module scenarios


joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/utils/AstGenRunner.scala
```diff
@@ -91,11 +96,128 @@ class AstGenRunner(config: Config, includeFileRegex: String = "") extends AstGen
...
+    outputModulePath: String,
+    modFilePath: Option[String],
+    parsedFiles: ListBuffer[String],
+    skippedFiles: ListBuffer[String],
+    childModules: ListBuffer[ModuleMeta]
+  ) {
+    def addModFile(modFile: String, inputPath: String, outPath: String): Unit = {
+      import scala.util.control.Breaks.*
+      var processed = false
+      breakable {
```
DavidBakerEffendi, line 150: Rather use `collectFirst` instead of introducing the break. Collect first can also then return an `Option` which can be used instead of `processed`

DavidBakerEffendi (changes requested):

DavidBakerEffendi (approved):

PR #4726, opened 2024-07-03
[ruby] Make `<body>` Call Static Dispatch

DavidBakerEffendi:
As the `<body>` call is synthetic and meant to be immediately deterministic, so there is no reason it should be re-determined.


ml86 (commented):
I would have expected some change indicating that there is no receiver specified anymore for the call to the body method. Am I missing this?

DavidBakerEffendi:
> I would have expected some change indicating that there is no receiver specified anymore for the call to the body method. Am I missing this?

So since it's static, we would get rid of the receiver, is that what you mean? For reference, this is the call:

graphviz

ml86:
Static calls do not have a receiver edge. They just have arguments and the call is dispatches via `methodFullName` property only.

ml86 (approved):

PR #4731, opened 2024-07-04
[c2cpg] Added typefullnames to all calls

max-leuthaeuser:
For: https://shiftleftinc.atlassian.net/browse/SEN-2777

`Any` there does not make much of a difference but it stops the DynamicCallLinker from crashing.


DavidBakerEffendi (approved):
@max-leuthaeuser An alternative is to make this the default value in `x2cpg`'s node builder for `callNode`

max-leuthaeuser:
@DavidBakerEffendi Yes definitely. Maybe we should do this additionally so nothing slips through.

ml86 (approved):

PR #4734, opened 2024-07-04
[ruby] Singleton methods on objects

### LINKED ISSUE #4721: [ruby] Singleton Methods on Objects
We have an existing test that has the following code
```ruby
class Animal; end
animal = Animal.new

class << animal
  def bark
    'Woof'
  end
end

animal.bark # => 'Woof'
```
Right now, this created a temporary class, and extends whatever type `animal` is and assigns that temp class the new method and gives `animal` the type hint to that new class.

The "correct" way would to rather assign a new property to `animal`, e.g.

```ruby
class Animal; end
animal = Animal.new

animal.bark = def bark; "Woof"; end # not valid ruby necessarily

animal.bark # => 'Woof'
```
The behaviour we instead want to model is assigning a property with the name of the method to a method ref to this method. We can keep the method as a lambda to avoid method full name clashes.

AndreiDreyer:
This PR changes the modelling of Singleton methods on objects. Ex:
```ruby
class Animal; end
animal = Animal.new

class << animal
  def bark
    "woof"
  end
end
```
Creates a lambda method for `animal.bark`, and assigns `animal.bark = methodRef(bark)`

Resolves #4721

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
@@ -866,7 +866,7 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
               case x => x
             }
 
-            StatementList(stmts)(ctx.toTextSpan)
+            SingletonStatementList(stmts)(ctx.toTextSpan)
         }
       case None =>
         SingletonClassDeclaration(freshClassName(ctx.toTextSpan), baseClass, body)(ctx.toTextSpan)
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
@@ -844,11 +844,26 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
   }
 
   override def visitSingletonClassDefinition(ctx: RubyParser.SingletonClassDefinitionContext): RubyNode = {
-    SingletonClassDeclaration(
-      freshClassName(ctx.toTextSpan),
-      Option(ctx.commandOrPrimaryValueClass()).map(visit),
-      visit(ctx.bodyStatement())
-    )(ctx.toTextSpan)
+    val baseClass = Option(ctx.commandOrPrimaryValueClass()).map(visit)
+    val body      = visit(ctx.bodyStatement()).asInstanceOf[StatementList]
+
+    baseClass match {
+      case Some(baseClass) =>
+        val stmts = body.statements.map {
+          case x: MethodDeclaration =>
+            val memberAccess =
+              MemberAccess(baseClass, ".", x.methodName)(x.span.spanStart(s"$baseClass.${x.methodName}"))
+            SingleAssignment(memberAccess, "=", x)(
+              ctx.toTextSpan.spanStart(s"${memberAccess.span.text} = ${x.span.text}")
+            )
+          case x => x
+        }
+
+        StatementList(stmts)(ctx.toTextSpan)
+      case None =>
+        logger.warn("No base class, creating normal SingletonClassDecl")
+        SingletonClassDeclaration(freshClassName(ctx.toTextSpan), baseClass, body)(ctx.toTextSpan)
+    }
   }
 
   private def findFieldsInMethodDecls(methodDecls: List[MethodDeclaration]): List[RubyNode & RubyFieldIdentifier] = {
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
@@ -849,20 +849,26 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
 
     baseClass match {
       case Some(baseClass) =>
-        val stmts = body.statements.map {
-          case x: MethodDeclaration =>
-            val memberAccess =
-              MemberAccess(baseClass, ".", x.methodName)(x.span.spanStart(s"${baseClass.span.text}.${x.methodName}"))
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
```diff
@@ -844,11 +844,26 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
...
+    val baseClass = Option(ctx.commandOrPrimaryValueClass()).map(visit)
+    val body      = visit(ctx.bodyStatement()).asInstanceOf[StatementList]
+
+    baseClass match {
+      case Some(baseClass) =>
+        val stmts = body.statements.map {
+          case x: MethodDeclaration =>
+            val memberAccess =
+              MemberAccess(baseClass, ".", x.methodName)(x.span.spanStart(s"$baseClass.${x.methodName}"))
+            SingleAssignment(memberAccess, "=", x)(
```
DavidBakerEffendi, line 856: If you do this, then you can mostly leave `astcreation` alone
```scala
val proc = ProcOrLambdaExpr(Block(x.parameters, x.body)(x.span))(x.span)
SingleAssignment(memberAccess, "=", proc)(
...
```

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ClassTests.scala
```diff
@@ -354,44 +354,64 @@ class ClassTests extends RubyCode2CpgFixture {
...
-          anonClass.name shouldBe "<anon-class-0>"
-          anonClass.fullName shouldBe "Test0.rb:<global>::program.<anon-class-0>"
-          // TODO: Attempt to resolve the below with the `scope` class once we're handling constructors
-          anonClass.inheritsFromTypeFullName shouldBe Seq("animal")
-          inside(anonClass.method.l) {
-            case defaultConstructor :: bark :: Nil =>
-              defaultConstructor.name shouldBe Defines.ConstructorMethodName
-              defaultConstructor.fullName shouldBe s"Test0.rb:<global>::program.<anon-class-0>:${Defines.ConstructorMethodName}"
+    "Create assignments to method refs for methods on singleton object" in {
+      inside(cpg.method.name(":program").astChildren.isBlock.astChildren.isBlock.l) {
```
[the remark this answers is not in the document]
DavidBakerEffendi: Also, I would expect the assignment to be directly under the block, not in a nested block, so we may need to make some special case handling somewhere, e.g., I would expect `cpg.method.isModule.assignment`

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ClassTests.scala
```diff
@@ -354,44 +354,64 @@ class ClassTests extends RubyCode2CpgFixture {
 
   }
 
-  // TODO: This should be remodelled as a property access `animal.bark = METHOD_REF`
-  "a basic singleton class" ignore {
+  "a basic singleton class" should {
```
DavidBakerEffendi, line 357:
```suggestion
  "a basic singleton class extending an object instance" should {
```

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ClassTests.scala
```diff
@@ -354,44 +354,64 @@ class ClassTests extends RubyCode2CpgFixture {
...
-          // TODO: Attempt to resolve the below with the `scope` class once we're handling constructors
-          anonClass.inheritsFromTypeFullName shouldBe Seq("animal")
-          inside(anonClass.method.l) {
-            case defaultConstructor :: bark :: Nil =>
-              defaultConstructor.name shouldBe Defines.ConstructorMethodName
-              defaultConstructor.fullName shouldBe s"Test0.rb:<global>::program.<anon-class-0>:${Defines.ConstructorMethodName}"
+    "Create assignments to method refs for methods on singleton object" in {
+      inside(cpg.method.name(":program").astChildren.isBlock.astChildren.isBlock.l) {
+        case singletonMethodsBlock :: Nil =>
+          inside(singletonMethodsBlock.astChildren.isCall.name(Operators.assignment).l) {
```
DavidBakerEffendi, line 377: As a tip, there is the `assignment` step
```suggestion
          inside(singletonMethodsBlock.astChildren.isCall.assignment.l) {
```

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ClassTests.scala
```diff
@@ -354,44 +354,64 @@ class ClassTests extends RubyCode2CpgFixture {
...
+            case barkAssignment :: legsAssignment :: Nil =>
+              inside(barkAssignment.argument.l) {
+                case (lhs: Call) :: (rhs: MethodRef) :: Nil =>
+                  val List(identifier, fieldIdentifier) = lhs.argument.l
```
DavidBakerEffendi, line 381: I speak under correction, but this might create a compilation warning, which can be suppressed as follows
```suggestion
                  val List(identifier, fieldIdentifier) = lhs.argument.l: @unchecked
```

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ClassTests.scala
```diff
@@ -354,44 +354,64 @@ class ClassTests extends RubyCode2CpgFixture {
...
-        case xs => fail(s"Expected a single anonymous class, but got [${xs.map(x => x.label -> x.code).mkString(",")}]")
+        case xs => fail(s"Expected one block for main program, got [${xs.code.mkString(",")}]")
       }
     }
 
-    "register that `animal` may possibly be an instantiation of the singleton type" in {
-      cpg.local("animal").possibleTypes.l should contain("Test0.rb:<global>::program.<anon-class-0>")
-    }
+    "Create lambda methods for methods on singleton object" in {
+      inside(cpg.method.name("<lambda>\\d+").l) {
```
DavidBakerEffendi, line 405:
```suggestion
      inside(cpg.method.isLambda.l) {
```

DavidBakerEffendi (changes requested):

DavidBakerEffendi (approved):

PR #4749, opened 2024-07-09
pull out changes from michael/flatgraph to minify diff

mpollmeier:


ml86 (approved):

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -230,7 +230,7 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this:
     registerType(typeFullName)
 
     val astParentType     = methodAstParentStack.head.label
-    val astParentFullName = methodAstParentStack.head.properties("FULL_NAME").toString
+    val astParentFullName = methodAstParentStack.head.properties(PropertyNames.FULL_NAME).toString
```
bbrehm, line 233: This is unfortunately quite bad: It assembles all properties into a dictionary before extracting the one it needs.

This is not a new problem, but it does suck :(

mpollmeier: we could define `NewNode.property(name)` in the codegen. after the flatgraph migration. sounds good?

PR #4754, opened 2024-07-10
[ruby] Singleton Methods on Objects Follow-Up

### LINKED ISSUE #4741: [ruby] Singleton Methods on Objects Follow-Up
Following from https://github.com/joernio/joern/pull/4734, the more natural modelling would be to keep these methods as method declarations instead of lambdas. This, however, comes at the catch of making the method full names unique, e.g.

```ruby
class Animal; end
animal = Animal.new

class << animal
  def bark
    "woof"
  end
end

def bark
  "meow"
end
```

For this purpose, the recommended method full name for the singleton method should look something like `:program.class<<animal.bark`. This conveys the kind of structure the method was declared in, and we no longer need to treat it as a lambda.

AndreiDreyer:
This PR handles
 * Modelling for singleton methods on objects changed to method declarations instead of lambdas

Resolves #4741


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -784,6 +785,34 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+        )
+    }
+
+    // Set span contents
+    methodAstsWithRefs.flatMap(_.nodes).foreach {
+      case m: NewMethodRef => DummyNode(m.copy)(node.block.span.spanStart(m.code))
+      case _               =>
+    }
+
+    val Seq(_, methodRef) = methodAstsWithRefs
```
DavidBakerEffendi, line 811: From this PR https://github.com/joernio/joern/pull/4753, you'll want to emit the `typeRef` instead
```suggestion
    val Seq(typeRef, _) = methodAstsWithRefs
```

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ClassTests.scala
```diff
@@ -371,6 +371,8 @@ class ClassTests extends RubyCode2CpgFixture {
         |animal.bark # => 'Woof'
         |""".stripMargin)
 
+    cpg.method.isModule.dotAst.l.foreach(println)
```
DavidBakerEffendi, line 374: Debugging code

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -36,12 +36,21 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
...
+    isClosure: Boolean = false,
+    singletonMethodFullName: Option[String] = None
+  ): Seq[Ast] = {
     val isInTypeDecl  = scope.surroundingAstLabel.contains(NodeTypes.TYPE_DECL)
     val isConstructor = (node.methodName == Defines.Initialize) && isInTypeDecl
     val methodName    = node.methodName
     // TODO: body could be a try
-    val fullName = computeMethodFullName(methodName)
+
+    val fullName = singletonMethodFullName match {
```
DavidBakerEffendi, line 49: Could we not instead have `isSingletonMethod: Boolean = false` and if true, use `computeSingletonObjectMethodFullName` instead of `computeMethodFullName`? Would keep things consistent with the `isClosure` parameter

DavidBakerEffendi (changes requested):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
AndreiDreyer: I tried it this way originally, but it does complicate things since the base class that the singleton method is being defined on isn't defined on the `MethodDecl` that is passed into `astForMethodDeclaration`. Only wya I can think of off the top of my head to make this possible is to add a new type that MethodDecl and SingletonObjectMethodDecl both extend and use that for the `astForMethodDeclaration`

DavidBakerEffendi: I believe I've added a `ProcedureDeclaration` trait which might be usable here, or you can extend `MethodDeclaration`, either way this would be preferable to be handled in a similar way to the closures/lambas

DavidBakerEffendi (approved):

PR #4756, opened 2024-07-11
[c2cpg] Fix string repr for type of dependent expressions

max-leuthaeuser:
We are not able to represent them with the CPG format.
See e.g.:
```
template <class charT, class traits>
inline re_syntax_base* basic_regex_creator<charT, traits>::append_set(
   const basic_char_set<charT, traits>& char_set)
{
   typedef mpl::bool_< (sizeof(charT) == 1) > truth_type;
   return char_set.has_digraphs() 
      ? append_set(char_set, static_cast<mpl::false_*>(0))
      : append_set(char_set, static_cast<truth_type*>(0));
}
```

The type of `append_set` is dependent to some boost library magic w.r.t. to the method template instantiation. I just set it to ANY here.


maltek:
you already got a minimal code example, can you make a test for that?

max-leuthaeuser:
I am not sure if it works without access to the boost library. But I will try it.

max-leuthaeuser:
As expected. With boost CDT parses that without any of these problematic depentend expression types.

maltek (approved):

PR #4763, opened 2024-07-11
[ruby] Remodel `yield` Calls

### LINKED ISSUE #4760: [ruby] Re-Model `yield` Calls
`yield` calls can be simply lowered from
```ruby
def foo()
  yield 1
end
```
to 
```ruby
def foo(&block)
  block.call(1)
end
```
Since we have implicit block parameters handled when a `yield` expression is present, we only need to replace it with the block invocation.

DavidBakerEffendi:
This PR remodels `yield` calls as explicit invocations of the implicit or explicit block parameter of the surrounding method.

Resolves #4760

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ProcParameterAndYieldTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ProcParameterAndYieldTests.scala
@@ -1,84 +1,96 @@
 package io.joern.rubysrc2cpg.querying
 
 import io.joern.rubysrc2cpg.testfixtures.RubyCode2CpgFixture
-import io.shiftleft.codepropertygraph.generated.Operators
-import org.scalatest.Inspectors
-import io.shiftleft.semanticcpg.language.*
 import io.shiftleft.codepropertygraph.generated.nodes.*
+import io.shiftleft.semanticcpg.language.*
+import org.scalatest.Inspectors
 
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ProcParameterAndYieldTests.scala
```diff
@@ -1,84 +1,96 @@
...
+    fooParam.code shouldBe "&<proc-param-0>"
+    fooParam.index shouldBe 1
+
+    val barParam = bar.parameter.last
+    barParam.name shouldBe "<proc-param-1>"
+    barParam.code shouldBe "&<proc-param-1>"
+    barParam.index shouldBe 1
+
+    foo.call.nameExact("call").argument.isIdentifier.name.l shouldBe List("<proc-param-0>")
+    bar.call.nameExact("call").argument.isIdentifier.name.l shouldBe List("<proc-param-1>")
```
ml86, line 60: the block/proc parameter is not passed as an argument into the block/proc itself. It is just the receiver. Also applies to your other tests.

ml86 (changes requested):

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ProcParameterAndYieldTests.scala
ml86: I was wrong about this. Lets keep it like it is.

ml86 (approved):

Issue #4762, opened 2024-07-11
[php2cpg] better representation of array unpacking


d1tto:
The current PHP frontend is not very friendly to handle array unpacking, for example `[$a, $b, $c] = $arr`, php2cpg will create an array of `$a, $b, $c` as the element, and put the array as the target of an assignment statement, which will cause the data flow analysis to fail. Similarly, there is also exists in the doubleArrow operator used in foreach statement, and the list operator

**Describe the solution you'd like**
pysrc2cpg handles this very well, which converts `[a, b, c] = arr` to `a = arr[0], b = arr[1], c = arr[2]`

PR #4767, opened 2024-07-13
[php2cpg] Lowering the init part of foreach statement

d1tto:
https://github.com/joernio/joern/issues/4762
In the previous implementation, the `doubleArrow` operator is used to wrap key-value-pari when foreach traverses both keys and values, which causes the data flow analysis to fail. In addition, `current` method can only get the element and not the key of it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name                     := "joern"
 ThisBuild / organization := "io.joern"
 ThisBuild / scalaVersion := "3.4.2"
 
-val cpgVersion = "1.6.18"
+val cpgVersion = "1.6.19"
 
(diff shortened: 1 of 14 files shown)


d1tto:
@DavidBakerEffendi @johannescoetzee Excuse me, could you help me review this pr. sorry, I didn't know that yesterday's commit would be appended here, just look at the last commit. Thanks again.

d1tto:
@DavidBakerEffendi @johannescoetzee Excuse me, do you think this change is reasonable

DavidBakerEffendi:
@d1tto Will look eventually, have other work on my plate right now

PR #4772, opened 2024-07-15
IF Cfg creation for if-statements with empty `then` block.

ml86:


joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/cfg/CfgCreationPassTests.scala
```diff
@@ -504,6 +504,26 @@ class CfgCreationPassTests extends CfgTestFixture(() => new CCfgTestCpg) {
       succOf("d") shouldBe expected(("RET", AlwaysEdge))
       succOf("e") shouldBe expected(("RET", AlwaysEdge))
     }
+
+    "be correct for empty 'then' block" in {
+      implicit val cpg: Cpg = code("if (cond()) {} else { foo() }")
```
maltek, line 509:
```suggestion
      implicit val cpg: Cpg = code("if (cond()) {} else { foo(); }")
```

maltek (approved):

Issue #4732, opened 2024-07-04
[Bug]Fail to get CPG for Ruby


spingARbor:
**Describe the bug**
I want to construct CPG for part of code for giltab.
When I try to parse the code, I get the following exception.
![joern-bug](https://github.com/joernio/joern/assets/57784180/2d250607-1077-4d05-98e8-350bae433532)
When I try to dump Graph, I get the following exception and no output was produced.
![joern-bug-1](https://github.com/joernio/joern/assets/57784180/1de21fc7-e3cf-49ed-acde-207c80d4f11f)

**To Reproduce**
1. Prepare a 256G memory Ubuntu server
2. Install the latest release of Joern
3. Download the source code of Gitlab at https://gitlab.com/gitlab-org/gitlab/-/archive/v16.10.4-ee/gitlab-v16.10.4-ee.tar.gz
4. Extract the file to /home/ubuntu/gitlab-v16.10.4-ee  
5. Parse the code of app folder
```
joern-parse /home/ubuntu/joern/gitlab-v16.10.4-ee/app
```
and you can get the exception
```
overflowdb.SchemaViolationException: IN edge with label AST to an adjacent TYPE_DECL is mandatory, but not defined for this MEMBER node with id=43243
```
6. Dump Graph
```
joern-export --repr cpg --format=dot
```
and you can get the exception
```
ERROR CpgPassBase: Pass io.joern.dataflowengineoss.passes.reachingdef.ReachingDefPass failed
overflowdb.SchemaViolationException: OUT edge with label REF to an adjacent METHOD is mandatory, but not defined for this METHOD_REF node with id=14929
```

**Expected behavior**
Get a CPG for Gitlab code.

**Desktop (please complete the following information):**
 - OS:ubuntu 24.04 server
 - Joern Version :2.0.433
 - Java version :21.0.3

DavidBakerEffendi:
@spingARbor Have you set `Xmx` properties somewhere so that Joern is making use of the available memory?

DavidBakerEffendi:
One way to do it is to add `-J-Xmx20G` or something to the Joern arguments. This schema violation thing can happen at low memory

spingARbor:
Thank you for your response!@DavidBakerEffendi 
I have set -Xmx240G in my server.And when I try to create CPG for a folder with a few ruby files, I also got the exception and without cpg dot file output.On the other hand,The running information will tell me that there is not enough memory if that happens as a result of low memory.
Sincerely，
Arbor

[...]

PR #4776, opened 2024-07-16
[ruby] Calls with reserved keywords

### LINKED ISSUE #4748: [ruby] Member Calls with EMARK
Ruby methods can end with `[?=!]`, but the parser doesn't appear to expect this on calls as members, e.g.
```ruby
batch.retry!
```
This may be because members have different expected identifiers, but the solution may require a bit of investigation.

AndreiDreyer:
An issue was picked up where having a reserved keyword as the call on a member (e.g. `batch.retry!`) was causing a syntax error due to lexer and parser issues.

This PR handles:
 * Changed precedence of keywords in the ANTLR Lexer to come after `LOCAL_VARIABLE_IDENTIFIER`
 * Added handling to `LOCAL_VARIABLE_IDENTIFIER` tokens to determine whether a token should be marked as reserved keyword or not. We check if the previous `nonWs` token is `.`, `:` or `::`, or if the next token is `.` or `:`. If this holds, we know it cannot be a reserved keyword and is instead marked as a `LOCAL_VARIABLE_IDENTIFIER`.
 * Added debug flag to ANTLR parser to print out the parse tree. Hidden and off by default.

Resolves #4748

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/antlr4/io/joern/rubysrc2cpg/parser/RubyLexer.g4
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/antlr4/io/joern/rubysrc2cpg/parser/RubyLexer.g4
@@ -32,6 +32,7 @@ options {
     superClass = RubyLexerBase;
 }
 
+
 // --------------------------------------------------------
 // Keywords
 // --------------------------------------------------------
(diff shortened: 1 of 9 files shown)


DavidBakerEffendi (approved):
Very nice!

DavidBakerEffendi:
cc @spingARbor: This bug affected a lot of GitLab files (https://github.com/joernio/joern/issues/4732), so once it's merged there will be a lot more content.

PR #4777, opened 2024-07-16
[ruby] Approximate Attribute Assignment

DavidBakerEffendi:
With Ruby attribute assignments being calls, this can lead to costly data-flow tracking and these setters are rarely overridden with custom logic. This PR simplifies this model by representing the attribute assignment from a setter call to a direct field assignment.

Misc: Simplified a parser bug workaround and linked it to an issue.


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/SetterTests.scala
```diff
@@ -1,29 +1,33 @@
...
+        fieldTarget.name shouldBe Operators.fieldAccess
+
+        inside(fieldTarget.argument.l) {
+          case (base: Identifier) :: (field: FieldIdentifier) :: Nil =>
+            base.name shouldBe "x"
+            field.canonicalName shouldBe "@y"
+          case xs => fail("Expected field access to have two targets")
+        }
+      case xs => fail("Expected a single assignment to the literal `1`")
+    }
```
ml86, line 30: Please also test if the call is static and whether its full name is set to operator assignment.

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/SetterTests.scala
```diff
@@ -1,29 +1,33 @@
...
 
 import io.joern.rubysrc2cpg.testfixtures.RubyCode2CpgFixture
+import io.shiftleft.codepropertygraph.generated.Operators
+import io.shiftleft.codepropertygraph.generated.nodes.{Call, FieldIdentifier, Identifier}
 import io.shiftleft.semanticcpg.language.*
 
 class SetterTests extends RubyCode2CpgFixture {
 
-  "`x.y=1` is represented by a `x.y=` CALL with argument `1`" in {
+  "`x.y=1` is approximated by a `x.y =` assigment with argument `1`" in {
```
ml86, line 10: Why is there only a test for the setter? This also needs to apply to the getter side.

DavidBakerEffendi: Would this still be a call or a field getter?

DavidBakerEffendi: I suppose let's keep it a call, otherwise we have to deal with parenthesis-less call ambiguity

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/AttributeAccessorTests.scala
```diff
@@ -0,0 +1,50 @@
...
+        val fieldTarget = xyAssign.target.asInstanceOf[Call]
+        fieldTarget.code shouldBe "x.@y"
+        fieldTarget.name shouldBe Operators.fieldAccess
+        fieldTarget.methodFullName shouldBe Operators.fieldAccess
+        fieldTarget.dispatchType shouldBe DispatchTypes.STATIC_DISPATCH
+
+        inside(fieldTarget.argument.l) {
+          case (base: Identifier) :: (field: FieldIdentifier) :: Nil =>
+            base.name shouldBe "x"
+            field.canonicalName shouldBe "@y"
```
ml86, line 29: Please add a check for `field.name shouldBe "y"`.

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/AttributeAccessorTests.scala
```diff
@@ -0,0 +1,50 @@
...
+    val cpg = code("""x = Foo.new
+                     |x.y = 1
+                     |""".stripMargin)
+    inside(cpg.assignment.where(_.source.isLiteral.codeExact("1")).l) {
+      case xyAssign :: Nil =>
+        xyAssign.lineNumber shouldBe Some(2)
+        xyAssign.code shouldBe "x.y = 1"
+
+        val fieldTarget = xyAssign.target.asInstanceOf[Call]
+        fieldTarget.code shouldBe "x.@y"
```
ml86, line 21: Code should be `x.y`

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/AttributeAccessorTests.scala
```diff
@@ -0,0 +1,50 @@
...
+            field.canonicalName shouldBe "@y"
+          case xs => fail("Expected field access to have two targets")
+        }
+      case xs => fail("Expected a single assignment to the literal `1`")
+    }
+  }
+
+  "`x.y` is represented by a CALL `x.y`" in {
+    val cpg = code("""x = Foo.new
+        |x.y
```
ml86, line 38: What happens for `a.b = x.y` ?
In that case the `x.y` needs to be a field access. Can you please write a test for that?

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/AttributeAccessorTests.scala
DavidBakerEffendi: There is no such field, but I will make this the case for `.code`

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/AttributeAccessorTests.scala
DavidBakerEffendi: My concern is that this is not true in all cases. e.g. 
```ruby
class Foo {
  def a
     puts "hello"
  end
}
x = Foo.new
y = SomeBar.new
y.b = x.a
```
In this case, `x.a` is in-fact a call, and not referring to the member access

ml86: Thats a valid concern but without this we wont have solved the issue of requiring to track back to the creation of x in order to learn about the function a.
Lets for now represent `x.a` as a field access if there are no parenthesis.
Its likely that we will change that later, but it is the better start.

ml86 (approved):

PR #4778, opened 2024-07-16
[c2cpg] Fix CFG creation for INLINE calls (LOCALS with multiple AST parents)

max-leuthaeuser:
No more LOCALs with multiple AST parents.


max-leuthaeuser:
Do not merge!
For https://github.com/ossrs/srs/ we would generate LOCALs without any AST parent. Needs more investigation.

ml86 (approved):

PR #4780, opened 2024-07-16
[php2cpg] array creation point marks and improved array unpacking

d1tto:
1. In the previous implementation, the local variable `$tmp` was accessed directly by `indexAccess` operator when handling the `PhpArrayExpr`. For example, `[1, 2]` will be lowered to `$tmp[0] = 1; $tmp[1] = 2; $tmp;`. However, this semantics was not complete. The correct flow would have been to assign an empty array first, i.e. `$tmp = []; $tmp[0] = 1; $tmp[1] = 2;`
2. array unpacking supports key in list, such as `["name" => $name] = $arr`


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
```diff
@@ -1408,19 +1414,30 @@ class AstCreator(filename: String, phpAst: PhpFile, fileContent: Option[String],
...
     val itemAssignments = expr.items.flatMap {
-      case Some(item) => Option(assignForArrayItem(item, tmpIdentifier.name, idxTracker))
+      case Some(item) => Option(assignForArrayItem(item, tmpName, idxTracker))
       case None =>
         idxTracker.next // Skip an index
         None
     }
     val arrayBlock = blockNode(expr)
 
     Ast(arrayBlock)
```
DavidBakerEffendi, line 1437: Every identifier needs a LOCAL variable to track that they refer to the same point in memory. So you would want to create a local variable for the tmp identifier and link REF edges from the identifier to the local. e.g.
```scala
Ast(/* the block */)
 .withChild(Ast(localNode))
 .withRefEdge(localNode, identifierNode)
```

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
```diff
@@ -1408,19 +1414,30 @@ class AstCreator(filename: String, phpAst: PhpFile, fileContent: Option[String],
   private def astForArrayExpr(expr: PhpArrayExpr): Ast = {
     val idxTracker = new ArrayIndexTracker
 
-    val tmpIdentifier = getTmpIdentifier(expr, Some(TypeConstants.Array))
+    val tmpName = getNewTmpName()
+
+    def newTmpIdentifier: Ast = Ast(identifierNode(expr, tmpName, s"$$$tmpName", TypeConstants.Array))
+
+    val tmpIdentifierAssignNode = {
+      val emptyArrayCall = callAst(newOperatorCallNode(PhpOperators.emptyArray, s"[]", line = line(expr)))
```
DavidBakerEffendi, line 1422: Instead of `emptyArray` could we refer to the constructor of the array class?

DavidBakerEffendi (changes requested):

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
maltek: php2cpg uses a separate step to create and link locals (`LocalCreationPass`) - so they shouldn't be added during AST creation.

PR #4786, opened 2024-07-19
Added `joern-slice` Data-Flow Script Test

### LINKED ISSUE #4783: bring back joern-slice test
Hey @DavidBakerEffendi, your name appears around the joern-slice code quite a bit... can you help me get the test back to work?

I commented it out as part of the flatgraph migration, but I'm pretty sure it was broken before:
https://github.com/joernio/joern/blob/c1919cfe5611665e433a4aff93989411f735d0b8/.github/workflows/pr.yml#L99-L103

On PR runs we create a tiny test cpg which most likely doesn't have any dataflows, and then we try and slice it? That doesn't make much sense, does it? The test doesn't assert anything either, apart from the exit code of joern-slice...

What's your thoughts on having a proper integration test for this, do we need some testcode that contains actual dataflows that can be sliced?

DavidBakerEffendi:
* Added a script that parses and creates a sensible string from a slice that can be tested against.
* Fixed a slicing bug where slicing direction was in the opposite direction of neighbour retrieval and ended up ignoring certain nodes.
* Added entry in the `test-scripts` job to run the slice script, parse the slice, and assert the expected flow with `grep`.

Resolves #4783

@mpollmeier, is this the direction we want to go? Is this test satisfactory to assert some expected behaviour with?

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -101,8 +101,10 @@ jobs:
           ./joern-scan /tmp/foo
           ./joern-scan --dump
       - run: |
-          mkdir /tmp/slice
+          mkdir /tmp/slice && ls -al
+          echo "Slicing!"
           ./joern-slice data-flow tests/code/javasrc/SliceTest.java -o /tmp/slice/dataflow-slice-javasrc.json
+          echo "Validating!"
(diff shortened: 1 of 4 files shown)


mpollmeier (approved):
nice one, thank you!

PR #4787, opened 2024-07-19
[c2cpg] Defer method decl handling

max-leuthaeuser:
For method de-duplication (CDT always copy methods from header files into the implementation files so we would end up generating CPG METHODs for both with the same fullName), we now skip over all method decls but remember their properties.
Afterwards, a new pass stubs them iff no actual method definition exists.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/CGlobal.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/CGlobal.scala
@@ -5,7 +5,7 @@ import java.util.concurrent.ConcurrentHashMap
 
 object CGlobal {
 
-  case class MethodInfo(
+  final case class MethodInfo(
     name: String,
     code: String,
     fileName: String,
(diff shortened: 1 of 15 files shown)


max-leuthaeuser:
FYI @ml86 You may want to have a look already as it is working fine.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/CGlobal.scala
```diff
@@ -0,0 +1,43 @@
...
+    lineNumber: Option[Int],
+    columnNumber: Option[Int],
+    lineNumberEnd: Option[Int],
+    columnNumberEnd: Option[Int],
+    signature: String,
+    offset: Option[(Int, Int)],
+    parameter: Seq[ParameterInfo],
+    modifier: Seq[String]
+  )
+  final class ParameterInfo(
```
ml86, line 24: Is there a reason why MethodInfo is a case class and ParameterInfo is not?

ml86 (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/CGlobal.scala
max-leuthaeuser: ParameterInfo has two mutable fields (to set var arg info). Scala case classes should never be mutable as it breaks some contracts when accessing them in a parallel environment (says the Scala doc).

ml86: Missed the mutability part. Makes sense.

PR #4791, opened 2024-07-22
workaround for scala completion bug

mpollmeier:
on stage: remove module-info.class from dependency jars - a hacky workaround for a
scala3 compiler bug: https://github.com/scala/scala3/issues/20421

Fixes https://github.com/joernio/joern/issues/4625

### DIFF
--- a/joern-cli/build.sbt
+++ b/joern-cli/build.sbt
@@ -131,4 +131,26 @@ generateScaladocs := {
 
 Universal / packageBin / mappings ++= sbt.Path.directory(new File("joern-cli/src/main/resources/scripts"))
 
+lazy val removeModuleInfoFromJars = taskKey[Unit]("remove module-info.class from dependency jars - a hacky workaround for a scala3 compiler bug https://github.com/scala/scala3/issues/20421")
+removeModuleInfoFromJars := {
+  import java.nio.file.{Files, FileSystems}
+  val logger = streams.value.log
+  val libDir = (Universal/stagingDirectory).value / "lib"
+
+  // remove all `/module-info.class` from all jars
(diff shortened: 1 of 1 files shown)


joern-cli/build.sbt
```diff
@@ -131,4 +131,26 @@ generateScaladocs := {
...
+  val logger = streams.value.log
+  val libDir = (Universal/stagingDirectory).value / "lib"
+
+  // remove all `/module-info.class` from all jars
+  Files.walk(libDir.toPath)
+    .filter(_.toString.endsWith(".jar"))
+    .forEach { jar =>
+      val zipFs = FileSystems.newFileSystem(jar)
+      zipFs.getRootDirectories.forEach { zipRootDir =>
+        Files.list(zipRootDir).filter(_.toString == "/module-info.class").forEach { moduleInfoClass =>
```
maltek, line 146: does this comparison with a `/` work on Windows?

maltek (commented):
are we certain that none of the JARs with such files will ever be signed? (If not, we'd also have to strip the signatures...)

mpollmeier:
> are we certain that none of the JARs with such files will ever be signed? (If not, we'd also have to strip the signatures...)

in that case we'd adapt the logic, but my hope is that the compiler bug is fixed by then...

joern-cli/build.sbt
mpollmeier: grafik
just updating crowdstrike, won't be long :)

mpollmeier: yes it works on windows

maltek (approved):

PR #4806, opened 2024-07-28
[x2cpg] Do not crash on .listRecursively

max-leuthaeuser:
This may throw (FileSystemLoopException, or any other reason why this dir may not be readable).
We use Files.walkFileTree now. The only solution where a safe continue mechanism can be implemented.

Fixes: https://shiftleftinc.atlassian.net/browse/SEN-2976

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
@@ -5,12 +5,48 @@ import better.files.*
 import org.slf4j.LoggerFactory
 
 import java.io.FileNotFoundException
+import java.nio.file.FileVisitor
+import java.nio.file.FileVisitResult
+import java.nio.file.Path
 import java.nio.file.Paths
+import java.nio.file.attribute.BasicFileAttributes
+import java.nio.file.Files
 import scala.util.matching.Regex
+
+import scala.jdk.CollectionConverters.SetHasAsJava
 
 object SourceFiles {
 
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
```diff
@@ -5,13 +5,49 @@ import better.files.*
...
   private val logger = LoggerFactory.getLogger(getClass)
 
+  /** Hack to have a FileVisitor in place that will continue iterating files even if an IOException happened during
+    * traversal.
+    */
+  private final class FailsafeFileVisitor extends FileVisitor[Path] {
+
+    private val seenFiles = scala.collection.mutable.Set.empty[Path]
+
+    def files(): Set[File] = seenFiles.map(File(_)).toSet
```
maltek, line 29: We might want to use linked hash sets here - or sort the results at the end - to get more predictable output.

Not that the order we traverse over files should materially affect the frontend behavior. But I imagine going over logs and such would be a lot more annoying when the files are parsed in effectively random order.

maltek (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
max-leuthaeuser: There is a .sorted call down below at the end.

maltek: ah, perfect :)

PR #4809, opened 2024-07-30
[ruby] Parser Tests

AndreiDreyer:
This PR handles:
 * Reworked parser tests to have a similar structure to python's `AstPrinter`. The `AstPrinter` follows the same structure as the `RubyNodeCreator` to mimic the AST.

TODO: Bug fixes for ignored tests.


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/testfixtures/RubyParserFixture.scala
```diff
@@ -13,7 +13,7 @@ import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Path}
 import scala.util.{Failure, Success, Using}
 
-class RubyParserFixture
+class RubyParserFixture(newMatch: Boolean = false)
```
ml86, line 16: I would really try to avoid such a setup. It is easy to forget setting newMatch to true and then the hole test suite is useless without anybody noticing it. 
Why do we need this in the first place?

ml86 (approved):

PR #4813, opened 2024-07-31
[ruby] String Array with Interpolations

### LINKED ISSUE #4810: [ruby][bug] `%W` String array with Interpolations not parsing correct
Given the following ruby code:
```ruby
%W(x#{1})
```
we end up with a `StaticLiteral(1)` as our final structure which is incorrect. This should give us a `ArrayLiteral` with the correct element (in this case `x1`).

AndreiDreyer:
This PR adds:
 * Handling for String Array Literal with Interpolations, denoted by `%W()`

TODO: Add parser test once #4809 is merged

Resolves #4810


ml86 (changes requested):
Please add the parser tests to this PR.

AndreiDreyer:
@ml86 
> Please add the parser tests to this PR.

Must I add the entire #4809 to this PR? There is already a parser test for this in the `ArrayParserTests`, but this is running on the old framework.

ml86:
@AndreiDreyer No, if #4809 contains the parser tests that is fine.

ml86 (approved):

PR #4815, opened 2024-07-31
[c2cpg] Name/FullName/Signature refactoring

max-leuthaeuser:


ml86 (commented):
The split between fullname related methods in AstCreatorHelper.scala and FullNameProvider.scala is unclear to me.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/FullNameProvider.scala
```diff
@@ -0,0 +1,337 @@
...
+          case f: CPPFunction if f.getDeclarations != null =>
+            f.getDeclarations.headOption.map(n => ASTStringUtil.getSimpleName(n.getName)).getOrElse(f.getName)
+          case f: CPPFunction if f.getDefinition != null => ASTStringUtil.getSimpleName(f.getDefinition.getName)
+          case other                                     => other.getName
+        }
+      case _ => lastNameOfQualifiedName(ASTStringUtil.getSimpleName(d.getName))
+    }
+  }
+
+  protected def shortName(node: IASTNode): String = {
```
ml86, line 183: Having two separate code path for short and full names brings the risk that the short names differ from the corresponding parts in the full name. 
Is there anything that speaks against a shared code path and only at the end we select if short or full name are taken?

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/FullNameProvider.scala
```diff
@@ -0,0 +1,337 @@
...
+          case cVariable: CVariable => Option(cVariable.getName)
+          case _                    => Option(declarator.getName.toString)
+        }
+      case definition: ICPPASTFunctionDefinition =>
+        Option(fullName(definition.getDeclarator))
+      case _ => None
+    }
+  }
+
+  private def fullNameForCPPASTIdExpression(d: CPPASTIdExpression): String = {
```
ml86, line 246: This method is called as fallback if fullNameFromBinding returns None. But it itself tries to work on bindings again. Does not seem to make sense.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/FullNameProvider.scala
max-leuthaeuser: One could do so but that would blow up that helper function both in code size as well as complexity. We have plenty of tests to ensure both are synced but its a good target for a future refactoring here.

max-leuthaeuser:
> The split between fullname related methods in AstCreatorHelper.scala and FullNameProvider.scala is unclear to me.

This has been fixed.
Sptests is also happy with many fixed fullnames (for lambdas).

ml86 (approved):

PR #4821, opened 2024-08-02
Bump cpg version and cleanup.

ml86:
Using the old SerializedCpg API hat no effect anymore.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/layers/LayerCreator.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/layers/LayerCreator.scala
@@ -36,17 +36,8 @@ abstract class LayerCreator {
     }
   }
 
-  protected def initSerializedCpg(outputDir: Option[String], passName: String, index: Int = 0): SerializedCpg = {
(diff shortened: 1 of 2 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/layers/LayerCreator.scala
```diff
@@ -36,17 +36,8 @@ abstract class LayerCreator {
...
-      case Some(dir) => new SerializedCpg((File(dir) / s"${index}_$passName").path.toAbsolutePath.toString)
-      case None      => new SerializedCpg()
-    }
-  }
-
   protected def runPass(pass: CpgPassBase, context: LayerCreatorContext, index: Int = 0): Unit = {
-    val serializedCpg = initSerializedCpg(context.outputDir, pass.name, index)
-    pass.createApplySerializeAndStore(serializedCpg)
-    serializedCpg.close()
+    pass.createAndApply()
```
maltek, line 40: Seems like this `runPass` method should be dropped next. (Not neccessarily now.)

maltek (approved):

PR #4829, opened 2024-08-07
[rubysrc2cpg] Fix implicit import handling.

ml86:
Symbols/identifiers used as based class where not taken into account for
the implicit import creation.


maltek (approved):

DavidBakerEffendi (commented):
Looks good

PR #4834, opened 2024-08-08
[ruby] Fix last parser tests

AndreiDreyer:
This PR fixes the last two parser test set:
 * Regex parser tests
 * Method defs in classes parser tests


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
```diff
@@ -1174,8 +1174,16 @@ class RubyNodeCreator extends RubyParserBaseVisitor[RubyNode] {
       if otherTypeDeclChildren.nonEmpty then "\n" + otherTypeDeclChildren.map(_.span.text).mkString("\n")
       else ""
 
+    val initMethodSpanText =
+      if initMethod.nonEmpty then "\n" + initMethod.head.span.text
```
DavidBakerEffendi, line 1178: For best practices, I would prefer string interpolation for string building here. Also, if we have a list we can match it like:
```scala
initMethod match {
  case head :: _ => s"\n${head.span.text}"
  case _ => ""
}
```

AndreiDreyer: done

DavidBakerEffendi (approved):

PR #4838, opened 2024-08-12
[ruby] `self` Param `REF` Edge

DavidBakerEffendi:
Fixed issue where no `REF` edges between self params and their identifiers.


maltek (commented):
seems like a good step forward, though for railsgoat I still see quite a few orphaned `self` identifiers:
```
cpg.identifier.name("self").not(_.refsTo).size
val res6: Int = 239
```

DavidBakerEffendi:
@maltek thanks, I'll verify against railsgoat to debug

DavidBakerEffendi:
@maltek Fixed the rest of the cases

DavidBakerEffendi:
@maltek FYI this is causing tests to fail in `codescience`. I'll chat with Markus when he's back from leave to understand why this may be an issue.

PR #4840, opened 2024-08-12
[ruby] Fixed Pseudo-Variables as Literals

DavidBakerEffendi:
Although the parser rule for picking up pseudo-variables seems fine, it may sometimes choose to represent it as a local variable. This fixes this and adds the related tests.


xavierpinho:
>  it may sometimes choose to represent it as a local variable

That's very likely due to some ambiguity in the grammar spec. It might be beneficial (in the future) to remove the pseudo-variable rule in the parser if we end up needing to test it in the AST building phase anyway, so that the parser loses this ambiguity and becomes more responsive.

maltek (approved):

PR #4841, opened 2024-08-12
[ruby] Fix Implicit Require Call Structure

DavidBakerEffendi:
The implicit require calls that are generated after the AST creation had an older/out-dated format and lead to identifiers not being linked via REF edges.


maltek (commented):
I would say that these `require` calls would ideally be simple static calls without any receiver at all. Just like e.g. `<operator>` calls.

(Even if the application code does something weird like define its own `require` variable, I think our modeling of autoloading would only suffer from dynamically resolving the implicitly added imports to that variable.)

DavidBakerEffendi:
@maltek I've made them static calls now as well as changed the name to `require_relative` as this only considers internal symbols

maltek (approved):

Issue #4822, opened 2024-08-02
[rubysrc] tracking bug for simpler misc issues


maltek:
I'll use this issue as a tracker bug for simpler issues. I might add more things as I come across them.

- [ ] identifiers for `self` parameters are missing REF edges to the corresponding parameters
- [ ] other identifiers are created without any matching local nodes (just run `cpg.identifier.not(_.refsTo).size` e.g. on railsgoat)
- [ ] the argument `0` of `<operator>.alloc` calls seems to always created get as an identifier, even if it's a complex expression (e.g. https://github.com/OWASP/railsgoat/blob/c1e8ff1e3b24a1c48fcfc9fbee0f65dc296b49d9/app/controllers/api/v1/mobile_controller.rb#L10)
- [ ] when a class is defined within a namespace, e.g. `class Api::V1::MobileController`, the type decl has a name field `Api`. The actual name `MobileController` is nowhere in the CPG, not even in the type's full name.
- [ ] parsing an empty file runs into a null pointer exception
- [ ] the detection of builtin method calls is a bit too eager, e.g. https://github.com/OWASP/railsgoat/blob/master/app/models/analytics.rb#L3 is translated as a call to `__core.Kernel.select` which it isn't
- [ ] the way that splitting receiver/arg 0 of calls was implemented, leads to a huge amount of duplication for chained method calls: `a().b().c().d()` has 1 call with name `a`, 2 calls with name `b`, 3x `c` and 4x `d`. Compare with e.g. jssrc: there, the receiver is an assignment `(_tmp_1 = a().b()).c` and argument 0 is simply a reference to that identifier `_tmp_1`

AndreiDreyer:
@maltek I'm struggling to recreate the issue for `detection of builtin method calls are a bit too eager`. If I use the snippet you linked in a test case in the Ruby frontend and run: 
```scala
    cpg.call.name("select").l.foreach(x => println(s"${x.code}: ${x.typeFullName} ${x.dynamicTypeHintFullName} ${x.possibleTypes}"))
```
then none of the calls to `select` has any types
![image](https://github.com/user-attachments/assets/dda3ef2b-7015-4dac-a0e0-2063b56b4bd9)

AndreiDreyer:
The line of the code snippet provided for the `<operator>.alloc` call issue doesn't have any calls to `<operator>.alloc`.
![image](https://github.com/user-attachments/assets/58a74f36-159a-4204-9b5e-64ebcc63ceaa)

[...]

PR #4845, opened 2024-08-13
[ruby] Constructor Lowering Fix 

DavidBakerEffendi:
* Fixed constructor lowering structure as per https://github.com/joernio/joern/issues/4822
* Tested and fixed issue in parenthesis-less `new` member calls (could use a better long-term fix in the parser)
* Handling cases where the constructor call is off of an expression instead of some class identifier

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
@@ -266,15 +266,13 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
   }
 
   protected def astForObjectInstantiation(node: RubyNode & ObjectInstantiation): Ast = {
-    val className  = node.target.text
-    val callName   = "new"
-    val methodName = Defines.Initialize
+    val className = node.target.text // fixme: This is not always a simple identifier
     /*
       We short-cut the call edge from `new` call to `initialize` method, however we keep the modelling of the receiver
       as referring to the singleton class.
      */
     val (receiverTypeFullName, fullName) = scope.tryResolveTypeReference(className) match {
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -318,8 +318,11 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
         x.arguments.map(astForMethodCallArgument) :+ typeRef
     }
 
-    val constructorCall    = callNode(node, code(node), callName, fullName, DispatchTypes.DYNAMIC_DISPATCH)
-    val constructorCallAst = callAst(constructorCall, argumentAsts, Option(tmpIdentifier))
+    val constructorCall =
+      callNode(node, Defines.Initialize, Defines.Initialize, Defines.Any, DispatchTypes.DYNAMIC_DISPATCH)
```
maltek, line 322: I suppose this is just nit-picking, but I don't like that the code field is simply `initialize`. Why not `code(node)`?

maltek (approved):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
DavidBakerEffendi: Related to the node I left about parser packing being a bit weird on parenthesis-less calls, it would set undesired code e.g. `A` or `a = A` instead of `A.new`. If preferred, I could construct a code block from the target and arguments.

DavidBakerEffendi: Actually nevermind, I've changed it to `code(node)` and things seem fine

maltek: perfect :)

PR #4861, opened 2024-08-19
[ruby] removed assignment operator restriction on BracketedAssignments

AndreiDreyer:
Removed `warning` log and `defaultResult` when using any assignment operator other than equals on `BracketAssignment`

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/SingleAssignmentTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/SingleAssignmentTests.scala
@@ -289,4 +289,84 @@ class SingleAssignmentTests extends RubyCode2CpgFixture {
       case xs => fail(s"Expected three lambdas, got ${xs.size} lambdas instead")
     }
   }
+
+  "Bracketed ||=" in {
+    val cpg = code("""
+        |hash[:id] ||= s[:id]
+        |""".stripMargin)
+
+    inside(cpg.call.name(Operators.assignmentOr).l) {
+      case assignmentCall :: Nil =>
+        assignmentCall.code shouldBe "hash[:id] ||= s[:id]"
+        assignmentCall.dispatchType shouldBe DispatchTypes.STATIC_DISPATCH
+
+        inside(assignmentCall.argument.l) {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/SingleAssignmentTests.scala
```diff
@@ -289,4 +289,84 @@ class SingleAssignmentTests extends RubyCode2CpgFixture {
       case xs => fail(s"Expected three lambdas, got ${xs.size} lambdas instead")
     }
   }
+
+  "Bracketed ||=" in {
+    val cpg = code("""
+        |hash[:id] ||= s[:id]
+        |""".stripMargin)
+
+    inside(cpg.call.name(Operators.assignmentOr).l) {
```
DavidBakerEffendi, line 298: I believe this is conditional assignment, not assign or https://stackoverflow.com/a/14697343
Probably the case that we need to lower this into:
```ruby
if hash[:id].nil? then
  hash[:id] = s[:id]
end
```

DavidBakerEffendi (changes requested):
One operator is incorrect, but could probably be lowered fairly easily. If it's more complicated, then we can separate the issue

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/SingleAssignmentTests.scala
```diff
@@ -289,4 +289,84 @@ class SingleAssignmentTests extends RubyCode2CpgFixture {
...
+            rhs.code shouldBe "s[:id]"
+          case xs => fail(s"Expected lhs and rhs arguments, got ${xs.code.mkString(",")}")
+        }
+      case xs => fail(s"Expected on assignmentOr call, got ${xs.code.mkString(",")}")
+    }
+  }
+
+  "Bracketed &&=" in {
+    val cpg = code("""
+        |hash[:id] &&= s[:id]
```
DavidBakerEffendi, line 335: `&&=` is actually the "truthy" version of `||=`, so 
```ruby
if !hash[:id].nil? then hash[:id] = s[:id] end
```

DavidBakerEffendi (approved):

PR #4862, opened 2024-08-19
[ruby] Visibility Access Modifiers

DavidBakerEffendi:
* Handles various access modifiers private/public/protected
* Also considers file-level methods are private and initialize methods cannot be anything other than private
* Accounts for nested types where their modifiers are separate from surrounding type

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
@@ -51,6 +51,7 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this: 
   ): Seq[Ast] = {
     val className    = nameIdentifier.text
     val inheritsFrom = node.baseClass.map(getBaseClassName).toList
+    pushAccessModifier(ModifierTypes.PUBLIC)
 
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -63,7 +63,8 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this:
       */
     def populateAstParentValues(typeDecl: NewTypeDecl, astParentFullName: String): NewTypeDecl = {
       val namespaceBlockFullName = s"${scope.surroundingScopeFullName.getOrElse("")}.$astParentFullName"
-      scope.pushNewScope(NamespaceScope(s"${scope.surroundingScopeFullName.getOrElse("")}.$astParentFullName"))
+      scope.pushNewScope(NamespaceScope(namespaceBlockFullName))
+      pushAccessModifier(ModifierTypes.PUBLIC)
```
AndreiDreyer, line 67: Why do we assume that the modfiier here is `PUBLIC`?

DavidBakerEffendi: Hmm, seems wrong to me, I think I meant to place that higher

AndreiDreyer (approved):

PR #4864, opened 2024-08-20
[ruby] Add `commandTernaryExpression` to `command`

AndreiDreyer:
Added `commandTernaryExpression` to `command` in the ANTLR grammar for situations like the given test case where we have a `ternaryExpression` as a parameter to a parenthesisless call


AndreiDreyer:
Seems I have fixed on problem on Railsgoat, but created 20 other problems. Will re-open once resolved

DavidBakerEffendi (commented):
Nice!

DavidBakerEffendi (approved):

PR #4867, opened 2024-08-21
[php2cpg] Improve performance.

ml86:
This php parser we use seems to have some warmup phase in which it
has very poor performance which makes it very expensive to feed
it single files like we so far did.
This change invokes the php parser with groups of 20 files which on my
machine give ~40% performance increase for the total frontend runtime on
large projects.
Bigger groups of files strangely only show marginal performance
increases despite the fact that profiling the frontend indicates that
there is still a lot of time wasted by the large amount of individual
parser invocations.
For now we keep it like this because for bigger groups we would run into
argument length limitations which are especially a pain on systems
like Windows. Sadly the php parser does not provide means to specify
the to be parsed files other than as a list on the command line.
So there would need to be some change there first before we could
increase the groups size in a meaningful way.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
@@ -24,8 +24,12 @@ class PhpParser private (phpParserPath: String, phpIniPath: String, disableFileC
   }
 
   def parseFiles(inputPaths: collection.Seq[String]): collection.Seq[(String, Option[PhpFile], String)] = {
+    // We need to keep a map between the input path and its canonical representation in
+    // order to map back the canonical file name we get from the php parser.
+    // Otherwise later on file name/path processing might get confused because the returned
+    // file paths are in no relation to the input paths.
     val canonicalToInputPath = mutable.HashMap.empty[String, String]
-    
+
     inputPaths.foreach { inputPath =>
       val canonicalPath = Path.of(inputPath).toFile.getCanonicalPath
       canonicalToInputPath.put(canonicalPath, inputPath)
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
@@ -28,7 +28,7 @@ class PhpParser private (phpParserPath: String, phpIniPath: String, disableFileC
       inputFile.canonicalPath
     }
 
-    val command = phpParseCommand(absoluteInputPaths)
+    val command = phpParseCommand(inputPaths)
 
     val (returnValue, output) = ExternalCommand.runWithMergeStdoutAndStderr(command, ".")
     returnValue match {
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
```diff
@@ -7,63 +7,115 @@ import io.joern.x2cpg.utils.ExternalCommand
...
+      inputFile.canonicalPath
     }
-  }
 
-  private def processParserOutput(output: Seq[String], filename: String): Option[PhpFile] = {
-    val maybeJson = linesToJsonValue(output, filename)
-    maybeJson.flatMap(jsonValueToPhpFile(_, filename))
+    val command = phpParseCommand(absoluteInputPath)
+
+    val (returnValue, output) = ExternalCommand.runWithMergeStdoutAndStderr(command, ".")
```
johannescoetzee, line 33: I don't know if this is something that could ever happen with how PhpParser handles output, but I'm wondering if it's possible to end up in a situation where an error occurs, printing something to stderr, but we still get JSON output (perhaps with error nodes or something like that) and the error and JSON output end up interleaved in some way. This could result in the JSON parsing to fail.

If we handle stdout and stderr as separate streams, then we could prepend stderr to stdout which would result in the error output being included in `infoLines` below.

With that being said, I might be misunderstanding the purpose of merging these, so please ignore if my comment doesn't make sense.

ml86: The problem is that the php parser gives information about the parsed file on stderr. This includes the file name and log messages about parsing errors and the actual json is provided on stdout.
This was easy to handle so far because only one file was given to the php parser. So the relation between file name, log messages and the json ouput was clear.
Now that we give multiple files as input this is not the case anymore, although the php parser seems to preserve the order, i did not want to (fully) rely on it. To a certain amount our code still does because we assume a certain timing between stderr and stdout which ensures that the file name comes first on stderr and than comes the json on stdout.
The hole interface provided by the php parsers is pretty bad to consume jsons for multiple file but sadly that is what it is. 
To improve we will like need to provide a PR upstream.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
```diff
@@ -7,63 +7,115 @@ import io.joern.x2cpg.utils.ExternalCommand
...
-    val command        = phpParseCommand(inputFilePath)
-    ExternalCommand.run(command, inputDirectory) match {
-      case Success(output) =>
-        val content = Option.unless(disableFileContent)(inputFile.contentAsString)
-        processParserOutput(output, inputFilePath).map((_, content))
-      case Failure(exception) =>
-        logger.error(s"Failure running php-parser with $command", exception.getMessage)
-        None
+  def parseFiles(inputPaths: collection.Seq[String]): collection.Seq[(String, Option[PhpFile], String)] = {
+    val absoluteInputPath = inputPaths.map { inputPath =>
```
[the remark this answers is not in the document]
ml86: Done.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
johannescoetzee: Ah, I see. In that case I don't see any alternative to this implementation until this is improved upstream

johannescoetzee (approved):

PR #4871, opened 2024-08-23
[php2cpg] Fixed the initialization of static/const members of class

d1tto:
1. The initialization of static members of class should be placed in `<clinit>` function, not in the constructor of the class. 
2. Access to static members should be prefixed with `self` keyword to distinguish them from normal class members. For example:
    ```php
    <?php
    class Foo {
        static $A = "A";
        const B = "B";
        function foo() {
            echo self::$A;
            echo self::B;
        }
    }
    ```

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
@@ -10,7 +10,7 @@ import io.joern.x2cpg.Defines.{StaticInitMethodName, UnresolvedNamespace, Unreso
 import io.joern.x2cpg.utils.AstPropertiesUtil.RootProperties
 import io.joern.x2cpg.utils.IntervalKeyPool
 import io.joern.x2cpg.utils.NodeBuilders.*
-import io.joern.x2cpg.{Ast, AstCreatorBase, AstNodeBuilder, ValidationMode}
+import io.joern.x2cpg.{Ast, AstCreatorBase, AstNodeBuilder, Defines, ValidationMode}
 import io.shiftleft.codepropertygraph.generated.*
 import io.shiftleft.codepropertygraph.generated.nodes.*
 import io.shiftleft.semanticcpg.language.types.structure.NamespaceTraversal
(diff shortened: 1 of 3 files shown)


d1tto:
@DavidBakerEffendi Excuse me, could you help me review this pr

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
```diff
@@ -885,11 +885,15 @@ class AstCreator(filename: String, phpAst: PhpFile, fileContent: Option[String],
...
     } else {
-      val identifierCode = memberNode.code.replaceAll("const ", "").replaceAll("case ", "")
-      val typeFullName   = Option(memberNode.typeFullName)
-      val identifier = newIdentifierNode(memberNode.name, typeFullName.getOrElse("ANY"))
-        .code(identifierCode)
-      Ast(identifier).withRefEdge(identifier, memberNode)
+      val selfIdentifier = {
+        val name = "self"
+        val typ  = scope.getEnclosingTypeDeclTypeName
+        newIdentifierNode(name, typ.getOrElse("ANY"), typ.toList, memberNode.lineNumber).code(name)
```
DavidBakerEffendi, line 891: from `io.joern.x2cpg.Defines`
```suggestion
        newIdentifierNode(name, typ.getOrElse(Defines.Any), typ.toList, memberNode.lineNumber).code(name)
```

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
```diff
@@ -885,11 +885,15 @@ class AstCreator(filename: String, phpAst: PhpFile, fileContent: Option[String],
...
+      }
+      val fieldIdentifier = newFieldIdentifierNode(memberNode.name, memberNode.lineNumber)
+      val code = s"self::${memberNode.code.replace("static ", "").replace("case ", "").replace("const ", "")}"
```
DavidBakerEffendi, line 894: I think should work, and if it does will be more concise
```suggestion
      val code = s"self::${memberNode.code.replaceAll("(static|case|const) ", "")}"
```

DavidBakerEffendi (commented):
I'll leave the final say with @johannescoetzee

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreator.scala
d1tto: Applied. Thanks

d1tto:
Excuse me, is there any progress on this pr 😛

DavidBakerEffendi (approved):
I'm happy to merge this, then if it breaks something for stakeholders we can always revert worst case

PR #4874, opened 2024-08-25
[cli] recover `help` functionality

xavierpinho:
Recently noticed that the `help` command was no longer working as expected, cf.

```
% ./joern

     ██╗ ██████╗ ███████╗██████╗ ███╗   ██╗
     ██║██╔═══██╗██╔════╝██╔══██╗████╗  ██║
     ██║██║   ██║█████╗  ██████╔╝██╔██╗ ██║
██   ██║██║   ██║██╔══╝  ██╔══██╗██║╚██╗██║
╚█████╔╝╚██████╔╝███████╗██║  ██║██║ ╚████║
 ╚════╝  ╚═════╝ ╚══════╝╚═╝  ╚═╝╚═╝  ╚═══╝
Version: 4.0.52
Type `help` to begin


joern> help
val res0: io.shiftleft.codepropertygraph.generated.help.type = io.shiftleft.codepropertygraph.generated.package$help$@1b2f18bd
```

---

This PR hides this specific `help` object from the REPL imports, thus recovering the original behaviour. Not sure this is the ideal approach, but not being familiar with this neck of the woods, didn't want to invest too much time without consulting first.

### DIFF
--- a/joern-cli/src/main/scala/io/joern/joerncli/console/Predefined.scala
+++ b/joern-cli/src/main/scala/io/joern/joerncli/console/Predefined.scala
@@ -9,7 +9,7 @@ object Predefined {
       "import _root_.io.joern.console.*",
       "import _root_.io.joern.joerncli.console.JoernConsole.*",
       "import _root_.io.shiftleft.codepropertygraph.cpgloading.*",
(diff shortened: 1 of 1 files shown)


mpollmeier (approved):
thank you for noticing and the fix!

Issue #4674, opened 2024-06-18
[C#]:Type recovery and data flow completion in CPG


chengkenyong:
When I use joern to scan C# projects, I encountered several issues:

1. For nodes of type 'fieldAccess', it cannot determine the type of the node. Additionally, there are occasional missing properties for other types of nodes.
![微信截图_20240618135636](https://github.com/joernio/joern/assets/82140257/f6324765-16f8-4365-8481-d3f712dde5e8)
![微信截图_20240618135942](https://github.com/joernio/joern/assets/82140257/91f228c1-0377-4d4a-bb3b-cfa14730ad75)
2. Joern seems to miss scanning content within Object Initializer (or anonymous method bodies) in C#.
![微信截图_20240618143512](https://github.com/joernio/joern/assets/82140257/1dd211a7-f0ef-4938-8ce4-2e2995c9e14c)
![微信截图_20240618142953](https://github.com/joernio/joern/assets/82140257/450c78b1-83af-4214-a941-c144f416742f)
3. There are instances where data flow is incomplete in certain nodes.
![微信截图_20240614162500](https://github.com/joernio/joern/assets/82140257/92a89fe1-5813-40df-bf57-8595e7f03c86)
I am particularly interested in understanding the principles behind attribute recovery, node recovery, and data flow completion. I would like to contribute to addressing these issues and hope you can provide me with guidance in this area. I understand you may be busy, but I sincerely hope you can offer me some relevant advice. I am committed to making a meaningful contribution. Thank you.

DavidBakerEffendi:
Hi @chengkenyong, development of C# frontend has been largely funded by Privado, but they have since scaled it back a bit due to priorities on other languages. The C# frontend relies on two components (1) the Roslyn parser wrapped in a neat binary which can be found here https://github.com/joernio/DotNetAstGen, and (2) the AstCreation in Joern.

For the missing AST features you've described, it could be that they need to be filtered in on the parser. So that change would be in DotNetAstGen. One way to check is to create a JSON AST dump of the file you're interested in, and see if it is present. 

If the entity is there, then it probably needs a handler in one of the classes within the `astcreation` package in csharpsrc2cpg. 

[...]

PR #4877, opened 2024-08-26
[c#] Implements the `forAst` Construct

DavidBakerEffendi:
The current `foreach` implementation is based on an older approach that doesn't capture the semantics quite as well.

This PR implements the newer `forAst` construct, which encourages de-sugaring and representing what is semantically happening underneath the `foreach` statement.

Resolves https://github.com/joernio/joern/issues/4674#issuecomment-2222254060

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForStatementsCreator.scala
@@ -6,10 +6,14 @@ import io.joern.csharpsrc2cpg.parser.ParserKeys
 import io.joern.csharpsrc2cpg.parser.DotNetJsonAst.*
 import io.joern.x2cpg.Ast
 import io.joern.x2cpg.ValidationMode
-import io.shiftleft.codepropertygraph.generated.ControlStructureTypes
-import io.shiftleft.codepropertygraph.generated.DispatchTypes
-import io.shiftleft.codepropertygraph.generated.nodes.NewControlStructure
-import io.shiftleft.codepropertygraph.generated.nodes.NewIdentifier
+import io.shiftleft.codepropertygraph.generated.{ControlStructureTypes, DispatchTypes, Operators}
+import io.shiftleft.codepropertygraph.generated.nodes.{
+  NewControlStructure,
+  NewFieldIdentifier,
+  NewIdentifier,
+  NewLiteral,
+  NewLocal
+}
 
 import scala.::
 import scala.util.Try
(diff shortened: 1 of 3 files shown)


AndreiDreyer (approved):

chengkenyong:
Hi!  @DavidBakerEffendi ,Thank you very much for solving my problem! Actually, I also made some rough modifications to the foreach beforehand. The specific modifications are as follows. 
```scala 
  private def astForForEachStatement(forEachStmt: DotNetNodeInfo): Seq[Ast] = {
    val forEachNode     = controlStructureNode(forEachStmt, ControlStructureTypes.FOR, forEachStmt.code)
    val iterableAst     = astForNode(forEachStmt.json(ParserKeys.Expression))

    // ----------------------------------Add the foreach iterator variable to the previous scope-------------------------------------
    val identifierValue = forEachStmt.json(ParserKeys.Identifier)(ParserKeys.Value).str
    val _localNode =
      localNode(
        node = createDotNetNodeInfo(forEachStmt.json(ParserKeys.Type)),
        name = identifierValue,
        code = identifierValue,
        typeFullName = nodeTypeFullName(createDotNetNodeInfo(forEachStmt.json(ParserKeys.Type)))
      )
    scope.addToScope(identifierValue, _localNode)
    // -----------------------------------------------------------------------------------------------------------------

    val forEachBlockAst = astForBlock(createDotNetNodeInfo(forEachStmt.json(ParserKeys.Statement)))

    val _identifierNode =
      identifierNode(
        node = createDotNetNodeInfo(forEachStmt.json(ParserKeys.Type)),
        name = identifierValue,
        code = identifierValue,
        typeFullName = nodeTypeFullName(createDotNetNodeInfo(forEachStmt.json(ParserKeys.Type)))
      )
    val iteratorVarAst = Ast(_identifierNode)

    // ------------------------------------------ Define a custom "in" operator--------------------------------------------------
    val args         = iteratorVarAst +: iterableAst
    val iterableCode = forEachStmt.json(ParserKeys.Expression)(ParserKeys.MetaData)(ParserKeys.Code).str
    val operatorCode = s"$identifierValue in $iterableCode"
    val operatorAst = callNode(
      forEachStmt,
      operatorCode,
      "<operator>.elementOf",
      "<operator>.elementOf",
      DispatchTypes.STATIC_DISPATCH,
      None,
      Option(_identifierNode.typeFullName)
    )
    val conditionAst = Seq(callAst(operatorAst, args))
    scope.moveToScope(identifierValue)//Remove identifierValue from the scope
    // -----------------------------------------------------------------------------------------------------------------

    Seq(
      Ast(forEachNode)
        .withChildren(conditionAst)
        .withChild(forEachBlockAst)
        .withConditionEdges(forEachNode, conditionAst.flatMap(_.root).toList)
    )
  }
```

However, during testing, I encountered many minor issues. For example, in the test code sample below, 
微信截图_20240829153939
there are data flow directions that violate the control flow.
graphviz (9)
 But when I tested your submitted code, I found that these issues didn't exist. I'm very curious about the reason behind this (I was actually looking into the specific CFG implementation logic in Joern). I would be very grateful if you could explain this to me.

DavidBakerEffendi:
Hi @chengkenyong, so while your implementation is quite close, one thing to keep in mind that unknown calls are overapproximated by Joern, and I think that may be what is causing the extra unrealizable paths. If a call is unknown, and doesn't have a respective semantic entry, it will consider all input and output arguments being defined and used by all input and outpu arguments. By adding `<operator>.elementOf` without specifying its semantic under `DefaultSemantics` (under the `dataflowengineoss` module) it may cause this. 

Another cause may be how the CFG generator interprets children on the `FOR` control structures, where each position has a special meaning in the scope of where the control flows (which then may result in data-flows being correct or incorrect).

chengkenyong:
Hi @DavidBakerEffendi ,I really appreciate your answer! Actually, I have already specified the semantics for <operator>.elementOf under the dataflowengineoss module, but the results are not ideal. Currently, I'm also looking into the logic of adding control flow edges to child nodes in the cfgForForStatements method in the CFG!Your suggestion has further clarified my direction.Thanks

DavidBakerEffendi:
@chengkenyong This is quite unusual, I wonder if this is caused by how field sources are retrieved at the start of the data-flow analysis. I know we've had to do some workarounds for handling global variables in dynamic languages on file-level, so that might be the cause. The code I'm thinking about is under `io.joern.dataflowengineoss.queryengine.SourcesToStartingPoints`

chengkenyong:
@DavidBakerEffendi , Thank you very much for your explanation. I will take a look at its implementation logic.

PR #4884, opened 2024-08-28
[ruby] Add handling for ArrayLiteral used directly as argument in call

AndreiDreyer:
Added handling where an `Array` is initialized and used directly as a paramter in a function call.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/antlr4/io/joern/rubysrc2cpg/parser/RubyParser.g4
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/antlr4/io/joern/rubysrc2cpg/parser/RubyParser.g4
@@ -216,10 +216,12 @@ argumentList
         # operatorsArgumentList
     |   associationList (COMMA NL* splattingArgument)? (COMMA NL* blockArgument)?
         # associationsArgumentList
+    |   LBRACK ((symbol|association) COMMA? NL*)* RBRACK
+        # arrayArgumentList
     |   command
         # singleCommandArgumentList
     ;
-    
+
 commandArgumentList
     :   associationList
     |   primaryValueList (COMMA NL* associationList)?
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/antlr4/io/joern/rubysrc2cpg/parser/RubyParser.g4
```diff
@@ -216,10 +216,12 @@ argumentList
         # operatorsArgumentList
     |   associationList (COMMA NL* splattingArgument)? (COMMA NL* blockArgument)?
         # associationsArgumentList
+    |   LBRACK ((symbol|association) COMMA? NL*)* RBRACK
```
DavidBakerEffendi, line 219: The following change also works. I see that `[1, 2, 3]` would work as an argument, and that the issue was the symbols/associations as this parser rules implies. I then followed why normal arrays weren't handling it and found there wasn't a rule for these in "indexingArgumentList"s. 

Perhaps my proposed solution will be a bit more general? As it covers array bodies elsewhere too

Screenshot 2024-08-28 at 08 18 04

DavidBakerEffendi: Might want to name it `symbolOrAssocationIndexingArgumentList`

joern-cli/frontends/rubysrc2cpg/src/main/antlr4/io/joern/rubysrc2cpg/parser/RubyParser.g4
```diff
@@ -177,6 +177,8 @@ indexingArgumentList
         # operatorExpressionListIndexingArgumentList
     |   operatorExpressionList COMMA splattingArgument
         # operatorExpressionListWithSplattingArgumentIndexingArgumentList
+    |   ((symbol|association) COMMA? NL*)*
+        #symbolOrAssociationIndexList
```
DavidBakerEffendi, line 181: Could we make the naming a bit more consistent, ie end in "IndexingArgumentList"

DavidBakerEffendi (approved):

PR #4895, opened 2024-09-04
[ruby] `Block Parameter` fixes

AndreiDreyer:
* Added `GroupedParameter` type and handling for new type
* Fixed ANTLR grammar rules for different combinations of block parameters
* Updated relevant parser tests


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -464,6 +461,15 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
                       case _                 =>
                     }
                     astForExpression(node.lhs)
+                  case x: SplattingRubyNode
```
DavidBakerEffendi, line 464: More concise version:
```scala
case SplattingRubyNode(nameNode: SimpleIdentifier) if scope.lookupVariable(code(nameNode)).isEmpty =>
  val name  = code(nameNode)
  val local = localNode(nameNode, name, name, Defines.Any)
```

DavidBakerEffendi (approved):

PR #4901, opened 2024-09-06
[rubysrc2cpg] Fix ImportPass.

ml86:
Only static calls with import related names 'require', 'load" and so on
are linked to import nodes.
Before that we ended up with import nodes and links for calls like
`Marshal.load(...)`.


maltek (approved):
Maybe we should switch to fullnames to be even more precise?

PR #4904, opened 2024-09-09
[ruby] All `require`-like Calls are Static

DavidBakerEffendi:
* All `require`-like call names are now a set under `ImportsPass`
* These calls are now static calls with only the arguments as children

Seems related to https://github.com/joernio/joern/pull/4901


maltek (approved):

DavidBakerEffendi:
> Please add tests for all versions of `require` and `load`. This is too important of a feature to not have tests.

There are tests using these keywords towards the bottom of `ImportsTest`, but I added some simple verification of these keywords, unless you mean something else?

ml86 (approved):

PR #4908, opened 2024-09-10
[ruby] add `methodInvocationWithoutParentheses` to RETURN argumentList

AndreiDreyer:
Added `methodInvocationWithoutParentheses` to `RETURN` arguments

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodReturnTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodReturnTests.scala
@@ -497,9 +497,12 @@ class MethodReturnTests extends RubyCode2CpgFixture(withDataFlow = true) {
         val List(ifReturnTrue: Return) = ifNode.whenTrue.isBlock.astChildren.isReturn.l
         ifReturnTrue.code shouldBe "return render json: {}, status: :internal_server_error"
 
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodReturnTests.scala
```diff
@@ -479,4 +479,35 @@ class MethodReturnTests extends RubyCode2CpgFixture(withDataFlow = true) {
...
+
+        val List(notCall: Call) = ifNode.condition.l: @unchecked
+        notCall.methodFullName shouldBe Operators.logicalNot
+
+        val List(ifReturnTrue: Return) = ifNode.whenTrue.isBlock.astChildren.isReturn.l
+        ifReturnTrue.code shouldBe "return render json: {}, status: :internal_server_error"
+
+        val List(_, jsonArg, statusArg) = ifReturnTrue.astChildren.isCall.argument.l: @unchecked
+        jsonArg.argumentName shouldBe Some("json")
+        jsonArg.code shouldBe "<empty>"
```
DavidBakerEffendi, line 502: Could you verify the call node types in this? If `jsonArg` is a call with name `<operator>.hashInitializer` then I'm happy, but if it's a block, then let's add a TODO to handle this differently (like we mentioned offline)

DavidBakerEffendi (approved):

PR #4916, opened 2024-09-11
[javasrc2cpg] Remove JavaSrc2CpgTestContext

xavierpinho:
Super low priority, but unless I'm missing something, `JavaSrc2CpgTestContext` can be removed without any loss, since all its features are already provided by the current `JavaSrcTestCpg`, thus consisting of duplicated code.


DavidBakerEffendi (approved):
I think this looks fine, good to deduplicate this

PR #4920, opened 2024-09-13
[dataflowengineoss] Turn `Semantics` into a node-directed trait

xavierpinho:
Presently, `Semantics` are exclusively based on method fullNames and, while that is perfectly reasonable for statically typed languages such as Java, they are not so great for dynamic ones.

The proposal here is to turn `Semantics` into a common trait equipped with a `forMethod :: Method -> Option[FlowSemantic]` operation, that can take into account more information (since it's given the node itself) other than just its fullName. In doing so, I also renamed the current `Semantics` class to `FullNameSemantics`, to make it clear what the latter operates on.

In particular, this would allow users of Joern as-a-library to provide their own custom `Semantics` instance, that can inform the dataflowengine on a node-by-node basis.

If this proposal is welcomed, I also intend to refactor `SemanticTestCpg` and friends to allow custom `Semantic` objects, instead of `FullNameSemantics` exclusively. Moreover, I'd also like to introduce a `NoCrossTaint` semantics -- essentially a lifted `PassThroughMapping`. This patch is already too large as is, so I refrained from doing so at this point.


mpollmeier (commented):
Sounds like a good idea from my POV, but since I'm not super familiar with Semantics I'll pass the torch to someone else, maybe @DavidBakerEffendi for approval..?

DavidBakerEffendi (approved):
Looks nice! Opens up a world of possibilities for this notation

DavidBakerEffendi:
FYI I've experimented with representing parameters undergoing fine-grained operations such as field accesses and calls within a method, and that used a similar approach as above, but was pretty difficult (not undoable though). This already makes the semantic DSL more readable though.

DavidBakerEffendi:
@xavierpinho do you have permissions to merge?

xavierpinho:
> @xavierpinho do you have permissions to merge?

I have, thanks for asking :) Was just giving it some time in case someone wanted to chime in any other feedback first.

PR #4930, opened 2024-09-17
[kotlin2cpg] Fix "nacked" call representation.

ml86:
Fix representation of "nacked" calls like `someFunc(someArg)`. These so
far have always being represented as statically dispatched which is
obviously not correct.

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForExpressionsCreator.scala
@@ -11,9 +11,11 @@ import io.joern.x2cpg.utils.NodeBuilders
 import io.shiftleft.codepropertygraph.generated.DispatchTypes
 import io.shiftleft.codepropertygraph.generated.Operators
 import io.shiftleft.codepropertygraph.generated.nodes.NewMethodRef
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForExpressionsCreator.scala
```diff
@@ -456,21 +458,52 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+      if (resolvedCall == null) {
+        DispatchTypes.STATIC_DISPATCH
+      } else {
+        if (resolvedCall.getDispatchReceiver == null) {
+          DispatchTypes.STATIC_DISPATCH
+        } else {
+          resolvedCall.getResultingDescriptor match {
+            case functionDescriptor: FunctionDescriptor
+                if functionDescriptor.getVisibility == DescriptorVisibilities.PRIVATE =>
+              DispatchTypes.STATIC_DISPATCH
```
maltek, line 475: From my reading of the logic below, the `this` identifier at argument 0 will be missing for calls of `private` instance methods. Or am I missing something there?

ml86: Good catch. Updated the PR. Please have another look.

maltek (approved):

PR #4932, opened 2024-09-18
[ruby] Add handling for `(&)` param

AndreiDreyer:
Added handling for unnamed proc parameter:
```ruby
def foo(&)
  bar(&)
end

def bar
  yield if block_given?
end

foo do
  puts "hello"
end
```

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
@@ -17,8 +17,10 @@ import scala.jdk.CollectionConverters.*
 
 /** Converts an ANTLR Ruby Parse Tree into the intermediate Ruby AST.
   */
-class RubyNodeCreator(variableNameGen: FreshNameGenerator[String] = FreshNameGenerator(id => s"<tmp-$id>"))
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyNodeCreator.scala
```diff
@@ -17,8 +17,10 @@ import scala.jdk.CollectionConverters.*
 
 /** Converts an ANTLR Ruby Parse Tree into the intermediate Ruby AST.
   */
-class RubyNodeCreator(variableNameGen: FreshNameGenerator[String] = FreshNameGenerator(id => s"<tmp-$id>"))
-    extends RubyParserBaseVisitor[RubyExpression] {
+class RubyNodeCreator(
+  variableNameGen: FreshNameGenerator[String] = FreshNameGenerator(id => s"<tmp-$id>"),
+  procParamGen: FreshNameGenerator[Left[String, Nothing]] = FreshNameGenerator(id => Left(s"<proc-param-$id>"))
```
DavidBakerEffendi, line 22: Does this still need to be given as a constructor?

DavidBakerEffendi: Versus how it used to be

DavidBakerEffendi (approved):

PR #4934, opened 2024-09-18
[Dependency Fetcher] Fix gradle dependency fetcher android detection and detect project and configuration names automatically

johannescoetzee:
Before this PR , android project detection only worked for some projects, which would cause conflicts between the `android` plugin in a (java, I think) android subproject and the `java` plugin added by the init script.

The second issue this PR addresses is needing to provide a project name and configuration to the dependency fetcher (without which it would fall back to some defaults ). This is problematic for enterprise customers who would need to do this per app uploaded, but also limited the dependency fetcher to fetching dependencies for a single subproject. The new version automatically detects all projects/subprojects with dependencies, filters configuration names based on some heuristics and fetches dependencies for all of these. 

I tested these manually on some projects I found on github, so it's possible I missed some cases.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/DependencyResolver.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/DependencyResolver.scala
@@ -3,7 +3,6 @@ package io.joern.x2cpg.utils.dependency
 import better.files.File
 import io.joern.x2cpg.utils.ExternalCommand
 import io.joern.x2cpg.utils.dependency.GradleConfigKeys.GradleConfigKey
-import io.joern.x2cpg.utils.dependency.GradleDependencies.{defaultGradleAppName, defaultGradleConfigurationName}
 import org.slf4j.LoggerFactory
 
 import java.nio.file.Path
@@ -32,7 +31,7 @@ object DependencyResolver {
         None
       else if (isGradleBuildFile(buildFile)) {
         // TODO: Don't limit this to the default configuration name
-        getCoordinatesForGradleProject(buildFile.getParent, defaultGradleConfigurationName)
+        getCoordinatesForGradleProject(buildFile.getParent, "compileClasspath")
       } else {
         logger.warn(s"Found unsupported build file $buildFile")
         Nil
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/DependencyResolver.scala
```diff
@@ -84,12 +83,14 @@ object DependencyResolver {
     projectDir: Path
   ): Option[collection.Seq[String]] = {
     logger.info("resolving Gradle dependencies at {}", projectDir)
-    val gradleProjectName = params.forGradle.getOrElse(GradleConfigKeys.ProjectName, defaultGradleProjectName)
-    val gradleConfiguration =
-      params.forGradle.getOrElse(GradleConfigKeys.ConfigurationName, defaultGradleConfigurationName)
+    val gradleProjectName   = params.forGradle.get(GradleConfigKeys.ProjectName)
```
max-leuthaeuser, line 86: Are you sure that does not throw if the key is not found?
Maybe we should make this mor fault tolerant here.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/GradleDependencies.scala
```diff
@@ -258,60 +410,62 @@ object GradleDependencies {
...
                   case Success(connection) =>
                     Using.resource(connection) { c =>
-                      runGradleTask(c, initScript, initScriptFile.pathAsString) match {
-                        case Some(deps) =>
-                          Some(deps.map { d =>
-                            if (!d.endsWith(Constants.aarFileExtension)) d
+                      projectInfo.subprojects.keys.flatMap { projectNameInfo =>
+                        val taskName = projectNameInfo.makeGradleTaskName(initScript.taskName)
+
+                        val x = 2
```
max-leuthaeuser, line 436: ?

max-leuthaeuser (commented):
Does it also work with CS/sptests? Afaik we also fetch deps there.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/DependencyResolver.scala
johannescoetzee: `params.forGradle` is a scala Map, so the `get` call returns an option. I'll rename the variables to make this clearer.

johannescoetzee:
> Does it also work with CS/sptests? Afaik we also fetch deps there.

We only have Maven projects for javasrc2cpg and it looks like the kotlin2cpg tests all use the gradle dependency fetcher plugin (https://github.com/ShiftLeftSecurity/gather-dependencies-gradle-plugin). I'll try these changes with a modified version of sptests

max-leuthaeuser (approved):

johannescoetzee:
This updated version seems to work on 3 of the 4 kotlin test projects with expectations. There was an issue with the springboot project where the project name (and therefore map key) in the init script contained a `-`. I didn't have this enclosed in `""`, so this caused an issue where regular identifiers don't. That has been fixed in the latest commit.

The android-vroom project doesn't work out of the box due to 
```
ould not determine the dependencies of task ':app:x2cpgCopyDeps_DgEt5lji_androidDeps'.
> SDK location not found. Define a valid SDK location with an ANDROID_HOME environment variable or by setting the sdk.dir path in your project's local properties file at
```

For some android projects, having the SDK set up is necessary. As far as I can tell, if `./gradlew properties` can be executed successfully, the dependency fetcher should work. In this PR, I already have a warning for this case which makes it clear what the problem is. An option to fix this automatically could be to use something like https://github.com/quittle/gradle-setup-android-sdk, but we would need to do so in a way that is secure and robust and doesn't run into licensing issues. That's likely to be a larger effort and is out of the scope of this PR.

johannescoetzee:
It's also worth noting that there are expectations differences in the kotlin projects, but with the large joern version jump and the fact that I'm running these locally, it's hard to say what actually causes this.

Issue #4796, opened 2024-07-24
[Bug] Type loss/erasure in domain classes severely hurts interoperability


tajobe:
**Describe the bug**
When invoking joern as a library (like in joernio/standalone-ext), the loss of type information across many domain classes breaks JVM-langauge interop, particularly with the switch to the EMT to provide property/traversal implicits.

For example, the generated `Method` node is compiled as (from `javap`, removing the fully qualified names for readability):
```
public class Method extends StoredNode implements MethodBase, CfgNode, Declaration
```

Because it doesn't have `StaticType[MethodEMT]` when compiled, it lacks all of those traits like `HasFilenameEMT` or `HasFullNameEMT`. This means there's no way for a `Method` node to satisfy the type constraints on traversals like
```
public final <NodeType extends StoredNode & StaticType<HasFullNameEMT>> Iterator<String> fullName$extension(Iterator);
```

---

There are other losses of type information, not exclusive to the EMTs.

For example, `AstNode` becomes
```
public interface AstNode extends AstNodeBase
```
...lacking `StoredNode`, and thus lacking the `.file` implicit/extension.

Also, there are several cases where traversals get elevated from `Iterator<Int>` or `Iterator<Boolean>` to `Iterator<Object>`:
```
  public Iterator<Object> offsetEnd(); // should return Iterator<int>...
  public Iterator<NodeType> offsetEnd(int); // param correctly an int
  public Iterator<NodeType> offsetEnd(Seq<Object>); // param is a seq of Objects?
```

**To Reproduce**
1. Look at classes in `codepropertygraph-domain-classes_3-1.7.1.jar` or generate classes using the flatgraph domain-classes-generator and compile
2. Use `javap` or an IDE which decompiles the classes to check signatures/types

**Expected behavior**
Enough information is available in the compiled classes to be able to reasonably use the nodes/traversals/etc

**Screenshots**
N/A

**Desktop (please complete the following information):**
Should be irrelevant here but:
 - OS: macOS 14.5
 - Joern Version: 4.0.2, 4.0.8
 - Java version: 17
 - Scala version: 3.4.2 / 3.5.1-RC1

**Additional context**

bbrehm:
That is a good observation: The new way the APIs are set up suck when used from java.

To give a little context on _why_ we set it up this way:

[...]

PR #4936, opened 2024-09-18
ExtendedCfgNodeMethods: rename type parameter

xavierpinho:
@tajobe has been doing some byte code surgery in order to create a Kotlin interoperability layer and found an interesting edge case when type parameters are shadowed.

---

Taking a simpler example, consider the following Scala code:

```scala
class Example[NodeType <: Object](node: NodeType) extends AnyVal{
  def foo[NodeType](x: List[NodeType]): List[NodeType] = x
}
```

When decompiling the `foo` method, we'll find the following:

```
public static <NodeType extends java.lang.Object, NodeType extends java.lang.Object> scala.collection.immutable.List<NodeType> foo$extension(java.lang.Object, scala.collection.immutable.List<NodeType>);
    descriptor: (Ljava/lang/Object;Lscala/collection/immutable/List;)Lscala/collection/immutable/List;
    flags: (0x0009) ACC_PUBLIC, ACC_STATIC
    Code:
      stack=3, locals=2, args_size=2
         0: getstatic     #18                 // Field Example$.MODULE$:LExample$;
         3: aload_0
         4: aload_1
         5: invokevirtual #25                 // Method Example$.foo$extension:(Ljava/lang/Object;Lscala/collection/immutable/List;)Lscala/collection/immutable/List;
         8: areturn
    Signature: #23                          // <NodeType:Ljava/lang/Object;NodeType:Ljava/lang/Object;>(Ljava/lang/Object;Lscala/collection/immutable/List<TNodeType;>;)Lscala/collection/immutable/List<TNodeType;>;
```

&nbsp;
Notice how `NodeType` (the type parameter) occurs twice in the definition of `foo$extension`. Apparently the JVM (assuming Oracle as canonical) is designed to accept this when loading/linking, but guarantees nothing if the byte code is interpreted through reflection/etc, cf. https://docs.oracle.com/javase/specs/jvms/se23/html/jvms-4.html#jvms-4.7.9:

> Oracle's Java Virtual Machine implementation does not check the well-formedness of Signature attributes during class loading or linking. Instead, Signature attributes are checked by methods of the Java SE Platform class libraries which expose generic signatures of classes, interfaces, constructors, methods, and fields. Examples include getGenericSuperclass in Class and toGenericString in java.lang.reflect.Executable.

This byte code poses a problem that is trivially fixed if we prevent shadowing by just renaming one of the type parameters, e.g.

```scala
class Example[CfgNodeType <: Object](node: CfgNodeType) extends AnyVal{
  def foo[NodeType](x: List[NodeType]): List[NodeType] = x
}
```

resulting in

```
public static <NodeType extends java.lang.Object, CfgNodeType extends java.lang.Object> scala.collection.immutable.List<NodeType> foo$extension(java.lang.Object, scala.collection.immutable.List<NodeType>);
    descriptor: (Ljava/lang/Object;Lscala/collection/immutable/List;)Lscala/collection/immutable/List;
    flags: (0x0009) ACC_PUBLIC, ACC_STATIC
    Code:
      stack=3, locals=2, args_size=2
         0: getstatic     #18                 // Field Example$.MODULE$:LExample$;
         3: aload_0
         4: aload_1
         5: invokevirtual #25                 // Method Example$.foo$extension:(Ljava/lang/Object;Lscala/collection/immutable/List;)Lscala/collection/immutable/List;
         8: areturn
    Signature: #23                          // <NodeType:Ljava/lang/Object;CfgNodeType:Ljava/lang/Object;>(Ljava/lang/Object;Lscala/collection/immutable/List<TNodeType;>;)Lscala/collection/immutable/List<TNodeType;>;
```

---

Compiling Joern with `-Wshadow:type-parameter-shadow` only warns about 2 sites: the one in this PR and another one buried in javasrc2cpg. The latter is not public-facing, so I won't even bother proposing to change it at this point.

This rename is purely syntactical and seems to easily solve a bigger hurdle. Wdyt?

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/nodemethods/ExtendedCfgNodeMethods.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/nodemethods/ExtendedCfgNodeMethods.scala
@@ -10,7 +10,7 @@ import io.shiftleft.semanticcpg.language.*
 import scala.collection.mutable
 import scala.jdk.CollectionConverters.*
 
-class ExtendedCfgNodeMethods[NodeType <: CfgNode](val node: NodeType) extends AnyVal {
+class ExtendedCfgNodeMethods[CfgNodeType <: CfgNode](val node: CfgNodeType) extends AnyVal {
 
   /** Convert to nearest AST node
     */


mpollmeier (approved):
I absolutely think we should merge this and also enable the type param shadowing warning by default (will create a separate PR shortly), for the sake of readability and disambiguity. 

That being said, I'm not sure I understand which specific problem this PR solves, maybe you can elaborate..?

From my POV, the two type parameters in your Example class are not related at all. They happen to have the same name, which only means that within `def foo` we cannot reference the `NodeType` type from the `class Example`, but apart from that there's no connection at all between them. 

Semantically it's similar to something like this on the value level: the two `bar` variables are not related at all, they just happen to have the same name, and therefor we cannot reference the class-level `bar` variable from within `def baz`. 
```scala
class Foo {
    val bar = 42

    def baz: Unit = {
      val bar = "123"
      ()
  }
}
```

xavierpinho:
Indeed, there's no ambiguity at Scala level: it's just another form of good ol' lexical scoping. It's only when it gets down to byte code that we find ourselves in a pickle, as we get 2 same-named type parameters in the same method, causing havoc with (I think) kotlin-reflect.

Briefly, and in the context of #4796, @tajobe has been working on a plugin to automatically generate ergonomic Kotlin bindings for the most relevant APIs. It starts by finding which methods to codegen for (based on their signature), and proceeds emitting a more ergonomic (Kotlin-wise) method definition for it. So for so good... until a _wild method signature with 2 same-named type parameters appeared_ and kotlin-reflect bailed out.

Changing the type parameter name was the trivial workaround. No one was expecting to find a method with 2 same-named type parameters. I couldn't find anything about this edge case in various JVM-related docs online: not sure if it's undefined behaviour or I just missed something. At any rate, if someone knows more about this, I'd love to hear!

mpollmeier:
Thanks for elaborating, yeah that all makes sense.

PR #4944, opened 2024-09-24
fix bug, importCpg goes wrong on windows,for this.config is null

1nhann:


johannescoetzee:
This is a good catch! It's not a windows or php specific issue and exists at least for `jssrc2cpg` as well. The issue is that there's a general assumption that the `generate` method will be called, so the config starts uninitialized and is never set if this is not the case. 

`JavaSrcCpgGenerator` gets around this with
```
if (enableTypeRecovery)
  javasrc2cpg.typeRecoveryPasses(cpg, typeRecoveryConfig).foreach(_.createAndApply())
```

but runs into the issue that `enableTypeRecovery` is set in the `generate` method, so it will always be `false` when only using `importCpg`.

As far as I can tell, there's also no way to provide frontend config to `importCpg`, so this error condition would always be hit. I could be missing something here, however.

PR #4951, opened 2024-09-25
[php] fix php-parser path lookup

mpollmeier:
On buildbot the installation path for php2cpg is
`/worker/sptestV2-php2cpg/build/x2cpg-internal/php2cpg/target/universal/stage`
which (prior to this PR) leads to an invalid derived php-parser name
and the following error:
```
2024-09-25 09:30:08.623 ERROR Invalid path for PhpParserBin: /worker/sptestV2-/php2cpg/bin/php-parser/php-parser.php
```

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
@@ -141,7 +141,7 @@ object PhpParser {
 
   private def defaultPhpParserBin: String = {
     val dir      = Paths.get(this.getClass.getProtectionDomain.getCodeSource.getLocation.toURI).toAbsolutePath.toString
-    val fixedDir = new java.io.File(dir.substring(0, dir.indexOf("php2cpg"))).toString
+    val fixedDir = new java.io.File(dir.substring(0, dir.lastIndexOf("php2cpg"))).toString
     Paths.get(fixedDir, "php2cpg", "bin", "php-parser", "php-parser.php").toAbsolutePath.toString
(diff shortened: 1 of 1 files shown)


mpollmeier:
Hmm, this hotfix doesn't really work either:
```
2024-09-25 12:45:04.996 ERROR PhpParser$: Invalid path for PhpParserBin: /home/runner/work/joern/joern/joern-cli/target/universal/stage/frontends/php2cpg/lib/io.joern./php2cpg/bin/php-parser/php-parser.php
```

Any ideas for a less-flakey way to handle this? If we use the second-to-last "php2cpg" then it works for the staged artifact, but other setups would fail again, e.g. a fat jar, or running it from within sbt or so...
Also note that we want to be able to execute php2cpg independent of the current working directory.

Or we leave it as is and configure the path override when we run it in buildbot. Still not ideal, because downstream users may run into the same issue and have a hard time debugging this...

johannescoetzee:
I'd guess whatever we do for `astgen` would be best. It looks like a generic AstGenRunner exists in x2cpg

mpollmeier:
Here's the part you mean. I guess that'll be the way forward then - a bit fiddly, but can't come up with a better way either...
https://github.com/joernio/joern/blob/cc7f787459d5591cefafee003ac740e810f78fce/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/astgen/AstGenRunner.scala#L49-L63

johannescoetzee:
> a bit fiddly, but can't come up with a better way either...

Yeah, it's not ideal, but it would be consistent, at least...

mpollmeier:
superseded by https://github.com/joernio/joern/pull/4956

PR #4952, opened 2024-09-25
[dataflowengineoss] remove overriding operator semantics

xavierpinho:
Conflicting (i.e. same methodFullNamed) `FlowSemantics` are not currently handled by `FullNameSemantics.fromList` and silently it's the last entry that takes precedence. (Will create another PR for this shortly, that combines two such `FlowSemantics` into a single one.)

Take for example `<operators>.assignmentModulo` (as in `x%=y`). We have 2 conflicting entries for it in `DefaultSemantics`:
* `F(Operators.assignmentModulo, List((2, 1), (1, 1), (2, -1)))`
* `F("<operators>.assignmentModulo", List((2, 1), (1, 1)))`

Coincidentally, `Operators.assignmentModulo` is `"<operators>.assignmentModulo"`, so it's the latter `FlowSemantic` that ends up being used, thus preventing the `FlowMapping(2,-1)` from being considered. As such, in an expression such as `call(x%=2)`, the literal `2` doesn't taint `call`.

The same happens for the remaining operators being removed in this PR.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/DefaultSemantics.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/DefaultSemantics.scala
@@ -63,7 +63,6 @@ object DefaultSemantics {
     //  some of those operators have duplicate mappings due to a typo
     //  - see https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1630
 
-    F("<operators>.assignmentExponentiation", List((2, 1), (1, 1))),
     F("<operators>.assignmentShiftLeft", List((2, 1), (1, 1))),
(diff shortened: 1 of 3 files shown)


mpollmeier (approved):

maltek:
I think the root of these dupes, and that comment above them, is that some frontends produce `<operators>.assignmentModulo` and some `<operator>.assignmentModulo`. (Look for the `s` before `>`). But the constant in `Operators` indeed has the `s`, so the entries as they are now don't help at all with that problem.

DavidBakerEffendi (approved):
Great catch, it would perhaps be valuable to have some kind of warning, similarly to how `reachableByFlows` warns about empty `src`/`sink` to avoid scratching one's head for too long

PR #4955, opened 2024-09-26
[dataflowengineoss] SemanticTestCpg doesn't rebuild semantics

xavierpinho:
As soon as `FullNameSemantics` started logging duplicate semantics (in #4953), it was observed that `SemanticTestCpg` was passing a fresh Semantics instance to `EngineContext` instead of passing the one used by `OssDataFlow`, cf. `SemanticTestCpg.scala:40`. By coincidence they were the same, so everything was good except for some duplicate work. Nevertheless, this patch guarantees they are the same, by exposing the semantics used by `OssDataFlow`.


DavidBakerEffendi (approved):

mpollmeier (approved):

xavierpinho:
Seems like the ubuntu check got stuck. Restarting it.

PR #4957, opened 2024-09-26
[dataflowengineoss] replace `extraFlows` with `semantics`

xavierpinho:
Continuing the abstraction work from `FullNameSemantics` towards an arbitrary `Semantics`, this PR refactors the test fixtures and `OssDataFlow` layer by replacing the traditional `extraFlows: List[FlowSemantic]` by a more general `semantics: Semantics`. Whereas previously we'd just pass the `FlowSemantic` fragment that we'd like to append to `DefaultSemantics`, the caller is now responsible for building it explicitly, which right now is as simple as `DefaultSemantics().plus(<fragment to append>)`.


mpollmeier (approved):
comes together nicely :)

PR #4974, opened 2024-09-27
[dataflowengineoss] composable semantics

xavierpinho:
* Allows one to compose arbitrary `Semantics` at runtime, cf. the newly introduced `Semantics.after`, with `NoSemantics` as the identity.
* Identifies `NilSemantics`, which under the current engine seems to work as "negative semantics", removing any potential flows that would otherwise be created. I can't say I have it fully covered yet, but fwiw that's my interpretation at the moment.
* Introduces `NoCrossTaintSemantics`, which can be built both extensionally/intensionally, allowing one to, for instance, specify that an entire external library should have this kind of semantics.

`pysrc2cpg` was the frontend I used for testing them, but only because that's the frontend I was playing with. There's no other reason for not testing with other frontends.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/semanticsloader/Semantics.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/semanticsloader/Semantics.scala
@@ -2,22 +2,86 @@ package io.joern.dataflowengineoss.semanticsloader
 
 import io.shiftleft.codepropertygraph.generated.Cpg
 import io.shiftleft.codepropertygraph.generated.nodes.Method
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi (approved):
> Identifies NilSemantics, which under the current engine seems to work as "negative semantics", removing any potential flows that would otherwise be created. I can't say I have it fully covered yet, but fwiw that's my interpretation at the moment.

The current engine assumes that, if a semantic flow entry is present, any combination of arg flows not explicitly stated is "killed". So I believe this is the correct interpretation, and a good distinction against `NoSemantics`. If semantics/CALL edges are not present, then all flows are assumed present for that call.

PR #4975, opened 2024-09-27
Set PHP static initalizer offsets to the whole file

johannescoetzee:
This is a hack to fix an autofix issue. The `offset` and `offsetEnd` properties aren't set for the static init method since there aren't always sensible values to set these to, but since they weren't external, autofix was expecting this.

Some backend passes may be affected by this too


max-leuthaeuser:
I guess a valid alternative to this could be: set the same offsets as for the surrounding class. Hence, the whole class code would become the context.

max-leuthaeuser (approved):

max-leuthaeuser (changes requested):
I guess a valid alternative to this could be: set the same offsets as for the surrounding class. Hence, the whole class code would become the context. Which is too much in general but may help autofix as more contextual info there is always better.

johannescoetzee:
I've set the offsets to the entire file as requested

max-leuthaeuser (approved):

PR #4982, opened 2024-10-01
[ruby] Ignore "Throwaway" AST Structures

DavidBakerEffendi:
Sometimes AST trees would be thrown away, specifically when long chained expressions would be compressed into temporary cache variables. Sometimes the ASTs that are thrown away, have nodes directly written to the diff graph. This only happens with expressions, and closures typically write directly to the diff graph.

This PR minimizes what is written directly to the diff graph from function creation, and extends `x2cpg.Ast` to also include `CAPTURE` edges. This also introduces `closureToRefs` map to track which `Block` nodes already have live method ASTs in the diff graph.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
@@ -157,18 +157,18 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
   /** Attempts to extract a type from the base of a member call.
     */
   protected def typeFromCallTarget(baseNode: RubyExpression): Option[String] = {
-    scope.lookupVariable(baseNode.text) match {
-      // fixme: This should be under type recovery logic
-      case Some(decl: NewLocal) if decl.typeFullName != Defines.Any             => Option(decl.typeFullName)
-      case Some(decl: NewMethodParameterIn) if decl.typeFullName != Defines.Any => Option(decl.typeFullName)
-      case Some(decl: NewLocal) if decl.dynamicTypeHintFullName.nonEmpty => decl.dynamicTypeHintFullName.headOption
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -157,18 +157,18 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
-      // fixme: This should be under type recovery logic
-      case Some(decl: NewLocal) if decl.typeFullName != Defines.Any             => Option(decl.typeFullName)
-      case Some(decl: NewMethodParameterIn) if decl.typeFullName != Defines.Any => Option(decl.typeFullName)
-      case Some(decl: NewLocal) if decl.dynamicTypeHintFullName.nonEmpty => decl.dynamicTypeHintFullName.headOption
-      case Some(decl: NewMethodParameterIn) if decl.dynamicTypeHintFullName.nonEmpty =>
-        decl.dynamicTypeHintFullName.headOption
+    baseNode match {
+      case literal: LiteralExpr => Option(literal.typeFullName)
       case _ =>
-        astForExpression(baseNode).nodes
```
AndreiDreyer, line 168: Why have we removed this?

AndreiDreyer (approved):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
DavidBakerEffendi: It can be attributed to quite the performance slowdown for these long chained calls, especially as `base` and `receiver` are often the same long tree for these chained calls. I initially found it to be the cause for generating about 250 TypeRefs that were thrown away, so I think we can simply rely on non-nested nodes to give us the type hint.

DavidBakerEffendi:
Seems this may have introduced `Implicits$: iterator was expected to have exactly one element, but it actually has more`

PR #4985, opened 2024-10-02
[ruby] Re-implemented "Ignore "Throwaway" AST Structures (#4982)"

DavidBakerEffendi:
This correctly prevents re-use of nodes that are already being used elsewhere by ensuring deep copies.


AndreiDreyer (approved):

maltek:
@DavidBakerEffendi with this PR, some lambda references seem to get mixed up:


for this file https://github.com/OWASP/railsgoat/blob/c1e8ff1e3b24a1c48fcfc9fbee0f65dc296b49d9/config/routes.rb

When I query for the lambda referenced in the `:dashboard` section, I get the lambda from the `:pay` section instead:

```
ocular> cpg.call.codeExact("""resources :dashboard do
     |     collection do
     |       get "home"
     |       get "change_graph"
     |     end
     |   end""").argument(2).isTypeRef.typ.referencedTypeDecl.code.l
val res62: List[String] = List(
  """do
      collection do
        post "update_dd_info"
        post "decrypted_bank_acct_num"
      end
    end"""
)
```

DavidBakerEffendi:
@maltek alright will have to revisit how I match "live" lambdas to references without regenerating said lambdas and duplicating them

DavidBakerEffendi:
@maltek Found the issue, `RubyExpression` nodes don't inherently consider the `span` in the calculation of its hash, so when put into any hashed context, nodes that only rely on `span` collide. A fix to come

PR #4986, opened 2024-10-02
[ruby] Implement `hashCode` for `RubyExpression`

DavidBakerEffendi:
`RubyExpression` nodes don't inherently consider the `span` in the calculation of its hash, so when put into any hashed context, nodes that only rely on `span` alone will collide in these contexts.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/RubyIntermediateAst.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/RubyIntermediateAst.scala
@@ -34,6 +34,13 @@ object RubyIntermediateAst {
     def text: String = span.text
 
     override def hashCode(): Int = Objects.hash(span)
+
+    override def equals(obj: Any): Boolean = {
+      obj match {
+        case o: RubyExpression => o.span == span
+        case _                 => false
+      }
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/RubyIntermediateAst.scala
```diff
@@ -33,6 +32,8 @@ object RubyIntermediateAst {
     def offset: Option[(Int, Int)] = span.offset
 
     def text: String = span.text
+
+    override def hashCode(): Int = Objects.hash(span)
```
maltek, line 36: Don't we also have to override `equals` here to uphold the invariants of that interface?

Also a potentially dumb question since I don't have the full context: the `TextSpan` class always doesn't contain any information about which file it is referring to. So there might be 'equal' `TextSpan`s referring to similar spans in different files. So are the places where these objects are hashed only ever in the context of a single file?

DavidBakerEffendi: Thanks for the catch on `equals`, you're right. Yes, there is only ever 1 AstCreator & friends per file, so this is safe in that sense.

PR #4990, opened 2024-10-04
[jssrc2cpg] Pass JVM  max memory setting to astgen

max-leuthaeuser:
For: https://shiftleftinc.atlassian.net/browse/SEN-3282

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala
@@ -30,7 +30,7 @@ object AstGenRunner {
   }
 
   private def nodejsSpaceSize: Int = {
-    Math.max(getMaxHeapSizeInMB.toInt, 8192) // as default we set 8G minimum
+    Math.max((0.75 * getMaxHeapSizeInMB).toInt, 8192) // as default we set 8G minimum
   }
 
   private val NODE_OPTIONS: Map[String, String] = Map("NODE_OPTIONS" -> s"--max-old-space-size=$nodejsSpaceSize")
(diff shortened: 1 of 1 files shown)


maltek (approved):

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala
```diff
@@ -23,7 +23,17 @@ object AstGenRunner {
...
 
-  private val NODE_OPTIONS: Map[String, String] = Map("NODE_OPTIONS" -> "--max-old-space-size=8192")
+  private def getMaxHeapSizeInMB: Long = {
+    val maxHeapSizeInBytes = Runtime.getRuntime.maxMemory()
+    val maxHeapSizeInMB    = maxHeapSizeInBytes / (1024 * 1024)
+    maxHeapSizeInMB // nodejs expects MB
+  }
+
+  private def nodejsSpaceSize: Int = {
+    Math.max((0.75 * getMaxHeapSizeInMB).toInt, 8192) // as default we set 8G minimum
```
mpollmeier, line 33: I understand that this isn't really a change, but I'm still wondering if setting a (rather high) minimum is the right idea, given that it may exceed the free memory, and/or even the physically available memory.

mpollmeier (approved):

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala
perrito666: I would assume this memory will be long freed when it needs to be reclaimed by java again. This is done under the assumption the end user will set up java mem thinking that this is all it takes for the cpg process

max-leuthaeuser:
With some local testing it turned out that the nodejs binary astgen we build using pkg does not accept any settings from env variables since this change here: https://github.com/vercel/pkg-fetch/commit/7f3403bd71d244850ef30a711c52e64f905e9423

PR #4998, opened 2024-10-11
[kotlin2cpg] Fix assertion error by locking descriptorRenderer options

tajobe:
DescriptorRendererImpl [asserts that the options are locked upon init](https://github.com/JetBrains/kotlin/blob/master/core/descriptors/src/org/jetbrains/kotlin/renderer/DescriptorRendererImpl.kt#L36). This means that the type rendering will fail anywhere with assertions enabled (`-ea` JVM option) as the options weren't being locked after setup.

In order to test this, I added `-ea` to the test javaOptions (which requires `Test/fork` to be `true`) and things fail. With the addition of `opts.lock()`, they pass. I'm not sure if there's a reason the tests were explicitly set to not fork or if there's a desire to keep the javaOpts "clean", but this was the easiest way to illustrate the error and fix.

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/build.sbt
+++ b/joern-cli/frontends/kotlin2cpg/build.sbt
@@ -23,4 +23,5 @@ libraryDependencies ++= Seq(
 
 enablePlugins(JavaAppPackaging, LauncherJarPlugin)
 trapExit    := false
-Test / fork := false
+Test / fork := true
+Test / javaOptions ++= Seq("-ea")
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi (approved):
Looks good on my end, if this causes issues downstream for others we can address it then

mpollmeier:
Sorry I missed this somehow. 
Things to look out for when changing the `Test/fork` setting is IDE support - in other words: did anyone check if the tests still run from within e.g. Intellij?

DavidBakerEffendi:
@mpollmeier let me give that a check quick

DavidBakerEffendi:
Running a test suite for a class via IDE seems good for `querying`, `io`, `dataflow` package tests

PR #5000, opened 2024-10-14
[dataflowengineoss] take into account frontends that use the `-1` index for named arguments

xavierpinho:
`-1` is used to denote a call's return value. However, some frontends (e.g. python) also use this index for named arguments. When specifying `FlowSemantic`s, the criteria for deciding if a `FlowSemantic` taints the return value was not taking this fact into account. This makes sure that `-1` only denotes the return value if it's not in the context of a named argument specification.


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/passes/reachingdef/EdgeValidator.scala
```diff
@@ -58,8 +58,10 @@ object EdgeValidator {
     flowSemantic.mappings.exists(explicitlyFlowsToReturnValue)
 
   private def explicitlyFlowsToReturnValue(flowPath: FlowPath): Boolean = flowPath match {
-    case FlowMapping(_, ParameterNode(dst, _)) => dst == -1
-    case PassThroughMapping                    => true
-    case _                                     => false
+    // Some frontends (e.g. python) denote named arguments using `-1` as the argument index. As such
+    // `-1` denotes the return value only if there's no argument name.
+    case FlowMapping(_, ParameterNode(-1, None)) => true
```
DavidBakerEffendi, line 63: Ah good catch. Yes `argumentIndex = -1` is the default value, it is "unset" but in this instance it'll clash with the return value logic

DavidBakerEffendi (approved):

PR #5006, opened 2024-10-17
[ruby_ast_gen] Handling for Singleton & Anon Classes 

DavidBakerEffendi:
Additionally, added handling for range operators and lowered hash arguments in calls to named arguments to be consistent with current ANTLR interpretation of these args.

Also, I noticed the runner may fail, so I added a retry mechanism.


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/astgen/AstGenRunner.scala
```diff
@@ -181,9 +181,13 @@ trait AstGenRunnerBase(config: X2CpgConfig[?] & AstGenConfig[?]) {
         val parsed  = filterFiles(srcFiles, out)
         val skipped = skippedFiles(in, result.toList)
         DefaultAstGenRunnerResult(parsed, skipped)
+      case Failure(f) if !isRetry =>
+        logger.error(s"\t- running ${metaData.name} failed! Retrying...", f)
+        execute(out, true)
```
max-leuthaeuser, line 186: This could run endlessly. We should have at least some max. number of runs value as param.

AndreiDreyer (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/astgen/AstGenRunner.scala
```diff
@@ -181,9 +181,13 @@ trait AstGenRunnerBase(config: X2CpgConfig[?] & AstGenConfig[?]) {
         val parsed  = filterFiles(srcFiles, out)
         val skipped = skippedFiles(in, result.toList)
         DefaultAstGenRunnerResult(parsed, skipped)
+      case Failure(f) if !isRetry =>
```
max-leuthaeuser, line 184: I do not think is a good idea. Everytime some astgen failing this will retry the invocation. In case of very large project scans failing at the very end (maybe even failing for a good reason) this may lead to very long runtimes.

max-leuthaeuser (changes requested):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/astgen/AstGenRunner.scala
DavidBakerEffendi: Ok will remove

PR #5016, opened 2024-10-21
ExternalCommand: more information in error conditions

mpollmeier:
* report exit code if it's non-zero
* pass on original error (if any) rather than disregarding it
* log.warn stderr output (if any)
* add tests


max-leuthaeuser (approved):
Not sure if the tests work on the Windows runner here.

PR #5032, opened 2024-10-28
add "isInline" method for "CallMethods"

canliture:


maltek:
thanks a lot!

Could you add a method for `Iterator[Call]` as well please? Like this one for `isStatic`: https://github.com/joernio/joern/blob/d9be1834841ad756e73b3e81f700996e097250d7/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/expressions/CallTraversal.scala#L12-L13

maltek (approved):

PR #5036, opened 2024-10-30
[javasrc2cpg] Generate TYPE_REF for implicit base of static field accesses.

ml86:
For static field access like `String x = staticField1;` we used to
generate a IDENTIFIER with the name of the class declaring the static
field as base for the field access operator.
Now we generate a TYPE_REF referencing the declaring type. This better
represents the actual situation where there is no local variable for
a class/type which in practice also resulted in a CPG format error
because the IDENTIFIFER node was at least not in all cases linked to a
LOCAL.
While implementing this i reorganized the code a bit to avoid
duplication.

Also changes the `equals` semantic for TrackedTypeRef. Before it was
based on the actual TYPE_REF node which does not make sense. Instead we
need to compare the referenced TYPE itself to check if two different
type references in code point to the same TYPE.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/expressions/AstForNameExpressionsCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/expressions/AstForNameExpressionsCreator.scala
@@ -1,12 +1,20 @@
 package io.joern.javasrc2cpg.astcreation.expressions
 
+import com.github.javaparser.ast.Node
 import com.github.javaparser.ast.expr.NameExpr
 import com.github.javaparser.resolution.declarations.ResolvedFieldDeclaration
 import io.joern.javasrc2cpg.astcreation.{AstCreator, ExpectedType}
+import io.joern.javasrc2cpg.scope.JavaScopeElement.TypeDeclScope
 import io.joern.javasrc2cpg.typesolvers.TypeInfoCalculator.TypeConstants
 import io.joern.javasrc2cpg.util.NameConstants
 import io.joern.x2cpg.{Ast, Defines}
-import io.shiftleft.codepropertygraph.generated.nodes.{NewLocal, NewMethodParameterIn}
+import io.shiftleft.codepropertygraph.generated.nodes.{
+  NewLocal,
+  NewMethodParameterIn,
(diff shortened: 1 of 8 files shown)


ml86:
@johannescoetzee Can you pleaes double check if `scope.enclosingTypeDecl.get` is ok. To me it looked like this is never empty. Please also check if the remove `TODO` would still apply.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/expressions/AstForNameExpressionsCreator.scala
```diff
@@ -65,28 +68,45 @@ trait AstForNameExpressionsCreator { this: AstCreator =>
...
+    )
+  }
+
   private def astForStaticImportOrUnknown(nameExpr: NameExpr, name: String, typeFullName: Option[String]): Ast = {
     tryWithSafeStackOverflow(nameExpr.resolve()) match {
       case Success(value) if value.isField =>
-        val identifierName = if (value.asField.isStatic) {
-          // TODO: v is wrong. Statically imported expressions can also be represented by just the name.
-          // A static field represented by a NameExpr must belong to the class in which it's used. Static fields
-          // from other classes are represented by a FieldAccessExpr instead.
```
johannescoetzee, line 74: This was very confusingly written. It's basically 2 comments in one. The original comment was
>           // A static field represented by a NameExpr must belong to the class in which it's used. Static fields
>           // from other classes are represented by a FieldAccessExpr instead.

which is wrong. Instead of removing this comment and replacing it with a todo, I decided (for some reason which I no longer remember or understand) to add the TODO

>  // TODO: v is wrong. Statically imported expressions can also be represented by just the name.

The idea I was trying to communicate with the TODO still applies, I think. My original assumption was that a name expression without  an explicit scope must refer to a field declared in the same class as the name expression, but that's not true because of static imports, as in the example below. We would actually need to check whether the name has been imported statically. I remember running into some issue when looking at this before, but not what that issue was. 

```
import static foo.Foo.FOO;

public class Bar {
  public static void main(String[] args) {
    System.out.println(FOO);
  }
}
```

I don't think fixing that should be a part of this PR, but keeping a TODO would be good, I think.

johannescoetzee (approved):

Issue #5037, opened 2024-10-30
[Bug] [rubysrc2cpg] some METHOD nodes are missing their mandatory AST edge to a BLOCK node


mpollmeier:
METHOD nodes must have an AST edge to a BLOCK node, according to the cpg schema. 

However sometimes that's always the case for rubysrc, which unfortunately currently breaks our [internal integration](https://buildbot.global.sltr.io/#/builders/35/builds/42) tests with rubysrc:
```
Exception in thread "main" flatgraph.SchemaViolationException: OUT edge with label AST to an adjacent BLOCK is mandatory, but not defined for this METHOD node with seq=190
	at io.shiftleft.codepropertygraph.generated.neighboraccessors.AccessNeighborsForMethod$.block$extension(Method.scala:365)
```

Here's a way to reproduce one such case:
```
git clone https://github.com/OWASP/railsgoat.git
joern-cli/frontends/rubysrc2cpg/rubysrc2cpg.sh path/to/railsgoat

./joern

importCpg("cpg.bin")
val method = cpg.method.fullName("app/models/benefits.rb:<main>.Benefits.silence_streams").head
method.outE("AST").map(_.dst.label).toSet
// no BLOCK neighbor...
val res18: Set[String] = HashSet(
  "METHOD_PARAMETER_IN",
  "METHOD_PARAMETER_OUT",
  "METHOD",
  "TYPE_DECL",
  "MODIFIER",
  "CONTROL_STRUCTURE",
  "METHOD_RETURN"
)
```

DavidBakerEffendi:
Good news that the new parser under `dave/ruby/ruby_ast_gen` does not display this behavior, I am aiming to get it into master this week after testing on `codescience`

mpollmeier:
fixed via https://github.com/joernio/joern/pull/5040
thank you!

PR #5040, opened 2024-10-30
[ruby] New `parser` Gem Parser Implementation

DavidBakerEffendi:
Due to the performance and stability issues with a 2-phase LLK parser approach with ANTLR, this branch, developed over the past two weeks, introduces the usage of a [`parser`](https://github.com/whitequark/parser/tree/master) based approach wrapped in [`ruby_ast_gen`](https://github.com/joernio/ruby_ast_gen). This Gem is used by many popular Ruby linters and static analysis tools, and is compatible with Ruby versions across 1-3. 

We bundle `ruby_ast_gen` with the frontend under `resources`, and execute the Ruby code using JRuby to maximize portability. ~This has a bit of an overhead for short-lived processes, so we have a disabled-by-default `--try-local-ruby` frontend arg that will attempt to detect and use a compatible native Ruby binary. This is enabled for tests to have them run faster.~ If the resource is detected to be in a JAR, `ruby_ast_gen` is written to a temporary directory, from where it is then executed.

Both JRuby and native Ruby runs of `ruby_ast_gen` far outperform the old ANTLR version, which would often have worst-case performance for some set of language constructs encountered.

Files worth reviewing if needed:
* `joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/*`
* `joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala`
* `joern-cli/frontends/rubysrc2cpg/build.sbt`

Changes include:
* Remove ANTLR implementation
* Execute `ruby_ast_gen` and parse JSON output 
* Map the JSON output to the same intermediate AST nodes that ANTLR used to avoid changing `AstCreation` logic
* Rename the method name for executable strings (`%x{}`-style strings) to `<operator>.backticks` instead of `exec` as `exec` technically has different semantics
* Trim `.code` properties to 60 characters to avoid large `.code` properties. (this is done by `ruby_ast_gen`)
* Added `db/migration` to default ignored directories

Tests are largely kept the same, and `ruby:test` is green on `codescience` after some small `.code` and `.lineNumber` tweaks. See `dave/ruby/prep-for-ruby-parse-gem`.

The `astGenResourceTask` task in `build.sbt` is tied to `Compile / resourceGenerators` as the task fetches the zipped Ruby code and places it under `resources` to be packaged alongside the classes in the JAR.

Co-authored by @AndreiDreyer


mpollmeier (commented):
I'd prefer if we handled the astgen like we do it in the other frontends, i.e. do not add those 17M binaries to the repo, and instead download them on demand by invoking the astGenDlTask. 
Most frontends invoke it before `compile`:
```
Compile / compile := ((Compile / compile) dependsOn astGenDlTask).value
```

DavidBakerEffendi:
> I'd prefer if we handled the astgen like we do it in the other frontends, i.e. do not add those 17M binaries to the repo, and instead download them on demand by invoking the astGenDlTask. Most frontends invoke it before `compile`:
> 
> ```
> Compile / compile := ((Compile / compile) dependsOn astGenDlTask).value
> ```

Bundling this in the resources directory was part of a discussion with Markus, as it would be easier to distribute, and running it with JRuby allows us to be platform independent (so no need to select specific binaries). However, is what you mean that we can still bundle it, but exclude it from Git and download this on compile time?

max-leuthaeuser:
Putting undiffable binaries in the repo will get us locked again on traffic until Fabs pulls out his credit card. We had this issue already.

mpollmeier:
> However, is what you mean that we can still bundle it, but exclude it from Git and download this on compile time?

By 'bundle' you probably refer something like `Universal/packageBin` or `stage`, right? You simply want to make it part of the result, rather than giving the user instructions to download and link it. 
So yes, you'll get all of that by just downloading it on demand, there's no need to add those files to the repo afaics.

DavidBakerEffendi:
> > However, is what you mean that we can still bundle it, but exclude it from Git and download this on compile time?
> 
> By 'bundle' you probably refer something like `Universal/packageBin` or `stage`, right? You simply want to make it part of the result, rather than giving the user instructions to download and link it. So yes, you'll get all of that by just downloading it on demand, there's no need to add those files to the repo afaics.

Rather, I mean under `src/main/resources`, none of `ruby_ast_gen` or its vendored dependencies are binaries.

mpollmeier:
Right, sorry, yes those are not binaries. But still they take up 17M and it's really not necessary to have them as part of this repo since we merely download them from a third party. So unless there's a benefit to having them as part of this repo I'd like to use the same pattern as for the other astgens and download them on compiler, rather than on git clone.

DavidBakerEffendi:
Ok I've moved to not checking them in, I agree it is a lot of strain on the git history and object in general.

For now, it's still going into `resources`, so the end-user doesn't need to expect additional downloads. It is about 14MB uncompressed (1.7 MB zip compressed) in `resources`, which may be a plausible argument to simply have it as the rest of the astGen programs instead of in the JAR.

mpollmeier:
Thanks! One point I forgot to mention: it would have been 17M for each new version of the astgen, being preserved unnecessarily in the git history forever. 

Looks good from my POV and I can approve it. Not sure if anyone else can/should have a review of the frontend code itself. Given that you worked together with @AndreiDreyer it might already be "reviewed-enough".

mpollmeier (approved):
there's a couple of tests still marked as `ignore`, is that intended?

DavidBakerEffendi:
> there's a couple of tests still marked as `ignore`, is that intended?

I think I'm going to double-check that they are not "newly" ignored - i.e., there were ignored ones from the last parser. Andrei and I are also meeting with Markus tomorrow to check if he is happy, so an approval won't necessarily mean a merge.

Also we are looking into the failing Window tests, seems to be our blocker right now.

maltek:
We're seeing regressions in [our integration tests](https://buildbot.global.sltr.io/#/builders/44/builds/50) due to changes around implicit field accesses. Until now, something like `params` would just be turned into a field access `self.params`. Now it's turned into a dynamic call `params` with receiver `self.params` and argument 0  `params`. This might be more correct on some level, but it's Kryptonite for Qwiet's dataflow engine.

DavidBakerEffendi:
@maltek I'm seeing it too on `ruby:test`, there was initially some logic to "guess" what certain identifiers were on the AstCreator side as everything coming from ANTLR was an identifier, however the new parser applies the Ruby logic to automatically determine if something is a call or identifier, and has taken the "right to guess" away from AstCreator. 

I think it's best to put off this merge a bit longer until we see green on Qwiet's dataflow engine, I'll have a look at how we can get the old behavior back.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/passes/Defines.scala
```diff
@@ -31,11 +35,12 @@ object Defines {
   def getBuiltInType(typeInString: String) = s"${GlobalTypes.kernelPrefix}.$typeInString"
 
   object RubyOperators {
-    val hashInitializer = "<operator>.hashInitializer"
-    val association     = "<operator>.association"
-    val splat           = "<operator>.splat"
-    val regexpMatch     = "=~"
-    val regexpNotMatch  = "!~"
+    val backticks: String = "<operators>.backticks"
```
maltek, line 38:
```suggestion
    val backticks: String = "<operator>.backticks"
```

We're already inconsistent with `<operator>` vs. `<operators>` in some places. But we should go with `<operator>` whenever we don't need to worry about compatibility.

DavidBakerEffendi:
@maltek I've figured out how to get this old behaviour back, and my local `ruby:test` is green again (had to update a few flow paths back to how they used to be). Of course, will wait for build bot to be green before going ahead, which may be something to look into next week.

maltek:
@DavidBakerEffendi thank you! That fixed all those regressions, and our results are the same again as before for our test projects. (Depth-first order of a few source/sink nodes has changed, and we'll need to adjust the expected values of - but that's totally fine.)

DavidBakerEffendi:
@maltek In that case, I'll merge here, so we can get the backlog of stability and speed fixes that this brings in sooner rather than later.

.github/workflows/pr.yml
```diff
@@ -21,10 +21,6 @@ jobs:
       - name: Install php
         if: matrix.os == 'macos-latest'
         run: brew install php
-      - name: Set up Ruby
```
max-leuthaeuser, line 24: This could be removed in the other GH action workflows as well right?

DavidBakerEffendi: Yes, good point

mpollmeier:
Looking good, the schema violation is gone (https://github.com/joernio/joern/issues/5037)
Just a small expectation bump and we'll be green again: https://github.com/ShiftLeftSecurity/codescience/pull/8035

Issue #4994, opened 2024-10-08
[gosrc2cpg] Missing function calls inside if statement


l1006986533:
**Describe the bug**
When I analyze a Go file with Joern, function calls inside an if statement are not correctly identified and listed.

**To Reproduce**
1. Create a go file
```
package main
import "fmt"

func myfunc() *string{
	var s string = "hello world"
	return &s
}

func main() {
	if s := myfunc(); s != nil {
    	fmt.Println(*s)
	}
}
```
2. import and output all function calls by `cpg.call.l`
```
joern> importCode(inputPath="a.go", projectName="a-go")

Using generator for language: GOLANG: GoCpgGenerator
Creating project `a-go` for code at `a.go`
Project with name a-go already exists - overwriting
=======================================================================================================
Invoking CPG generator in a separate process. Note that the new process will consume additional memory.
If you are importing a large codebase (and/or running into memory issues), please try the following:
1) exit joern
2) invoke the frontend: /opt/joern/joern-cli/gosrc2cpg a.go -o /Users/xxx/xxx/workspace/a-go/cpg.bin.zip
3) start joern, import the cpg: `importCpg("path/to/cpg")`
=======================================================================================================

moving cpg.bin.zip to cpg.bin because it is already a database file
Creating working copy of CPG to be safe
Loading base CPG from: /Users/xxx/xxx/workspace/a-go/cpg.bin.tmp
Code successfully imported. You can now query it using `cpg`.
For an overview of all imported code, type `workspace`.
Adding default overlays to base CPG
The graph has been modified. You may want to use the `save` command to persist changes to disk.  All changes will also be saved collectively on exit
The graph has been modified. You may want to use the `save` command to persist changes to disk.  All changes will also be saved collectively on exit
val res0: io.shiftleft.codepropertygraph.generated.Cpg = Cpg[Graph[64 nodes]]
                                                                                                                                       
joern> cpg.call.l

val res1: List[io.shiftleft.codepropertygraph.generated.nodes.Call] = List(
  Call(
    argumentIndex = -1,
    argumentName = None,
    code = "s string = \"hello world\"",
    columnNumber = Some(value = 17),
    dispatchType = "STATIC_DISPATCH",
    dynamicTypeHintFullName = IndexedSeq(),
    lineNumber = Some(value = 5),
    methodFullName = "<operator>.assignment",
    name = "<operator>.assignment",
    order = 2,
    possibleTypes = IndexedSeq(),
    signature = "",
[...]

PR #5046, opened 2024-10-31
[gosrc2cpg] Fixed an issue with skipping the statement before the if condition

ch0a3:
I fix the issue seems like [gosrc2cpg] [#4994](https://github.com/joernio/joern/issues/4994)
when I handle the code like:
`if a := 1 (a < b)`
in  results of Json, the code `a := 1` was put in init block, and the method skip this block.

Resolves https://github.com/joernio/joern/issues/4994


DavidBakerEffendi (approved):

DavidBakerEffendi:
Thanks for the fix!

PR #5052, opened 2024-11-04
[kotlin2cpg] Some cleanup and refactoring

max-leuthaeuser:
Most changes are due to the removal of the implicitely passed instance of TypeInfoProvider which is totally unnecessary.


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/jar4import/UsesService.scala
```diff
@@ -4,15 +4,15 @@ import com.squareup.tools.maven.resolution.ArtifactResolver
...
-import java.net.{MalformedURLException, URL}
+import java.net.{MalformedURLException, URI}
 
 trait UsesService { this: Kotlin2Cpg =>
 
   private val logger = LoggerFactory.getLogger(getClass)
 
   protected def reachableServiceMaybe(serviceUrl: String): Option[Service] = {
     try {
-      val url            = new URL(serviceUrl)
```
max-leuthaeuser, line 15: URL constructor is deprecated.

mpollmeier (approved):
If TypeInfoProvider is totally unnecessary (I can't really say), then 👍🏻

max-leuthaeuser:
Passing it around as an implicit argument is unnecessary. Sorry, if I framed it a bit misleading.
It's now a simple class variable.

PR #5071, opened 2024-11-06
[kotlin2cpg] Dispose the environment as soon as it's not needed

xavierpinho:
We were observing some OOM exceptions when providing large .jars to kotlin2cpg's classpath. Profiling the memory usage, it was clear it kept growing between multiple same-process invocations.

Below are two such graphs. They were obtained by:
* including a ~350MB .jar in kotlin2cpg's test resources
* setting the flag `withDefaultJars=true` by default
* running the kotlin2cpg tests

Before:
Screenshot 2024-11-06 at 14 22 40

After:
Screenshot 2024-11-06 at 14 23 31


max-leuthaeuser (approved):

max-leuthaeuser:
Nice finding 👍

Issue #5049, opened 2024-11-01
[Bug] Flow lost for member access between 4.0.127 and 4.0.128


xavierpinho:
Consider the following test case (for `javasrc2cpg`):

```scala
class FlowFromMemberToMethod extends JavaDataflowFixture {
  override val code =
    """
      |import org.hibernate.cfg.Configuration;
      |public class Main {
      |    private static String HIBERNATE_URL_PROP = "hibernate.connection.url";
      |    public static void Main(String[] args) {
      |        Configuration cfg = new Configuration()
      |                .configure()
      |                .setProperty(HIBERNATE_URL_PROP, "jdbc:mysql://localhost/Test");
      |    }
      |}
      |""".stripMargin
  it should "find flow from field to its usage" in {
    val sink = cpg.call("setProperty").argument(1).l
    val source = cpg.literal("\"hibernate.connection.url\"").l
    sink.reachableByFlows(source).map(flowToResultPairs).l shouldBe List(
      List(
        ("String Main.HIBERNATE_URL_PROP = \"hibernate.connection.url\"", Some(4)),
        ("configure().setProperty(HIBERNATE_URL_PROP, \"jdbc:mysql://localhost/Test\")", Some(8))))
  }
}
```

As of v.4.0.128, `reachableByFlows` is empty, whereas in v.4.0.127 it was passing.

PR #5075, opened 2024-11-07
[javasrc2cpg] add basic static field test

xavierpinho:
Apropos #5049, includes an ignored test that stopped working after #5036. Unfortunately, such a test wasn't there before to flag this regression.

Previously, in the example's `System.out.println(CONST)` we were lowering `CONST` as `Bar.CONST`, with `Bar` an IDENTIFIER. Presently, `Bar` is a TYPE_REF and we lose this dataflow. 

Another thing to point out regarding this design change is that it's not consistent throughout. For instance, if we had written `System.out.println(Bar.CONST)` , `Bar` is here not a TYPE_REF but again an IDENTIFIER.

I'm willing to contribute patch(es) to the recovery of this flow, but would need some guidance as I've not been able to find a workaround after some initial experiments. (Thanks in advance)


johannescoetzee (approved):
Good catch! I don't have much experience with the OSS dataflow engine so can't say much there, but I agree we should update the explicit `Bar.CONST` case as well. I'll add it to my TODO list for when I'm done with pattern matching support

PR #5076, opened 2024-11-07
[javasrc2cpg] support flows for field accesses with a TYPE_REF base node

### LINKED ISSUE #5049: [Bug] Flow lost for member access between 4.0.127 and 4.0.128

xavierpinho:
Resolves #5049 and fixes the ignored unit-test introduced in #5075

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
@@ -147,7 +147,12 @@ class SourceToStartingPointsInMethod(
   private def usageInOtherClasses(m: Method, usageInputs: List[UsageInput]): List[StartingPointWithSource] = {
     usageInputs.flatMap { case UsageInput(src, typeDecl, astNode) =>
       m.fieldAccess
-        .where(_.argument(1).isIdentifier.typeFullNameExact(typeDecl.fullName))
+        .where(
+          _.or(
+            _.argument(1).isIdentifier.typeFullNameExact(typeDecl.fullName),
+            _.argument(1).isTypeRef.typeFullNameExact(typeDecl.fullName)
+          )
+        )
         .where { x =>
           astNode match {
             case identifier: Identifier =>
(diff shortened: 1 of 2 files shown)


DavidBakerEffendi (approved):

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
```diff
@@ -147,7 +147,12 @@ class SourceToStartingPointsInMethod(
   private def usageInOtherClasses(m: Method, usageInputs: List[UsageInput]): List[StartingPointWithSource] = {
     usageInputs.flatMap { case UsageInput(src, typeDecl, astNode) =>
       m.fieldAccess
-        .where(_.argument(1).isIdentifier.typeFullNameExact(typeDecl.fullName))
+        .where(
+          _.or(
+            _.argument(1).isIdentifier.typeFullNameExact(typeDecl.fullName),
+            _.argument(1).isTypeRef.typeFullNameExact(typeDecl.fullName)
+          )
+        )
```
maltek, line 155: slight nit-pick: we can simplify `.where(_.or(...))` to just `.or(...)`
```suggestion
        .or(
          _.argument(1).isIdentifier.typeFullNameExact(typeDecl.fullName),
          _.argument(1).isTypeRef.typeFullNameExact(typeDecl.fullName)
        )
```

maltek (approved):

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/SourcesToStartingPoints.scala
xavierpinho: Thanks! Good to know!

PR #5079, opened 2024-11-08
[ruby] Field Call Target, Fixed Some NPE Exceptions

### LINKED ISSUE #5063: [Bug] Field Identifiers as Targets of Calls Unrecognized
**Describe the bug**
Seems the old parser may have lowered field identifiers before `AstCreator`, however these are making their way to `AstCreator` now and are not being accounted for.

```
2024-11-06 11:24:30.469 WARN AstCreator: Unrecognized target of call: __callee__ (lib/active_record/with_recursive.rb), skipping
2024-11-06 11:24:30.488 WARN AstCreator: Could not represent expression: __callee__ (ClassFieldIdentifier) (lib/active_record/with_recursive.rb), skipping
```

**To Reproduce**
1. `git clone git@github.com:mastodon/mastodon.git --depth 1`
2. `joern-parse mastodon`
3. See error

### LINKED ISSUE #5064: [Bug] Key not found `children` on `mastodon`
**Describe the bug**
During method creations, the arguments array appears to not have a children value at times. It would be worth finding out what is there, and to safely handle this error.

```scala
  private def visitMethodDefinition(obj: Obj): RubyExpression = {
    val name       = obj(ParserKeys.Name).str
    val parameters = obj(ParserKeys.Arguments).asInstanceOf[ujson.Obj].visitArray(ParserKeys.Children) // <-- here
    val body = obj
      .visitOption(ParserKeys.Body)
      .map {
        case x: StatementList => x
        case x                => StatementList(List(x))(x.span)
      }
      .getOrElse(StatementList(Nil)(obj.toTextSpan.spanStart("<empty>")))
    MethodDeclaration(name, parameters, body)(obj.toTextSpan)
  }
```

**To Reproduce**
1. `git clone git@github.com:mastodon/mastodon.git --depth 1`
2. `joern-parse mastodon`
3. See error

### LINKED ISSUE #5065: [Bug] `KwRestArg` value is sometimes not `ujson.Str`
**Describe the bug**
On `mastodon`, we see the `value` field not being a string, and this is not being handled gracefully.
```scala
// RubyJsonToNodeCreator
  private def visitKwRestArg(obj: Obj): RubyExpression = {
    val name = obj(ParserKeys.Value).str // <-- here
    HashParameter(name)(obj.toTextSpan)
  }
```

**To Reproduce**
1. `git clone git@github.com:mastodon/mastodon.git --depth 1`
2. `joern-parse mastodon`
3. See error

DavidBakerEffendi:
* Convert `FieldIdentifier` nodes to `MemberAccess` if they reach AstCreator for call targets
* Handle forwarded args which were a cause for some null pointer exceptions
* Fixed `kwrestarg` null value issue, which turned out to be a `**` parameter

Resolves #5063
Resolves #5064
Resolves #5065

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyJsonToNodeCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyJsonToNodeCreator.scala
@@ -478,11 +478,16 @@ class RubyJsonToNodeCreator(
     ForExpression(forVariable, iterableVariable, doBlock)(obj.toTextSpan)
   }
 
-  private def visitForwardArg(obj: Obj): RubyExpression = defaultResult(Option(obj.toTextSpan))
+  private def visitForwardArg(obj: Obj): RubyExpression = {
+    println("forward args")
+    defaultResult(Option(obj.toTextSpan))
+  }
 
-  private def visitForwardArgs(obj: Obj): RubyExpression = defaultResult(Option(obj.toTextSpan))
+  // Note: Forward args should probably be handled more explicitly, but this should preserve flows if the same
+  // identifier is used in latter forwarding
+  private def visitForwardArgs(obj: Obj): RubyExpression = MandatoryParameter("...")(obj.toTextSpan)
 
-  private def visitForwardedArgs(obj: Obj): RubyExpression = defaultResult(Option(obj.toTextSpan))
+  private def visitForwardedArgs(obj: Obj): RubyExpression = SimpleIdentifier()(obj.toTextSpan)
 
   private def visitGlobalVariable(obj: Obj): RubyExpression = {
     val span     = obj.toTextSpan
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyJsonToNodeCreator.scala
```diff
@@ -478,11 +478,16 @@ class RubyJsonToNodeCreator(
     ForExpression(forVariable, iterableVariable, doBlock)(obj.toTextSpan)
   }
 
-  private def visitForwardArg(obj: Obj): RubyExpression = defaultResult(Option(obj.toTextSpan))
+  private def visitForwardArg(obj: Obj): RubyExpression = {
+    println("forward args")
```
AndreiDreyer, line 482: println?

AndreiDreyer (approved):
Just that one println

PR #5083, opened 2024-11-10
[c#] Bump `DotNetAstGen`

DavidBakerEffendi:
Uses new version of `DotNetAstGen` that uses a parallelized version. Credits to @ricekot

https://github.com/joernio/DotNetAstGen/pull/26


DavidBakerEffendi:
Sorry, I see a PR is already uphttps://github.com/joernio/joern/pull/5082

PR #5087, opened 2024-11-11
[ruby] Bitwise Assignment Operators

### LINKED ISSUE #5067: [Bug] [ruby] `<<=` Unrecognized Operator
**Describe the bug**
It appears `<<=` is being determined as an operator. While this may be a call in practice, this is worth investigating before determining if this should be a call or not. This is reproducible on `mastodon`.
```scala
 Unrecognized assignment operator: id <<= 16 (lib/mastodon/snowflake.rb)
```

**To Reproduce**
1. `git clone git@github.com:mastodon/mastodon.git --depth 1`
2. `joern-parse mastodon`
3. See error

DavidBakerEffendi:
Added handling for `|=`, `&=`, `<<=`, and `>>=`

Resolves #5067

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
@@ -222,7 +222,7 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
       "|"   -> Operators.or,
       "^"   -> Operators.xor,
 //      "<<"  -> Operators.shiftLeft,  Note: Generally Ruby abstracts this as an append operator based on the LHS
-      ">>" -> Operators.logicalShiftRight
+      ">>" -> Operators.arithmeticShiftRight
     )
 
   protected val AssignmentOperatorNames: Map[String, String] = Map(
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -233,8 +233,9 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
     "/="  -> Operators.assignmentDivision,
     "%="  -> Operators.assignmentModulo,
     "**=" -> Operators.assignmentExponentiation,
-    // Strictly speaking, `a ||= b` means `a || a = b`, but I reckon we wouldn't gain much representing it that way.
-    "||=" -> Operators.assignmentOr,
```
AndreiDreyer, line 237: Was this actually invalid Ruby code?

AndreiDreyer (approved):

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstCreatorHelper.scala
DavidBakerEffendi: This is detected as a separate entity by the parser - a unique structure. So this is valid, but the visitor for it lies elsewhere in the `RubyJsonNodeCreator` and it never pings this map

PR #5094, opened 2024-11-11
[ruby] Fix Call Type for Block

### LINKED ISSUE #5069: [Bug] [ruby] Unexpected call type used for block class on `gitlab`
**Describe the bug**
Seems a modifier call is taking a block, this is probably legal and would be best to test this. This warning occurs on `gitlab`.
```
Unexpected call type used for block class io.joern.rubysrc2cpg.astcreation.RubyIntermediateAst$ProtectedModifier, ignoring block
```
**To Reproduce**
1. `git clone https://gitlab.com/gitlab-org/gitlab.git --depth 1`
2. `joern-parse gitlab`
3. See error (after a while)

AndreiDreyer:
Fixed call type for block when using `protected` keyword.

Resolves #5069

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyJsonToNodeCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyJsonToNodeCreator.scala
@@ -13,7 +13,8 @@ import ujson.*
 
 class RubyJsonToNodeCreator(
   variableNameGen: FreshNameGenerator[String] = FreshNameGenerator(id => s"<tmp-$id>"),
-  procParamGen: FreshNameGenerator[Left[String, Nothing]] = FreshNameGenerator(id => Left(s"<proc-param-$id>"))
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyJsonToNodeCreator.scala
```diff
@@ -13,7 +13,8 @@ import ujson.*
 
 class RubyJsonToNodeCreator(
   variableNameGen: FreshNameGenerator[String] = FreshNameGenerator(id => s"<tmp-$id>"),
-  procParamGen: FreshNameGenerator[Left[String, Nothing]] = FreshNameGenerator(id => Left(s"<proc-param-$id>"))
+  procParamGen: FreshNameGenerator[Left[String, Nothing]] = FreshNameGenerator(id => Left(s"<proc-param-$id>")),
+  fileName: String = ""
```
AndreiDreyer, line 17: Added this for future debugging while we fix warnings. Makes it easier to track down the piece of code in a repo that's causing the warning to be triggered

DavidBakerEffendi (approved):

PR #5099, opened 2024-11-12
[ruby] Handle `yield` in Constructor

### LINKED ISSUE #5090: [Bug] [ruby] Yield Expression Outside of Method Scope
**Describe the bug**
On `forem` we're seeing
```
2024-11-11 13:23:02.283 WARN AstCreator: Yield expression outside of method scope: yield(self) (app/services/audit/event/payload.rb), skipping
2024-11-11 13:23:02.283 WARN AstCreator: Could not represent expression: yield(self) (YieldExpr) (app/services/audit/event/payload.rb), skipping
```
This should be investigated as to how we are meant to handle this, and if it's a true error or FP. If it's not a yield in a traditional sense, then it should simply be a call.

**To Reproduce**
Parse `git@github.com:forem/forem.git`

AndreiDreyer:
Added handling for `yield` in a constructor.

Resolves #5090


DavidBakerEffendi (changes requested):
Could we make `procParam` and `hasYield` properties of `MethodLikeScope` to reduce the redundancy?

DavidBakerEffendi (approved):

PR #5106, opened 2024-11-14
[ruby] Handling for `Splat` field declarations

AndreiDreyer:
Added handling for `*field` declarations. We attempt to determine the elements in the `*field` by looking at `SingleAssignments` also found in the `Field` statements, if we manage to evaluate the collection we create a `FieldsDeclaration` with all the elements from the collection otherwise we create a `SimpleCall` with the target being the `attr_(reader|writer|accessor)` and the `splat` being the `arg`,


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyJsonHelpers.scala
```diff
@@ -137,19 +152,85 @@ object RubyJsonHelpers {
...
+      * @param fieldStmts
+      *   List of all the field statements
+      * @param splattingFields
+      *   List of splatting fields
+      * @return
+      *   List of:
+      *   - Some(_) => if splattingField either is evaluated to a list of FieldDeclarations, otherwise a SimpleCall
+      *   - None => if splattingField cannot be evaluated to either FieldsDeclaration or SimpleCall
+      */
+    def lowerSplattingFieldDecl(
```
DavidBakerEffendi, line 166: Is it not possible to move this lowering closer to AstCreator?

AndreiDreyer: I can't think of a way. This determines whether something is a FieldsDeclaration at the top of a class, or whether it is a SimpleCall that's going in the `<body>` method. I can't see a way that we have this somewhere in the AstCreator and make it work without making it extremely convoluted

DavidBakerEffendi: Yeah that's fair. I'm keen to get this in so the issue wil be fixed then we can revisit the location of the lowering

DavidBakerEffendi (approved):

PR #5119, opened 2024-11-19
[jssrc2cpg] Fixed lowering of tagged template expressions

max-leuthaeuser:


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -399,13 +403,70 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+            val fieldAccessAst =
+              createFieldAccessCallAst(tmpAssignmentAst, memberNode, callee.lineNumber, callee.columnNumber)
+            val thisTmpNode = identifierNode(callee, tmpVarName)
+            scope.addVariableReference(tmpVarName, thisTmpNode)
+
+            (fieldAccessAst, thisTmpNode, member.code)
+        }
+      case _ =>
+        val receiverAst = astForNodeWithFunctionReference(callee.json)
+        val thisNode    = identifierNode(callee, "this").dynamicTypeHintFullName(typeHintForThisExpression())
```
maltek, line 465: I realize this is already done in `astForCallExpression`, and so I wouldn't change it for this PR. But I don't believe JS semantics ever call for an implicit `this` reference for the base. Under some circumstances the global object (e.g. `window`) will implicitly be used as the base / arg 0. But I think that's mostly a legacy thing in non-strict mode.

maltek (approved):
Just playing around trying to break things, I came up with `x.y()`a ${1+1} b ${2+2} c`;` where we set the `methodFullName` of the call to the unknown `tmp.js::program:x.y()`, which will get stubbed and linked by the dynamic call linker. Instead of the simple `<unknownFullName>` like we  This has also already been the case for something like `x.y()()`, so I suppose it's a rare edge-case that we can fix separately. 

Also wondering if we can't share more code between `astForTemplateExpression` and `astForCallExpression`. (But I'm not the one who will maintain this code.)

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForExpressionsCreator.scala
max-leuthaeuser: W.r.t. code sharing and the legacy this: yes, I need to do a round of refactoring there but I want to have the basic functionality in to unblock you and the POC.

PR #5122, opened 2024-11-20
[ruby] `public` in a Method body is lowered to `SimpleCall`

### LINKED ISSUE #5109: [Bug] [ruby] Unhandled `public` modifier in expression position
**Describe the bug**
While parsing `forem`, this warning is emitted

```
2024-11-18 10:04:37.527 WARN AstCreator:  not supported yet: public (PublicModifier), only generating statement
```

**To Reproduce**
Parse `git@github.com:forem/forem.git`

AndreiDreyer:
Fixed warning where `public` was a call to a user-defined method instead of to the Ruby reserved keyword.

resolves #5109

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyJsonToNodeCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyJsonToNodeCreator.scala
@@ -672,19 +672,8 @@ class RubyJsonToNodeCreator(
     val body = obj
       .visitOption(ParserKeys.Body)
       .map {
-        case x: StatementList =>
-          x.copy(statements = x.statements.map {
-            case _: AccessModifier =>
-              val ident = SimpleIdentifier()(x.span)
-              SimpleCall(ident, List.empty)(x.span)
-            case y => y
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/parser/RubyJsonToNodeCreator.scala
```diff
@@ -672,8 +672,19 @@ class RubyJsonToNodeCreator(
...
-        case x: StatementList => x
-        case x                => StatementList(List(x))(x.span)
+        case x: StatementList =>
+          x.copy(statements = x.statements.map {
+            case _: AccessModifier =>
+              val ident = SimpleIdentifier()(x.span)
+              SimpleCall(ident, List.empty)(x.span)
+            case y => y
+          })(x.span)
+        case x: AccessModifier =>
```
DavidBakerEffendi, line 682: This may not be the best place to handle this, as there may be control structures too, etc. Could we not handle this in `AstCreator`? If the parent scope is method then make it a call instead of doing the usual access control.

DavidBakerEffendi (changes requested):

DavidBakerEffendi (approved):

PR #5124, opened 2024-11-21
Revert "[kotlin2cpg] JvmClasspathRoot without an actual temporary file (#5097)"

ml86:
This reverts commit 16f9541eb84605053e9c5484830b4b05e72364a7.

See source code comment for further details.

cc @xavierpinho


xavierpinho (approved):
My bad! Although I didn't notice any issue - I assume it's silently breaking something?

max-leuthaeuser:
If we have org.apache.commons.io already as dependency we could do `IOUtils.copy(resourceStream, tmpFile)`
or with plain java something like `Files.copy(resourceStream, tmpFile.path, StandardCopyOption.REPLACE_EXISTING)`.
So no manual while loop stuff.

ml86:
@xavierpinho Yes, the compiler library silently drops the triggered `NoSuchFileException`.
@max-leuthaeuser Optimization will be done separately.

mpollmeier (approved):

PR #5131, opened 2024-11-25
upgrade scala-repl-pp

mpollmeier:
https://github.com/mpollmeier/scala-repl-pp now has two separate concepts: 
1) predefFiles which are being compiled and loaded into the classpath - for these we will now get correct line numbers if there's any compilation errors. 
2) runBefore code which is executed before the console and/or script is executed - this is e.g. for things like import statements etc. which we need to bring the joern dsl into scope


maltek (approved):

DavidBakerEffendi:
Giving this a look wrt the slicing test

tests/test-dataflow-slice.sc
DavidBakerEffendi, line 1: Shouldn't the `test-scripts` job for slicing be modified for this move?
```
      - run: |
          mkdir /tmp/slice
          ./joern-slice data-flow tests/code/javasrc/SliceTest.java -o /tmp/slice/dataflow-slice-javasrc.json
          ./joern --script "./tests/test-dataflow-slice.sc" --param sliceFile=/tmp/slice/dataflow-slice-javasrc.json | grep -q 'List(boolean b, b, this, s, "MALICIOUS", s, new Foo("MALICIOUS"), s, s, "SAFE", s, b, this, this, b, s, System.out)'
 ```
 
 Note the last line ` ./joern --script "./tests/test-dataflow-slice.sc"`

mpollmeier: I'll also ensure that next time we get a proper error message for something like this

mpollmeier: the 'ensure we get a proper error message' isn't all that easy :(

Issue #4999, opened 2024-10-13
[Bug][Python] Fail to import CPG via Joern command line option


UESuperGate:
**Describe the bug**
After saving and exiting the current Joern work context, a `cpg.bin` is generated under the project's work directory. However, I failed to get the cpg details using `joern <path_to_cpg_bin>` to reload the cpg file. Typing `cpg` to Joern's console reports error. This fails on Python but works fine on Java and C/C++.

**To Reproduce**

poc.py
```python
# (c) 2012-2014, Michael DeHaan <michael.dehaan@gmail.com>
#
# This file is part of Ansible
#
# Ansible is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.
#
# Ansible is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License
# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
__version__ = '1.6'
__author__ = 'Michael DeHaan'
```

1. Open the Joern's console via `joern`
2. Import the above `poc.py` to this work context via `importCode("poc.py", "test")`
3. Enter `ctrl+d` to save & exit in Ubuntu.
4. Import the saved cpg file via `joern ./workspace/test/cpg.bin`.
5. Type `cpg.all.l` and see error report.

**Expected behavior**
`cpg.all.l` will print out all the nodes in the cpg.

**Screenshots**
![image](https://github.com/user-attachments/assets/7e37d797-00a2-4cca-8f78-37b64f832959)

**Desktop (please complete the following information):**
 - OS Ubuntu 22.04
 - Joern Version 4.0.98
 - Java version openJDK 17.0.11

mpollmeier:
It is indeed reproducable, thanks for reporting! 
Looks like the predef/runBefore code execution is broken when you pass the 'cpg-to-load' on startup. 

I'll get this fixed, but until then, you can just start `joern` and run `open` - it opens the last project by default:

```
➜  joern git:(master) ✗ ./joern

     ██╗ ██████╗ ███████╗██████╗ ███╗   ██╗
     ██║██╔═══██╗██╔════╝██╔══██╗████╗  ██║
[...]

PR #5137, opened 2024-11-27
fixup initialisation order of cpg generators

mpollmeier:
We trigger the post-processing directly if a cpg is loaded on startup
(e.g. `./joern cpg.bin`). In that case, the typeRecoveryConfig has not
been initialized. There's a reason why in Scala you have to explicitly
make an effort to start with `null`, because more often than not
you'll be missing out some random edge case... lazy val is the better
alternative

fixes https://github.com/joernio/joern/issues/4999 (again...)


max-leuthaeuser (approved):
👍

PR #5148, opened 2024-11-29
[querydb] Add scan rule for Twig Template Injection detection

piggyctf:
Add scan rule in querydb to detect potential Twig Template Injection (PHP) vulnerabilities

- Introduced a new rule to identify potential Twig template injection issues.
- Tested with compiled source code (using some juicy public CTF sample code) and confirmed successful detection.


piggyctf:
> Thanks again @piggyctf!
> 
> Can you share some test code? :)
The test code was added in the commit [2a31e79](https://github.com/joernio/joern/pull/5148/commits/2a31e79a2ab1baa0998da0d7366252e8da220ce1). Let me know if anything is needed 😄

querydb/src/main/scala/io/joern/scanners/php/TwigTemplateInjection.scala
```diff
@@ -0,0 +1,122 @@
...
+      score = 5,
+      withStrRep({ cpg =>
+
+        def source =
+          cpg.call.name(Operators.assignment).argument.code("(?i).*request.*").l
+
+        def sink =
+          cpg.call.name("createTemplate").l.filter(call => call.methodFullName.toLowerCase.contains("twig")).argument
+
+        sink.reachableBy(source).l.iterator
```
maltek, line 35: could you remove all the intermediate `.l` please? They are convenient in the REPL or debugger to look at intermediate results, but otherwise these linked lists are just overhead.

querydb/src/main/scala/io/joern/scanners/php/TwigTemplateInjection.scala
```diff
@@ -0,0 +1,122 @@
...
+          |This doesn't necessarily indicate a Twig template injection, but if the input is not sanitized and the escape settings are disabled in the application, it could potentially lead to a template injection vulnerability.
+          |""".stripMargin,
+      score = 5,
+      withStrRep({ cpg =>
+
+        def source =
+          cpg.call.name(Operators.assignment).argument.code("(?i).*request.*").l
+
+        def sink =
+          cpg.call.name("createTemplate").l.filter(call => call.methodFullName.toLowerCase.contains("twig")).argument
```
maltek, line 33: the `filter()` here can be more idiomatically expressed as `.methodFullName("(?i).*twig.*")`

piggyctf: Hi maltek, thanks for your comments. I've updated in https://github.com/joernio/joern/pull/5148/commits/7471b8a27fc50c2f1ce034654841befc698685f5 as you recommended, however, there is one test failed at `io.joern.rubysrc2cpg.io.RubySrc2CpgHTTPServerTests` after the change. My code has nothing to do with `rubysrc2cpg` module, this error happened once before when I added a new rule for detecting a Java vulnerability, and then it suddenly disappeared without any further changes.. Could you please take a look at it?

maltek: - you've also deleted the `.argument` at the end. Could you add that back please? Even if it seems to work, it's not really correct: in a dataflow query a call stands for the return value of that call. And for a template we're not interested in the output of template expansion, but only in the input that's used as a template. So we want to use the argument of the call as the sink, not its return value. (Sorry if I wasn't clear about what to change.)
- Github shows that all checks have passed, so the test seems to work in our CI environment. I can't say much about why it'd fail on your machine.

piggyctf:
The sink should be fixed in [dc4cf9f](https://github.com/joernio/joern/pull/5148/commits/dc4cf9f3d72926e56dc7f40e8ca58bf67a445723). Thanks for pointing that out. Please let me know if any concerns :)

maltek (approved):

maltek:
Thanks again @piggyctf. I'm sorry it took over a month of back-and-forth to get this merged. I hope this experience wasn't too discouraging.

PR #5149, opened 2024-12-02
[pysrc2cpg] Create member writes for variables writes in class bodies.

ml86:
In class body functions, assignments to variables represent assignments
to members in the meta class object.
This change adds the additional statement to the class body method to
reflect this.

Also: Made `topLevelExpressions` step available for individual methods.

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/dataflow/DataFlowTests.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/dataflow/DataFlowTests.scala
@@ -1187,7 +1187,7 @@ class RegexDefinedFlowsDataFlowTests
     "be found" in {
       val src = cpg.identifier("Foo").l
       val snk = cpg.call("print").l
-      snk.reachableByFlows(src).size shouldBe 2
+      snk.reachableByFlows(src).size shouldBe 3
     }
   }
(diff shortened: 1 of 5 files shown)


ml86:
@DavidBakerEffendi I bumped the number of flows for one test up to 3. (`snk.reachableByFlows(src).size shouldBe 3`) 

The way those tests are written i have no clue what the actual expectation is. Is this ok find 3 instead of 2? I meant 2 already seems strange.

maltek (approved):

DavidBakerEffendi:
Hi @ml86. Sorry for the late reply, spent the day moving! The test itself is written incorrectly, seems the sink is attached to the call, not its 1st argument. There should be fewer calls if argument(1) is the sink

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/dataflow/DataFlowTests.scala
```diff
@@ -1187,7 +1187,7 @@ class RegexDefinedFlowsDataFlowTests
     "be found" in {
       val src = cpg.identifier("Foo").l
       val snk = cpg.call("print").l
```
DavidBakerEffendi, line 1189: This should produce a more representative flow
```suggestion
      val snk = cpg.call("print").argument(1).l
```

DavidBakerEffendi (approved):

PR #5153, opened 2024-12-03
[c2cpg] Added MSVC support

max-leuthaeuser:
This adds support for MSVC system header discovery via `--with-include-auto-discovery`.

1) Ask `vswhere.exe` for the actual MSVC installation path (https://github.com/microsoft/vswhere, included in every installation as of Visual Studio 2017).
2) Call `vcvars64.bat` to set the `INCLUDE` environment variable.
3) Use that to configure CDT.

Also in this PR (disovered during testing the MSVC stuff):
- fixed the `DefaultIgnoredFolders` regex for paths containing dot folders.
- proper thread-safe header file caching (caching the actual parse result and not only the plain code) using `createForExternalFileLocation` and `resetForTranslationUnit()`.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/utils/IncludeAutoDiscovery.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/utils/IncludeAutoDiscovery.scala
@@ -6,6 +6,7 @@ import org.slf4j.LoggerFactory
 
 import java.nio.file.Path
 import java.nio.file.Paths
+import scala.collection.mutable
 import scala.util.Failure
 import scala.util.Success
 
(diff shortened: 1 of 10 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/utils/IncludeAutoDiscovery.scala
```diff
@@ -66,36 +75,77 @@ object IncludeAutoDiscovery {
...
+        None
+    }
+  }
+
+  private def extractMSVCIncludePaths(resolvedInstallationPath: String): Set[Path] = {
+    ExternalCommand.run(VcVarsCommand, resolvedInstallationPath, Map("VSCMD_DEBUG" -> "3")) match {
+      case Success(results) =>
+        val includesLine   = results.find(_.startsWith("INCLUDE="))
+        val includesString = includesLine.map(_.replaceFirst("INCLUDE=", ""))
+        val includes       = includesString.map(_.split(";").toSet.map(p => Paths.get(p.trim).toRealPath()))
```
maltek, line 93: Include path order can sometimes matter, so I think we want to use a linked hash set here, not just a plain set.

maltek (approved):

PR #5154, opened 2024-12-03
[c2cpg] Guard instanceOf casts

max-leuthaeuser:
For: https://shiftleftinc.atlassian.net/browse/SEN-3467

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
@@ -79,11 +79,11 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) { 
     val functionNameExpr = call.getFunctionNameExpression
     val typ              = functionNameExpr.getExpressionType
     typ match {
-      case pointerType: IPointerType =>
+      case _: IPointerType =>
         createPointerCallAst(call, cleanType(safeGetType(call.getExpressionType)))
       case functionType: ICPPFunctionType =>
         functionNameExpr match {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -79,11 +79,11 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
     val typ              = functionNameExpr.getExpressionType
     typ match {
-      case pointerType: IPointerType =>
+      case _: IPointerType =>
         createPointerCallAst(call, cleanType(safeGetType(call.getExpressionType)))
       case functionType: ICPPFunctionType =>
         functionNameExpr match {
-          case idExpr: CPPASTIdExpression if idExpr.getName.getBinding.isInstanceOf[ICPPFunction] =>
+          case idExpr: CPPASTIdExpression if idExpr.getName.resolveBinding().isInstanceOf[ICPPFunction] =>
             val function = idExpr.getName.getBinding.asInstanceOf[ICPPFunction]
```
maltek, line 87: I don't know the difference between the two... is it intentional that you're using `resolveBinding()` in the match guard, and `getBinding` in the body?

maltek (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
max-leuthaeuser: resolveBinding actually calculates the binding (its null beforehand) and getBinding simply returns the result (which is stored as instance variable at, e.g., CASTName).

Ferada (approved):

PR #5157, opened 2024-12-03
[ruby] Fix Call as TypeDecl AST Child

### LINKED ISSUE #5155: [Bug] [ruby] Calls Directly under Type Decl
**Describe the bug**
On the `zammad` project, there are a number of call nodes directly under a Type Decl which is invalid
```
joern> cpg.typeDecl.where(_.astOut.isCall).size
val res2: Int = 929
```

**To Reproduce**
Parse https://github.com/zammad/zammad

DavidBakerEffendi:
Fixes issue where `<body>` calls from nested types were being given as direct children to type decls. The solution pushes `<body>` calls to the next method.

Another issue was that access modifier calls did not check that the argument was actually a method. Sometimes methods could be referred to via an identifier in some data structure. This PR currently emits a warning for these cases.

Resolves #5155

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
@@ -203,7 +203,10 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
         pushAccessModifier(ModifierTypes.PUBLIC)
     }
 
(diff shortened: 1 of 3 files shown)


AndreiDreyer:
> Another issue was that access modifier calls did not check that the argument was actually a method. Sometimes methods could be referred to via an identifier in some data structure. This PR currently emits a warning for these cases.
> 

I might have missed this, but is this something that is already happening or something that should have been implemented since I see no new warnings being emitted in this PR?

DavidBakerEffendi:
@AndreiDreyer probably best if I actually add the test case - will do shortly, there seems to be a fire drill at the office

AndreiDreyer (approved):

PR #5169, opened 2024-12-06
[c#] add `external-summary-paths` option

xavierpinho:
Introduces an `external-summary-paths` option, allowing users to provide custom directories for type summaries à la `builtin_types`. Some minor refactoring otherwise.


DavidBakerEffendi (approved):

xavierpinho:
Need to take into account the newer SourceFiles revisions. On it...

PR #5173, opened 2024-12-10
[c2cpg] Recognize more source file extensions

max-leuthaeuser:
To be consistent with latest cmake and CDT (https://github.com/eclipse-cdt/cdt/pull/422).

### DIFF
--- a/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/HeaderAstCreationPassTests.scala
+++ b/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/HeaderAstCreationPassTests.scala
@@ -40,9 +40,9 @@ class HeaderAstCreationPassTests extends C2CpgSuite {
         case Seq(bar, foo, m, printf) =>
           // note that we don't see bar twice even so it is contained
           // in main.h and included in main.c and we do scan both
-          bar.fullName shouldBe "bar"
+          bar.fullName shouldBe "bar:void()"
           bar.filename shouldBe "main.h"
-          foo.fullName shouldBe "foo"
+          foo.fullName shouldBe "foo:int()"
           foo.filename shouldBe "other.h"
           // main is also deduplicated. It is defined within the header file,
           // and has an actual implementation in the source file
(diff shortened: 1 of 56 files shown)


mpollmeier (approved):

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/HeaderAstCreationPassTests.scala
```diff
@@ -40,9 +40,9 @@ class HeaderAstCreationPassTests extends C2CpgSuite {
         case Seq(bar, foo, m, printf) =>
           // note that we don't see bar twice even so it is contained
           // in main.h and included in main.c and we do scan both
-          bar.fullName shouldBe "bar"
+          bar.fullName shouldBe "bar:void()"
```
maltek, line 43: This looks like it would be a regression for C code, in order to improve C++ code. I guess there just is inherent ambiguity with `.h` files, whether they contain C or C++ code. But I don't think we can get away with this right now.

Maybe we do header files in a second pass after the regular files, so we can know whether they got included from C or C++ files? Or maybe CDT has some guess-the-file-type magic since the IDE runs into the same problem?

maltek (changes requested):

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/HeaderAstCreationPassTests.scala
max-leuthaeuser: > Or maybe CDT has some guess-the-file-type magic since the IDE runs into the same problem?

Sadly no, they also simply use the C++ parser in all cases.

I will test the two-passes approach.

max-leuthaeuser: Or to phrase it differently:
The behaviour without this PR is definitely wrong as it makes parsing CPP code in .h files impossible.
With this PR will are able to parse such code. "Wrong" fullnames for C method declarations that are never used in any source file (because there we will create the correct fullname and de-duplicate correctly) should be no issue or do I miss something there?

max-leuthaeuser: Latest commit to this PR does not handle *.h files as CPP files anymore.
Just want to get the additional source file extensions in.

maltek (approved):

PR #5175, opened 2024-12-11
[c2cpg] Fixed logging via CCorePlugin

max-leuthaeuser:
This PR adds a stripped-down version of the original `org.eclipse.cdt.core.CCorePlugin` shadowing it and providing only the functionality to get it running without all the Eclipse OSGI context.

Sadly, some parser internal components (e.g., the ambiguity resolving) log via this class. Without a proper OSGI instantiation (which we do not have/want in Joern) we are running into all kind of exceptions due to non-initialized entities (e.g., the said logging utils).

Discovered during testing c2cpg on https://github.com/chromium/chromium where we have a lot of ambiguous nodes during parsing that can't be resolved and are logged causing the said exceptions.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/java/org/eclipse/cdt/core/CCorePlugin.java
+++ b/joern-cli/frontends/c2cpg/src/main/java/org/eclipse/cdt/core/CCorePlugin.java
@@ -1,0 +1,100 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2020 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *     Markus Schorn (Wind River Systems)
+ *     Andrew Ferguson (Symbian)
+ *     Anton Leherbauer (Wind River Systems)
+ *     oyvind.harboe@zylin.com - http://bugs.eclipse.org/250638
+ *     Jens Elmenthaler - http://bugs.eclipse.org/173458 (camel case completion)
+ *     Sergey Prigogin (Google)
+ *     Alexander Fedorov (ArSysOp) - Bug 561992
+ *******************************************************************************/
+
+package org.eclipse.cdt.core;
+
+import org.eclipse.cdt.core.model.CModelException;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Plugin;
+import org.eclipse.core.runtime.Status;
+import org.osgi.framework.Version;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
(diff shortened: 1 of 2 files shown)


mpollmeier:
I have some doubts about the shadowing... with the proposed setup we get two versions of `org.eclipse.cdt.core.CCorePlugin.class` in our classpath, and some unforseen magic in the classloader decides at runtime which one is being loaded... Am I missing something?

Not sure if this works with cdt, but have you considered shading instead of shadowing? I.e., copy the source to our repo and put it into a different namespace, similar to what I did in the repl: https://github.com/mpollmeier/scala-repl-pp/tree/main/shaded-libs/src/main/scala/replpp/shaded

Another option: given that we already ship a redeployed version of cdt, we could also fiddle with the jar before publishing it, i.e. remove the original .class file. Doesn't sound very nice though.

max-leuthaeuser:
> and some unforseen magic in the classloader decides at runtime which one is being loaded

Hm, might be true. I was hoping the class files from c2cpg always takes precedence.

> Not sure if this works with cdt, but have you considered shading instead of shadowing?

Yes, but I wasn't able to get it running. If you want to try it feel free to alter this PR here directly.

mpollmeier:
> I was hoping the class files from c2cpg always take precedence.

No, there's nothing special about the project's sources once it's all packaged up...
See e.g. after `sbt stage`:
```
➜  lib git:(max/fixCCorePluginLogging) unzip -l io.joern.c2cpg-4.0.182+1-d55c4699.jar|grep CCorePlugin
     3049  2010-01-01 00:00   org/eclipse/cdt/core/CCorePlugin.class
➜  lib git:(max/fixCCorePluginLogging) unzip -l io.joern.eclipse-cdt-core-8.5.0.202410191453+2.jar |grep CCorePlugin
    39710  2024-12-09 10:25   org/eclipse/cdt/core/CCorePlugin.class
     4077  2024-12-09 10:25   org/eclipse/cdt/core/CCorePlugin$2.class
     1104  2024-12-09 10:25   org/eclipse/cdt/core/CCorePlugin$NullConsole$1.class
     2472  2024-12-09 10:25   org/eclipse/cdt/core/CCorePlugin$1.class
     1290  2024-12-09 10:25   org/eclipse/cdt/core/CCorePlugin$NullConsole.class
     5944  2024-12-09 10:25   org/eclipse/cdt/internal/core/CCorePluginResources.properties
```

mpollmeier:
> Note: CCorePlugin.log is called from within org.eclipse.cdt.foo (e.g., org.eclipse.cdt.internal.core.dom.parser) and not our code in reality.

In that case shading doesn't seem to be a good option either. We'd need to build the entire cdt project and publish it separately. 

Another option would be to remove the class from the original jar on `sbt stage`, but that's not a good idea either because it needs to be repeated for every stage/assembly config downstream as well, and it means we're lying about the project's dependencies. 

Sounds like there's no good solution here, and out of the bad solutions I lean towards removing the class from the jar when we republish it... thoughts?

max-leuthaeuser:
> Sounds like there's no good solution here, and out of the bad solutions I lean towards removing the class from the jar when we republish it... thoughts?

I was thinking about this in first place, yes. Only downside: without a custom CCorePlugin like we do in c2cpg this republished jar is useless.

mpollmeier:
That's true, although maybe not the end of the world. 

That being said, we could move the simplified CCorePlugin into the custom build that we create anyway for republishing. Like so: https://github.com/joernio/joern/pull/5178

max-leuthaeuser:
Ok, let's do it that way.

PR #5178, opened 2024-12-12
[c2cpg] Fixed logging via CCorePlugin

mpollmeier:
Adapted from https://github.com/joernio/joern/pull/5175

This PR adds a stripped-down version of the original
`org.eclipse.cdt.core.CCorePlugin` and providing only the functionality to
get it running without all the Eclipse OSGI context. To avoid classpath
issues we removes the original class when we repulish the jar...

Sadly, some parser internal components (e.g., the ambiguity resolving) log
via this class. Without a proper OSGI instantiation (which we do not
have/want in Joern) we are running into all kind of exceptions due to
non-initialized entities (e.g., the said logging utils).

Discovered during testing c2cpg on https://github.com/chromium/chromium
where we have a lot of ambiguous nodes during parsing that can't be
resolved and are logged causing the said exceptions.


max-leuthaeuser (approved):

mpollmeier:
Not as easy as I thought - there's a few catch 22 type dilemmas in the publishing process... 
I'm still fiddling with it..

max-leuthaeuser:
I got around it locally when disabling javadoc generation entirely (which is not needed in the jar I guess).

mpollmeier:
javadoc is required for maven central publishing...

mpollmeier:
got it working though

PR #5187, opened 2024-12-16
[c#] support for `global using` directives in `CSharpProgramSummary`

xavierpinho:
Adds support for handling `global using` directives. These are distinguished both during summarisation and in `CSharpProgramSummary`, so that `findGlobalTypes` accounts for them.

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpProgramSummary.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpProgramSummary.scala
@@ -29,8 +29,7 @@ type NamespaceToTypeMap = mutable.Map[String, mutable.Set[CSharpType]]
 case class CSharpProgramSummary(namespaceToType: NamespaceToTypeMap, imports: Set[String], globalImports: Set[String])
     extends ProgramSummary[CSharpType, CSharpMethod, CSharpField] {
 
-  def findGlobalTypes: Set[CSharpType] =
-    (globalImports ++ Set(Constants.Global)).flatMap(namespaceToType.getOrElse(_, Set.empty))
+  def findGlobalTypes: Set[CSharpType] = namespaceToType.getOrElse(Constants.Global, Set.empty).toSet
 
   @targetName("appendAll")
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/datastructures/CSharpProgramSummary.scala
```diff
@@ -27,14 +26,19 @@ type NamespaceToTypeMap = mutable.Map[String, mutable.Set[CSharpType]]
   * @see
   *   [[CSharpProgramSummary.jsonToInitialMapping]] for generating initial mappings.
   */
-case class CSharpProgramSummary(namespaceToType: NamespaceToTypeMap, imports: Set[String])
+case class CSharpProgramSummary(namespaceToType: NamespaceToTypeMap, imports: Set[String], globalImports: Set[String])
     extends ProgramSummary[CSharpType, CSharpMethod, CSharpField] {
 
-  def findGlobalTypes: Set[CSharpType] = namespaceToType.getOrElse(Constants.Global, Set.empty).toSet
```
DavidBakerEffendi, line 33: So I believe the original design choice was to assign globals to the global namespace. Does this not work for your use-case?

xavierpinho: Good question. From what I've seen, this `Global` namespace is being used to store top-level user-defined types found during pre-analysis. For `using global XXX`, we want to have `XXX` available in `typesInScope` but `XXX` is to be found in the program summary's `NamespaceToTypeMap`. Perhaps the best patch would be to add `globalImports` directly to `typesInScope` -- currently it's set to this `findGlobalTypes`. What do you reckon?

DavidBakerEffendi: Thanks! Yes this looks more along what I had in mind originally. Of course, if this ends up not suiting your needs, we can revisit the earlier changes in this branch.

DavidBakerEffendi (approved):

PR #5188, opened 2024-12-17
[c2cpg] Fix aliasing of named anonymous typedefs

max-leuthaeuser:


maltek (approved):
Using (hash) sets with `headOption` always worries me a little, because the set basically shuffles the contained elements. 
I can't find any cases where something actually goes wrong - not sure how I could define a type where these sets contain multiple elements. Still, changing the code to read like `...find(_ != fullName)` would just avoid the problem class as a whole, without any loss in readability or brevity.

PR #5189, opened 2024-12-18
[ruby] Cyclic Singleton Method AST & Bracket Invocation Fix 

DavidBakerEffendi:
* Fixed a bug where singleton methods defined on types unresolvable in scope were assigned themselves as AST parents
* Fixed a bug where bracket invocations on unresolvable types were interpreted as index accesses without an index

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/CallTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/CallTests.scala
@@ -568,12 +568,19 @@ class CallTests extends RubyCode2CpgFixture(withPostProcessing = true) {
     }
   }
 
-  "A Set instantiated with a 'brackets' call" should {
-    val cpg = code("x = Set[]")
+  "A Set instantiation with a 'brackets' call" should {
+    val cpg = code("Set[]")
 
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/CallTests.scala
```diff
@@ -567,4 +567,13 @@ class CallTests extends RubyCode2CpgFixture(withPostProcessing = true) {
       case xs => fail(s"Expected two params, got [${xs.code.mkString(",")}]")
     }
   }
+
+  "A Set instantiated with a 'brackets' call" should {
+    val cpg = code("x = Set[]")
+
+    "not be interpreted as an index access" in {
+      cpg.call.nameExact("[]").size shouldBe 1
+      cpg.call.nameExact(Operators.indexAccess).size shouldBe 0
```
ml86, line 576: Please only state our expectation. But in a more precise manner. 
Do we need the `x =` part for the test?
Also check for the `Set` call and then test for the child call to `[]`.

We are not expecting a hole universe of things so only checking for one of those does not make too much sense. So please remove the 'cpg.call.nameExact(Operators.indexAccess).size shouldBe 0'.

ml86 (changes requested):

PR #5190, opened 2024-12-18
[c#] support for top-level statements

xavierpinho:
Adds support for top-level statements by imitating the lowering performed by Roslyn, i.e. by wrapping them inside a code block like
```csharp
internal class Program
{
  private static void <Main>$(string[] args)
  {
    <top-level-statements>
  }
}
```

I used the same names (`<Main>$` and `Program`) here, but I'm not sure if the usage of `<` or `$` is reserved in full names? If so, what would the recommended names be?

Caveat: there can only be one such file (with top-level statements), which implicitly becomes the program's entry point. In an arbitrary repo we might find more than one since this restriction is tied to a `csproj`. In other words, the "compilation unit" would ideally be a .csproj, but that would mean a large refactoring.

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstCreator.scala
@@ -1,15 +1,23 @@
 package io.joern.csharpsrc2cpg.astcreation
 
 import io.joern.csharpsrc2cpg.{CSharpDefines, Constants}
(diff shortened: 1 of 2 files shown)


maltek (approved):
> I used the same names (<Main>$ and Program) here, but I'm not sure if the usage of < or $ is reserved in full names? If so, what would the recommended names be?

Not that I'd be aware of. Frontends tend to use these characters for generated names in order to guarantee that these names don't clash with names chosen by the programmer. Presumably, the C# compiler chose the `<Main>$` name for the same reason. The only requirement following from that is that the frontend should be internally consistent. I.e. ensure that it doesn't generate a clashing name under for a different construct.

Another consideration is that fullnames are supposed to be unique. If we think it's likely that people will include multiple such files with top-level statements in the same scan, then we should handle that. E.g. by incorporating the file name in the method fullname.

PR #5203, opened 2025-01-06
[c#] create implicit ctor for field inits

xavierpinho:
Fields with initial assignments are now moved to the `<init>` (constructor) method, even when a default constructor is not explicitly declared. Previously, this behavior was supported only for explicitly declared constructors. This patch applies exclusively to dynamic fields -- support for moving static field initializations will be addressed in a future PR.

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForDeclarationsCreator.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForDeclarationsCreator.scala
@@ -1,6 +1,6 @@
 package io.joern.csharpsrc2cpg.astcreation
 
-import io.joern.csharpsrc2cpg.CSharpModifiers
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForDeclarationsCreator.scala
```diff
@@ -84,6 +83,50 @@ trait AstForDeclarationsCreator(implicit withSchemaValidation: ValidationMode) {
...
+
+      val methodNode_ = NewMethod()
+        .name(name)
+        .fullName(fullName)
+        .signature(signature)
+        .filename(relativeFileName)
+
+      val parameterNodes = Seq(
+        NewMethodParameterIn()
+          .name("this")
```
DavidBakerEffendi, line 116: A good place to use `io.joern.csharpsrc2cpg.Constants.This`

DavidBakerEffendi (approved):

PR #5207, opened 2025-01-06
Fix names of anonymous classes defined in lambdas

johannescoetzee:
When an anonymous class is defined, the name of the enclosing scope is prepended to the anonymous class name to ensure uniqueness. I previously used an empty method node for lambda methods, however, so the enclosing scope name for anonymous classes in lambdas would just be `<empty>`. This resulted in duplicate anonymous class names being possible and in an edge case where orphan locals were created in anonymous classes of the same name where one is defined in a lambda in a static scope and one in a non-static scope.

This PR fixes this issue by using the actual lambda method node in the scope.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/expressions/AstForLambdasCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/expressions/AstForLambdasCreator.scala
@@ -54,6 +54,7 @@ private[expressions] trait AstForLambdasCreator { this: AstCreator =>
     variablesInScope: Seq[ScopeVariable],
     expectedLambdaType: ExpectedType
   ): (NewMethod, LambdaBody) = {
+
     val implementedMethod    = implementedInfo.implementedMethod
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/expressions/AstForLambdasCreator.scala
```diff
@@ -119,8 +123,8 @@ private[expressions] trait AstForLambdasCreator { this: AstCreator =>
     val containsEmptyType   = maybeParameterTypes.exists(_.contains(ParameterDefaults.TypeFullName))
 
     (returnType, maybeParameterTypes) match {
-      case (Some(returnTpe), Some(parameterTpes)) if !containsEmptyType =>
-        composeMethodLikeSignature(returnTpe, parameterTpes)
+      case (Some(returnType), Some(parameterTypes)) if !containsEmptyType =>
+        composeMethodLikeSignature(returnType, parameterTypes)
```
johannescoetzee, line 127: This is just a typo I saw while working on this. I could move this to a separate PR, but I thought I'd include it here since the PR is relatively small anyways

maltek (approved):

PR #5215, opened 2025-01-09
[c2cpg] Improvements for range-based for-statement and local code fields

max-leuthaeuser:
- range-based for-statement blocks had multiple outgoing CFG edges because the children were not wrapped into blocks.
- code fields for locals from NamedTypeSpecifiers now preserve static and const modifier

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
@@ -142,14 +142,16 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
       "class"
     )
 
-  private val KeepTypeKeywords: List[String] = List("unsigned", "volatile")
+  private val KeepTypeKeywords: List[String] = List("unsigned", "volatile", "const", "static")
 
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -142,14 +142,16 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
       "class"
     )
 
-  private val KeepTypeKeywords: List[String] = List("unsigned", "volatile")
+  private val KeepTypeKeywords: List[String] = List("unsigned", "volatile", "const", "static")
```
maltek, line 145: I'd rename this, since `static` is decidedly not part of the type. (I think it's called a storage modifier.)

maltek (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
max-leuthaeuser: Well, it's a list of keywords to keep at the string representation of a type. I keep it that way for now.

PR #5222, opened 2025-01-13
[php2cpg] Try multiple charsets when reading file content

johannescoetzee:
There's a customer issue where reading the file content was resulting in a `MalformedInputException` being thrown. With this PR, we just try all of the Java standard charsets before emitting a warning instead of crashing: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/charset/StandardCharsets.html

I would be very surprised if we ever need the specific endianness UTF-16 variants, but since this only tries encodings until something works, I also don't think adding them to the list has much of a downside.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/MethodTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/MethodTests.scala
@@ -146,7 +146,7 @@ class MethodTests extends PhpCode2CpgFixture {
         fooMethod.file.head.content.substring(offsetStart, offsetEnd) shouldBe
           """function foo() {
             |  // ⦝
-            |  $x = "??⨌????????????";
+            |  $x = "🙂⨌🙂𐇐🙂🙂🙂🙂";
             |}""".stripMargin
       }
     }
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/MethodTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/MethodTests.scala
@@ -146,7 +146,7 @@ class MethodTests extends PhpCode2CpgFixture {
         fooMethod.file.head.content.substring(offsetStart, offsetEnd) shouldBe
           """function foo() {
             |  // ⦝
-            |  $x = "🙂⨌🙂𐇐🙂🙂🙂🙂";
+            |  $x = "??⨌????????????";
             |}""".stripMargin
       }
     }
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ name                     := "joern"
 ThisBuild / organization := "io.joern"
 ThisBuild / scalaVersion := "3.5.2"
 
-val cpgVersion = "1.7.19"
+val cpgVersion = "1.7.21"
 
 lazy val joerncli          = Projects.joerncli
 lazy val querydb           = Projects.querydb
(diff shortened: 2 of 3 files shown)


max-leuthaeuser:
I think thats why we have https://github.com/ShiftLeftSecurity/codepropertygraph/blob/master/codepropertygraph/src/main/scala/io/shiftleft/utils/IOUtils.scala

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/MethodTests.scala
```diff
@@ -146,7 +146,7 @@ class MethodTests extends PhpCode2CpgFixture {
         fooMethod.file.head.content.substring(offsetStart, offsetEnd) shouldBe
           """function foo() {
             |  // ⦝
-            |  $x = "🙂⨌🙂𐇐🙂🙂🙂🙂";
+            |  $x = "??⨌????????????";
```
johannescoetzee, line 149: Using `IOUtils.readEntireFile` introduces a regression in the characters we can actually parse, but I suspect it's unlikely we'll encounter many such characters in actual code and is outweighed by the benefit of using a uniform approach across frontends

max-leuthaeuser: Hm, stange. It works here: io.joern.jssrc2cpg.io.CodeDumperFromContentTests

johannescoetzee: I get the same results when pasting that string into one of the jssrc tests:

```
...
  x = 'foo[??⨌????????????]';
}" was not equal to "...ass Foo
{
  x = 'foo[🙂⨌🙂𐇐🙂🙂🙂🙂]';
...
```

I did a hex dump of a file containing only `🙂` and it matches the UTF-8 encoding from https://www.compart.com/en/unicode/U+1F642

```
❯ hexdump /tmp/smile
0000000 9ff0 8299 000a                         
0000005
❯ cat /tmp/smile
🙂
```

nvim could've done some conversions on write or such though. If not, it looks like it should be a valid UTF-8 character

max-leuthaeuser: Maybe its time to move IOUtils to x2cpg (which wasnt a thing back then) and fix the utf8 handling. I'll do a draft PR.

max-leuthaeuser: With this: https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1804 the test case works again. 
I was not able to (re-) discover some code with actual unpaired surrogates (which was the reason I implemented it in first place). Not really sure on how to proceed there.

maltek: does this table here help with crafting such a file? https://simonsapin.github.io/wtf-8/#surrogates-byte-sequences

johannescoetzee:
Upgrading to https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1804 fixed the test. Thanks @max-leuthaeuser!

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/MethodTests.scala
max-leuthaeuser: Interesting! We should add some testing for that.

max-leuthaeuser (approved):

PR #5232, opened 2025-01-17
some minor performance improvements for LinkingUtil. improve logging for CdgPass

bbrehm:


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/codepencegraph/CdgPass.scala
```diff
@@ -22,6 +22,8 @@ import org.slf4j.{Logger, LoggerFactory}
 class CdgPass(cpg: Cpg) extends ForkJoinParallelCpgPass[Method](cpg) {
   import CdgPass.logger
 
+  val hasLogged = java.util.concurrent.atomic.AtomicInteger(10)
```
mpollmeier, line 25: please add a comment why you chose 10..

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/codepencegraph/CdgPass.scala
```diff
@@ -39,16 +41,11 @@ class CdgPass(cpg: Cpg) extends ForkJoinParallelCpgPass[Method](cpg) {
         case postDomFrontierNode =>
           val nodeLabel  = postDomFrontierNode.label
           val containsIn = postDomFrontierNode._containsIn
-          if (containsIn == null || !containsIn.hasNext) {
-            logger.warn(s"Found CDG edge starting at $nodeLabel node. This is most likely caused by an invalid CFG.")
-          } else {
-            val method = containsIn.next()
+          if(hasLogged.get() > 0 && hasLogged.decrementAndGet() > 0) {
```
mpollmeier, line 44: why these two checks? the first one doesn't make a difference, does it?

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/codepencegraph/CdgPass.scala
```diff
@@ -39,16 +41,11 @@ class CdgPass(cpg: Cpg) extends ForkJoinParallelCpgPass[Method](cpg) {
...
+            val method = containsIn.nextOption().map{_.toString}.getOrElse("NA")
             logger.warn(
-              s"Found CDG edge starting at $nodeLabel node. This is most likely caused by an invalid CFG." +
-                s" Method: ${method match {
-                    case m: Method => m.fullName;
-                    case other     => other.label
-                  }}" +
+              s"Found CDG edge starting at $nodeLabel node ${node.toString} <-> ${postDomFrontierNode.toString}. This is most likely caused by an invalid CFG." +
```
mpollmeier, line 47:
```suggestion
              s"Found CDG edge starting at $nodeLabel node $node <-> ${postDomFrontierNode}. This is most likely caused by an invalid CFG." +
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/codepencegraph/CdgPass.scala
bbrehm: I don't particularly care about the precise number -- it's just that, if a graph is broken in a way that triggers this log message often, then it tends to trigger this message _very_ often, which takes forever (and spams the logs).

mpollmeier: Makes sense, and I was aware that 10 is just an arbitrary number, but whenever the next reader sees an arbitrary number they might wonder 'why that number', so a short comment is helpful IMO

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/codepencegraph/CdgPass.scala
bbrehm: It doesn't matter a lot, it's just a performance micro-optimization. I wouldn't add that to anybodies code unless it makes a measurable difference, but I write that way automatically without thinking.

To explain: This is a typical fastpath optimization. Suppose that 16 threads all try to atomically decrement the counter -- then 16 cores will play tug-of-war with the cache-line, spam whatever fabric interconnects them with whatever messages used for their cache coherency protocol, etc.

However, if hasLogged is already exceeded, then we don't need to decrement. Hence, the core doesn't need to communicate with the others, and the atomic read is about as fast as a non-atomic read.

If you dislike this style of coding, we can remove the double read here -- the other stuff in context is expensive enough that this will be unmeasureable anyways.

bbrehm: I wouldn't even bat an eye at a simple `volatile` that could emit e.g. 15 log messages if threads race...

If you think that this confuses future readers, I can change it, your call on readability.

mpollmeier: Ah right yeah makes sense. I'll suggest a comment

mpollmeier, line 44:
```suggestion
          // duplicate check looks (and is) superfluous, but it's a fastpath micro optimization
          if(hasLogged.get() > 0 && hasLogged.decrementAndGet() > 0) {
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/codepencegraph/CdgPass.scala
mpollmeier, line 25:
```suggestion
  // 10 is just an arbitrary number - we merely want to log 'a few times' but no more than that
  val hasLogged = java.util.concurrent.atomic.AtomicInteger(10)
```

mpollmeier (approved):

PR #5234, opened 2025-01-17
[c2cpg] Refactored code string calculation for locals

max-leuthaeuser:
The local code field is now the actual source code for that declaration. Not something we reconstruct from its type. Never lose static again.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
@@ -142,18 +142,14 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
       "class"
     )
 
-  private val KeywordsAtTypesToKeep: List[String] = List("unsigned", "volatile", "static", "const")
+  private val KeywordsAtTypesToKeep: List[String] = List("unsigned", "volatile")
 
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -142,18 +142,14 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
-      ReservedKeywordsAtTypes.foldLeft(rawType) { (cur, repl) =>
-        if (cur.startsWith(s"$repl ") || cur.contains(s" $repl ")) {
-          cur.replace(s" $repl ", " ").replace(s"$repl ", "")
-        } else cur
-      }
-    } else {
-      rawType
+    val tpe = ReservedKeywordsAtTypes.foldLeft(rawType) { (cur, repl) =>
+      if (cur.startsWith(s"$repl ") || cur.contains(s" $repl ")) {
+        cur.replace(s" $repl ", " ").replace(s"$repl ", "")
```
maltek, line 151:
```suggestion
        cur.replace(s" $repl ", " ").stripPrefix(s"$repl ")
```

maltek (commented):
how does this end up looking when multiple variabled are declared at the same time? say `int x, *y, z[1];`

maltek (approved):

PR #5237, opened 2025-01-19
[c#] basic support for ImplicitUsings

xavierpinho:
Introduces an initial mechanism to handle implicit global imports enabled by `<ImplicitUsings>` in `.csproj` files. Behind the scenes, Roslyn generates a set of `global using XXX` directives based on the project type. In this PR these directives are globally collected and included as `globalImports` to `CSharpProgramSummary`. Remaining changes are just minor refactoring tweaks.


DavidBakerEffendi (approved):
Interesting config/use case, nice!

PR #5242, opened 2025-01-21
[c2cpg] Remaining C++20 features

max-leuthaeuser:


joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/features20/Cpp20FeaturesTests.scala
```diff
@@ -210,59 +211,103 @@ class Cpp20FeaturesTests extends AstC2CpgSuite(fileSuffix = FileDefaults.CppExt)
...
-          |    // return [=]() { return n; };
-          |
-          |    // GOOD:
           |    return [=, *this]() { return n; };
           |  }
           |};
           |""".stripMargin)
-      ???
+      // TODO: we can not express these lambda types in the current schema
+      // We would need to add a new type for lambdas that capture `this`
```
maltek, line 306: At the end of the day, we'll want a unique type for every lambda / method reference just like we do it for lambdas in other frontends. I believe differently to those other languages this actually matches C++ semantics, where the return type of `getter_fn` is an "unnamable" lambda type that's unique for each lambda. (But we'd want to do it anyway, because we can't resolve dynamic calls of such functions otherwise.)

But that's just side-commentary, not a blocker.

maltek (approved):

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/features20/Cpp20FeaturesTests.scala
max-leuthaeuser: Yes, that's the plan. Just need to figure out how to get the name/fullname for such a lambda during the traversal of the method scope it's located in. At this point during the traversal we have not seen the lambda yet.

PR #5243, opened 2025-01-22
[javasrc2cpg] Fixed logged debug command suggestion.

ml86:
A '$' was missing to be able to directly copy-paste the logged command
suggestion.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/MavenDependencies.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/MavenDependencies.scala
@@ -42,7 +42,7 @@ object MavenDependencies {
         "The compile class path may be missing or partial.\n" +
         "Results will suffer from poor type information.\n" +
         "To fix this issue, please ensure that the below command can be executed successfully from the project root directory:\n" +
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/MavenDependencies.scala
```diff
@@ -42,7 +42,7 @@ object MavenDependencies {
         "The compile class path may be missing or partial.\n" +
         "Results will suffer from poor type information.\n" +
         "To fix this issue, please ensure that the below command can be executed successfully from the project root directory:\n" +
-        s"mvn $MavenCliOpts " + fetchArgs.mkString(" ") + "\n\n"
+        s"mvn $$MavenCliOpts " + fetchArgs.mkString(" ") + "\n\n"
```
johannescoetzee, line 45: We want this to output `$MAVEN_CLI_OPTS`, so I think the correct fix would be:
```suggestion
        s"mvn $$$MavenCliOpts " + fetchArgs.mkString(" ") + "\n\n"
```

johannescoetzee (changes requested):

johannescoetzee (approved):

PR #5247, opened 2025-01-23
[c#] cleaning up around "this" and composeMethodFullName

xavierpinho:
Making sure we (a) use `composeMethodFullName` every time we build a `methodFullName`; (b) use `Constants.This`.


max-leuthaeuser (approved):

DavidBakerEffendi (approved):

xavierpinho:
Re-running jobs. Failed on php2cpg:
```
[info] Detecting minified files
[info] - should skip minified files by name correctly
[info] - should skip minified files by content correctly
[info] - should parse all namespaces and move the files to the correct locations *** FAILED ***
[info]   0 was not greater than 0 (PhpDownloadDependenciesTest.scala:86)
[info] a PHP project with dependencies with a non-semver like version
[info] - should resolve and still get all namespaces
```

PR #5256, opened 2025-01-27
[sarif] Sarif Conversion from Finding Node 

DavidBakerEffendi:
* Implemented an extensible SARIF 2.1.0 conversion from Finding Nodes. 
* The actual design consists of SARIF-themed traits with the minimal properties we "want", followed by a version-specific case class implementation which may extend upon the base properties.
* The query step has an implicit configuration which allows one to rename the tool, version, description, etc. related component of what is exported. 
* Test posts a sample SARIF file generated by `joern-scan` + `toSarifJson` to an online Microsoft SARIF validator.

Question to reviewers: Do we want the structure of a trait which is extended upon by versioned case classes? Or do we only want to support one version at a given time?


DavidBakerEffendi:
Example file to be viewed at https://microsoft.github.io/sarif-web-component/
[Uploading outfile.sarif.json…]()

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifConfig.scala
```diff
@@ -0,0 +1,26 @@
...
+import org.json4s.Serializer
+
+import java.net.URI
+
+case class SarifConfig(
+  toolName: String = "Joern",
+  toolFullName: String = "Joern - The Bug Hunter's Workbench",
+  toolInformationUri: URI = URI("https://joern.io"),
+  organization: String = "Joern.io",
+  semanticVersion: String = "4.x.x",
```
DavidBakerEffendi, line 14: I see the Joern REPL has a `version` call, which I can use to dynamically load this - one sec...

mpollmeier: So `semanticVersion` refers to the joern version? 
If you get that from joern via e.g. `io.joern.joerncli.console.JoernConsole.version`, then ocular would also refer to the joern version, which is probably fine. 
That can be changed, but at the same time we should keep things simple... 
just to make sure I understand correctly, can you elaborate on what that version means?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifSchema.scala
```diff
@@ -0,0 +1,315 @@
...
+
+    /** @return
+      *   The SARIF format version of this log file.
+      */
+    def version: String
+
+    /** @return
+      *   The URI of the JSON schema corresponding to the version.
+      */
+    def `$schema`: String
```
mpollmeier, line 24: I guess this beautiful member name stems from the name it needs to have in the json? might be worth documenting the fact that a future refactoring must not change these names, unless the serialization is adapted.

tests/finding-to-sarif-test.sh
```diff
@@ -0,0 +1,19 @@
...
+set -euo pipefail
+
+SCRIPT_ABS_PATH=$(readlink -f "$0")
+JOERN_TESTS_DIR=$(dirname "$SCRIPT_ABS_PATH")
+JOERN="$JOERN_TESTS_DIR"/..
+
+./joern-scan "$JOERN_TESTS_DIR/code/sarif-test" --store
+./joern --script "$JOERN_TESTS_DIR/test-sarif.sc" --param cpgFile="$JOERN/workspace/sarif-test/cpg.bin" --param outFile="/tmp/sarif/test.sarif"
+
+mkdir -p /tmp/sarif
```
mpollmeier, line 12:
```suggestion
mkdir -p /tmp/sarif
./joern-scan "$JOERN_TESTS_DIR/code/sarif-test" --store
./joern --script "$JOERN_TESTS_DIR/test-sarif.sc" --param cpgFile="$JOERN/workspace/sarif-test/cpg.bin" --param outFile="/tmp/sarif/test.sarif"
```

mpollmeier: just moved mkdir two lines up

tests/test-sarif.sc
```diff
@@ -0,0 +1,7 @@
+// to test, run e.g.
+// ./joern --script test-sarif.sc --param cpgFile=workspace/src/cpg.bin --param outFile=test.sarif
+
+@main def exec(cpgFile: String, outFile: String) = {
+  importCpg(cpgFile)
+  cpg.finding.toSarifJson() |> outFile
```
mpollmeier, line 6:
```suggestion
  assert(cpg.finding.nonEmpty, "no findings in this cpg - please check the setup")
  cpg.finding.toSarifJson() |> outFile
```

mpollmeier (commented):
this looks nice and clean 👍🏻 
left a few remarks / suggestions

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifConfig.scala
DavidBakerEffendi: This version refers to the tool. I have a branch on `Ocular` which implements this work and runs this configuration before the REPL begins: https://github.com/ShiftLeftSecurity/ocular/compare/master...dave/ocular-sarif#diff-36b4f70cbc3441ca2033a3f58698e662fabe8f4e3fcaae662f7ba3c414782d00R55

I should likely name `organization` and `semanticVersion` more consistently with the `tool` prefix

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifSchema.scala
DavidBakerEffendi: Yes this is required by the JSON schema. I can opt for a more sane name and adapt the serialization too.

mpollmeier: if it's straightforward, I'd be 👍🏻 
otherwise just add a comment

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/SarifExtension.scala
```diff
@@ -0,0 +1,66 @@
...
+  *   the findings
+  * @see
+  *   https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html
+  */
+class SarifExtension(val traversal: Iterator[Finding]) extends AnyVal {
+
+  @Doc(info = "execute this traversal and convert findings to SARIF format")
+  def toSarif(implicit config: SarifConfig = SarifConfig()): Sarif = {
+
+    def generateSarif(results: List[SarifSchema.Result] = Nil, baseUri: Option[URI] = None): Sarif = {
```
ml86, line 26: The default parameter values are not really necessary here right? We have two call sites for this method. For clarity we can just spell out the arguments.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/SarifExtension.scala
```diff
@@ -0,0 +1,66 @@
...
+            v2_1_0.Schema.Run(
+              tool = v2_1_0.Schema.Tool(driver =
+                v2_1_0.Schema.ToolComponent(
+                  name = config.toolName,
+                  fullName = config.toolFullName,
+                  organization = config.organization,
+                  semanticVersion = config.semanticVersion,
+                  informationUri = config.toolInformationUri
+                )
+              ),
```
ml86, line 39: Please introduce a local variable for `tool`. This is hard to read.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifSchema.scala
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+  val serializers: List[Serializer[?]] = List(
+    new CustomSerializer[SarifSchema.Sarif](implicit format =>
+      (
+        { case _ =>
+          ???
+        },
+        { case sarif: SarifSchema.Sarif =>
+          Extraction.decompose(Map("version" -> sarif.version, "$schema" -> sarif.`$schema`, "runs" -> sarif.runs))
```
ml86, line 300: Why do we need the ugly `$schema` member if we have this custom decompose step which can do and already has a mapping?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifSchema.scala
```diff
@@ -0,0 +1,315 @@
...
+        case score =>
+          logger.error(s"Score '$score' is not a valid CVSS score! Defaulting to 'warning' SARIF level.")
+          Warning
+      }
+    }
+
+  }
+
+  val serializers: List[Serializer[?]] = List(
+    new CustomSerializer[SarifSchema.Sarif](implicit format =>
```
ml86, line 294: Why do we only need customer serializers for some of the schema traits?

tests/finding-to-sarif-test.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
ml86, line 1: I think we should also have at least one proper scalatest unit test.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifConfig.scala
```diff
@@ -0,0 +1,26 @@
+package io.shiftleft.semanticcpg.sarif
+
+import io.shiftleft.semanticcpg.sarif.SarifConfig.SarifVersion
+import io.shiftleft.semanticcpg.sarif.v2_1_0.JoernScanResultToSarifConverter
+import org.json4s.Serializer
+
+import java.net.URI
+
+case class SarifConfig(
```
ml86, line 31: Do we really need a separate `SarifConfig` classs if we already have a "Sarif converter" which could easily contain the values from the config class in addition to the implementation it currently has.

ml86 (changes requested):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifConfig.scala
DavidBakerEffendi: My idea is to be consistent with the data-flow engine style of having a single constant config that the queries pick up on in the REPL. The configuration itself rarely changes, and is only necessary to configure once for compulsory tool-related meta-data. 

In the scope of a REPL, the configuration will generally not change, but one may call `toSarifJson` multiple times for possibly different scans.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifConfig.scala
DavidBakerEffendi: @mpollmeier I see that `joerncli` is not accessible here - where do you suggest I initialize this kind of config value for the REPL? I see there are some `runBefore`-style hooks in the BridgeBase interface but I'm wondering where specifically may be best

ml86 (approved):
Please also add further unit tests which check not just the SARIF case classes but instead the resulting SARIF json.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifConfig.scala
mpollmeier: 👍🏻 sounds good

mpollmeier (approved):

Issue #5254, opened 2025-01-26
[Bug] Exceptions on blender database


Bazmundi:
**Describe the bug**
I download blender database which is a mix of c++ (cc) and python.  Joern broke on the import of cc code, so I create a python code only copy.

Running against python only I am getting:

```
2025-01-26 12:14:38.429 WARN CodeToCpg$: Failed to convert file examples/ocd_addon_1_8_2/__MACOSX/ocd_addon/.___init__.py
io.joern.pythonparser.TokenMgrError: Lexical error at line 1, column 1.  Encountered: '0' (0),
	at io.joern.pythonparser.PythonParserTokenManager.getNextToken(PythonParserTokenManager.java:3692) ~[io.joern.pysrc2cpg-4.0.226.jar:4.0.226]
	at io.joern.pythonparser.PythonParser.jj_ntk_f(PythonParser.java:7599) ~[io.joern.pysrc2cpg-4.0.226.jar:4.0.226]
	at io.joern.pythonparser.PythonParser.module(PythonParser.java:179) ~[io.joern.pysrc2cpg-4.0.226.jar:4.0.226]
	at io.joern.pythonparser.PyParser.parse(PyParser.scala:23) ~[io.joern.pysrc2cpg-4.0.226.jar:4.0.226]
	at io.joern.pythonparser.PyParser.parse(PyParser.scala:15) ~[io.joern.pysrc2cpg-4.0.226.jar:4.0.226]
	at io.joern.pysrc2cpg.CodeToCpg.runOnPart(CodeToCpg.scala:25) [io.joern.pysrc2cpg-4.0.226.jar:4.0.226]
	at io.joern.pysrc2cpg.CodeToCpg.runOnPart(CodeToCpg.scala:20) [io.joern.pysrc2cpg-4.0.226.jar:4.0.226]
	at io.shiftleft.passes.ForkJoinParallelCpgPass$$anon$2.accept(CpgPass.scala:124) [io.shiftleft.codepropertygraph_3-1.7.23.jar:1.7.23]
	at io.shiftleft.passes.ForkJoinParallelCpgPass$$anon$2.accept(CpgPass.scala:123) [io.shiftleft.codepropertygraph_3-1.7.23.jar:1.7.23]
	at java.util.stream.ReduceOps$4ReducingSink.accept(ReduceOps.java:220) [?:?]
	at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:992) [?:?]
[...]

PR #5259, opened 2025-01-27
[c2cpg] Multiple crash fixes for blender sources

max-leuthaeuser:
Seen during testing on https://github.com/blender/blender

For: https://github.com/joernio/joern/issues/5254


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -78,7 +80,10 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
 
   private def astForCppCallExpression(call: ICPPASTFunctionCallExpression): Ast = {
     val functionNameExpr = call.getFunctionNameExpression
-    val typ              = functionNameExpr.getExpressionType
+    val typ = Try(functionNameExpr.getExpressionType) match {
+      case Failure(_)   => return astForCppCallExpressionUntyped(call)
```
mpollmeier, line 84: a shortpath `return statement` isn't considered 'nice' scala code and may even result in a new compiler warning - consider rewriting into something like 

```
Try(functionNameExpr.getExpressionType).map { typ => 
...
}.getOrElse(astForCppCallExpressionUntyped(call))
```

mpollmeier (approved):

PR #5274, opened 2025-01-30
[javasrc2cpg] Populate generic signatures roughly following the class file signature format

johannescoetzee:
The generic signatures added by this PR follow the format described in https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.7.9.1 and largely correspond to the generic signatures in class files with some notable differences:

### Class type signatures 
* In most cases, only the simple name for the class will be used (so `LString;` will be used instead of `Ljava/lang/String`)
* Where a qualified name is used in source, that name is used verbatim in the signature, for example `Ljava.util.List` (note the `.` were not substituted for `/`. 
* For local classes, the name of the class as it appears in the CPG is used in the signature for instances of that class (we don't follow the JVM naming scheme for these), for example `Ltestpackage.TestClass.testMethod.LocalClass;`

### Type parameter bounds
From the language specification:
```
TypeParameter:
  Identifier ClassBound {InterfaceBound}

ClassBound:
 : [ReferenceTypeSignature]

InterfaceBound:
  : ReferenceTypeSignature
```
If a type parameter only has interface bounds I1, I2, ..., then the signature should contain `<T::LI1;:...>` (note the empty class bound), but in general we won't know if a type is a class or interface without resolving the it, so the signature in the CPG will contain `<T:LI1;:...>` instead.

### Unspecified types
Where no type name is specified, the special `L__unspecified_type;` type is used in generic signatures. This happens in a few places:
* For lambda return types and lambda parameters which do not have explicit type annotations
* For locals with a `var` type, for example `var x = 42`
* For synthetic locals created for `foreach` loops, for example in `for (String item : items())`, we create a temporary `String[] $iterLocal0 = items()` local which will have an unspecified signature (`item` will still have the signature `LString;` as expected)
* For synthetic locals created for the LHS of `instanceof` expressions with pattern matching, for example in `foo() instanceof String s`, we create an `Object o = foo()` local (since the type depends on the return type of `foo`).

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/GenericSignatureTests.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/GenericSignatureTests.scala
@@ -1,0 +1,1000 @@
+package io.joern.javasrc2cpg.querying
+
+import io.joern.javasrc2cpg.testfixtures.JavaSrcCode2CpgFixture
+import io.shiftleft.semanticcpg.language.*
+
+class GenericSignatureTests extends JavaSrcCode2CpgFixture {
+
+  "a simple example with primitive types" should {
+    val cpg = code("""
+        |package test;
+        |
+        |class Test {
+        |  char charMember;
+        |
+        |  public void test(boolean b) {
+        |    int x;
+        |  }
+        |}
+        |""".stripMargin)
+
+    "have the correct generic signature for locals" in {
+      cpg.local.genericSignature.l shouldBe List("I")
+    }
+
+    "have the correct generic signature for a void method with a boolean arg" in {
+      cpg.method.name("test").genericSignature.l shouldBe List("(Z)V")
+    }
+
+    "have the correct generic signature for the parameter" in {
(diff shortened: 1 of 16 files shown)


joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/GenericSignatureTests.scala
```diff
@@ -0,0 +1,1042 @@
...
+        |  }
+        |}
+        |""".stripMargin)
+
+    "have the correct generic signature for the lambda method" in {
+      cpg.method.name(".*lambda.*").genericSignature.l shouldBe List("(L__unspecified_type;)L__unspecified_type;")
+    }
+
+    "have an empty generic signature for the lambda type decl" in {
+      cpg.typeDecl.name(".*lambda.*").genericSignature.l shouldBe List("<empty>")
```
ml86, line 269: Isnt a single `L__unspecified_type` more appropriate?

ml86 (approved):
Please put the description from the PR to into the GenericSignatureTests file so that we are aware of the deviations.

PR #5281, opened 2025-02-03
[JsSrc2Cpg] - Handle runtime exception gracefully when filtering astGen result

khemrajrathore:


max-leuthaeuser:
How did an invalid path end up there? What do you mean by "invalid"?
Could you provide any test case?

khemrajrathore:
My bad, the PR title was incorrect. This PR is to avoid running into any runtime exception if any and instead handle the exception gracefully. Have updated the PR

max-leuthaeuser (approved):

khemrajrathore:
Can we get this merged?

PR #5282, opened 2025-02-03
[JsSrc2Cpg] : Increase robustness of Js frontend

khemrajrathore:
The PR have the following changes
- In case of failure in AST Gen step, we should continue with partial result at that time
- The exclusion regex should work and only send files to AST Gen which are not excluded by the exclusion rule


max-leuthaeuser:
This is not needed at all. astgen already respects exclusions since https://github.com/joernio/astgen/pull/24 and jssrc2cpg passes them as well: https://github.com/joernio/joern/blob/master/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala#L180

khemrajrathore:
Can we get this merged?

PR #5290, opened 2025-02-04
[c#] initial support for replacing setter assignments with setter calls

xavierpinho:
Renders setter assignments as their corresponding `set_PropertyName` calls, e.g. `x.MyProperty = y` roughly becomes `x.set_MyProperty(y)`. There are still some edges to be addressed in future PRs; in particular, we should have a general solution for building signatures so we don't keep falling for them.


joern-cli/frontends/csharpsrc2cpg/src/main/scala/io/joern/csharpsrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -117,47 +196,61 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
     callAst(callNode, argsAst) :: Nil
   }
 
-  protected def astForBinaryExpression(binaryExpr: DotNetNodeInfo): Seq[Ast] = {
+  private def translateBinaryOperatorName(operatorToken: String): Option[String] = {
+    Map(
```
max-leuthaeuser, line 200: This map could statically defined once in the companion object.

DavidBakerEffendi (approved):

PR #5294, opened 2025-02-05
[c#] fix synthetic set_* method call signature

xavierpinho:
By re-using `createInvocationAst`, we can start fixing the remaining incorrect signatures.


xavierpinho:
Merged in #5295

PR #5307, opened 2025-02-13
[ruby] Persist JRuby Script Container with Environment 

DavidBakerEffendi:
Persisting the lazily created JRuby execution environment and loaded global variables with cleaning up only happening at shutdown.

This removes the overhead of JRuby instantiating a Ruby environment for each RubySrc2Cpg.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/Main.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/Main.scala
@@ -21,9 +21,9 @@ final case class Config(downloadDependencies: Boolean = false, useTypeStubs: Boo
   override val astGenConfigPrefix: String       = "rubysrc2cpg"
   override val multiArchitectureBuilds: Boolean = true
 
-  this.defaultIgnoredFilesRegex = List("spec", "tests?", "vendor", "db(\\\\|/)([\\w_]*)migrate([_\\w]*)").flatMap {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/RubySrc2Cpg.scala
```diff
@@ -44,7 +45,12 @@ class RubySrc2Cpg extends X2CpgFrontend[Config] {
 
   private def createCpgAction(cpg: Cpg, config: Config): Unit = {
     File.usingTemporaryDirectory("rubysrc2cpgOut") { tmpDir =>
-      val astGenResult = RubyAstGenRunner(config).execute(tmpDir)
+      val astGenResult = persistedRubyAstGenRunner match {
+        case Some(astGenRunner) =>
+          astGenRunner.execute(tmpDir, config)
+        case None =>
+          Using.resource(RubyAstGenRunner(config))(_.execute(tmpDir))
+      }
```
ml86, line 53: If we create the runner here on demand for the non server scenario, when not always create it here on demand?

And i suggest we let X2CpgFrontend extend `AutoClosable` so that the `close` call can be handled be X2CpgFrontend and not in the individual frontend Main classes.

ml86 (changes requested):

ml86 (approved):
Please add a comment on the RubyAstGenRunner which documents why we do it this way ( dont forget to mention the Windows problems).

PR #5308, opened 2025-02-13
fix tab-completion in distribution

mpollmeier:
the `removeModuleInfoFromJars` hack only get's triggered for
`Universal/stage` - our release uses `Universal/packageBin` and that
doesn't seem to reuse anything related that we could trigger on
unfortunately...
So I thought it's easiest to just add the same hack to the start
script...

fixes https://github.com/joernio/joern/issues/4625

### DIFF
--- a/joern-cli/build.sbt
+++ b/joern-cli/build.sbt
@@ -133,15 +133,15 @@ Universal/mappings ++= sbt.Path.directory(new File("joern-cli/src/main/resources
 
 // remove module-info.class from dependency jars - a hacky workaround for a scala3 compiler bug
 // see https://github.com/scala/scala3/issues/20421
-val moduleInfoLocation = "/module-info.class"
+val moduleInfoLocation = "module-info.class"
 Universal/mappings := (Universal/mappings).value.collect {
   case (jar, location)
       if location.startsWith("lib")
       && location.endsWith(".jar")
-      && FileUtils.jarContainsEntry(jar, moduleInfoLocation, onlyAtRootLevel = true) =>
+      && FileUtils.jarContainsEntryInRoot(jar, moduleInfoLocation) =>
     val newJar = target.value / "without-module-info" / jar.getName()
     IO.copyFile(jar, newJar)
-    FileUtils.removeJarEntry(newJar, moduleInfoLocation, onlyAtRootLevel = true)
+    FileUtils.removeJarEntryFromRoot(newJar, moduleInfoLocation)
     streams.value.log.info(s"workaround for scala completion bug: including a modified version of $jar without the $moduleInfoLocation entry: $newJar")
     newJar -> location
   case other =>
--- a/project/FileUtils.scala
+++ b/project/FileUtils.scala
@@ -19,27 +19,22 @@ object FileUtils {
     }
   }
 
-  /** checks if the given jar contains the given entry */
-  def jarContainsEntry(jar: File, entry: String, onlyAtRootLevel: Boolean = false): Boolean = {
+  /** checks if the given jar contains the given entry in the root directory
+    */
+  def jarContainsEntryInRoot(jar: File, entry: String): Boolean = {
     val zipFs = FileSystems.newFileSystem(jar.toPath)
     val result = zipFs.getRootDirectories.asScala.exists { zipRootDir =>
-      val iterable =
-        if (onlyAtRootLevel) Files.list(zipRootDir)
-        else Files.walk(zipRootDir)
-      iterable.iterator.asScala.exists(_.toString == entry)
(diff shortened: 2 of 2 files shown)


joern-cli/src/universal/joern
```diff
@@ -13,5 +13,11 @@ if [ ! -f "$SCRIPT" ]; then
     exit 1
 fi
 
+for jar in lib/*.jar; do
```
maltek, line 16: this should be relative to `SCRIPT_ABS_DIR`, I think

maltek (changes requested):
This doesn't cover the Windows `joern.bat` :( And it's probably much harder to implement there... maybe the Github Actions script would be a most convenient place to do this step? (Not a fan of putting too much logic there, but where else could we do it if `sbt` can't do it?)

Seems that this same script is also used for local runs after `sbt stage`, so if we keep doing these rewrites in the launcher script we should then remove the equivalent hacks from `build.sbt`.

mpollmeier:
Good points, I had a better idea and changed the entire approach: now we're adapting the mappings, which affects both `stage` and `Universal/packageBin`. Tested locally with both. 
Please re-review

project/FileUtils.scala
```diff
@@ -19,4 +19,30 @@ object FileUtils {
...
   }
 
+  /** checks if the given jar contains the given entry */
+  def jarContainsEntry(jar: File, entry: String, onlyAtRootLevel: Boolean = false): Boolean = {
+    val zipFs = FileSystems.newFileSystem(jar.toPath)
+    val result = zipFs.getRootDirectories.asScala.exists { zipRootDir =>
+      val iterable =
+        if (onlyAtRootLevel) Files.list(zipRootDir)
+        else Files.walk(zipRootDir)
+      iterable.iterator.asScala.exists(_.toString == "/module-info.class")
```
maltek, line 29:
```suggestion
      iterable.iterator.asScala.exists(_.getFileName == entry)
```

`onlyAtRootLevel=false` would break with `toString`. (Or we just remove the flag, since the code path is not used?)

project/FileUtils.scala
```diff
@@ -19,4 +19,30 @@ object FileUtils {
...
+  }
+
+  /** removes the given entry from the given jar */
+  def removeJarEntry(jar: File, entry: String, onlyAtRootLevel: Boolean = false): Unit = {
+    val zipFs = FileSystems.newFileSystem(jar.toPath)
+    zipFs.getRootDirectories.forEach { zipRootDir =>
+      val iterable =
+        if (onlyAtRootLevel) Files.list(zipRootDir)
+        else Files.walk(zipRootDir)
+      iterable.filter(_.toString == entry).forEach(Files.delete(_))
```
maltek, line 42:
```suggestion
      iterable.filter(_.getFileName == entry).forEach(Files.delete(_))
```

joern-cli/build.sbt
```diff
@@ -129,28 +129,23 @@ generateScaladocs := {
...
-          Files.delete(moduleInfoClass)
-        }
-      }
-      zipFs.close()
-    }
+Universal/mappings ++= sbt.Path.directory(new File("joern-cli/src/main/resources/scripts"))
+
+// remove module-info.class from dependency jars - a hacky workaround for a scala3 compiler bug
+// see https://github.com/scala/scala3/issues/20421
+val moduleInfoLocation = "/module-info.class"
```
maltek, line 136: with the other changes, this then needs to be a plain file name without path (which also makes me more confident it won't break on Windows)

```suggestion
val moduleInfoLocation = "module-info.class"
```

maltek (changes requested):
I like this approach much more 👍

project/FileUtils.scala
mpollmeier: Not quite - if we use 'getFileName' we cannot delete files only from select directories. 

Your suggestion as-is would not delete anything, because the filename is actually `module-info.class` (without the `/`). 
And if we use your suggestion and delete `module-info.class` then we'd touch many more jars, because they contain e.g. a `META-INF/9/module-info.class` which doesn't actually harm. And given that this is quite hacky I'd rather keep the changes as small as possible.

joern-cli/build.sbt
mpollmeier: this is the path in java's ZipFilesystem which is identical for windows.

project/FileUtils.scala
maltek: then the `onlyAtRootLevel` code path is incredibly confusing, and should be deleted on those grounds :)

mpollmeier:
updated to simplify, verified locally, please re-review

maltek (approved):

PR #5310, opened 2025-02-14
[c#] support chained `?.` expressions

xavierpinho:
`?.` expressions were being handled as if they were `.`, but the implementation was still in its infancy, e.g. `x?.Foo()` wasn't handled, and in chained expressions `x?.Foo?.Bar`, we would only take the 1st and last members, yielding `x.Bar`.

This PR extends this support to handle those scenaria. That said, in the long run, we might actually want to lower `?.` as an `<operator>.safeNavigation`, or rewrite them using `if` expressions, like Roslyn would do. 

Question to the reviewers: rewriting Json Asts is no fun. Would a PR providing a dedicated C# AST be welcomed? (Similar to `rubysrc2cpg`'s)


DavidBakerEffendi (approved):
I would generally want to go the "lower-to-if-statement" route - the semantics are much clearer here. I can foresee this being a headache, so if adding an intermediate AST layer would help, then by all means go ahead. Another route would be a dedicated `AstForLoweringCreator` to at least separate lowering code, but yeah not much less painful.

PR #5312, opened 2025-02-17
Remove afterEvaluate in gradle dependency fetcher init script

johannescoetzee:
We ran into the following gradle error when fetching dependencies for a customer project:
```
Failed to apply plugin 'org.gradle.java-test-fixtures'.    

> Project#afterEvaluate(Action) on project '<project-name>' cannot be executed in the current context
```

This PR attempts to avoid this issue by removing the `afterEvaluate {...}` call in the init script we use and doing the task registration whenever `allprojects { project -> ... }` would do it. I wasn't able to reproduce the issue, so I don't know if this will fix the customer's issue or not. I'm also no gradle expert, so I don't know what the full implications of removing `afterEvaluate` are, but the test projects I've tried have all worked equally well with and without it. Most of the logic should be handled lazily, as far as I understand, but there could be issues with certain project configurations that I just haven't found yet.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/GradleDependencies.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/GradleDependencies.scala
@@ -90,44 +90,42 @@ object GradleDependencies {
     val dependsOnAndroidTask = Option.when(projectInfo.hasAndroidSubproject)("dependsOn androidDepsCopyTaskName")
 
     s"""
-       |allprojects {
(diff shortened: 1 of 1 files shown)


mpollmeier (approved):
not a gradle expert either, so just rubber stamping really...
re reproduction: do we know the exact command and the gradle version they used?

johannescoetzee:
> not a gradle expert either, so just rubber stamping really... re reproduction: do we know the exact command and the gradle version they used?

We do! Gradle is invoked via the API in `GradleDependencies.scala` and they were using 
```
Gradle 7.5.1
Java version: 11.0.21
```
I did try this Gradle version with Java 11, but no luck

PR #5313, opened 2025-02-17
[c2cpg] Closure/Capturing rework

max-leuthaeuser:
So basically we are now using the rather simple `createVariableReferenceLinks` that runs after the AST generation just like we do in JS and very similar in Python. No more special handling for lambdas. ClosureBindings and Captures are done on one step there.

- sptests diff is looking good.
- one codescience C test fails because of the now broken isGlobal check in https://github.com/ShiftLeftSecurity/codescience/blob/680fdafa1eb4a9272810a42746129427e6717fd5/securityprofile/src/main/scala/io/shiftleft/securityprofile/passes/finalcustom/c/generic/DeleteGlobalNoReassignment.scala#L31

Also: fixed they way names for anonymous elements are created to be more in line with other frontends.


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -24,13 +25,19 @@ import scala.collection.mutable
 
 trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { this: AstCreator =>
 
-  protected def astForBlockStatement(blockStmt: IASTCompoundStatement, order: Int = -1): Ast = {
-    val codeString = code(blockStmt)
-    val blockCode  = if (codeString == "{}" || codeString.isEmpty) Defines.Empty else codeString
-    val node = blockNode(blockStmt, blockCode, registerType(Defines.Void))
+  protected def astForBlockStatement(blockStmt: IASTCompoundStatement, blockNode: NewBlock, order: Int = -1): Ast = {
+    val codeString  = code(blockStmt)
+    val blockCode   = if (codeString == "{}" || codeString.isEmpty) Defines.Empty else codeString
```
maltek, line 30: I see that it's existing behavior, but I don't think this is helpful: there is a difference between a unknown/missing value (`<empty>`), and an actual empty value. So we should just use `codeString` here universally.

maltek (approved):
today's fixes could use some regression tests, I think. At least I I couldn't tell just from the code which exact C++ construct is handled there.

LGTM otherwise.

max-leuthaeuser:
I tried for hours to replicate the issue from the FPs in sptests in standalone unit tests but failed. It looks like a combination of includes in these projects that lead to the problematic type resolution in CDT that trigger the missing method and member references. I guess we have to rely on sptests.

mpollmeier:
@max-leuthaeuser the CS integrationtests are currently red, and local bisecting points to this PR as the root cause. Steps to reproduce:

```bash
cd joern
git checkout 3f735ff
sbt c2cpg/stage

cd ../codescience
sbt "securityprofileTests/C/testOnly  io.shiftleft.securityprofile.passes.finalcustom.c.generic.DeleteGlobalNoReassignmentTest"
# [info] All tests passed.
# [success] Total time: 5 s, completed 10 Mar 2025, 12:32:50

# now make it red
cd ../joern
git checkout 21e550d
sbt c2cpg/stage

cd ../codescience
sbt "securityprofileTests/C/testOnly  io.shiftleft.securityprofile.passes.finalcustom.c.generic.DeleteGlobalNoReassignmentTest"

# output:
[info] DeleteGlobalNoReassignment pass
[info] - should find exactly one thing in a basic case *** FAILED ***
[info]   Array() did not contain the same elements in the same (iterated) order as Array("┌─────┬────────┬────┬─────────────────┬─────────┬─────────┬─────┬────┐
[info]   │stack│tracked │line│method           │file     │trackedAP│extAP│tags│
[info]   ├─────┼────────┼────┼─────────────────┼─────────┼─────────┼─────┼────┤
[info]   │|<   │ret     │N/A │<operator>.delete│<unknown>│         │     │    │
[info]   │|    │delete x│8   │bar              │foo.cc   │         │     │    │
[info]   └─────┴────────┴────┴─────────────────┴─────────┴─────────┴─────┴────┘") (DeleteGlobalNoReassignmentTest.scala:55)
[info] Run completed in 2 seconds, 220 milliseconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 0, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
```

just like in https://ci.shiftleftsecurity.com/job/TEST-Integration-Codescience/6066/

max-leuthaeuser:
Yes, this is expected and the fix is here: https://github.com/ShiftLeftSecurity/codescience/pull/8174 (see the comments there)

PR #5314, opened 2025-02-17
[c#] fix `.code` property for calls with comments surrounding them

xavierpinho:
Fixes `.code` for the following statement:

```c#
// some comment
System.Console.Out.WriteLine("X");
```

in which we currently get

```scala
cpg.call.nameExact("WriteLine").head.code == "// some comment"
```


max-leuthaeuser (commented):
What about calls with comments where the call itself spans multiple lines? 
Can other types of line breaks appear? (Windows)

PR #5316, opened 2025-02-19
[ruby] Fix Regex Lowering Bug

DavidBakerEffendi:
Fixed a bug introduced by converting all regex match methods by the `~=` binary statements. The bug is introduced is that this replaces ordinary calls (expressions) with binary statements leading to unexpected structures during AST creation.

This was done to prevent infinite lowering whenever a regex method is created (as a `match` call is part of this lowering, which in of itself is a regex method).

This change adds a "tag" to show when a call needs this lowering.


AndreiDreyer (approved):

DavidBakerEffendi:
Note to self: this is quite hacky, will follow up with a better process, i.e., creating a special RubyAst type

PR #5317, opened 2025-02-20
[WIP]: Replace `scala.sys.Process`

AndreiDreyer:
* Replaced instances of `scala.sys.Process` with `java.lang.ProcessBuilder`

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ExternalCommand.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ExternalCommand.scala
@@ -5,6 +5,7 @@ import io.shiftleft.utils.IOUtils
 import java.io.File
 import java.nio.file.{Path, Paths}
 import scala.jdk.CollectionConverters.*
+import scala.language.implicitConversions
 import scala.util.{Failure, Success, Try}
 import scala.util.control.NonFatal
 
@@ -26,16 +27,24 @@ object ExternalCommand {
     }
   }
 
-  def run(
+  def runWithDir(
     command: Seq[String],
-    cwd: String,
-    mergeStdErrInStdOut: Boolean = false,
-    extraEnv: Map[String, String] = Map.empty
+    cwd: Option[String],
+    mergeStdErrInStdOut: Boolean,
+    extraEnv: Map[String, String]
   ): ExternalCommandResult = {
-    val builder = new ProcessBuilder()
-      .command(command.toArray*)
-      .directory(new File(cwd))
-      .redirectErrorStream(mergeStdErrInStdOut)
+    val builder = cwd match {
+      case Some(dir) =>
+        new ProcessBuilder()
+          .command(command.toArray*)
+          .directory(new File(dir))
+          .redirectErrorStream(mergeStdErrInStdOut)
+      case _ =>
+        new ProcessBuilder()
+          .command(command.toArray*)
+          .redirectErrorStream(mergeStdErrInStdOut)
+    }
+
     builder.environment().putAll(extraEnv.asJava)
 
     val stdOutFile = File.createTempFile("x2cpg", "stdout")
@@ -59,6 +68,13 @@ object ExternalCommand {
       stdErrFile.foreach(_.delete())
     }
   }
+
+  def run(
+    command: Seq[String],
+    cwd: String,
+    mergeStdErrInStdOut: Boolean = false,
+    extraEnv: Map[String, String] = Map.empty
+  ): ExternalCommandResult = runWithDir(command, Some(cwd), mergeStdErrInStdOut, extraEnv)
 
   /** Finds the absolute path to the executable directory (e.g. `/path/to/javasrc2cpg/bin`). Based on the package path
     * of a loaded classfile based on some (potentially flakey?) filename heuristics. Context: we want to be able to
--- a/project/ExternalCommandUtil.scala
+++ b/project/ExternalCommandUtil.scala
@@ -1,0 +1,62 @@
+import java.io.{BufferedReader, File, InputStreamReader}
+import java.nio.file.{Path, Paths}
+import scala.jdk.CollectionConverters.*
+import scala.language.implicitConversions
+import scala.util.{Failure, Success, Try}
+import scala.util.control.NonFatal
+
+object ExternalCommandUtil {
+  case class ExternalCommandOutput(exitCode: Int, stdOut: Seq[String], stdErr: Seq[String]) {
+    def toTry: Try[ExternalCommandOutput] = {
+      exitCode match {
+        case 0 => Success(this)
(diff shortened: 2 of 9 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ExternalCommand.scala
```diff
@@ -26,16 +27,24 @@ object ExternalCommand {
     }
   }
 
-  def run(
+  def runWithDir(
```
DavidBakerEffendi, line 30: This looks like the below function, just that `cwd` is optional. Wouldn't it make sense to make this implementation function private, then have a version of this that excludes the `cwd` parameter? I'm also aware that some of these defaults have to be excluded because they look the same to the compiler. Tricky...

project/ExternalCommandUtil.scala
```diff
@@ -0,0 +1,62 @@
+import java.io.{BufferedReader, File, InputStreamReader}
+import java.nio.file.{Path, Paths}
+import scala.jdk.CollectionConverters.*
+import scala.language.implicitConversions
+import scala.util.{Failure, Success, Try}
+import scala.util.control.NonFatal
+
+object ExternalCommandUtil {
+  case class ExternalCommandOutput(exitCode: Int, stdOut: Seq[String], stdErr: Seq[String]) {
```
DavidBakerEffendi, line 9: Given that all of the usages here use the `option`/`try` variant, it may make sense to simplify the result of `run` to always return the `Option[ExternalCommandOutput]` varaint without `stdErr` (since I don't see `stdErr` used)

project/ExternalCommandUtil.scala
```diff
@@ -0,0 +1,62 @@
...
+        case _ => Failure(new RuntimeException(stdErr.mkString("\n")))
+      }
+    }
+  }
+
+  def run(cmd: String, cwd: Option[File] = None, isShellCmd: Boolean = true): ExternalCommandOutput = run(cmd.split(" ").toIndexedSeq, cwd, isShellCmd)
+
+  def run(cmd: Seq[String], cwd: Option[File], isShellCmd: Boolean): ExternalCommandOutput = {
+    val externalCmd = if (isShellCmd) {
+      Seq("sh", "-c") ++ cmd
```
DavidBakerEffendi, line 22: Might need to double-check the Windows compatibility for this

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ExternalCommand.scala
AndreiDreyer: Was still playing around with this one, but will give that a try.

max-leuthaeuser (commented):
I don't quite get it sorry.

- We now carry two almost identical implementation around. (semanticcpg and sbt)
- User downstream need to change the dependency from x2cpg to semanticcpg or at it additionaly
- Scala process is a wrapper around Java processbuilder anyway 
- Every invocation is now more complicated, e.g., if std out/err is not used and one only is interested in the return value
- Especially at the sbt side of things simply using Scala process looks cleaner

project/ExternalCommandUtil.scala
```diff
@@ -0,0 +1,48 @@
...
+
+    stdOutReader.close()
+
+    if (exitCode == 0) {
+      Some(ExternalCommandOutput(exitCode, out))
+    } else {
+      None
+    }
+  }
+}
```
ml86, line 48: This is confusing since this implementation is almost the same but not the same as `ExternalCommand`. 
I am with @max-leuthaeuser that we can stick to using the scala ProcessBuilder in the sbt files to avoid redefining out own java ProcessBuilder wrapper.
So only inside the actual Joern code the scala ProcessBuilder should be replaced not in the build files.
Sorry for not making that clear in the first place.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommand.scala
```diff
@@ -1,10 +1,11 @@
-package io.joern.x2cpg.utils
+package io.shiftleft.semanticcpg.utils
```
ml86, line 1: In order to not make this a breaking change, make `object ExternalCommand` a trait or class and with e.g. an `Impl` suffix and have two `object`s extend it, one in the old and one in the new namespace. Please put a deprecation warning on the old namespace version. That way we dont break anything without a headsup.

ml86 (changes requested):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/utils/GccSpecificExternalCommand.scala
```diff
@@ -11,7 +11,7 @@ object GccSpecificExternalCommand {
   private val IsWin = scala.util.Properties.isWin
 
   def run(command: Seq[String], cwd: String, extraEnv: Map[String, String] = Map.empty): Try[Seq[String]] = {
-    ExternalCommand.run(command, cwd, mergeStdErrInStdOut = true, extraEnv) match {
+    ExternalCommand.run(command, Some(cwd), mergeStdErrInStdOut = true, extraEnv) match {
```
DavidBakerEffendi, line 14: Just to avoid the `Some(null)` edge case
```suggestion
    ExternalCommand.run(command, Option(cwd), mergeStdErrInStdOut = true, extraEnv) match {
```

joern-cli/frontends/gosrc2cpg/src/main/scala/io/joern/gosrc2cpg/passes/DownloadDependenciesPass.scala
```diff
@@ -27,13 +27,13 @@ class DownloadDependenciesPass(cpg: Cpg, parentGoMod: GoModHelper, goGlobal: GoG
       parentGoMod
         .getModMetaData()
         .foreach(mod => {
-          ExternalCommand.run(Seq("go", "mod", "init", "joern.io/temp"), projDir).toTry match {
+          ExternalCommand.run(Seq("go", "mod", "init", "joern.io/temp"), Some(projDir)).toTry match {
```
DavidBakerEffendi, line 30: More `Some` instances here

DavidBakerEffendi: Safe to ignore the `Some` calls where you specify the string directly

DavidBakerEffendi: Found more elsewhere too

DavidBakerEffendi (approved):
I'm happy if everyone else is happy

ml86 (approved):

PR #5322, opened 2025-02-22
[c2cpg] Fixed constructor initializer handling

max-leuthaeuser:
For: https://shiftleftinc.atlassian.net/browse/SEN-3782


maltek (approved):
The RHS of the assignments still go nowhere, which isn't great. But it's progress and it fixes the immediate problem we saw.

PR #5323, opened 2025-02-23
[jimple2cpg]add final static member tag of class to member's code.

sfncat:
If a class has a member variable defined as final static, such a private static final String finalPrivateString = "PRIVATE_STATIC_FINAL"; when processed by Soot, a Tag with the value "PRIVATE_STATIC_FINAL" will be added to the Member finalPrivateString. In jimple2cpg AstForTypeDeclsCreator astForField,the tag not be handled. and the value also not be changed to an assignment call,the variable not be handled as local node or a literal node. so in the cpg, we can find the member variable, but can't find anything about it. Now add the variable contantValueTag to the member code, so we can get something about the variable. I add an issue(https://github.com/ShiftLeftSecurity/codepropertygraph/issues/1811) to suggest add a Property "TAG" for Member nodes. if Member nodes have Property "TAG", I can add the tag to the property tag.

### DIFF
--- a/joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/astcreation/declarations/AstForTypeDeclsCreator.scala
+++ b/joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/astcreation/declarations/AstForTypeDeclsCreator.scala
@@ -61,7 +61,19 @@ trait AstForTypeDeclsCreator(implicit withSchemaValidation: ValidationMode) { th
   private def astForField(field: SootField): Ast = {
     val typeFullName = registerType(field.getType.toQuotedString)
     val name         = field.getName
-    val code         = if (field.getDeclaration.contains("enum")) name else s"$typeFullName $name"
+    val code: String = if (field.getDeclaration.contains("enum")) {
+      name
+    } else if (field.isStatic && field.isFinal) {
+      val constantValue = field.getTags.asScala.collectFirst { case tag: ConstantValueTag =>
+        tag.getConstant.toString
+      }
+      constantValue match {
+        case Some(value) => s"$typeFullName $name = $value"
+        case None        => s"$typeFullName $name"
+      }
+    } else {
+      s"$typeFullName $name"
+    }
     val annotations = field.getTags.asScala
       .collect { case x: VisibilityAnnotationTag => x }
       .flatMap(_.getAnnotations.asScala)
--- a/joern-cli/frontends/jimple2cpg/src/test/scala/io/joern/jimple2cpg/querying/FieldAccessTests.scala
+++ b/joern-cli/frontends/jimple2cpg/src/test/scala/io/joern/jimple2cpg/querying/FieldAccessTests.scala
@@ -78,4 +78,66 @@ class FieldAccessTests extends JimpleCode2CpgFixture {
     identifier.typeFullName shouldBe "Foo"
     fieldIdentifier.canonicalName shouldBe "value"
   }
+
+  val cpg2 = code("""
+    |import java.util.Objects;
+    |class TestStaticFinal {
+    |    private static final String finalPrivateString = "PRIVATE_STATIC_FINAL";
+    |    public static final String finalPublicString = "PUBLIC_STATIC_FINAL";
+    |    public String publicString;
+    |    public static String staticString = "PUBLIC_STATIC";
+    |    private static final Object obj = new Object();
+    |    public static final int FLAG = 1;
+    | public static int staticInt = 100;
+    |    public TestStaticFinal(String publicString) {
+    |        if (Objects.equals(publicString, "")) {
+    |            this.publicString = finalPrivateString+"-"+ staticString+"-"+ finalPublicString;
+    |        }else{
+    |            this.publicString = publicString;
+    |        }
+    |    }
+    |    public void setPublicString(String publicString) {
+    |        this.publicString = publicString;
+    |    }
+    |    public String getFinal(){
+    |        return finalPublicString.concat(finalPrivateString.toLowerCase());
+    |    }
+    |    public String getPublicString(){
+    |        return publicString;
+    |    }
+    |}
+    |class Test {
+    |    public static void test(){
+    |        TestStaticFinal test = new TestStaticFinal("");
+    |        String x = test.getPublicString();
+    |        System.out.println(x);
+    |        test.setPublicString("Change");
+    |        String y = test.getPublicString() + "|" + TestStaticFinal.staticString+ "|"+ TestStaticFinal.finalPublicString;
+    |        System.out.println(y);
+    |        String finalS = test.getFinal();
+    |        System.out.println(finalS);
+    |    }
+    |}
+    |""".stripMargin).cpg
+
+  "should handle static final field access" in {
+    val staticFields = cpg2.typeDecl.name("TestStaticFinal").member.l
+    println("Static fields:")
+    staticFields.foreach(f => println(s"${f.name}: ${f.code}"))
+
+    val clinitAssignments = cpg2.method("<clinit>").call(".*assignment").code.l
+    println("Clinit assignments:")
+    clinitAssignments.foreach(println)
+    val staticStringAssignment = clinitAssignments.find(_.contains("staticString")).getOrElse(null)
+    staticStringAssignment shouldBe "TestStaticFinal.staticString = \"PUBLIC_STATIC\""
+
+    val finalPrivateStringField = staticFields.find(_.name == "finalPrivateString").get
+    finalPrivateStringField.code shouldBe "java.lang.String finalPrivateString = \"PRIVATE_STATIC_FINAL\""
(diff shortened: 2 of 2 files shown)


joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/astcreation/declarations/AstForTypeDeclsCreator.scala
```diff
@@ -61,7 +61,19 @@ trait AstForTypeDeclsCreator(implicit withSchemaValidation: ValidationMode) { th
...
     val name         = field.getName
-    val code         = if (field.getDeclaration.contains("enum")) name else s"$typeFullName $name"
+    val code: String = if (field.getDeclaration.contains("enum")) {
+      name
+    } else if (field.isStatic && field.isFinal) {
+      val constantValue = field.getTags.asScala.collectFirst { case tag: ConstantValueTag =>
+        tag.getConstant.toString
+      }
+      constantValue match {
+        case Some(value) => s"$typeFullName $name = $value"
```
DavidBakerEffendi, line 71: If I understand correctly, you want a way to show the value of some definition on the member node? From what I understand, aren't these assigned within the `<clinit>` method?

Or are you referring to a modifier of some sort? Modifiers are usually available to attach to member nodes..

DavidBakerEffendi (commented):
Just left a comment quick to ask for clarification

joern-cli/frontends/jimple2cpg/src/main/scala/io/joern/jimple2cpg/astcreation/declarations/AstForTypeDeclsCreator.scala
sfncat: In my use of joern to process the apk process, I found that some complex apks have static members of the class definition referenced as parameters when calling functions, but in cpg, I cannot find any relevant information about this static variable except for this static variable itself.
I directly use Soot to export the corresponding jimple, and there is a definition of static member variable in the exported jimple.

```
public class d4.a extends java.lang.Object
{
    private static final java.lang.Object v;
    private static java.lang.String w;
    private static java.lang.String x;
    static void <clinit>()
    {
    java.lang.Object $r0;

    code = """ static void <clinit>()

    <d4.a: java.lang.String x> = "RESTART";

    <d4.a: java.lang.String w> = "11.0.10";

    $r0 = new java.lang.Object;

    specialinvoke $r0.<java.lang.Object: void <init>()>();

    <d4.a: java.lang.Object v> = $r0;
    return;

    }
```

```
$r4 = <d4.a: java.lang.String x>;
virtualinvoke $r1.<android.content.Context: android.content.Intent registerReceiver(android.content.BroadcastReceiver,android.content.IntentFilter,java.lang.String,android.os.Handler,int)>($r2, $r3, $r4, null, 2);

```


The source code is similar to

```
private static String w = "11.0.10";

private static String x = "RESTART";
```


However, there are no relevant definitions in <clinit> in joern-transformed cpg, including code and call statements

```
{
java.lang.Object $r0;

$r0 = new java.lang.Object;

specialinvoke $r0.<java.lang.Object: void <init>()>();

<d4.a: java.lang.Object v> = $r0;

return;
}
""",
"$r0 = new java.lang.Object",
"new java.lang.Object",
"$r0.java.lang.Object()",
"d4.a.v = $r0",
"d4.a.v"
```



During the debugging process, I found that the tag attributes that are mainly related to static variables passed to Soot to Joern, but there is no such attribute in the member node of cpg, so I had to add the tag of the variable to the code of the variable

The best solution is to <clinit>add the definition statement of such variables in, I tried to add, although it can be added, but it is not quite correct, there are problems, so give up.

sfncat: I find Member has the io.shiftleft.codepropertygraph.generated.nodes.Tag Property. So maybe can add the constantValue to the property "tag".

sfncat: I can not find any api to change the nodes.properties.
I can make a NewTag node as the Member's chidren, so in this way ,code of Member will not be change.
it is better?
```
    if (field.isStatic && field.isFinal){
      val constantValue = field.getTags.asScala.collectFirst { case tag: ConstantValueTag =>
        tag.getConstant.toString
      }.getOrElse("")
      val annotations = field.getTags.asScala
        .collect { case x: VisibilityAnnotationTag => x }
        .flatMap(_.getAnnotations.asScala)
      Ast(memberNode(field, name, code, typeFullName))
        .withChildren(annotations.map(astsForAnnotations(_, field)).toSeq)
        .withChildren(Seq(Ast(NewTag().name("ConstantValue").value(constantValue))))
```

DavidBakerEffendi: Unfortunately, the `Tag` node is not used by the data-flow engine. So if we were to make `"Constant"` value a source, it would not be detected during the taint analysis. 

Generally, `<clinit>` is where this should go since these are static fields. I've put the variables
```java
private static String w = "11.0.10";

private static String x = "RESTART";
```
And I see they are initialized in the `<clinit>` method. 
Screenshot 2025-02-25 at 10 27 47
Is your request then specifically to add the constant in the code? I think that is a good idea.

Or is there another feature you'd like? 

PS: If you want to navigate from a field access to the member, you can use something like:
```
cpg.fieldAccess.dedupBy(_.member).map(fieldAccess => fieldAccess.code -> fieldAccess.member.code.l).foreach(println)
/*
(Foo.w,List(java.lang.String w))
(Foo.x,List(java.lang.String x))
*/
(for better code, you can use `fieldAccess.astParent.code` which is the assignment )
/*
(Foo.w = "11.0.10",List(java.lang.String w))
(Foo.x = "RESTART",List(java.lang.String x))
*/
```

sfncat: I add a test case in FieldAccessTests
the source code :
```
class TestStaticFinal {
    private static final String finalPrivateString = "PRIVATE_STATIC_FINAL";
    public static final String finalPublicString = "PUBLIC_STATIC_FINAL";
    public String publicString;
    public static String staticString = "PUBLIC_STATIC";
    private static final Object obj = new Object();
    public static final int FLAG = 1;
    public static int staticInt = 100;
```
and the joern method <clinit> code:
```
    static void <clinit>()
    {
        java.lang.Object $stack0;
        <TestStaticFinal: java.lang.String staticString> = "PUBLIC_STATIC";
        $stack0 = new java.lang.Object;
        specialinvoke $stack0.<java.lang.Object: void <init>()>();
        <TestStaticFinal: java.lang.Object obj> = $stack0;
        <TestStaticFinal: int staticInt> = 100;
        return;
    }
```
the staticString and staticInt init in the <clinit>method,but finalPrivateString,finalPublicString,FLAG not init .and in the cpg2, i can't find the define value of finalPrivateString.
I tried to add the value to the propertiedMap of member node, but failed, member node doesn't have the "Value" property. so i add the value to the field code.
```
      val constantValue = field.getTags.asScala.collectFirst { case tag: ConstantValueTag =>
        tag.getConstant.toString
      }.getOrElse("")      
      val m = memberNode(field, name, code, typeFullName)
      m.propertiesMap.put(PropertyNames.VALUE, constantValue)
```

DavidBakerEffendi: It might be optimized away, have you compiled it with debugging flags enabled?

sfncat: I found this issue because when using joern to analyze many actual APKs, I encountered a situation where a function called a static variable of a class as a parameter. Both the source code and the reverse code show that it is defined with a specific value, but it cannot be found in the cpg.

DavidBakerEffendi: If you could provide a minimal reproducible example that would really help

sfncat: I try to write a example, but failed, I will find a real apk example.

sfncat: I send a email to you, include the example

PR #5332, opened 2025-02-28
Remove last instances of `better.files.File.newTemporaryDirectory`

AndreiDreyer:
* Removed last instances of `better.files.File.newTemporaryDirectory`, replaced with appropriate `java.nio` classes
* Implemented required methods in `FileUtil` for replacement
* Updated required methods to use `Path` instead of `File`


DavidBakerEffendi (approved):

max-leuthaeuser:
If these are really the last instances we could remove `better files` from all `build.sbt` and also remove all `import better.files.*` from all Scala files.

AndreiDreyer:
> If these are really the last instances we could remove `better files` from all `build.sbt` and also remove all `import better.files.*` from all Scala files.

These are just the last instances of the `better.files.File.newTemporaryDirectory`, there are still a few more instances of different `better.files` functions that I need to replace in coming PRs

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/FileUtil.scala
```diff
@@ -69,60 +89,97 @@ object FileUtil {
...
-      linkOptions: File.LinkOptions = File.LinkOptions.default,
-      copyOptions: File.CopyOptions = File.CopyOptions(overwrite = false)
-    ): Unit = {
-      require(directory.isDirectory(linkOptions), s"$directory must be a directory")
-
-      f.walk().filter(_.isRegularFile).foreach { x =>
-        val relativePath = x.pathAsString.stripPrefix(s"${f.pathAsString}${java.io.File.separator}")
+    def mergeDirectory(
+      directory: Path
+    )(implicit copyOptions: StandardCopyOption = StandardCopyOption.COPY_ATTRIBUTES): Unit = {
```
AndreiDreyer, line 127: Note - move `copyOptions` to be normal parameter instead of implicit

PR #5334, opened 2025-02-28
parallelize delombok-ing

jbellis:


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/Delombok.scala
```diff
@@ -99,6 +100,8 @@ object Delombok {
       case Success(tempDir) =>
         PackageRootFinder
           .packageRootsFromFiles(inputPath, fileInfo)
+          .toList
```
maltek, line 103:
```suggestion
```

`packageRootsFromFiles()` already returns a list, so let's just skip this step

maltek (commented):
Thank you for your contribution!

maltek (approved):

Issue #5338, opened 2025-03-03
[CFG][c2cpg?] CFG in try-catch statements is broken for multiple returns


max-leuthaeuser:
```
> importCode.cpp.fromString("""int main() {
     |   try {
     |     if (1+1) {
     |       return foo();
     |     }
     |     return bar();
     |   } catch (...) {
     |   }
     | }""")
> cpg.call.name("foo").cfgNext.l
val res1: List[io.shiftleft.codepropertygraph.generated.nodes.CfgNode] = List(
  Return(
    argumentIndex = -1,
    argumentName = None,
    code = "return foo();",
    columnNumber = Some(value = 7),
    depthFirstOrder = 14,
    internalFlags = 4,
    lineNumber = Some(value = 4),
    order = 1
  ),
  Call(
    argumentIndex = -1,
    argumentName = None,
    code = "bar()",
    columnNumber = Some(value = 12),
    depthFirstOrder = 12,
    dispatchName = "",
    dispatchType = "STATIC_DISPATCH",
    dynamicTypeHintFullName = IndexedSeq(),
    internalFlags = 4,
    lineNumber = Some(value = 6),
    methodFullName = "<unresolvedNamespace>.bar:<unresolvedSignature>(0)",
    methodInstFullName = None,
    name = "bar",
    order = 1,
    possibleTypes = IndexedSeq(),
    resolved = None,
    signature = "<unresolvedSignature>",
    typeFullName = "ANY"
  )
)
```

There is a CFG edge from the `foo()` call to the `bar()` call, even though it’s impossible for the execution to take that path. A CFG-successor in the catch block would make some sense, but a CFG successor within the same try block doesn’t.

@DavidBakerEffendi 
You did some work w.r.t. returns in try blocks: https://github.com/joernio/joern/blob/master/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/controlflow/cfgcreation/CfgCreator.scala#L88
I just want to double check: is is change indented / validated for multiple returns there?
I am unable to find any issue with c2cpg' representation of the try-catch there (its the same as e.g., in javasrc2cpg). 
Do you have any idea whats wrong?

DavidBakerEffendi:
I recall my fix trying to solve an issue where an explicit return node within a try-catch would not "flow out" of the try-catch, thus killing flows to callers.

This is an interesting test case, and maybe requires room for finesse on the heuristic

PR #5339, opened 2025-03-04
[CFG] Fix CFG in try-catch statements for multiple returns

max-leuthaeuser:
@DavidBakerEffendi You may want to have a look. The tests you added for this case for javasrc2cpg are still working as expected.

Fixes: https://github.com/joernio/joern/issues/5338


max-leuthaeuser:
CS dataflow and integration tests are also fine for all frontends.

PR #5340, opened 2025-03-04
[ruby] Ruby `match` to use `DynamicCallUnknownFullName`

DavidBakerEffendi:
To be consistent with other dynamic calls, adds a dynamic type hint to literals + sets `match` to be `DynamicCallUnknownFullName` to enable dynamic call resolution strategies to kick in.


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/RegexTests.scala
```diff
@@ -21,7 +22,7 @@ class RegexTests extends RubyCode2CpgFixture(withPostProcessing = false) {
       val tmpSource = tmpInit.source.asInstanceOf[Call]
       tmpSource.code shouldBe s"/h(el)lo/.match($expectedSubject)"
       tmpSource.name shouldBe "match"
-      tmpSource.methodFullName shouldBe "__core.Regexp.match"
+      tmpSource.methodFullName shouldBe Defines.DynamicCallUnknownFullName
```
AndreiDreyer, line 25: Should we not include a check to the type hint somewhere to ensure it is also working properly?

AndreiDreyer (approved):

PR #5342, opened 2025-03-04
replace `better.files.File.newTemporaryFile`

AndreiDreyer:
Replaces all usages of `better.files.File.newTemporaryFile` in `io.joern` with `FileUtil.newTemporaryFile`


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/FileUtil.scala
```diff
@@ -2,14 +2,27 @@ package io.joern.x2cpg.utils
...
 import scala.annotation.tailrec
 import scala.jdk.CollectionConverters.*
 
 object FileUtil {
+  def newTemporaryFile(prefix: String = "", suffix: String = ""): Path = {
+    Files.createTempFile(prefix, suffix)
+  }
+
+  // Similar to `.touch` from better.files
+  def createTempFileIfNotExistsAndSetLastModificationDate(tmpDirName: String): Path = {
```
ml86, line 18: I do not get why we would need this:
1. It combines two things: New temp file creation and setting the last modification date.
2. Does a new file not already have the last modified date set to the current time.

Then there is the issue of calling `createWithParentsIfNotExists` after creating the temp file. That looks superfluous. 
Then there is the `tmpDirName` parameter given as `prefix` to to `newTemporaryFile` looks wrong.
Then there is the issue that this hole function only severs a two test cases where i am not even sure if the `touch` feature is needed. I am almost sure that it is not because a temp file is always new and thus we should not change anything meaningful by a touch operation.

ml86 (changes requested):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/FileUtil.scala
AndreiDreyer: I think I got a bit too caught up in better files. I've removed this function and just replaced it with a call to `newTemporaryFile`. All tests seem to still pass locally with this update.

ml86 (approved):

PR #5343, opened 2025-03-04
[c2cpg] Proper types for lambda expressions

max-leuthaeuser:


max-leuthaeuser:
Not needed, we do not put template tags into the type names.

PR #5344, opened 2025-03-04
Replace `better.files.File.usingTemporary(Directory|File)`

AndreiDreyer:
Replaced all instances of `better.files.File.usingTemporary(File|Directory)`

The next PRs will deal with the removal of `better.files.DSL.*` and `better.files.*`


joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/io/JSONCompilationDatabaseParserTests.scala
```diff
@@ -111,10 +111,10 @@ class JSONCompilationDatabaseParserTests extends AnyWordSpec with Matchers {
           |    "file": "file2.cc" }
           |]""".stripMargin
 
-      BetterFile.usingTemporaryFile("compile_commands.json") { commandJsonFile =>
-        commandJsonFile.writeText(content)
+      FileUtil.usingTemporaryFile("compile_commands.json") { commandJsonFile =>
+        Files.writeString(commandJsonFile, content, Charset.defaultCharset())
```
ml86, line 115: Please create a `write` or `writeString` named function in FileUtil to wrap `Files.writeString`. The goal is to not always require people to give `Charset.defaultCharset()` as argument.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/FileUtil.scala
```diff
@@ -68,6 +99,16 @@ object FileUtil {
     }
   }
 
+  def writeBytes(file: Path, content: Iterator[Byte], bufferSize: Int = 8192): Unit = {
```
ml86, line 102: We do not have a single use case where we really need to work on `Iterator[Bytes]`. 
So why not make our live easier and just use `Iterable[Byte]` as parameter type?

ml86 (changes requested):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/FileUtil.scala
```diff
@@ -330,16 +268,25 @@ object FileUtil {
...
+   * Ensures that a resource is properly closed after use, even if an exception occurs.
+   * Similar to Java's try-with-resources but with more control.
+   *
+   * @param resource The resource to use
+   * @param f The function to apply to the resource
+   * @tparam R The resource type
+   * @tparam A The return type
+   * @return The result of applying f to the resource
+   */
+  def using[R <: AutoCloseable, A](resource: R)(f: R => A): A = {
```
max-leuthaeuser, line 285: We can't simply use https://www.scala-lang.org/api/current/scala/util/Using$.html?

AndreiDreyer: I should have marked this PR as a draft again. I've been struggling with getting windows tests to pass due to files still being in use causing errors to happen on the console tests. Some of things in this PR is just trying to get the tests to pass on the runners, I will try removing this and using standard scala `Using` and see if it still passes all the tests.

AndreiDreyer: Reverted to using standard `Using`, seems to have still worked.

ml86 (approved):

PR #5347, opened 2025-03-06
ghidra: latest release which includes mac_x86_64 binaries

mpollmeier:


DavidBakerEffendi (approved):
Looks good, we're now as OS-agnostic as the forked project :D

PR #5348, opened 2025-03-07
Move gradle dependency fetching logic into init.gradle

johannescoetzee:
Before this PR, a lot of the logic for gradle dependency fetching was happening from Scala code through the gradle connection via the API. This has a lot of overhead as the project needs to be reloaded/initialized/correct verb each time. For example, just listing the available projects with their configurations for the elasticsearch project took ~30 minutes on my machine since there's a configuration step that had to happen for each request. Doing this from a gradle task defined in the build or init file is almost instantaneous.

With this PR, all of that logic is moved into the gradle init script. A separate task to fetch dependencies is created for each subproject and then one task is created in the root project which either depends on all of these, or only on that of the `:app` project if it exists to stay consistent with how we filtered projects before. Only this root project task has to be executed to fetch all of the dependencies, so we avoid the overhead associated with multiple API calls.


ml86 (approved):
Approved so you can go ahead quickly.
But i think putting the init script in a separate resource file might make more sense then this huge inline string. Please explore that for a follow up PR.

PR #5350, opened 2025-03-07
Replaced all instances of `better.files.*` functions

AndreiDreyer:
* Replaced all instances of `better.files.File` with `java.nio.file.Path` equivalent
* Created helper methods in `FileUtil` where required

Note: We are still allowing `better.files` to be used in the `build.sbt` files, as we don't have access to any of the functions in `FileUtil`.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/typestub/TypeStubUtil.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/typestub/TypeStubUtil.scala
@@ -23,7 +23,7 @@ object TypeStubUtil {
         "."
       }
     }
-    Paths.get(fixedDir, "/type_stubs").toAbsolutePath.normalize()
+    Paths.get(fixedDir, "type_stubs")
   }
 
 }
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/typestub/TypeStubUtil.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/typestub/TypeStubUtil.scala
@@ -1,8 +1,6 @@
 package io.joern.x2cpg.typestub
 
-import better.files.File
-
-import java.nio.file.Paths
+import java.nio.file.{Path, Paths}
 
 object TypeStubUtil {
 
@@ -12,7 +10,7 @@ object TypeStubUtil {
     * @return
     *   the directory where type stubs are.
     */
-  def typeStubDir(implicit metaData: TypeStubMetaData): File = {
+  def typeStubDir(implicit metaData: TypeStubMetaData): Path = {
     val dir        = metaData.packagePath.toString
     val indexOfLib = dir.lastIndexOf("lib")
     val fixedDir = if (indexOfLib != -1) {
@@ -25,7 +23,7 @@ object TypeStubUtil {
         "."
       }
     }
-    File(Paths.get(fixedDir, "/type_stubs").toAbsolutePath)
+    Paths.get(fixedDir, "/type_stubs").toAbsolutePath
   }
 
 }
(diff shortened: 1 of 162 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/FileUtil.scala
```diff
@@ -117,12 +117,12 @@ object FileUtil {
     def absolutePathAsString: String = p.toAbsolutePath.toString
 
     def /(child: String): Path = {
-      p.resolve(child)
+      p.resolve(child).toAbsolutePath.normalize()
```
ml86, line 120: Why to we need the absolute path conversion and the normalization?
To do this always even if people only want the `resolve` functionality seems odd.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/typestub/TypeStubUtil.scala
```diff
@@ -25,7 +23,7 @@ object TypeStubUtil {
         "."
       }
     }
-    File(Paths.get(fixedDir, "/type_stubs").toAbsolutePath)
+    Paths.get(fixedDir, "/type_stubs").toAbsolutePath.normalize()
```
ml86, line 26: `normalize` only seems necessary because of the extra leading `/` in `/type_stubs`

ml86 (commented):
I am missing the removal of better.files from the build.sbt files.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/FileUtil.scala
AndreiDreyer: This took me a while to track down - in `better.files.File` we have the following:
```scala
def /(child: String): File = path.resolve(child)

implicit def apply(path: Path): File = new File(path.toAbsolutePath.normalize())

def apply(path: String, fragments: String*): File = Paths.get(path, fragments: _*)
```

From my understanding, since the `/` function returns a `File` type, the `implicit apply` is going to be called which in turn calls the `path.toAbsolutePath.normalize()`. This is what lead me to then updating the `/` function to call the `toAbsolutePath.normalize()`.

There were some issues with resolving directories with an included `./` (e.g. `./someDir`) that this ended up fixing which is why I kept it in.

AndreiDreyer:
> I am missing the removal of better.files from the build.sbt files.

I thought we are allowing `better.files` to stay in `build.sbt` files?

ml86:
@AndreiDreyer Yes, but `better.files` can be removed from the dependency lists in the build.sbt files. Only in `project/meta-build.sbt` it is still required as dependency.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/FileUtil.scala
AndreiDreyer:
```scala
def /(child: String): File = path.resolve(child)
implicit def apply(path: Path): File = new File(path.toAbsolutePath.normalize())
```

`path.resolve(child)` above returns type `java.nio.file.Path`, which then calls the `implicit apply(path)` which returns the `File`

```scala
val foo = File(someDir) // `better.files.File`
val bar = foo / "cpg.bin" // equivalent to calling the `implicit apply(foo.resolve("cpg.bin"))`
```

AndreiDreyer:
> @AndreiDreyer Yes, but `better.files` can be removed from the dependency lists in the build.sbt files. Only in `project/meta-build.sbt` it is still required as dependency.

@ml86 The only reference I can find to `better-files` is in the `meta-build.sbt`.

ml86 (approved):

PR #5353, opened 2025-03-07
Deprecate newAnnotationLiteral node

johannescoetzee:
This is the first in a series of PRs which will deprecate the node builder methods in `NodeBuliders.scala` in favour of the `AstNodeBuilder` methods which follow the pattern of taking an origin node and calculating line/column numbers and offsets from that. 

I think that breaking these PRs up by method will probably be the easiest to review, but am happy to do it by frontend instead if that's better.


max-leuthaeuser (approved):

max-leuthaeuser:
> I think that breaking these PRs up by method will probably be the easiest to review
Think so too.

PR #5354, opened 2025-03-07
Deprecate newLocalNode method

johannescoetzee:
Ideally we wouldn't need the `localNodeWithExplicitPositionInfo` method at all, but I don't see an obvious way to avoid it in the `AstCreatorHelper` classes. @max-leuthaeuser you're more familiar with this part of code, so if you have any ideas please let me know. Otherwise I think keeping moving this local creation method into `AstNodeBuilders` and clearly distinguishing it from `localNode` by name is fine.


max-leuthaeuser:
Hm, I also just had a look into c2cpg. Due to the fact, that the whole closure binding / local creation happens after the AST traversal there we do not have an IASTNode by hand. Same for jssrc and swiftsrc.
I guess your solution is just fine.
We could refactor the whole scope datastructure to carry around the originating ASTNode until we need it for a local but thats something for later I guess.

max-leuthaeuser (approved):

PR #5358, opened 2025-03-11
[c2cpg] Fix-up types on calls, identifier, and locals for lambda expressions

max-leuthaeuser:


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -30,16 +31,30 @@ trait AstForTypesCreator { this: AstCreator =>
...
     declaration match {
       case d if isTypeDef(d) && shortName(d.getDeclSpecifier).nonEmpty =>
         val filename = fileName(declaration)
-        val typeDefName = if (name.isEmpty) {
-          safeGetBinding(declarator.getName).map(b => registerType(b.getName))
-        } else {
-          Option(registerType(name))
-        }
+        val typeDefName = if (name.isEmpty) { safeGetBinding(declarator.getName).map(b => registerType(b.getName)) }
+        else { Option(registerType(name)) }
```
johannescoetzee, line 57: Is this a result of auto-formatting. Personally I like the original version more, or doing the if/else on 2 lines as it is here, but getting rid of the `{}`. Just a minor style nit

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/CallTests.scala
```diff
@@ -302,17 +302,17 @@ class CallTests extends C2CpgSuite {
       val cpg = code(
         """
           |void outer() {
-          |  [](int a) {}(1);
+          |  [](int a) -> int { return a; }(1);
```
johannescoetzee, line 305: Wouldn't the original test still be a valid test? I think adding this as a new test would be better if so.

johannescoetzee (approved):
Approved with minor comments that aren't blocking IMO

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/CallTests.scala
max-leuthaeuser: Right, I will change that back and add an additional test.

PR #5362, opened 2025-03-11
Joern Project Linting: Checking Restricted Imports

DavidBakerEffendi:
As we aim to reduce our class path and restrict libraries that have been removed from being re-introduced, this PR adds a `scalafix` based rule module to lint against imports. This makes use of `scalameta` to parse imports and emit a linter error when a violating import is detected. 

This is added to the PR which should fail as an example.


.github/workflows/pr.yml
```diff
@@ -73,6 +73,30 @@ jobs:
...
+          java-version: 21
+          cache: sbt
+      - uses: sbt/setup-sbt@v1
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.sbt
+            ~/.coursier
+          key: ${{ runner.os }}-sbt-${{ hashfiles('**/build.sbt') }}
+      - name: Check formatting
```
mpollmeier, line 95:
```suggestion
      - name: Check for usage of restricted imports
```

build.sbt
```diff
@@ -24,6 +24,11 @@ lazy val gosrc2cpg         = Projects.gosrc2cpg
 lazy val swiftsrc2cpg      = Projects.swiftsrc2cpg
 lazy val csharpsrc2cpg     = Projects.csharpsrc2cpg
 
+//lazy val projectLinter = Projects.projectLinter
```
mpollmeier, line 27:
```suggestion
```

project-linter/rules/src/main/scala/fix/RestrictedImports.scala
```diff
@@ -0,0 +1,63 @@
...
+  /** If a file is within any of the packages defined below, it will not be checked for restricted imports.
+    */
+  private val ignoredPackages: Seq[String] = Seq("io.shiftleft.semanticcp")
+
+  /** The rules for restricted imports.
+    */
+  private val rules: Map[String, ImportRule] = Seq(
+    ImportRule(
+      "BetterFiles",
+      "Avoid the 'Better Files' library, use `io.joern.x2cpg.utils.FileUtil` or `java.nio.file` instead.",
```
mpollmeier, line 18:
```suggestion
      "Avoid the 'Better Files' library, use `io.joern.x2cpg.utils.FileUtil`, `os-lib` or `java.nio.file` instead.",
```

mpollmeier (commented):
I like the idea and the outcome, i just wonder if we can simplify the setup. E.g. do we really need to define a `dependsOn` in every project? And does the linter project really need to be a standalone project with it's own sbt version?

If you already tried something in that direction let me know, otherwise I'll see if I can simplify it this afternoon.

project-linter/rules/src/main/scala/fix/RestrictedImports.scala
AndreiDreyer: Also update `io.joern.x2cpg.utils.FileUtil` to `io.shiftleft.semanticcpg.utils.FileUtil`

DavidBakerEffendi:
@mpollmeier So there are other ways to do this for sure, including having this Joern rule project be brought in as a maven dependency and then things may be a bit simpler. However, in a multi-module setup, I largely followed this guide: https://scalacenter.github.io/scalafix/docs/developers/local-rules.html

In terms of `dependsOn`, I've tried simply having this on the root project alone, but then you start to see errors such as:
```
(csharpsrc2cpg / Compile / scalafix) scalafix.sbt.InvalidArgument: Unknown rule 'RestrictedImports'
```
Which I've only managed to fix by having each submodule explicitly state the dependency

ml86 (commented):
@DavidBakerEffendi Looks good to me, i leave the final review to @mpollmeier . Only thing which would be better is to avoid the need to add the linter project to every sub project. I am sure we will forget to add this for sub projects...

DavidBakerEffendi:
@ml86 The linter attempts to run over all subprojects, the dependency loads the rule, so the linter pass will fail if the dependency is missing since it won't be able to find the rule for that project. So forgetting will result in a failure ;)

mpollmeier:
I'm still fighting with this, for some reason some projects react differently to my global config attempts... will keep you updated

mpollmeier:
ok, I just pushed the simplifications that do work, most importantly 
* removed `project-linter`
* moved settings for `linter-rules` into `linter-rules/build.sbt`
* removed `sbt-projectmatrix` experimental plugin
* created a root aggregate project because we don't want `linter-rules` to be involved in tasks like `publish` etc. 

As mentioned in the previous comment, unfortunately we have to define the `dependsOn` clause in every single build.sbt...

I pushed them onto this branch, so please take a good look :)

mpollmeier (approved):

DavidBakerEffendi:
@mpollmeier Sounds relatable, in my initial implementations I had noticed some modules inherited the settings, others not. Very strange.

PR #5363, opened 2025-03-11
Deprecate newOperatorCallNode and remove uses from multiple frontends

johannescoetzee:
There's one use still left in the `fieldAccessAst` method in `AstCreatorBase`, but that method doesn't seem too useful so I think deprecating it as well and unwrapping it would make the most sense. I'll do that in a separate PR though.


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/NodeBuilders.scala
```diff
@@ -128,6 +128,10 @@ object NodeBuilders {
       .lineNumber(line)
   }
 
+  @deprecated(
+    "Deprecated in favour of the corresponding method io.joern.x2cpg.AstNodeBuilder and will be removed in a future version",
+    "4.0.284"
```
johannescoetzee, line 133: I'll probably need to adjust the deprecation version, but I'll update that once I know which version this will be released in

max-leuthaeuser (approved):

PR #5369, opened 2025-03-12
Make AstNodeBuilder available to AstCreatorBase and deprecate newMethodReturn

johannescoetzee:
This PR touches a bunch of files since I needed to make `AstCreatorBase` extend `AstNodeBuilder` to be able to use the `methodReturnNode` method.

There is still one use of the deprecated method in `c2cpg.passes.TypeDeclNodePass`. The pattern we're following here doesn't work for non-ASTCreator passes, so some sort of static node builder method may be required, potentially in a companion object for `AstNodeBuilder`. I'm open to other suggestions for how we could handle this case though.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/TypeDeclNodePass.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/TypeDeclNodePass.scala
@@ -53,7 +53,7 @@ class TypeDeclNodePass(cpg: Cpg, config: Config) extends CpgPass(cpg) {
         .astParentType(NodeTypes.NAMESPACE_BLOCK)
         .astParentFullName(fullName)
     val blockNode    = NewBlock().typeFullName(Defines.Any)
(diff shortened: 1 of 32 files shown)


max-leuthaeuser:
+1 for a companion object for AstNodeBuilder

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/TypeDeclNodePass.scala
```diff
@@ -53,7 +53,7 @@ class TypeDeclNodePass(cpg: Cpg, config: Config) extends CpgPass(cpg) {
         .astParentType(NodeTypes.NAMESPACE_BLOCK)
         .astParentFullName(fullName)
     val blockNode    = NewBlock().typeFullName(Defines.Any)
-    val methodReturn = AstNodeBuilder.methodReturnNodeWithExplicitPositionInfo(Defines.Any, line = None, column = None)
+    val methodReturn = AstNodeBuilder.methodReturnNodeWithExplicitPositionInfo(Defines.Any, lineNumber = Option(1))
```
max-leuthaeuser, line 56: Why does it have a specific line number now?

max-leuthaeuser (approved):

PR #5371, opened 2025-03-13
[c2cpg] Simplify for-loop lowering

max-leuthaeuser:
More in line with javasrc2cpg and kotlin2cpg now. Was too jssrc-ish.


max-leuthaeuser:
Do not merge yet; might need some adjustments for DF tracking.

PR #5375, opened 2025-03-14
Put column numbers into Location

bbrehm:
This puts column numbers into Location, and does some minor cleanup of the code to generate Location nodes.

The old `LocationCreator.apply` method stays for a short time, in order to facilitate CS compat (i.e. it should be possible to use this without any CS adjustments). It will get thrown out in the near future, though.


mpollmeier (approved):
The general idea is good, I just had a few code style remarks. Instead of having a lengthy conversation here I simply pushed them to your branch. Hope that's ok, please take a look. Main things:

* use `Try` instead of `try` (more readable, doesn't catch FatalErrors like OutOfMemoryError)
* refactor to use `val` insteaad of `var`

mpollmeier:
timeout -> merging. @bbrehm please let me know if you want to get anything changed

PR #5376, opened 2025-03-14
Fix some issues in methodStubCreator

bbrehm:
This:
1. Never causes fullname collisions during method stubbing
2. gets the arguments of stubbed methods correct
3. warns if there are mismatches

The main point is: In order to make anything work, each argument of a call to the method must have a matching parameter. Matching means: argumentIndex or argument equals index of parameter.

Unfortunately the old code looks hopelessly confused about all that, and it is used from other components. These components need to deconfuse themselves, and then we need to clean this up.


max-leuthaeuser (approved):

max-leuthaeuser:
With this PR CS  DF tests do work now: https://github.com/ShiftLeftSecurity/codescience/pull/8189

PR #5377, opened 2025-03-17
Add `.strippedCode` step for String literals

AndreiDreyer:
* Added a new `.innerText` step for String literals to get the value inside the quotes
* Returns `Some(string)` if the literal is a `string` and has surrounding quotes (language specific), otherwise returns `None`
* Only works on `LITERAL` nodes, not on interpolated strings that are transformed into calls in the AST.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/LiteralMethods.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/LiteralMethods.scala
@@ -8,9 +8,16 @@ import io.shiftleft.semanticcpg.language.*
 import io.shiftleft.semanticcpg.language.nodemethods.LiteralMethods.delimiters
 import org.apache.commons.lang3.StringUtils
 
+import scala.collection.immutable.HashMap
+
 class LiteralMethods(val literal: Literal) extends AnyVal with NodeExtension with HasLocation {
   def innerText: Option[String] = {
-    delimiters(literal)
+    val stringDelimiter = Cpg(literal.graph).metaData.language.headOption match {
+      case Some(language) => delimiters.getOrElse(language, List("\""))
+      case _              => "\"" :: Nil
+    }
+
+    stringDelimiter
       .filter(literal.code.startsWith(_))
       .map(delimiter =>
         val start =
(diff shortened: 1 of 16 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/types/structure/LiteralTraversal.scala
```diff
@@ -0,0 +1,9 @@
+package io.shiftleft.semanticcpg.language.types.structure
+
+import io.shiftleft.codepropertygraph.generated.nodes.Literal
+import io.shiftleft.semanticcpg.language.toLiteralMethods
+
+class LiteralTraversal(val traversal: Iterator[Literal]) extends AnyVal {
+  def innerText: Iterator[Option[String]] =
```
DavidBakerEffendi, line 7: I think the ideal behaviour is that this should always be a `String`, so for non-string literals we just return `.code` basically. Is that correct, @ml86  or do you prefer that this returns an `Option`?

AndreiDreyer: Would we want the same behavior on the `.innerText` function in `LiteralMethods`?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/LiteralMethods.scala
```diff
@@ -1,16 +1,43 @@
...
 import io.shiftleft.codepropertygraph.generated.nodes.{Literal, NewLocation}
 import io.shiftleft.semanticcpg.NodeExtension
 import io.shiftleft.semanticcpg.language.*
 
 class LiteralMethods(val literal: Literal) extends AnyVal with NodeExtension with HasLocation {
-  def value(typeFullName: String): String = {
-    if (literal.typeFullName == typeFullName) {
-      return "asd"
+  def innerText: Option[String] = {
+    val delimiters = Cpg(literal.graph).metaData.language.headOption match {
```
max-leuthaeuser, line 11: This mapping should be defined in a static field outside, maybe in the companion object.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/LiteralMethods.scala
```diff
@@ -1,16 +1,43 @@
...
-    return ""
+    delimiters
+      .filter(literal.code.startsWith(_))
+      .map { delimiter =>
+        val start =
+          if (delimiter == "\"\"\"" || delimiter == "'''") then literal.code.indexOf(delimiter) + 3
+          else literal.code.indexOf(delimiter) + 1
+        val end = literal.code.lastIndexOf(delimiter)
+
+        literal.code.substring(start, end)
```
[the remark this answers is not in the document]
AndreiDreyer: Just tried this, reverted changes. Trailing quotes that are escaped in a string is also being stripped with `StringUtils.strip`, so
```java
String a = "abc\"";
```
ends up being `abc\` instead of `abc\"`

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/LiteralMethods.scala
AndreiDreyer: @max-leuthaeuser moved this into a `Map` in `object LiteralMethods`. Let me know if this is what you wanted or if I should change it to something else.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/LiteralMethods.scala
```diff
@@ -1,20 +1,60 @@
...
+import scala.collection.immutable.HashMap
 
 class LiteralMethods(val literal: Literal) extends AnyVal with NodeExtension with HasLocation {
-  def value(typeFullName: String): String = {
-    if (literal.typeFullName == typeFullName) {
-      return "asd"
+  def innerText: Option[String] = {
+    val stringDelimiter = Cpg(literal.graph).metaData.language.headOption match {
+      case Some(language) => delimiters.getOrElse(language, List("\""))
+      case _              => "\"" :: Nil
```
ml86, line 17: These two fallbacks just hide errors.
`metaData.language` always needs to exist.
`delimiters` needs to be updated for each frontend otherwise we use a maybe non sense default.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/LiteralMethods.scala
```diff
@@ -1,20 +1,60 @@
...
     }
 
-    return ""
+    stringDelimiter
+      .filter(literal.code.startsWith(_))
+      .map(delimiter =>
+        val start =
+          if (delimiter == "\"\"\"" || delimiter == "'''") then literal.code.indexOf(delimiter) + 3
+          else literal.code.indexOf(delimiter) + 1
```
ml86, line 25: Use the length of the machted delimiter instead of fixed values 1 and 3.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/LiteralMethods.scala
```diff
@@ -1,20 +1,60 @@
...
+        val end = literal.code.lastIndexOf(delimiter)
+
+        /* Windows uses `\r\n` for new line characters, but this can cause issues on different operating systems. So
+        we replace all `\r\n` to have consistency across all OS
+         */
+        literal.code.substring(start, end).replaceAll("\r\n", "\n")
```
ml86, line 31: If this causes issues why do we address this in such a special purpose function as this one?

ml86 (changes requested):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/LiteralMethods.scala
```diff
@@ -1,20 +1,66 @@
...
-    return ""
+    val code = literal.code
+
+    stringDelimiters
+      .filter(literal.code.startsWith(_))
+      .map(delimiter =>
+        val delimiterLength = delimiter.length
+        val start =
+          if (delimiter == "\"\"\"" || delimiter == "'''") then literal.code.indexOf(delimiter) + delimiterLength
+          else literal.code.indexOf(delimiter) + delimiterLength
```
ml86, line 31: If-else is not required anymore since both do the same.

ml86 (approved):

Issue #5382, opened 2025-03-19
[Bug] Method return type is always ANY within a class definition for imported types


Freeman-Trader:
**Describe the bug**
When looking at methods that come from a C++ file, function definitions within a class definition will have their return type dropped in favor of ANY while the same function definition within the global scope will maintain its original return type. This behavior only occurs when the return type is imported.

**To Reproduce**
1. Create a C++ file with the following contents

![Image](https://github.com/user-attachments/assets/3af6e6a4-a1de-4c54-939a-e75d71ca4bcf)

2. Start Joern and load the previously created C++ file using ```importCode()```
3. Run the following command ```cpg.method.filter(_.name.contains("foo")).methodReturn.typeFullName.l```
4. It will return the following

![Image](https://github.com/user-attachments/assets/c4c75a6f-b279-44b0-8536-377186f7d895)

**Expected behavior**
Joern should be returning two instances of "vector<int>" rather than the one and the "ANY" that replaces it.
```List[String] = List("int", "vector<int>", "vector<int>")```

**Desktop (please complete the following information):**
OS: "Ubuntu 24.04.2 LTS"
Joern Version: 4.0.287
Java version: "21.0.6" 2025-01-21

max-leuthaeuser:
Thanks for the report, will investigate.

max-leuthaeuser:
Fixed with: https://github.com/joernio/joern/pull/5383

Issue #5381, opened 2025-03-18
[Bug] Function definitions showing up twice in methods


Freeman-Trader:
**Describe the bug**
I am not sure if this is meant to occur, but when defining a function within a global scope in C++, and then calling that function, two methods will be produced for that single function definition. One of those 2 methods will look correct while the other one will be missing most of its information and have its isExternal value set to true, even though everything is locally defined.

**To Reproduce**

1. Create a C++ file with the following contents 

![Image](https://github.com/user-attachments/assets/e4a044c3-45d8-4a72-bde9-f49c276c56db)

2. Start Joern, and import the previously created C++ file using ```importCode()```
3. Then run the command ```cpg.method.filter(_.name == "multiplyLong").l```
4. This will return the following

![Image](https://github.com/user-attachments/assets/81697969-d20e-43cd-b381-13bd57b1f2d4)

**Expected behavior**
I expected to see only one method with the name of "multiplyLong" since there is only one function definition with that name.

**Desktop (please complete the following information):**
OS: "Ubuntu 24.04.2 LTS"
Joern Version: 4.0.287
Java version: "21.0.6" 2025-01-21

max-leuthaeuser:
Thanks for the report, will investigate.

PR #5383, opened 2025-03-19
[c2cpg] Some fixes for calculating type string representations

max-leuthaeuser:
For: https://github.com/joernio/joern/issues/5381 and https://github.com/joernio/joern/issues/5382


joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/MethodTests.scala
```diff
@@ -330,6 +330,55 @@ class MethodTests extends C2CpgSuite {
...
+    }
+
+    "be correct for C++ method in class" in {
+      val cpg = code(
+        """
+          |#include <vector>
+          |
+          |class A {
+          |  public:
+          |    int foo1(void) { return; }
```
maltek, line 361:
```suggestion
          |    int foo1(void) { return 1; }
```

I was confused for a bit there thinking this was a `void` function

DavidBakerEffendi (approved):

maltek (commented):
`short [int]` and `long long [int]` should probably get the same treatment as `long` here

maltek (changes requested):
`long long` is different from `long` (e.g. on linux amd64 `long` would be 64 bit, `long long` 128 bit)

max-leuthaeuser:
Totally right. Because we may have problems with spaces in fullnames / signatures and types I reduced the type to `longlong` then.

maltek (approved):

PR #5392, opened 2025-03-25
[c2cpg] Extract type string calculations into separate trait

max-leuthaeuser:
Type string calculation is quite a mess right now and to enable further refactorings and clean-ups I started with extracting everything related to it into its own trait.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreator.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreator.scala
@@ -2,14 +2,18 @@ package io.joern.c2cpg.astcreation
 
 import io.joern.c2cpg.Config
 import io.joern.c2cpg.parser.HeaderFileFinder
+import io.joern.x2cpg.Ast
+import io.joern.x2cpg.AstCreatorBase
+import io.joern.x2cpg.ValidationMode
 import io.joern.x2cpg.datastructures.Stack.*
-import io.joern.x2cpg.{Ast, AstCreatorBase, ValidationMode, AstNodeBuilder as X2CpgAstNodeBuilder}
 import io.shiftleft.codepropertygraph.generated.NodeTypes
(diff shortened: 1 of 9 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreator.scala
```diff
@@ -41,8 +46,6 @@ class AstCreator(
 
   protected val scope: C2CpgScope = new C2CpgScope()
 
-  protected val usingDeclarationMappings: mutable.Map[String, String] = mutable.HashMap.empty
```
max-leuthaeuser, line 44: Was never used anyway as CDT does everything for us.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForPrimitivesCreator.scala
```diff
@@ -49,7 +50,11 @@ trait AstForPrimitivesCreator { this: AstCreator =>
           case Some(binding: (CPPVariable | CVariable)) =>
             Try(binding.getScope).toOption
               .collect { case n: IASTInternalScope => n.getPhysicalNode }
-              .exists(_.isInstanceOf[IASTTranslationUnit])
+              .exists {
+                case _: IASTTranslationUnit        => true
+                case n: ICPPASTNamespaceDefinition => n.getParent.isInstanceOf[IASTTranslationUnit]
```
max-leuthaeuser, line 55: Stuff directly under file-global namespaces is also global.

maltek: do we explicitly need to handle nested namespaces here, or does CDT collapse those?

```c++
namespace N { namespace M { const int global = 2; }; };
```

maltek (approved):

PR #5394, opened 2025-03-25
[c2cpg] Added fullname uniqueness for type decls, bindings, and namespace blocks

max-leuthaeuser:


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/FullNameUniquenessPass.scala
```diff
@@ -0,0 +1,134 @@
...
+    nodes.sortBy { node =>
+      val fileName = node.filename
+      val lineNr   = node.lineNumber.getOrElse(-1)
+      val columnNr = node.columnNumber.getOrElse(-1)
+      s"$fileName:$lineNr:$columnNr"
+    }
+  }
+
+  private def logDuplicates[T <: AffectedNodeType](nodes: List[T], fullName: String): Unit = {
+    val fromFiles = nodes.map(_.filename).mkString("[", ", ", "]")
```
maltek, line 97: maybe wrap this whole function in a `if (logger.isDebugEnabled)`

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/FullNameUniquenessPass.scala
```diff
@@ -0,0 +1,134 @@
...
+      // sort for stable output
+      .sortBy(_._1)
+  }
+
+  private def sortNodesByLocation[T <: AffectedNodeType](nodes: List[T]): List[T] = {
+    nodes.sortBy { node =>
+      val fileName = node.filename
+      val lineNr   = node.lineNumber.getOrElse(-1)
+      val columnNr = node.columnNumber.getOrElse(-1)
+      s"$fileName:$lineNr:$columnNr"
```
maltek, line 92: I think sorting by a tuple should also work, and it avoids problems like `10 < 2` in lexicographical sorting
```suggestion
      (fileName, lineNr, columnNr)
```

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/FullNameUniquenessPass.scala
```diff
@@ -0,0 +1,134 @@
...
+    handleNamespaceBlocks(dstGraph)
+  }
+
+  private def handleMethods(dstGraph: DiffGraphBuilder): Unit = {
+    val methodFullNameMap = generateStableFullNameMapping(cpg.method.nameNot(NamespaceTraversal.globalNamespaceName))
+    methodFullNameMap.foreach { case (fullName, methods) =>
+      logDuplicates(methods, fullName)
+      val sortedMethods = sortNodesByLocation(methods)
+      lazy val callsAffected =
+        cpg.call.methodFullNameExact(fullName).filterNot(_.file.exists(_.name == sortedMethods.head.filename)).l
```
maltek, line 52: I think `call.file` uses `CONTAINS` edges which won't be available at the time this is running.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/FullNameUniquenessPass.scala
```diff
@@ -0,0 +1,134 @@
...
+          s"$fullNameWithoutSignature${Defines.DuplicateSuffix}$index:$signature"
+        }
+        dstGraph.setNodeProperty(method, Method.PropertyNames.FullName, newFullName)
+        // fixup bindings
+        bindingsAffected.filter(_.refOut.contains(method)).foreach { binding =>
+          dstGraph.setNodeProperty(binding, Binding.PropertyNames.MethodFullName, newFullName)
+        }
+        // fixup calls to static methods in the same compilation unit via the naive namespace-by-filename approach
+        if (method.isStatic.nonEmpty) {
+          val callCandidates = callsAffected.filter(_.file.exists(_.name == method.filename))
```
maltek, line 70: same problem with `.file`

maltek (commented):
I was going to caution that we need to preserve the links from a class with forward-declared member functions (usually in header files) to the implementations of those functions (usually not in headers). But I can't actually find any such links (e.g. method bindings) as of yet. So we'll just have to be careful about all that once we implement it.

```c++
class A {
    int foo();
}
```

```c++
#include "a.hpp"

int A::foo() {
    return 1;
}
```

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/FullNameUniquenessPass.scala
max-leuthaeuser: It uses astIn which is fine.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -197,7 +197,7 @@ trait AstForExpressionsCreator { this: AstCreator =>
 
             // TODO This wont do if the name is a reference.
             val name          = fieldRefExpr.getFieldName.toString
-            val signature     = functionTypeToSignature(functionType)
+            val signature     = s"${functionTypeToSignature(functionType)}$constFlag"
             val classFullName = cleanType(safeGetType(fieldRefExpr.getFieldOwnerType))
             val fullName      = s"$classFullName.$name$constFlag:$signature"
```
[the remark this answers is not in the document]
max-leuthaeuser: For C methods (that do not carry the signature in the fullname) that won't work.

The PolicyLinker compares bindings by name + signature so we need it there.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/FullNameProvider.scala
```diff
@@ -44,11 +44,12 @@ trait FullNameProvider { this: AstCreator =>
   }
 
   protected def methodFullNameInfo(methodLike: MethodLike): MethodFullNameInfo = {
-    val returnType_       = returnType(methodLike)
-    val signature_        = signature(returnType_, methodLike)
     val name_             = shortName(methodLike)
     val fullName_         = fullName(methodLike)
-    val sanitizedFullName = sanitizeMethodLikeFullName(name_, fullName_, signature_, methodLike)
+    val constSuffix       = if fullName_.contains(Defines.ConstSuffix) then Defines.ConstSuffix else ""
```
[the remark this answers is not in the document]
max-leuthaeuser: Ok, I'll reformat that.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/FullNameUniquenessPass.scala
maltek: Huh, that's weird. But indeed makes it fine for us here.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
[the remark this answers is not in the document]
max-leuthaeuser: Right, that does not make any sense. I will fix it.

max-leuthaeuser:
> I was going to caution that we need to preserve the links from a class with forward-declared member functions

We never create methods or bindings for forward-declared methods if we encounter an actual implementation later on.

maltek:
> We never create methods or bindings for forward-declared methods if we encounter an actual implementation later on.

Yes, that's what I was trying to say. We should at some point (once we also emit proper dynamic calls), but that's for another PR.

maltek (approved):

PR #5395, opened 2025-03-26
[c2cpg] Make bindings also unique to the PolicyLinker

max-leuthaeuser:
The PolicyLinker compares them  by name + signature so adjusting the fullname was not enough. We set the name now also. That should not break things as the REFed  method stays the same (REF edge is already there).


maltek (approved):
we should also adjust the `Name` property for the `callCandidates`

PR #5400, opened 2025-03-28
[x2cpg] Pass environment variable to gradle connection

max-leuthaeuser:


gacevedo:
Johannes is on vacation. @DavidBakerEffendi do you mind reviewing, please? Thank you.

max-leuthaeuser:
He will return tomorrow. I asked him already to do testing / a review via Slack. But if it's urgent I guess David can step in.

DavidBakerEffendi (approved):
Looks sane to me, seems like some dead code removed too in the PR

max-leuthaeuser:
The question is whether the env vars are actually passed. Do you have a test project that requires that to resolve the gradle dependencies?

DavidBakerEffendi:
Will create one now, sec

DavidBakerEffendi:
@max-leuthaeuser Here is a zip that requires `MY_BUILD_VAR` to be defined, or it throws an exception. There is also a variable that defines a dependency version but falls back on the default. You can use it to instead fall back toa blank string thus failing the dependency resolution. [Uploading sample-app.zip…]()

max-leuthaeuser:
Thanks, will test that.

johannescoetzee:
Our dependency fetching task doesn't depend on the `checkEnvVar` or `printJunitVersion` tasks (I'm guessing since we never have to build the project), so just fetching dependencies for `sample-app` succeeds regardless of whether `MY_BUILD_VAR` is set or not.

I tried moving the 
```
        def requiredVar = 'MY_BUILD_VAR'
        
        if (!System.getenv(requiredVar)) {
            throw new GradleException(
                "Build failed: Required environment variable '${requiredVar}' is not set.\n" +
                "Please set this variable before building, for example:\n" +
                "export ${requiredVar}=some_value"
            )
        }
```
check to the `dependency-fetcher-init.gradle` script which does fail if the var is not set, but if I set the var in my shell, then gradle picks it up even without setting environment variables in the builder.

As a last test, I added
```
          .setEnvironmentVariables(
            List(("MY_BUILD_VAR", "hello")).toMap.asJava
          )
```
to the builder which works as expected, so as long as `System.getEnv()` does return the required variables, this PR should work.

max-leuthaeuser:
Ok, thanks for testing that @johannescoetzee 

So we'll take it so.

PR #5405, opened 2025-04-01
[c2cpg] Header file parser detection

max-leuthaeuser:
This PR adds header file parser detection for header files with the following rules:

1) header file is not included from any source file: we will parse it based on the header files extension (just like without this PR)
2) included from at least one C source file: parse it as with the C parser
3) included from at least one C++ source file:  parse it as with the C++ parser
4) included from at least one C and one C++ source file: parse it with both the C parser and C++ parser

Transitive includes are supported.


max-leuthaeuser:
W.r.t. sptests: polarphp profits from this PR and got a couple of new findings in .h header files which are parsed correctly as C++ header files now.

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/HeaderFileParserDetectionTests.scala
```diff
@@ -0,0 +1,76 @@
...
+        "main.cpp"
+      ).moreCode(
+        """
+          |#include "util.h"
+          |int fooA(int a) { return 0; }
+          |""".stripMargin,
+        "main.c"
+      )
+      cpg.typeDecl.nameNot("<global>").internal.map(t => (t.fullName, t.file.name.head)).sorted.l shouldBe List(
+        // even class Foo is defined in a .h file we parse it as C++ because it's included in a .cpp file:
```
maltek, line 36:
```suggestion
        // even though class Foo is defined in a .h file we parse it as C++ because it's included in a .cpp file:
```

PR #5408, opened 2025-04-02
[x2cpg] Deprecate newBindingNode

johannescoetzee:


max-leuthaeuser:
Small hint: I did some minor refactoring for c2cpg w.r.t. stuff from AstNodeBuilder in https://github.com/joernio/joern/pull/5405. Its custom functions are gone now.

PR #5411, opened 2025-04-03
[kotlin2cpg] Another NPE fix

max-leuthaeuser:
Even so PR #5399 fixed the retrieval of a ClassDescriptor for a given function w.r.t. null safety we may still end up in a situation where the name or signature calculation throws a NPE in case of missing types.


maltek (approved):
if it's just about null, I'd prefer something that doesn't also catch and ignore all other kinds of exceptions. (E.g. `Option.flatMap(Option(...))` or a `for` comprehension)

max-leuthaeuser:
Some kotlin compiler api calls (like e.g., getName here) forward functionality to the underlying object of the AST node which may or may not be null (depending on the type resolving). I do not have access to them.

maltek:
ugh, I see.

PR #5420, opened 2025-04-07
Fix scaladoc warnings

max-leuthaeuser:
We had some `Scala3doc: No DRI found for query ` warnings which are now fixed by providing the correct scaladoc. (see: https://github.com/scala/scala3/issues/14212)

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifConfig.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifConfig.scala
@@ -25,8 +25,7 @@ import java.net.URI
   * @param resultConverter
   *   A transformer class to map from Finding nodes to a SARIF `Result`.
   * @param customSerializers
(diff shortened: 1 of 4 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/sarif/SarifConfig.scala
```diff
@@ -25,8 +25,7 @@ import java.net.URI
   * @param resultConverter
   *   A transformer class to map from Finding nodes to a SARIF `Result`.
   * @param customSerializers
-  *   Additional JSON serializers for any additional properties for [[io.shiftleft.semanticcpg.sarif.Sarif]] derived
```
max-leuthaeuser, line 28: Linking to Sarif did not work at all. Its a sealed trait inside `io.shiftleft.semanticcpg.sarif.SarifSchema` but can't be addressed.
I tried:
`io.shiftleft.semanticcpg.sarif.SarifSchema.Sarif` and also `io.shiftleft.semanticcpg.sarif.SarifSchema$$Sarif`

mpollmeier (approved):
🤷🏻

PR #5424, opened 2025-04-08
[x2cpg] Deprecate NodeBuilders

johannescoetzee:
Deprecate the `NodeBuilders` companion object along with 2 left-over uses


max-leuthaeuser (approved):
👍

PR #5434, opened 2025-04-11
[php] Add <tmp> assignment chain for call chains / property accesses

AndreiDreyer:
Attempting this issue revealed a number of things that were incorrectly implemented and that had to be fixed as a prerequisite:
* Fixed shadowing in the test suite of `builtin_functions.txt` in the resources
* Changed how calls are represented - these include a base and receiver in a similar manner to other interpreted languages in Joern
* Due to the nature of PHP, unless a dynamic call is dispatched within the scope of a class, there will not be a `$this` receiver.
* Built-in functions are statically dispatched
* Field accesses were found to use an identifier in place of a field identifier; this was rectified
* Chained calls use aliasing with temporary variables to ensure that field accesses and calls only occur once, as per the source code.

TODO:
* This broke some type recovery tests and may impact downstream users.
* The way receivers may often be an identifier referring to the call name, the local pass created a local for these - investigate if this is canonical
* Classes need to be split into an instance class and a singleton "meta" class, due to how object instances carry references to static functions. This needs to be referred to when static calls are made on object instances.
* Call bindings need to be created where possible.


joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -4,17 +4,39 @@ import io.joern.php2cpg.astcreation.AstCreator.NameConstants
...
+      arg.argumentIndex shouldBe 1
+    }
+
+    val contentsCall = cpg.call.nameExact("contents").head
+    contentsCall.code shouldBe "$a->contents()"
+    contentsCall.dispatchType shouldBe DispatchTypes.DYNAMIC_DISPATCH
+    inside(contentsCall.astChildren.l) { case (base: Call) :: (receiver: Identifier) :: Nil =>
+      base.name shouldBe Operators.fieldAccess
+      base.code shouldBe "$a->contents"
+      base.argumentIndex shouldBe -1
```
ml86, line 33: There must have been a misunderstanding in our last talk.
The base is the "this" argument and this is just an IDENTIFIER `a` in this case.
You can reference the same IDENTIFIER node as receiver. 
I just red through the PHP docu and played around with it a little bit and i think we can model the PHP classes and the call dispatch in a classic vtable style.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -4,17 +4,39 @@ import io.joern.php2cpg.astcreation.AstCreator.NameConstants
...
+
+      receiver.name shouldBe "a"
```
ml86, line 35: Please check explicitly for the existence of the receiver.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -187,4 +219,38 @@ class CallTests extends PhpCode2CpgFixture {
     val call = cpg.call("test").head
     call.code shouldBe "test(...$args)"
   }
+
+  "chained calls should alias calls in receivers and bases" in {
```
ml86, line 223: The hole tmp variable creation can be removed because in PHP the receiver and the this argument are always the same and you can just reference the same AST subtree.

ml86 (changes requested):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
DavidBakerEffendi: So, should we remove the `base` altogether? From what I understand, if the base and receiver are the same AST subtree the `tmp` stuff is there to reduce duplication between the two.

ml86: Please add one tree for the this argument and point to one argument and one receiver edge to it. No extra base tree.

PR #5436, opened 2025-04-13
[php] Refactor and Re-Organize

DavidBakerEffendi:
Removed dead code, organized `AstCreator` into traits. It's clear the following still need to be handled:
* `astForAnonymousClass`, `astforTraitUseStmt`, `astForAnonymousClassInstantiation` require implementations
* `astForCatchStmt` needs a local variable
* `astForUseStmt` and `astForGroupUseStmt` require local handling


max-leuthaeuser (approved):

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -0,0 +1,163 @@
...
+  protected def astForClassLikeStmt(stmt: PhpClassLikeStmt): Ast = {
+    stmt.name match {
+      case None       => astForAnonymousClass(stmt)
+      case Some(name) => astForNamedClass(stmt, name)
+    }
+  }
+
+  private def astForAnonymousClass(stmt: PhpClassLikeStmt): Ast = {
+    // TODO
+    Ast()
```
max-leuthaeuser, line 20: Could we return a NewUnknown and a debug log message just like the other frontends do for AST elements not yet implemented / supported?

max-leuthaeuser: This applies to all other similar code locations as well of course

PR #5438, opened 2025-04-15
[php] Add handling for anonymous classes

AndreiDreyer:
Added handling for anonymous class instantiation in PHP


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/FreshNameGenerator.scala
```diff
@@ -1,4 +1,4 @@
-package io.joern.rubysrc2cpg.utils
+package io.joern.x2cpg.utils
```
ml86, line 1: I do not think somethink like this is worth existing in x2cpg as a shared component. This is just an automatic `int + 1` and a function call.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -15,9 +15,9 @@ import java.nio.charset.StandardCharsets
 trait AstCreatorHelper(disableFileContent: Boolean)(implicit withSchemaValidation: ValidationMode) { this: AstCreator =>
 
   protected val globalNamespace: NewNamespaceBlock = globalNamespaceBlock()
-  private val tmpKeyPool                           = new IntervalKeyPool(first = 0, last = Long.MaxValue)
 
-  protected def getNewTmpName(prefix: String = "tmp"): String = s"$prefix${tmpKeyPool.next.toString}"
+  protected val tmpGen: FreshNameGenerator[String]   = FreshNameGenerator[String](i => s"tmp-$i")
+  protected val classGen: FreshNameGenerator[String] = FreshNameGenerator[String](i => s"anon-class-$i")
```
ml86, line 20: Both of these naming schemas do not work. We currently have problems in go2cpg which uses the exact same naming schema.
Reason: The naming changes if surrounding unrelated code changes. So lets say you have an anon class at the end of a file nested in a function. You then add a new anon class somewhere at the start of that file and the name of the anon class at the end of the file would change.
Why is that a problem? Because it makes it harder to correlate unchanged code parts between different scans which we need in order to correlated findings between different scans.

So please use a naming schema which only uses an incremental counter as a last resort and use structural information (mostly nesting comes to mind) in order to compose a unique name.

ml86 (changes requested):
I will review the rest once the naming schema change is in.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
```diff
@@ -369,13 +369,13 @@ class TypeDeclTests extends PhpCode2CpgFixture {
...
     "parse methods in classes correctly" in {
-      inside(cpg.typeDecl.name("anon-class-\\d+").l) {
+      inside(cpg.typeDecl.name("Test0.php:<global>@anon-class-\\d+").l) {
         case anonClass0 :: anonClass1 :: Nil =>
           val List(memberX) = anonClass0.member.l
           memberX.code shouldBe "$x"
 
           val List(anonConstructor0) = anonClass0.method.name("__construct").l
-          anonConstructor0.fullName shouldBe "anon-class-0->__construct"
+          anonConstructor0.fullName shouldBe "Test0.php:<global>@anon-class-0->__construct"
```
ml86, line 378: Is there some sort of special meaning of `@` and `->` in php? Why not stick to the `.` separator we use for most of the other frontends?

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
```diff
@@ -350,4 +350,158 @@ class TypeDeclTests extends PhpCode2CpgFixture {
...
+    inside(cpg.typeDecl.name("C->D@anon-class-\\d+").l) {
+      case anonClass :: Nil =>
+        anonClass.fullName shouldBe s"C->D@anon-class-0"
+        anonClass.member.code.l shouldBe List("$x")
+      case xs => fail(s"Expected one anonymous class, got ${xs.code.mkString("[", ",", "]")}")
+    }
+
+    inside(cpg.method.name("D").body.astChildren.l) {
+      case (localNode: Local) :: (bodyBlock: Block) :: Nil =>
+        localNode.code shouldBe "$C->D@tmp-0"
```
ml86, line 489: Sorry my last comment was not quite correct/clear. 
The temp variables cannot be just numbered via an incremented suffix on a per file basis. So the suffix needs to be "per scope" but since the locals are locals and do not need a globally unique name, it is ok to just name them "tmp0" and so on.

ml86 (changes requested):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
AndreiDreyer: From my understanding this `local` node's name is based on the name of the identifier in the method body so I don't think we should modify this in any way as this links to the identifier in the method. From my understanding this local node is being generated in the `LocalCreationPass`

image

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
AndreiDreyer: I just used the `@` as a separator between the namespace and the anon-class name so we have `namespace@anon-class`. I can change this to be `.`

The `->` is what is always used when creating the fullName for the `init` call (i.e. the call to `__construct`) so that I would say we keep as is (`AstForExpressionCreator:847`)

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
ml86: Yes, the name of a local an the name of the corresponding identifiers should be the same.
Please add tests for the name.
From the graph picture i cannot see what is name and what is code.
For now the local names can stay in this full name style but we will need to change that once the local generation is reworked.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
ml86: So there is no apparent reason as it looks. Please make a followup PR where only `.` instead of `@` and `->`.
This would also require adjusting the policy full names and the sptest expectation files.

ml86 (approved):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
maltek: PHP uses `->` for member/method access. Since the `.` operator was already taken to mean string concat when OOP support was added. Makes sense to me to use the same in our names.

AndreiDreyer: @maltek @ml86 - should I go ahead with a new PR and make the changes to the `@` and `->`?

ml86: @AndreiDreyer Please wait with such a PR. There are other questions regarding the full names which need to get sorted out first.

PR #5442, opened 2025-04-17
[c2cpg] Option-safe scope implementation

max-leuthaeuser:
No .get calls on Option anymore.

Follow-up to: https://github.com/joernio/joern/pull/5439


mpollmeier (approved):
nice one!

PR #5447, opened 2025-04-23
[ruby] Add Support for ERB files

AndreiDreyer:
* Added support for basic ERB files
* Added `templateOutRaw` (`<%== %>`) and `templateOutEscape` (`<%= %>`) operator calls
* Added return for the generated fmtString from the ERB processing
* Added `.html.erb` files as config files

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
@@ -6,6 +6,7 @@ import io.joern.rubysrc2cpg.parser.RubyJsonHelpers
 import io.joern.rubysrc2cpg.passes.Defines
 import io.joern.rubysrc2cpg.passes.GlobalTypes
 import io.joern.rubysrc2cpg.passes.Defines.{RubyOperators, prefixAsKernelDefined}
+import io.joern.x2cpg.frontendspecific.rubysrc2cpg.Constants
 import io.joern.x2cpg.{Ast, ValidationMode, Defines as XDefines}
 import io.shiftleft.codepropertygraph.generated.nodes.*
 import io.shiftleft.codepropertygraph.generated.{
@@ -85,17 +86,29 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
   protected def astForDynamicLiteral(node: DynamicLiteral): Ast = {
     val fmtValueAsts = node.expressions.collect {
       case stmtList: StatementList if stmtList.size == 1 =>
-        val expressionAst = astForExpression(stmtList.statements.head)
-        val call = callNode(
-          node = stmtList,
-          code = stmtList.text,
-          name = Operators.formattedValue,
-          methodFullName = Operators.formattedValue,
-          dispatchType = DispatchTypes.STATIC_DISPATCH,
-          signature = None,
-          typeFullName = Some(node.typeFullName)
-        )
-        callAst(call, Seq(expressionAst))
+        stmtList.statements.head match {
+          case x: SimpleCall if x.span.text.startsWith("joern__") =>
+            val argAsts = x.arguments.map(astForExpression)
+            val (opName, callCode) = if (x.span.text.startsWith("joern__template_out_escape")) {
+              (RubyOperators.templateOutEscape, s"<%= ${x.arguments.headOption.map(_.span.text).getOrElse("")} %>")
+            } else {
+              (RubyOperators.templateOutRaw, s"<%== ${x.arguments.headOption.map(_.span.text).getOrElse("")} %>")
+            }
+            val opNode = callNode(x, callCode, opName, opName, DispatchTypes.STATIC_DISPATCH)
+            callAst(opNode, argAsts)
+          case x =>
+            val expressionAst = astForExpression(stmtList.statements.head)
+            val call = callNode(
+              node = stmtList,
+              code = stmtList.text,
+              name = Operators.formattedValue,
+              methodFullName = Operators.formattedValue,
+              dispatchType = DispatchTypes.STATIC_DISPATCH,
+              signature = None,
+              typeFullName = Some(node.typeFullName)
+            )
+            callAst(call, Seq(expressionAst))
+        }
       case stmtList: StatementList if stmtList.size > 1 =>
         logger.warn(
           s"Interpolations containing multiple statements are not supported yet: ${stmtList.text} ($relativeFileName), skipping"
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
@@ -83,7 +83,7 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
   protected def astForNilBlock: Ast = blockAst(NewBlock(), List(astForNilLiteral))
 
   protected def astForDynamicLiteral(node: DynamicLiteral): Ast = {
-    val fmtValueAsts = node.expressions.map {
+    val fmtValueAsts = node.expressions.collect {
       case stmtList: StatementList if stmtList.size == 1 =>
         val expressionAst = astForExpression(stmtList.statements.head)
         val call = callNode(
@@ -101,7 +101,7 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
           s"Interpolations containing multiple statements are not supported yet: ${stmtList.text} ($relativeFileName), skipping"
         )
         astForUnknown(stmtList)
-      case node =>
+      case node if (!isLineFeed(node.text)) =>
         val call = callNode(
           node = node,
           code = node.text,
@@ -1087,4 +1087,6 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
   private def getUnaryOperatorName(op: String): Option[String] = UnaryOperatorNames.get(op)
 
   private def getAssignmentOperatorName(op: String): Option[String] = AssignmentOperatorNames.get(op)
+
+  private def isLineFeed(text: String): Boolean = text == "\n" || text == "\r" || text == "\r\n"
 }
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ErbTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ErbTests.scala
@@ -1,0 +1,196 @@
+package io.joern.rubysrc2cpg.querying
+
+import io.joern.x2cpg.Defines
+import io.joern.x2cpg.frontendspecific.rubysrc2cpg.Constants
+import io.joern.rubysrc2cpg.testfixtures.RubyCode2CpgFixture
+import io.shiftleft.codepropertygraph.generated.{ControlStructureTypes, DispatchTypes, Operators}
+import io.shiftleft.codepropertygraph.generated.nodes.{Block, Call, FieldIdentifier, Identifier, Literal}
+import io.shiftleft.semanticcpg.language.*
+
+class ErbTests extends RubyCode2CpgFixture {
+  "Complete ERB processing" should {
+    val cpg = code(
+      """
+        |app_name: <%= ENV['APP_NAME'] %>
+        |version: <%= ENV['APP_VERSION'] %>
+        |
+        |database:
+        |  host: <%= ENV['DB_HOST'] %>
+        |  port: <%= ENV['DB_PORT'] %>
+        |
+        |<% if ENV['USE_REDIS'] == 'true' %>
+        |redis:
+        |  host: <%= ENV['REDIS_HOST'] %>
+        |  port: <%= ENV['REDIS_PORT'] %>
+        |<% end %>
+        |""".stripMargin,
+      "test.erb"
+    )
+
+    "Contain a call to <operator>.conditional" in {
+      inside(cpg.call.name(Operators.conditional).l) {
+        case _ :: Nil =>
+        // Do nothing - contains call to operator.conditional
+        case _ => fail("AST Should contain one if statement")
+      }
+    }
+
+    "Contain a call to fmtString in the true branch of the <operator>.conditional" in {
+      inside(cpg.call.name(Operators.conditional).argument.l) {
+        case (condition: Call) :: (trueBranch: Block) :: _ :: Nil =>
+          condition.code shouldBe "ENV['USE_REDIS'] == 'true'"
+          condition.methodFullName shouldBe Operators.equals
+
+          inside(trueBranch.astChildren.isCall.l) {
+            case fmtStringCall :: Nil =>
+              fmtStringCall.methodFullName shouldBe Operators.formatString
+              fmtStringCall.argument.l.size shouldBe 4
+              val List(fmtVal1: Call, fmtVal2: Call, fmtVal3: Call, fmtVal4: Call) =
+                fmtStringCall.argument.l: @unchecked
+
+              fmtVal1.methodFullName shouldBe Operators.formattedValue
+              fmtVal1.code shouldBe
+                """redis:
+                  |  host: """.stripMargin
+
+              fmtVal2.methodFullName shouldBe Operators.formattedValue
+              fmtVal2.code shouldBe "#{ENV['REDIS_HOST']}"
+
+              fmtVal3.methodFullName shouldBe Operators.formattedValue
+              fmtVal3.code shouldBe "  port: "
+
+              fmtVal4.methodFullName shouldBe Operators.formattedValue
+              fmtVal4.code shouldBe "#{ENV['REDIS_PORT']}"
+            case xs => fail(s"Expected one call to fmtString, got [${xs.mkString(",")}]")
+          }
+        case xs => fail(s"Expected three arguments, got ${xs.size}: [${xs.mkString(",")}] instead")
+      }
+    }
+
+    "Contains calls to formattedValue" in {
+      cpg.call.name(Operators.formattedValue).l.size shouldBe 14
+    }
+  }
+
+  "ERB With If-Elsif contains IF-Struct" in {
+    val cpg = code(
+      """
+        |app_name: <%= ENV['APP_NAME'] %>
+        |version: <%= ENV['APP_VERSION'] %>
+        |
+        |database:
+        |  host: <%= ENV['DB_HOST'] %>
+        |  port: <%= ENV['DB_PORT'] %>
+        |
+        |<% if ENV['USE_REDIS'] == 'true' %>
+        |redis:
+        |  host: <%= ENV['REDIS_HOST'] %>
+        |  port: <%= ENV['REDIS_PORT'] %>
+        |<% elsif ENV['USE_RABBITMQ'] == 'true' %>
+        |rabbitmq:
+        |  host: <%= ENV['RABBITMQ_HOST'] %>
+        |  port: <%= ENV['RABBITMQ_PORT'] %>
+        |<% end %>
+        |""".stripMargin,
+      "test.erb"
+    )
+
+    inside(cpg.controlStructure.controlStructureType(ControlStructureTypes.IF).l) {
+      case ifStruct :: _ :: Nil =>
+        inside(ifStruct.condition.l) {
+          case (cond: Call) :: Nil =>
+            cond.methodFullName shouldBe Operators.equals
+            val List(lhs: Call, rhs: Literal) = cond.argument.l: @unchecked
+            lhs.methodFullName shouldBe Operators.indexAccess
+            rhs.code shouldBe "'true'"
+          case xs => fail(s"Expected one condition, got [${xs.code.mkString(",")}]")
+        }
+
+        inside(ifStruct.whenTrue.isBlock.astChildren.isCall.l) {
+          case fmtStringCall :: Nil =>
+            fmtStringCall.methodFullName shouldBe Operators.formatString
+            fmtStringCall.argument.l.size shouldBe 4
+            val List(fmtVal1: Call, fmtVal2: Call, fmtVal3: Call, fmtVal4: Call) = fmtStringCall.argument.l: @unchecked
+
+            fmtVal1.methodFullName shouldBe Operators.formattedValue
+            fmtVal1.code shouldBe
+              """redis:
+                |  host: """.stripMargin
+
+            fmtVal2.methodFullName shouldBe Operators.formattedValue
+            fmtVal2.code shouldBe "#{ENV['REDIS_HOST']}"
+
+            fmtVal3.methodFullName shouldBe Operators.formattedValue
+            fmtVal3.code shouldBe """  port: """
+
+            fmtVal4.methodFullName shouldBe Operators.formattedValue
+            fmtVal4.code shouldBe "#{ENV['REDIS_PORT']}"
+
+          case xs => fail(s"Expected one body for true branch, got [${xs.code.mkString(",")}]")
+        }
+
+        inside(ifStruct.whenFalse.isBlock.astChildren.isControlStructure.l) {
+          case elsifStruct :: Nil =>
+            inside(elsifStruct.condition.l) {
+              case (cond: Call) :: Nil =>
+                cond.methodFullName shouldBe Operators.equals
+                val List(lhs: Call, rhs: Literal) = cond.argument.l: @unchecked
+                lhs.methodFullName shouldBe Operators.indexAccess
+                rhs.code shouldBe "'true'"
+              case xs => fail(s"Expected one condition, got [${xs.code.mkString(",")}]")
+            }
+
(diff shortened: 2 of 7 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -83,25 +84,37 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
-          methodFullName = Operators.formattedValue,
-          dispatchType = DispatchTypes.STATIC_DISPATCH,
-          signature = None,
-          typeFullName = Some(node.typeFullName)
-        )
-        callAst(call, Seq(expressionAst))
+        stmtList.statements.head match {
+          case x: SimpleCall if x.span.text.startsWith("joern__") =>
+            val argAsts = x.arguments.map(astForExpression)
+            val (opName, callCode) = if (x.span.text.startsWith("joern__template_out_escape")) {
```
DavidBakerEffendi, line 92: Might want to associate constants with the joern__ stuff

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -83,25 +84,37 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+              (RubyOperators.templateOutEscape, s"<%= ${x.arguments.headOption.map(_.span.text).getOrElse("")} %>")
+            } else {
+              (RubyOperators.templateOutRaw, s"<%== ${x.arguments.headOption.map(_.span.text).getOrElse("")} %>")
```
DavidBakerEffendi, line 95: Do we expect more arguments possibly? Could make this `x.arguments.map(_.span.text).mkString`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -83,25 +84,37 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+            )
+            callAst(call, Seq(expressionAst))
+        }
       case stmtList: StatementList if stmtList.size > 1 =>
         logger.warn(
           s"Interpolations containing multiple statements are not supported yet: ${stmtList.text} ($relativeFileName), skipping"
         )
         astForUnknown(stmtList)
-      case node =>
+      case node if (!isLineFeed(node.text)) =>
```
DavidBakerEffendi, line 117:
```suggestion
      case node if !isLineFeed(node.text) =>
```

DavidBakerEffendi (commented):
Nice

maltek:
I'm still missing a lot of .erb files. E.g. for https://github.com/chatwoot/chatwoot:

```
ocular> cpg.file.name(".*.html.erb").size
val res18: Int = 40

ocular> cpg.configFile.name(".*.html.erb").size
val res19: Int = 88
```

The frontend spits out some warnings for the project, e.g. `Yield expression outside of method scope: yield(:head)` and `Type 'String' is considered a 'core' type, not a 'Kernel-contained' type` - but not enough to explain 48 missing files.

AndreiDreyer:
@maltek these are the new results on chatwoot with all of the changes made for ERB handling. 
```
joern> cpg.configFile.name(".*.html.erb").l.map(_.name).l.diff(cpg.file.name(".*.html.erb").map(_.name).l)
val res1: List[String] = List()

joern> cpg.file.name(".*.html.erb").l.size
val res2: Int = 88

joern> cpg.configFile.name(".*.html.erb").l.size
val res3: Int = 88
```

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ErbTests.scala
```diff
@@ -212,4 +189,162 @@ class ErbTests extends RubyCode2CpgFixture {
...
     }
   }
+
+  // Test for lowered do-block
+  "Do-block in <%= %> tags" should {
+    /* The code gets lowered to:
+     * ```
+     *   joern__buffer = ""
+     *   joern__buffer << form_with(url: some_url)
+     *   rails_lambda_0 = lambda do |form|
```
maltek, line 199: both `joern__buffer` and `rails_lambda_0` should have names that are invalid ruby identifiers so they can't possibly clash with real names from the program (e.g. `<output_buffer>` and `<rails_lambda>0`)

AndreiDreyer: We can adjust this in the `RubyJsonToNodeCreator` then. It isn't possible to give these variables invalid identifiers during the `ruby_ast_gen` phase, otherwise we won't get an AST from `ruby_ast_gen`

maltek: No, at that point we'd just be fooling ourselves about whether clashes with names in the analyzed program are possible or not.

If there are technical reasons that we must parse these names as Ruby identifiers, then we just have to live with the potential name clashes as a limitation.

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ErbTests.scala
```diff
@@ -0,0 +1,350 @@
...
+        case erbReturn :: Nil =>
+          val List(buffer: Identifier) = erbReturn.astChildren.l: @unchecked
+          buffer.code shouldBe "joern__buffer"
+        case xs => fail(s"Expected one method return, got ${xs.code.mkString("[", ",", "]")}")
+      }
+    }
+
+    "Contains 6 calls to append (<<)" in {
+      cpg.call.name("<<").l.size shouldBe 6
+    }
```
ml86, line 37: Please stop writing such tests. This does not tell us anything.

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ErbTests.scala
```diff
@@ -0,0 +1,350 @@
...
+
+import io.joern.rubysrc2cpg.passes.Defines.RubyOperators
+import io.joern.x2cpg.Defines
+import io.joern.x2cpg.frontendspecific.rubysrc2cpg.Constants
+import io.joern.rubysrc2cpg.testfixtures.RubyCode2CpgFixture
+import io.shiftleft.codepropertygraph.generated.{ControlStructureTypes, DispatchTypes, Operators}
+import io.shiftleft.codepropertygraph.generated.nodes.{Block, Call, FieldIdentifier, Identifier, Literal, TypeRef}
+import io.shiftleft.semanticcpg.language.*
+
+class ErbTests extends RubyCode2CpgFixture {
```
ml86, line 11: Why are you writing tests that include the lowering of a ruby AST into CPG?
The ERB support takes input ERB files and emits ruby code that is then fed into rubyAstGen. For the later part we already have plenty of tests. Here it would be much better readable if you test the first step ( ERB -> ruby code) the added benefit would be much better readability.

ml86 (changes requested):

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ErbTests.scala
DavidBakerEffendi: ERB -> Ruby code is tested on the `ruby_ast_gen` project e.g., https://github.com/joernio/ruby_ast_gen/blob/main/spec/ruby_ast_gen_spec.rb#L155

Albeit, one could argue for more tests there perhaps, but we haven't designed the RubyAstGen hook to give us the lowered ERB code as-is, only via AST

ml86 (approved):

maltek:
all the ruby code in
https://github.com/forem/forem/blob/0772f2d49b18d94f3b982b39420ea31235c1c8aa/app/views/layouts/application.html.erb#L88-L92 exists in the CPG only as `Literal` nodes, instead of the actual calls and control structures. Also the lines are off-by-one (code from line 90 has line 91 in the CPG).

maltek:
current problem:

```ruby
class UsersController < ActionController::Base
  def show
    respond_to do |format|
      format.json { render partial: "foo" }
    end
  end
end
```

﻿some `self` identifiers refer to multiple locals:

```scala
cpg.identifier.name("self").filter(_.refsTo.size != 1).l
```

Identifiers should only ever have REF edges to the single variable of that name in the exact same method.

maltek:
@TNSelahle the problem with the `self` variable is fixed with your last change. But there is still a related problem: the `format` identifiers are not referencing the `format` `MethodParameterIn` node.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -191,44 +185,73 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+      }
       val argumentAsts = n.arguments.map(astForMethodCallArgument)
       val dispatchType = if (isStatic) DispatchTypes.STATIC_DISPATCH else DispatchTypes.DYNAMIC_DISPATCH
 
       val callCode = if (baseCode.contains("<tmp-")) {
         val rhsCode =
           if (n.methodName == "new") n.methodName
+          else if (isErbCall && hasStringArguments)
+            s"${n.methodName} \"${n.arguments.map(code(_).replaceAll("\n", "\\\\n")).mkString(", ")}\""
+          else if (isErbCall) s"${n.methodName} ${n.arguments.map(code(_).replaceAll("\n", "\\\\n")).mkString(", ")}"
```
maltek: I don't think we have to try this hard to make it look like the code field of the calls is valid ruby code. I'd just remove this special case.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -300,14 +323,19 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
           scope.tryResolveTypeReference(x.typeName).map(_.name).getOrElse(Defines.Any)
       }
       .orElse(Option(Defines.Any))
+    val typeFullName = if (node.memberName == "joern__buffer" || node.memberName == "joern__inner_buffer") {
+      Constants.stringPrefix
+    } else {
+      Defines.Any
+    }
```
maltek: I don't think we should handle this special case in the general `astForFieldAccess` method. Can we make the type an optional parameter that's handed down here from the surrounding context if it's known?

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -205,22 +222,30 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
...
     // Determine which locals are captured
     val capturedLocalNodes = baseStmtBlockAst.nodes
-      .collect { case x: NewIdentifier if x.name != Defines.Self => x } // Self identifiers are handled separately
+      .collect { case x: NewIdentifier if x.name != Defines.Self => x }
       .distinctBy(_.name)
-      .map(i => scope.lookupVariableInOuterScope(i.name))
-      .filter(_.nonEmpty)
+      .map {
+        case i if i.name == Defines.Self => scope.lookupSelfInOuterScope
+        case i                           => scope.lookupVariableInOuterScope(i.name)
```
maltek: this is more complicated than it needs to be - `lookupSelfInOuterScope` is merely a shortcut for `lookupVariableInOuterScope(Defines.Self)`

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -229,18 +254,13 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
...
-          (local, local.name, local.code, closureBindingId)
-        case param: NewMethodParameterIn =>
-          val closureBindingId = scope.variableScopeFullName(param.name).map(x => s"$x.${param.name}")
-          (param, param.name, param.code, closureBindingId)
-      }
+
+    createClosureBindingInformation(capturedNodes)
       .collect { case (capturedLocal, name, code, Some(closureBindingId)) =>
-        val capturingLocal =
+        val selfCapturingLocal = if name == Defines.Self then scope.lookupSelfInCurrentScope else None
```
maltek: we still use classic `if (...) { } else {}` style in this repo, not `if ... then ... else`. In this case, `Option.when` is an equally valid choice.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/RubyIntermediateAst.scala
```diff
@@ -685,4 +689,5 @@ object RubyIntermediateAst {
       extends RubyExpression(span)
       with AllowedTypeDeclarationChild
 
+  final case class EmptyExpression()(span: TextSpan) extends RubyExpression(span)
```
maltek: drop the double parameter list (unless it does something that I'm not getting)

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/datastructures/RubyScope.scala
```diff
@@ -152,6 +152,20 @@ class RubyScope(summary: RubyProgramSummary, projectRoot: Option[String])
...
+    }
+  }
+
+  def lookupSelfInCurrentScope: Option[DeclarationNew] = {
+    stack.headOption.collectFirst {
+      case scopeElement if scopeElement.variables.contains(RubyDefines.Self) =>
+        scopeElement.variables(RubyDefines.Self)
+    }
+  }
+
```
maltek: I think these would be clearer as simple wrappers for the `lookupVariableIn*Scope()` methods

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/LiteralTests.scala
```diff
@@ -262,10 +261,11 @@ class LiteralTests extends RubyCode2CpgFixture {
         |-> (a, *b, &c) {}
         |""".stripMargin)
 
+    cpg.method.name(Constants.Main).dotAst.l.foreach(println)
```
maltek: stray debug output

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/frontendspecific/rubysrc2cpg/Constants.scala
```diff
@@ -4,9 +4,15 @@ object Constants {
...
   val builtinPrefix = "__core"
   val kernelPrefix  = s"$builtinPrefix.Kernel"
+  val stringPrefix  = s"$builtinPrefix.String"
   val Initialize    = "initialize"
   val Main          = "<main>"
 
+  val joernErbTemplatePrefix        = "joern__"
+  val joernErbTemplateOutRawName    = s"${joernErbTemplatePrefix}template_out_raw"
+  val joernErbTemplateOutEscapeName = s"${joernErbTemplatePrefix}template_out_escape"
+  val joernErbBufferAppend          = s"${joernErbTemplatePrefix}buffer_append"
```
maltek: it's confusing that these use snake case, but in our lowering we use camel case, e.g. `<operator>.joernBufferAppend`. Our operators are all camel case. So I think we should leave those and change the astgen output to match that.

TNSelahle: I suspect it was done like this because snake_case is being used in the Ruby code generated from the ERB files. I think the implementation in `ruby_ast_gen` should try to be free from influence of its users, like joern. `ruby_ast_gen` doesn't know about joern operators and the writing style required there. It just converts the ERBs into Ruby, which typically uses snake_case for symbols and methods, then into an AST. 

I'm happy to change this though as I'm only speculating here and don't have the prior context :D

maltek: The method is called `joernBufferAppend`. It's very much only made for joern :). The spelling may stick out in ruby. But the name leaks through e.g. in the `code` field of those calls. And I very much dislike the inconsistency: I can already see myself wasting a whole day wondering why some query doesn't work. Only to find that it was correct all along - except I spelled it like in the code field, not like in the method full name.

TNSelahle: That's fair :D the inconsistency is not nice. I had a chat with @ml86 earlier this week about the `joern` part being in the generated Ruby code, but we concluded that we could just leave it as is for now but given your point, I think then it's definitely worth changing. I also see myself being confused by the inconsistency between the `code` field and the method full name xD. 

I could implement your suggestion of just changing the writing style to `camelCase` in `ruby_ast_gen`. The other option is to do that along with removing the `joern` reference, but I think if we change the writing style, we might as well leave in the `joern` part.

TNSelahle: @maltek I've made the changes in `ruby_ast_gen`. Please see PR https://github.com/joernio/ruby_ast_gen/pull/35

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/datastructures/RubyScope.scala
TNSelahle: I left `lookupSelfInCurrentScope` as is since there's no current use case for looking up a variable in the current scope, except this.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForFunctionsCreator.scala
TNSelahle: The initial case was not even required cos identifiers with `name == "self"` were filtered out in the `collect` up top. I removed the match.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -191,44 +190,70 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
   protected def astForMemberCall(node: RubyCallWithBase, isStatic: Boolean = false): Ast = {
 
     def createMemberCall(n: RubyCallWithBase): Ast = {
-      val receiverAst = astForFieldAccess(MemberAccess(n.target, ".", n.methodName)(n.span), stripLeadingAt = true)
+      val isErbCall = n.methodName == Constants.joernErbBufferAppend
+
+      val (nodeOp, hasStringArguments, receiverAst) = if (isErbCall) {
```
maltek: `hasStringArguments` is now unused

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
maltek: It's sort of the same special case now, just one layer up. I thought we could do it from around here where we've already determined to be dealing with the ERB lowering. But I just read over the diff in the browser, so I could very well be mistaken.

https://github.com/joernio/joern/blob/bfd14fcfde74a9b224322a7165c5b7b11616d1e4/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala#L200-L201

TNSelahle:
> https://github.com/joernio/joern/pull/5447#discussion_r2440321788

@maltek the snippet you highlighted applies to `astForFieldAccess` for the `self.joernBufferAppend` call. The special case for you identified in the `astForFieldAccess` method was for field access for `self.joernBuffer` and `self.joernInnerBuffer`. 

I'll rename `isErbCall` to `isErbBufferApppendCall` to make it clearer.

maltek:
@TNSelahle The only field access in the AST for `<operator>.joernBufferAppend` calls that we have in the CPG are field accesses for `self.joern__buffer`. But now I see that this location I pointed to creates the "receiver ast" - and then just discards it in the `joernBufferAppend` case. Please change this, so the nodes for the receiver are only created when it's going to be used. (A frontend that creates AST nodes but then fails to actually add them to the AST is a somewhat common source of runtime errors.)

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -223,12 +231,21 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+      }
+      if (isErbBufferAppendCall) {
+        callAst(call, argumentAsts)
       } else {
-        callAst(call, argumentAsts, base = Option(baseAst), receiver = Option(receiverAst))
+        if (isStatic) {
+          callAst(call, argumentAsts, base = Option(baseAst)).copy(receiverEdges = Nil)
+        } else {
+          callAst(call, argumentAsts, base = Option(baseAst), receiver = receiverAstOpt)
+        }
```
maltek, line 248: this whole `if`-cascade should also simplify to a single call `callAst(call, argumentAsts, base = Option(baseAst), receiver = receiverAstOpt)` - after adjusting the condition used for creating `receiverAstOpt`

TNSelahle: I tried that before and found that `.copy(receiverEdges = Nil)` was done to remove the receiver edges coming from the `baseAst`.

maltek: can't we do the same as with the receiver? Conditionally create it only in case we want to actually make use of it

TNSelahle: Seems like the creation of the receiver stores some state somewhere that's then used in the creation of `baseAst`. I'm getting a couple of failed tests if I don't create the receiver when `isStatic` is true. The difference I spotted is that `baseAst` has more nodes in it when `astForFieldAccess` is not called for the receiver. I'm not 100% sure what state it is that's at play here but I'm busy debugging to try figure that out.

TNSelahle: Might have something to do with this `TODO` -> https://github.com/joernio/joern/blob/39947b29fb76b2a7139f011aa4c84bcc37af0ec5/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForExpressionsCreator.scala#L345-L363

`handleTmpGen` gets called during the creation of the receiver (`astForFieldAccess`) and triggers this flow. It's then also triggered when the `baseAst` is created. I don't understand enough yet to know why these side-effects exist yet the result of `astForFieldAccess` isn't used when `isStatic` is true :/. 

I'll see if I can figure out how to make this better. I have a call with @ml86 later and will probably ask for some guidance on this.

TNSelahle: @ml86 on closer inspection, I think this implementation is okay. The way the method `createMemberCall` was doing its operations was actually incorrect. It was resulting in some `tmp` assignments being omitted. You can have a look at my diff here to double check if what I did makes sense: https://github.com/joernio/joern/pull/5447/files/5319cc841d17e36954830f09782ecbf2438c99f7..f0fdf28d92ae5f751eedb93d76379c9d078e159b

ml86 (approved):

PR #5455, opened 2025-04-28
[PHP] Add query to detect magic hash

peng-hui:
Magic hash exploits the type juggling in PHP, for example, the following conditional statement would be evaluated as true. 
```php
<php
if(md5("240610708")== md5("QLTHNDT"))
    echo "good";
```
This is because 
- `md5("240610708")= "0e462097431906509019562988736854"`
-  `md5("QLTHNDT")="0e405967825401955372549139051580"`
- Both form the scientific notation of 0*10^xxx, then 0;

This PR query identifies flows from 1. source -> hash functions, and 2. hash functions -> loose comparison.


maltek (commented):
thank you, this looks good. I just have some small suggestions.

maltek (approved):

PR #5456, opened 2025-04-28
[PHP Case Sensitivity] PHP is case-insensitive for namespaces, function names, class names, and keywords, but case-sensitive for variable names.

peng-hui:
PHP APIs are case-insensitive, like `echo` is the same as `eCHo`. Add `(?i)` in the queries to realize this.


DavidBakerEffendi:
I think the solution for this should be in the frontend itself. e.g., the `name` property of method and call nodes should be lowercased. Will incorporate that in upcoming PRs for PHP

Issue #5443, opened 2025-04-21
[Bug] Error while joern-export for cpg


wildoranges:
**Describe the bug**
I tried to use joern to parse openssl and export its cpg using the following commands:

```bash
$ git clone https://github.com/openssl/openssl
$ cd openssl
$ git checkout 871182d29dc24bbb51ff7a4d952ff9d9868f14b5
$ joern-parse --max-num-def 99999  ./ -o cpg.bin --frontend-args --with-include-auto-discovery
$ joern-export --repr cpg --format dot --out ./cpg
```

however, `joern-export` failed with the following error:

```
Exception in thread "main" java.nio.file.AccessDeniedException: /__DATE__.dot
        at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:90)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
        at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:261)
        at java.base/java.nio.file.spi.FileSystemProvider.newOutputStream(FileSystemProvider.java:482)
        at java.base/java.nio.file.Files.newOutputStream(Files.java:227)
        at java.base/java.nio.file.Files.newBufferedWriter(Files.java:2988)
        at java.base/java.nio.file.Files.newBufferedWriter(Files.java:3036)
        at flatgraph.formats.dot.DotExporter$.runExport(DotExporter.scala:43)
        at io.joern.joerncli.JoernExport$.$anonfun$1(JoernExport.scala:155)
        at scala.collection.Iterator$$anon$9.next(Iterator.scala:584)
        at scala.collection.IterableOnceOps.reduceLeft(IterableOnce.scala:861)
        at scala.collection.IterableOnceOps.reduceLeft$(IterableOnce.scala:851)
        at scala.collection.AbstractIterator.reduceLeft(Iterator.scala:1303)
        at scala.collection.IterableOnceOps.reduce(IterableOnce.scala:812)
        at scala.collection.IterableOnceOps.reduce$(IterableOnce.scala:812)
        at scala.collection.AbstractIterator.reduce(Iterator.scala:1303)
        at io.joern.joerncli.JoernExport$.exportWithFlatgraphFormat(JoernExport.scala:157)
[...]

PR #5459, opened 2025-04-29
joern export: fix file sanitization: never try to write to `/`...

mpollmeier:
fixes https://github.com/joernio/joern/issues/5443


max-leuthaeuser (approved):

max-leuthaeuser:
Due to empty filenames coming from c2cpg that are fixes with https://github.com/joernio/joern/pull/5460
But good to have this PR here anyway 👍

PR #5467, opened 2025-05-05
[php] Pre-Parse and Summarize Symbols

DavidBakerEffendi:
To resolve symbols and types that are brought into the current context via `use` functions, the approach of pre-parsing the AST for high-level symbols is implemented. A similar approach is used in Ruby, and this follows on that.

As this "summary" object is pretty similar to that in Ruby, we may want to unify these at some point with a "default" implementation of the summary objects to reduce the number of classes.


DavidBakerEffendi:
Superseded by https://github.com/joernio/joern/pull/5480

PR #5468, opened 2025-05-05
[swiftsrc2cpg] Refactor synthetic name generation

max-leuthaeuser:
To be more in line with the other frontends.


maltek (changes requested):
The goal must always be that the generated names should be as stable as possible, even when some of the code in the file changes. So we should be using separate counters for separate things (methods vs. `<iterator>` variables vs. `<wildcard>` variables... ). And the scope of each of them should be as small as possible, e.g. for variables it should be a per-method counter, not per-file.

(I know the other frontends aren't perfect about this either. That's bad there, too. Changing it there is a compatibility hazard itself, so if it isn't completely unstable we will live with it.)

max-leuthaeuser:
I made this PR actually just for the  naming itself. The index is another story and should be generated with individual counters and method-local, totally agree with that. At least for c2cpg we have the exact same issue.

maltek (approved):

PR #5470, opened 2025-05-06
[c2cpg] Implement scope-local unique names

max-leuthaeuser:
Just like in https://github.com/joernio/joern/pull/5468

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
@@ -76,7 +76,7 @@ trait AstCreatorHelper { this: AstCreator =>
 
   protected def scopeLocalUniqueName(name: String, fullName: String, targetName: String): (String, String) = {
     if (name.isEmpty && (fullName.isEmpty || fullName.endsWith("."))) {
-      val newName           = scopeLocalUniqueName(targetName)
+      val newName           = scopeLocalUniqueName(targetName, fullName)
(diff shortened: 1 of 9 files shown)


max-leuthaeuser:
sptests adaptation is here: https://github.com/ShiftLeftSecurity/codescience/pull/8259

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -63,12 +63,14 @@ trait AstCreatorHelper { this: AstCreator =>
     ast
   }
 
-  protected def fileLocalUniqueName(name: String, fullName: String, targetName: String = ""): (String, String) = {
+  protected def scopeLocalUniqueName(name: String, fullName: String, targetName: String = ""): (String, String) = {
```
maltek, line 66: I'm wondering if this method could be made to automatically retrieve the fullname from the scope. I'm a little worried that the scope object and the provided full name might not always align, and we end up with clashing full names.

I think just for clarity it also wouldn't hurt to have multiple overloads of this method for the case where a fullname is needed or not (empty string).

maltek (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstCreatorHelper.scala
max-leuthaeuser: I changed it so we are now re-using the provided fullname (if any) to build the map key.
Only if the fullname is empty we fallback to the scope path.

PR #5471, opened 2025-05-07
[c2cpg] Added operator method ref to fold expression lowering 

max-leuthaeuser:


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/passes/TypeDeclNodePass.scala
```diff
@@ -32,11 +31,20 @@ class TypeDeclNodePass(cpg: Cpg, config: Config) extends CpgPass(cpg) {
...
         .astParentFullName(fullName)
       dstGraph.addNode(newTypeDecl)
+      createOperatorBinding(newTypeDecl, dstGraph)
       hadMissingTypeDecl = true
     }
     if (hadMissingTypeDecl) Ast.storeInDiffGraph(createGlobalAst(), dstGraph)
   }
 
+  private def createOperatorBinding(typeDecl: NewTypeDecl, dstGraph: DiffGraphBuilder): Unit = {
+    if (typeDecl.fullName.startsWith("<operator>.") && typeDecl.fullName != Defines.OperatorUnknown) {
```
maltek, line 41:
```suggestion
    if (typeDecl.fullName.startsWith("<operator") && typeDecl.fullName != Defines.OperatorUnknown) {
```

We unfortunately have some operators that are defined as `<operators>.something` (note the additional 's')

maltek (approved):

PR #5480, opened 2025-05-12
[php] Pre-Parsing Directly on `PhpNode` objects

DavidBakerEffendi:
Implemented a leaner version of https://github.com/joernio/joern/pull/5467 by operating directly on `PhpNode`s and parsing code twice to keep memory low. Handles namespace collisions gracefully.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/passes/PhpSymbolSummaryPassTest.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/passes/PhpSymbolSummaryPassTest.scala
@@ -115,6 +115,85 @@ class PhpSymbolSummaryPassTest extends AnyWordSpec with Matchers {
     )
   }
 
+  "pre-parsing a file with a nested namespace declaration should be summarized and deduplicated accordingly" in {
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/passes/PhpSymbolSummaryPassTest.scala
```diff
@@ -0,0 +1,144 @@
...
+            ) :: Nil =>
+          globalName shouldBe NamespaceTraversal.globalNamespaceName
+          fooName shouldBe "Foo"
+          barName shouldBe "Bar"
+          bazName shouldBe "Baz"
+          fazName shouldBe "Faz"
+      }
+    )
+  }
+
```
ml86, line 117: Missing a test about functions nested in namespace.
If that is possible we should also have a test of namespace in namespace.

DavidBakerEffendi: Added a couple of tests, and improved deduplication when separated namespaces are adjacent to one another

PR #5481, opened 2025-05-12
factor out 'release-github' as a separate build.

mpollmeier:
Required because we no longer create a release if there's no changes - see
the currently release build.

Triggered by version tags, which are set by the regular release build.

Also: use the tag from gh actions env


maltek (commented):
don't we need to run `sbt createDistribution` at some point in the new file?

maltek (approved):

max-leuthaeuser (approved):

PR #5482, opened 2025-05-12
Switch to lazy/less storage `.loc` and `Loc` over `.location` and `{New}Location`

Flandini:
- Adds `Loc` class as replacement for `LocationCreator` and the `location` node step. `Loc` is a replacement that is meant to compute and store less up-front compared to `LocationCreator` and `location`. IIUC, scala lazy values also take up space, so I eyeballed the operations that might take more steps to compute and made those lazy and those that are just lookups on the underlying node and made these methods.
- Adds a duplicate of one of the `c2cpg` test files to use `Loc` rather than replacing the other tests for now
- Add deprecation warnings to the previous location code

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Loc.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Loc.scala
@@ -1,0 +1,66 @@
+package io.shiftleft.semanticcpg.language
+
+import io.shiftleft.codepropertygraph.generated.nodes.*
+import io.shiftleft.semanticcpg.language.*
+
+import scala.annotation.tailrec
+
+class Loc(val node: StoredNode) {
+  def symbol: String = {
+    node match {
+      case call: Call                   => call.code
+      case method: Method               => method.name
+      case inParam: MethodParameterIn   => inParam.name
+      case ident: Identifier            => ident.name
+      case lit: Literal                 => lit.code
+      case local: Local                 => local.name
+      case outParam: MethodParameterOut => outParam.name
+      case methodRef: MethodRef         => methodRef.code
+      case methodReturn: MethodReturn   => "$ret"
+      case _                            => defaultString
+    }
+  }
+
+  def nodeLabel: String = node.label
+
+  def lineNumber: Option[Int] = node match {
+    case astNode: AstNode => astNode.lineNumber
+    case cfgNode: CfgNode => cfgNode.lineNumber
+  }
+
+  def methodFullName: String = method.fullName
+
+  def methodShortName: String = method.name
+
+  def packageName: String = namespaceOption.getOrElse(defaultString)
+
+  def className: String = typeOption.map(_.fullName).getOrElse(defaultString)
+
+  def classShortName: String = typeOption.map(_.name).getOrElse(defaultString)
+
+  def filename: String = if (method.filename.isEmpty) "N/A" else method.filename
+
+  private val defaultString = "<empty>";
+
+  private lazy val typeOption = findAncestor[TypeDecl](method)
+
+  private lazy val namespaceOption = for {
+    tpe            <- typeOption
+    namespaceBlock <- tpe.namespaceBlock
+    namespace      <- namespaceBlock._namespaceViaRefOut.nextOption()
+  } yield namespace.name
+
+  private lazy val method: Method = node match {
+    case cfgNode: CfgNode => cfgNode.method
+    case local: Local     => local.method.head
+  }
+
+  @tailrec
+  private def findAncestor[TargetNodeType <: StoredNode](node: StoredNode): Option[TargetNodeType] = {
+    node._astIn.iterator.nextOption() match {
+      case Some(head) if head.isInstanceOf[TargetNodeType] => Some(head.asInstanceOf[TargetNodeType])
+      case Some(head)                                      => findAncestor[TargetNodeType](head)
+      case None                                            => None
+    }
+  }
+}
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CallMethods.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CallMethods.scala
@@ -5,7 +5,7 @@ import io.shiftleft.codepropertygraph.generated.DispatchTypes
 import io.shiftleft.semanticcpg.NodeExtension
 import io.shiftleft.semanticcpg.language.*
 
-class CallMethods(val node: Call) extends AnyVal with NodeExtension with HasLocation {
+class CallMethods(val node: Call) extends AnyVal with NodeExtension with HasLocation with HasLoc {
 
   def isStatic: Boolean =
     node.dispatchType == DispatchTypes.STATIC_DISPATCH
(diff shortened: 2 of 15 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Loc.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  private lazy val method: Method = node match {
+    case cfgNode: CfgNode => cfgNode.method
+    case local: Local     => local.method.head
+  }
+
+  @tailrec
+  private def findAncestor[TargetNodeType <: StoredNode](node: StoredNode): Option[TargetNodeType] = {
+    node._astIn.iterator.nextOption() match {
+      case Some(head) if head.isInstanceOf[TargetNodeType] => Some(head.asInstanceOf[TargetNodeType])
```
ml86, line 61: I would be surprised if this works because `TargetNodeType` will be erased to `StoredNode` at runtime. This should give a compile warning.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/NodeMethods.scala
```diff
@@ -6,10 +6,17 @@ import io.shiftleft.semanticcpg.language.*
...
   def location(implicit finder: NodeExtensionFinder): NewLocation =
     node match {
       case storedNode: StoredNode => LocationCreator(storedNode)
       case _                      => LocationCreator.emptyLocation("", None)
     }
 
+  def loc: Loc = {
+    node match {
+      case storedNode: StoredNode => Loc(storedNode)
+    }
```
ml86, line 19: In order to make this a replacement for `def location` please also add the empty location equivalent. Otherwise we will get match exceptions.

fabsx00:
Why do we need to deprecate the `.location` step and replace it with `.loc`? The query language is long-standing public API and `.location` can be spoken while `.loc` is more of an abbreviation.

fabsx00 (changes requested):
I don't think we should deprecate `.location` in favour of `.loc`. Is it not possible to simply adapt the old implementation but keep the name? Also, can you motivate why this change is necessary in the first place?

Flandini:
> I don't think we should deprecate .location in favour of .loc. Is it not possible to simply adapt the old implementation but keep the name? Also, can you motivate why this change is necessary in the first place?

@fabsx00, sorry, I did not see your earlier mesage. 

This is an attempt at a more efficient implementation of `LOCATION` and the `.location` node step. These implementations compute and store all of the fields of the `Location`/`NewLocation`, but users may not use all of the computed information.

I had an internal issue assigned to me, IIUC deprecating instead of replacing the `location` implementation is because there is maybe some dislike for defining location in the CPG schema. I will send you a link here and maybe we can all discuss.

Flandini:
After some discussion: there are understandably some problems with the name `loc` and `Loc`: it could means 'lines of code' or something else and isn't directly related to `location` unless you know already. Maybe we will change the name to `.newLocation` or `.locationLazy` or something besides `.loc`.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Loc.scala
Flandini: Since this is only used once after recent changes, I removed the type parameter so it is now `findParentTypeDecl(node: StoredNode) : Option[TypeDecl]`

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/NodeMethods.scala
Flandini: Added, see `Location.apply` and `LocationCreator` trait

Flandini:
It was decided that we will replace `Location` and `NewLocation` with the new lazily computed location info (previously `Loc` and `.loc` in this PR thread). This also means deleting these from the CPG spec: https://github.com/ShiftLeftSecurity/codepropertygraph/pull/1819.

Flandini:
Pushed some changes to make this nicer to use as a consumer of joern that extends the codepropertygraph / location types. It is nicer in that both the joern location types and any extended location types can both be wildcard imported, but the extended location types will be used by the [implicits because of the way extending the location information uses subclassing](https://scala-lang.org/files/archive/spec/3.4/06-expressions.html#overloading-resolution).

Say that you have a package, `io.shiftleft.extendedcpg` that extends the codepropertygraph and location nodes. For these extended nodes, you subclass `LazyLocation` with something called like `ExtendedLazyLocation` and provide the implicit `apply`. Then, in the consuming packages, you now do this:

```
package io.shiftleft.extendedcpg;

def doSomething(node: AbstractNode): Unit = {
  ...
  val location = node.location
  ...
}
```
and everything works as expected even if node is an extended node type.

Similarly, if you import both the location types in joern and the location types in the extending and consuming library, then the implicit location creator in the extending and consuming library will be chosen since it will be more specific with this design. So this works as expected:

```
package io.shiftleft.someotherpackage;
import io.shiftleft.extendedcpg.*
import import io.shiftleft.semanticcpg.language.*

def doSomething(node: AbstractNode): Unit = {
  ...
  val location = node.location
  ...
}
```
The location will be `ExtendedLazyLocation` and the location creator in the extended library will be used, not the one in the joern repo.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Location.scala
```diff
@@ -0,0 +1,124 @@
...
+// Previously, the non-stored Location node generated
+// from the CPG schema was named NewLocation. The
+// addition of lazily computed location info in this
+// file brought the deletion of these CPG nodes. This
+// is not directly used in joern, but intended to not
+// break usage in users of the API. Users directly
+// importing NewLocation from the CPG package rather
+// will still have a break.
+@deprecated("NewLocation is deprecated, prefer LocationInfo")
+type NewLocation = LocationInfo
```
ml86, line 31: So the scenario here is that people only have to adjust den import but not the other references to `NewLocation` in there code?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CallMethods.scala
```diff
@@ -45,6 +45,7 @@ class CallMethods(val node: Call) extends AnyVal with NodeExtension with HasLoca
     node.astChildren.isBlock.maxByOption(_.order).iterator.expressionDown
   }
 
-  override def location: NewLocation =
-    LocationCreator.defaultCreateLocation(node)
+  override def location: LocationInfo = {
+    node.location
```
ml86, line 49: This should be a recursive call to `location`, at least my IDE agrees with me.
Same applies to the other `XXXMethods.scala` files except for `NodeMethods.scala`.
I think we do not even need all those overrides anymore. It is enough to have `location` defined in `NodeMethods`.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Location.scala
```diff
@@ -0,0 +1,124 @@
...
+
+trait HasLocation extends Any {
+  def location: LocationInfo
+}
```
ml86, line 35: Do we still need this? I have not seem a usage in Joern or CS repo outside of declarations.

PR #5483, opened 2025-05-13
release-github.yml didn't trigger - try another pattern

mpollmeier:


max-leuthaeuser (approved):
We do not need to specify the branch right?

mpollmeier:
If we add `master` to the `on -> push` list it'll get triggered for both, changes on master and version tags, so that's not an option. 
We could add a filter later on to ensure it doesn't run on branches other than master, but since debugging this triggering isn't easy let's start with this as is.

PR #5485, opened 2025-05-13
[jssrc2cpg] Added support for Parameter Properties

max-leuthaeuser:
https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties


maltek (commented):
we also need to add assignments like `this.x = x` inside the constructor

maltek (approved):

PR #5486, opened 2025-05-13
[c2cpg] Proper C++ constructor calls

max-leuthaeuser:


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
```diff
@@ -163,31 +163,75 @@ trait TypeNameProvider { this: AstCreator =>
     }
   }
 
+  private val FundamentalTypeKeywords = List(
+    "void",
+    "bool",
+    "true",
+    "false",
```
maltek, line 170: this is... weird? If it is correct / intended, please add some comments explaining what's going on with these being designated as type keywords.

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/features17/Cpp17FeaturesTests.scala
```diff
@@ -30,6 +54,38 @@ class Cpp17FeaturesTests extends AstC2CpgSuite(fileSuffix = FileDefaults.CppExt)
...
+        "this->val = val",
+        "c1 = new MyContainer.MyContainer(1)",
+        "c2 = new MyContainer.MyContainer()"
+      )
+      cpg.call.isAssignment.argument(1).isIdentifier.typeFullName.l shouldBe List("MyContainer", "MyContainer")
+
+      val List(c1NewCall, c2NewCall) = cpg.call.isAssignment.argument(2).isCall.l
+
+      c1NewCall.code shouldBe "new MyContainer.MyContainer(1)"
+      c1NewCall.methodFullName shouldBe Defines.OperatorNew
```
maltek, line 75: Where is this coming from? There is no `new` in the code. The `new` operator results in a heap allocation. It's very important that we don't mix those up with stack/inline allocations.

(Also in other tests, which I'm not all flagging individually.)

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/features17/Cpp17FeaturesTests.scala
```diff
@@ -413,7 +469,7 @@ class Cpp17FeaturesTests extends AstC2CpgSuite(fileSuffix = FileDefaults.CppExt)
           |byte d = byte{1};
           |""".stripMargin)
       cpg.local.nameExact("b").typeFullName.l shouldBe List("byte")
-      cpg.identifier.nameExact("b").typeFullName.l shouldBe List("byte")
+      cpg.identifier.nameExact("b").typeFullName.l shouldBe List("byte", "byte")
```
maltek, line 472: This is unexpected to me. What are the two identifiers doing here? Could you assert something in the test here that explains / shows what each of them is doing?

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/AstCreationPassTests.scala
```diff
@@ -1406,29 +1406,113 @@ class AstCreationPassTests extends AstC2CpgSuite {
...
-      hi.code shouldBe "\"hi\""
-      buf.argumentIndex shouldBe 3
-      buf.code shouldBe "buf"
+      val List(newCall)         = cpg.call.methodFullNameExact(Defines.OperatorNew).l
+      val List(constructorCall) = newCall.argument.isCall.l
+      constructorCall.code shouldBe """new (buf) Foo("hi")"""
+      constructorCall.methodFullName shouldBe "Foo.Foo:void(char[3],char[80])"
+      constructorCall.signature shouldBe "void(char[3],char[80])"
+      constructorCall.typeFullName shouldBe Defines.Void
+      constructorCall.argument.code.l shouldBe List(""""hi"""", "buf")
```
maltek, line 1449: The arguments seem wrong here: `buf` is a pointer to the memory where `this` will be stored. So it should not be part of the signature. And it should be promoted to argument 0, not used as the last argument.

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/AstCreationPassTests.scala
```diff
@@ -1406,29 +1406,113 @@ class AstCreationPassTests extends AstC2CpgSuite {
...
-      cpg.call.nameExact("<operator>.new").code("new int\\[n\\]").argument.code("int").size shouldBe 1
+      val List(newCall)         = cpg.call.methodFullNameExact(Defines.OperatorNew).l
+      val List(constructorCall) = newCall.argument.isCall.l
+      constructorCall.code shouldBe "new int[n]"
+      constructorCall.methodFullName shouldBe Operators.alloc
+      constructorCall.typeFullName shouldBe Defines.Any
+      constructorCall.argument.code.l shouldBe List("int", "n")
     }
 
     "be correct for 'new' with explicit identifier" in {
```
maltek, line 1433:
```suggestion
    "be correct for 'placement new'" in {
```

maltek (commented):
Some notes/questions from reading over the diff. I haven't had the time to play around with the CPGs yet - some more things might come up once I do.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -138,6 +258,14 @@ trait AstForTypesCreator { this: AstCreator =>
 
   protected def astForASMDeclaration(asm: IASTASMDeclaration): Ast = Ast(unknownNode(asm, code(asm)))
 
+  protected def isCPPClass(decl: IASTSimpleDeclaration): Boolean = {
+    decl.getDeclSpecifier match {
+      case t: ICPPASTNamedTypeSpecifier =>
+        safeGetBinding(t.getName).exists { binding => binding.isInstanceOf[ICPPClassType] }
```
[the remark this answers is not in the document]
max-leuthaeuser: > struct and class in C++

Yes, that's exactly what that type signals.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
max-leuthaeuser: That comes from the official C++ docs.

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/features17/Cpp17FeaturesTests.scala
max-leuthaeuser: One from the variable decl, one as base argument to the constructor call.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -288,15 +364,17 @@ trait AstForFunctionsCreator { this: AstCreator =>
           case _ => None
         }
         maybeOwner.toSeq.map { owner =>
+          val tpe = if (isConstructor) { cleanType(owner) }
+          else { s"${cleanType(owner)}*" }
```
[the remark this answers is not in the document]
max-leuthaeuser: I tried so many things to get this passed as `this` pointer. But then the function gets a `parameterout` node that is also a pointer and DF in CS stops working when the calling code tries to access members with `fieldAccess` calls. That's also the reason why a synthetic return is implemented (see your comment below). I was not able to get the DF running when doing the .alloc; constructor call block idoim as in javasrc.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -159,22 +138,44 @@ trait AstForFunctionsCreator { this: AstCreator =>
...
+    val methodBodyAst = astForMethodBody(Option(funcDef.getBody), methodBlockNode)
+
+    if (isConstructor) {
+      implicitThisParam.foreach { param =>
+        val thisIdentifier = identifierNode(funcDef, Defines.This, Defines.This, param.typeFullName)
+        scope.addVariableReference(Defines.This, thisIdentifier, param.typeFullName, EvaluationStrategies.BY_SHARING)
+        val cpgReturn = returnNode(funcDef, "return this")
+        val returnAst = Ast(cpgReturn).withChild(Ast(thisIdentifier)).withArgEdge(cpgReturn, thisIdentifier)
+        Ast.storeInDiffGraph(returnAst, diffGraph)
+        diffGraph.addEdge(methodBlockNode, cpgReturn, EdgeTypes.AST)
```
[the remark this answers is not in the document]
max-leuthaeuser: See comment above.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForFunctionsCreator.scala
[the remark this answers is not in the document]
max-leuthaeuser: But I will give it another try tomorrow.

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/AstCreationPassTests.scala
max-leuthaeuser: This is not quite right. See: https://en.cppreference.com/w/cpp/language/new (Placement new)

>If placement-args are provided, they are passed to the allocation function as additional arguments. Such allocation functions are known as "placement new", after the standard allocation >function void* [operator new](http://en.cppreference.com/w/cpp/memory/new/operator_new([std::size_t](http://en.cppreference.com/w/cpp/types/size_t), void*), which simply returns its second argument unchanged. This is used to construct objects in allocated storage:

```C++
// within any block scope...
{
    // Statically allocate the storage with automatic storage duration
    // which is large enough for any object of type “T”.
    alignas(T) unsigned char buf[sizeof(T)];
 
    T* tptr = new(buf) T; // Construct a “T” object, placing it directly into your 
                          // pre-allocated storage at memory address “buf”.
 
    tptr->~T();           // You must **manually** call the object's destructor
                          // if its side effects is depended by the program.
}                         // Leaving this block scope automatically deallocates “buf”.
```

maltek: they are passed to the allocator, not to the constructor

max-leuthaeuser: Hm right. So what to do now with them?
Pass them as args under <operator>.new?
Something like:
```
new (buf, ...) Foo(1, 2, 3)
--> lowered to:
<operator>.new -arg1-> constructor call block <tmp>0 = .alloc; Foo.Foo(1, 2, 3); <tmp>0
|
arg2 ... argN -> buf, ...
```

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForTypesCreator.scala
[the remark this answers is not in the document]
maltek: could you also rename the helper method? e.g. `isCPPComposite` or even just `isCPPClassLike`. To make sure readers know that it's not looking for the `class` keyword exactly

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
maltek: what link? I don't see one for `FundamentalTypeKeywords`. The one below for `isFundamentalTypeKeywords` does mention `true` and `false` - but only as the possible values for `bool` not as types themselves.

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/features17/Cpp17FeaturesTests.scala
```diff
@@ -23,13 +47,47 @@ class Cpp17FeaturesTests extends AstC2CpgSuite(fileSuffix = FileDefaults.CppExt)
...
+      tmp1.name shouldBe "<tmp>1"
+      tmp1.typeFullName shouldBe "MyContainer"
+      tmp0.name shouldBe "<tmp>0"
+      tmp0.typeFullName shouldBe "MyContainer"
       // We are unable to express this template argument deduction in the current schema
       cpg.typeDecl.member.nameExact("val").typeFullName.l shouldBe List("T")
+
+      val List(c1Method, c2Method) = cpg.typeDecl.nameExact("MyContainer").astChildren.isMethod.isConstructor.l
+      c1Method.fullName shouldBe "MyContainer.MyContainer:void()"
+      c1Method.signature shouldBe "void()"
```
maltek, line 64: afaiu the `c1Method` does call the default-constructor of `val`. But that's not showing up in the CPG. Can we make that happen?

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
max-leuthaeuser: Ah you meant true/false explicitely. Right, thats a copy/paste error.

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/features17/Cpp17FeaturesTests.scala
max-leuthaeuser: No, because `val` is of type `T` and that is `<typename T = float>` and `float` it not a class like but its in `FundamentalTypeKeywords`

maltek: But `{}` initialising such a type does initialize it with 0 (
[godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:16,endLineNumber:7,positionColumn:16,positionLineNumber:7,selectionStartColumn:16,selectionStartLineNumber:7,startColumn:16,startLineNumber:7),source:'%0Atemplate+%3Ctypename+T+%3D+float%3E%0Astruct+MyContainer+%7B%0A++T+val%3B%0A++MyContainer()+:+val%7B%7D+%7B%7D%0A%7D%3B%0AMyContainer+c2%3B'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:41.46341463414633,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g151,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+15.1+(Editor+%231)',t:'0')),k:50.40650406504065,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+15.1',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+15.1+(Compiler+%231)',t:'0')),k:8.130081300813007,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)). Apparently that's 
https://en.cppreference.com/w/cpp/language/aggregate_initialization Maybe that's something for another PR?

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/features17/Cpp17FeaturesTests.scala
```diff
@@ -281,24 +345,26 @@ class Cpp17FeaturesTests extends AstC2CpgSuite(fileSuffix = FileDefaults.CppExt)
           |}
           |""".stripMargin)
```
maltek, line 346: I'm missing the constructor call for `std::unordered_map<std::string, int> mapping` in the calls below. Is that because the definition for that type is missing in this test?

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/features17/Cpp17FeaturesTests.scala
max-leuthaeuser: Right, but currently CDT actually resolves it to the class field `val` of type `T` (template type parameter). I can't query the actual type there for whatever reason so I can't decide wheter to initialize it with 0 or '' or a construtor call for example.

maltek: would it be possible to emit an assignment `val = {}` where `{}` is just an `Unknown` node? Just so that there's something there in the CPG that shows that the member actually does get initialized.

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/features17/Cpp17FeaturesTests.scala
max-leuthaeuser: That one was missing because the type can't be resolved. I worked around that and now its there.
(as always: "problem bindings" need to be handeled manually)

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/cpp/features17/Cpp17FeaturesTests.scala
max-leuthaeuser: Good Point, did that.

maltek (approved):

PR #5493, opened 2025-05-15
[php] Updated Call Representation for Correctness

DavidBakerEffendi:
Replacement for https://github.com/joernio/joern/pull/5434

Changes:

* Fixed shadowing in the test suite of builtin_functions.txt in the resources
* Changed how calls are represented - these include a receiver in the case of dynamic dispatches
* Due to the nature of PHP, unless a dynamic call is dispatched within the scope of a class, there will not be a $this receiver.
* Built-in functions are statically dispatched

TODO:

* Method bindings need to be created where possible.


DavidBakerEffendi:
This is painful to get working for type recovery, which is a pass that should be phased out and replaced by more purpose-built type recovery anyway.

I'm feeling like we should likely comment out the failing type recovery tests and move on

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -53,12 +72,19 @@ class CallTests extends PhpCode2CpgFixture {
...
-      fooCall.receiver.isEmpty shouldBe true
-      fooCall.dispatchType shouldBe DispatchTypes.STATIC_DISPATCH
+      fooCall.receiver.isEmpty shouldBe false
+      fooCall.receiver.foreach { recvNode =>
+        val recv = recvNode.asInstanceOf[Identifier]
+        recv.code shouldBe "foo"
+        recv.name shouldBe "foo"
+      }
+
+      fooCall.dispatchType shouldBe DispatchTypes.DYNAMIC_DISPATCH
```
ml86, line 82: The call to `foo` should be static. The only thing to do is check the summary in order to find out whether this references a method from the global or local namespace.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -55,77 +55,96 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+     * case of a normal top-level call like foo() that is not possible. There is no corresponding foo identifier which
+     * one could assign another function to. The only uncertainty in such a call is whether this is a foo in the global
+     * namespace or from the namespace surrounding the call-site. But this we cannot decide without knowing which php
+     * files are loaded. In terms of lambdas, they will be assigned to a variable which always begins with `$`.
+     *
+     * This is why we use pre-parse summary idea where we check which functions exist in which namespaces and make a
+     * lookup in it to figure out whether a foo call-site needs to have a global method full name or one from the
+     * current namespace. The `scope` object then delegates what has been defined or imported to determine calls in
+     * scope.
+     */
```
ml86: This is not detailed enough. In which case do we emit a static call to a global function and in which case do we assume the call goes to a namespaced function? 
I do not fully remember the outcome of our discussion about this, so i am wondering how calls to library functions are represented? Please also include this explicitly.

ml86: Some more test cases detailing this would also help.

ml86 (changes requested):

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
DavidBakerEffendi: The way I've initially addressed this is to describe how full names are interpreted from external calls. Also added a test to show static and dynamic calls to show in which case we see a method full name or unresolved method full name. 

Added more to the description in this direction, also mentioning builtin calls (what I assume you mean by global functions) and imports.

ml86 (approved):

PR #5494, opened 2025-05-16
 DSL: make param <-> argument steps respect named arguments 

maltek:
At first I had changed the numbered `call.arguments(i)` query to not return named arguments. That's why I changed some frontend tests to query by argument name. In the end, I felt that went a bit too far as a potentially breaking change. So that DSL step continues to just look at the value of the `argumentIndex` property without considering whether there is a `argumentName` or not.

(I've kept the changes to the tests anyway, because 1) it's more correct in the context of the test to query for a named argument and 2) it provides some test coverage for the DSL step.)

One Ruby dataflow is now ignored that wasn't previously - it sort of accidentally worked until now, since the first (named) argument happened to get matched to the name-variadic parameter that also happens to have index 1. Change the order of the named arguments, and the test would have always failed. I'd say ignoring the test is a more honest assessment of what works and what doesn't wrt to flows through name-variadic parameters. Otoh, the oss dataflow engine now does find flows through named arguments when a parameter with the same name exists.

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/TaskCreator.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/queryengine/TaskCreator.scala
@@ -78,12 +78,9 @@ class TaskCreator(context: EngineContext) {
     }
   }
 
-  private def paramToArgsOfCallers(param: MethodParameterIn): List[Expression] =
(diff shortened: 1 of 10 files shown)


ml86 (approved):
I agree to not change the `argument(i: Int)` function to not match named arguments and leave it to the using code to prefer named arguments if needed.

PR #5499, opened 2025-05-19
[php] Incorporate Imports into Scope

DavidBakerEffendi:
* Modifies the scope object to incorporate summaries and knowledge of imports into the scope 
* Using import information, it is able to map a symbol or alias to an imported entity that has appeared from the summary pass
* Simplified the summary pass to not be nested, but rather flattened, and use string paths similar to PHP notation
* When a symbol is resolved, the kind of symbol is also returned, i.e., class, function, namespace, etc.

Note: Excluded class member summarization (anything nested under a class) as PHP class members cannot be imported directly.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/utils/Scope.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/utils/Scope.scala
@@ -1,15 +1,24 @@
 package io.joern.php2cpg.utils
 
 import io.joern.php2cpg.astcreation.AstCreator.NameConstants
+import io.joern.php2cpg.passes.SymbolSummaryPass.*
 import io.joern.x2cpg.Ast
-import io.joern.x2cpg.datastructures.{ScopeElement, NamespaceLikeScope, Scope as X2CpgScope}
+import io.joern.x2cpg.datastructures.{NamespaceLikeScope, ScopeElement, Scope as X2CpgScope}
 import io.shiftleft.codepropertygraph.generated.NodeTypes
-import io.shiftleft.codepropertygraph.generated.nodes.{NewBlock, NewMethod, NewNamespaceBlock, NewNode, NewTypeDecl}
+import io.shiftleft.codepropertygraph.generated.nodes.{
+  NewBlock,
+  NewImport,
+  NewMethod,
+  NewNamespaceBlock,
+  NewNode,
+  NewTypeDecl
+}
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/utils/Scope.scala
```diff
@@ -18,6 +20,7 @@ class Scope(implicit nextClosureName: () => String) extends X2CpgScope[String, N
   private val anonymousMethods                                    = mutable.ArrayBuffer[Ast]()
   private var tmpVarCounter                                       = 0
   private var tmpClassCounter                                     = 0
+  private var importedSymbols                                     = Map.empty[String, Seq[SymbolSummary]]
```
ml86, line 23: If the imported symbols are the same in a hole file and e.g. are not different from one method body to another, why does this need to live in the Scope object then?

ml86 (changes requested):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/passes/PhpSymbolSummaryPassTest.scala
```diff
@@ -202,7 +153,7 @@ object PhpSymbolSummaryPassTest {
 
   case class ConfigAndParser(config: Config, parser: PhpParser)
 
-  def assertAgainstTempFile(code: Seq[String], assertion: Seq[SymbolSummary] => Unit): Unit = {
+  def assertAgainstTempFile(code: Seq[String], assertion: Seq[(String, Seq[SymbolSummary])] => Assertion): Unit = {
```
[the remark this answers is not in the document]
DavidBakerEffendi: For this test suite, yes, this is redundant, so I'll update it. 

For context: The design of this is that it can handle namespace conflicts, where if the `Seq` is sorted, the head item has the same precedence as how the Zend engine would resolve namespace conflicts, i.e., prefer classes, functions, namespaces in that order.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/utils/Scope.scala
DavidBakerEffendi: My plan for this is that the scope object can hide the logic that decides whether an identifier is a local variable or an import. In the former case, the scope object will be consulted so I figured I'd plan to combine the logic here instead.

DavidBakerEffendi:
@ml86 addressed reviews

PR #5500, opened 2025-05-19
[php] Create meta `TYPE_DECL` (.<class>) for class-like structures

AndreiDreyer:
* Create a meta `TYPE_DECL` node, which is a sibling to the normal `TYPE_DECL` node for named classes, anonymous classes and enums (handled like class structures in `php2cpg`)
* Move `<clinit>` (static init method), all static methods, all static variables and consts to the meta `TYPE_DECL`
* Moved the AST generation for the global type decl into it's own function, was creating too much convolution in the `astForNamedClass` function
* updated required test expectations to account for the new `.<class>` type decl that is introduced.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
@@ -54,7 +54,7 @@ trait AstCreatorHelper(disableFileContent: Boolean)(implicit withSchemaValidatio
       globalNamespace.fullName
     } else {
       val className = if (appendClass) {
-        getTypeDeclPrefix.map(name => s"${name}.<class>")
+        getTypeDeclPrefix.map(name => s"${name}$MetaTypeDeclExtension")
       } else {
         getTypeDeclPrefix
       }
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
@@ -49,11 +49,16 @@ trait AstCreatorHelper(disableFileContent: Boolean)(implicit withSchemaValidatio
     identifierNode(originNode, name, s"$$$name", typeFullName)
   }
 
-  protected def composeMethodFullName(methodName: String, isStatic: Boolean): String = {
+  protected def composeMethodFullName(methodName: String, isStatic: Boolean, appendClass: Boolean = false): String = {
     if (methodName == NamespaceTraversal.globalNamespaceName) {
       globalNamespace.fullName
     } else {
-      val className       = getTypeDeclPrefix
+      val className = if (appendClass) {
+        getTypeDeclPrefix.map(name => s"${name}.<class>")
+      } else {
+        getTypeDeclPrefix
+      }
+
       val methodDelimiter = if (isStatic) StaticMethodDelimiter else InstanceMethodDelimiter
 
       val nameWithClass = List(className, Some(methodName)).flatten.mkString(methodDelimiter)
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForTypesCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForTypesCreator.scala
@@ -16,16 +16,48 @@ import io.shiftleft.semanticcpg.language.types.structure.NamespaceTraversal
 
 trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this: AstCreator =>
 
-  protected def astForClassLikeStmt(stmt: PhpClassLikeStmt): Ast = {
+  protected def astForClassLikeStmt(stmt: PhpClassLikeStmt): List[Ast] = {
+    val (staticStmts, dynamicStmts) = stmt.stmts.partition {
+      case x: PhpPropertyStmt if x.modifiers.contains(ModifierTypes.STATIC) => true
+      case x: PhpMethodDecl if x.modifiers.contains(ModifierTypes.STATIC)   => true
+      case x: PhpConstStmt                                                  => true
+      case _                                                                => false
+    }
+
     stmt.name match {
-      case None                                           => astForAnonymousClass(stmt)
-      case Some(name) if name.name.contains("anon-class") => astForAnonymousClass(stmt)
-      case Some(name)                                     => astForNamedClass(stmt, name)
+      case None => astForAnonymousClass(stmt, dynamicStmts, staticStmts) :: Nil
+      case Some(name) if name.name.contains("anon-class") =>
+        astForAnonymousClass(stmt, dynamicStmts, staticStmts) :: Nil
+      case Some(name) => astForNamedClass(stmt, name, dynamicStmts, staticStmts)
     }
   }
 
-  private def astForAnonymousClass(stmt: PhpClassLikeStmt): Ast = {
+  protected def astForGlobalDecl(stmt: PhpClassLikeStmt, name: PhpNameExpr): Ast = {
     val inheritsFrom = (stmt.extendsNames ++ stmt.implementedInterfaces).map(_.name)
+    val code         = codeForClassStmt(stmt, name)
+
+    val fullName = globalNamespace.fullName
+
+    val typeDecl = typeDeclNode(stmt, name.name, fullName, relativeFileName, code, inherits = inheritsFrom)
+    val createDefaultConstructor = stmt.hasConstructor
+
+    scope.pushNewScope(typeDecl)
+    val bodyStmts      = astsForClassLikeBody(stmt, stmt.stmts, createDefaultConstructor)
+    val clinitAst      = astForStaticAndConstInits(stmt).getOrElse(Ast())
+    val modifiers      = stmt.modifiers.map(modifierNode(stmt, _)).map(Ast(_))
+    val annotationAsts = stmt.attributeGroups.flatMap(astForAttributeGroup)
+    scope.popScope()
+
+    Ast(typeDecl).withChildren(modifiers).withChildren(bodyStmts :+ clinitAst).withChildren(annotationAsts)
+  }
+
+  private def astForAnonymousClass(
+    stmt: PhpClassLikeStmt,
+    dynamicStmts: List[PhpStmt],
+    staticStmts: List[PhpStmt]
+  ): Ast = {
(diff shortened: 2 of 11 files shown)


d1tto:
hello, I have a question for this pr. what is the benefit of moving static members into meta typedecl. It seems like it would increase workload for downstream static analysis

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -72,7 +72,7 @@ class CallTests extends PhpCode2CpgFixture {
     "have the correct method node defined" in {
       inside(cpg.call.l) { case List(fooCall) =>
         fooCall.name shouldBe "foo"
-        fooCall.methodFullName shouldBe s"Foo::foo"
+        fooCall.methodFullName shouldBe s"Foo.<class>::foo"
```
ml86, line 75: Please remove the `.` as separator from the naming scheme.

ml86: I find the `<class>` suffix to be unclear because i always wonder: "Is this the type of the Foo class object or the type of the Foo class objects. If there is no php specific name for this or if `<class>` is the php specific name, please name this `<metaclass>` at least that is unabiguous.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
```diff
@@ -507,4 +529,128 @@ class TypeDeclTests extends PhpCode2CpgFixture {
...
+        |  public $baz = 'baz';
+        |  const BAZZ = 'bazz';
+        |  public static function bar() {}
+        |}
+        |""".stripMargin)
+
+    "create a singleton type decl" in {
+      inside(cpg.typeDecl.name("Foo.<class>").l) {
+        case fooTypeDecl :: Nil =>
+          fooTypeDecl.modifier.modifierType.l shouldBe List(ModifierTypes.STATIC)
```
ml86, line 546: I see no reason to put the STATIC modifier on the meta class type decl. That is not indicated by the php code and not needed for our modeling.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
```diff
@@ -507,4 +529,128 @@ class TypeDeclTests extends PhpCode2CpgFixture {
...
+      }
+    }
+
+    "contain the <clinit> static constructor" in {
+      inside(cpg.typeDecl.name("Foo.<class>").method.name(Defines.StaticInitMethodName).l) {
+        case staticConstructor :: Nil =>
+          inside(staticConstructor.body.astChildren.l) {
+            case (self: Local) :: (assignmentBazz: Call) :: (assignmentFoo: Call) :: Nil =>
+              // TODO: Self in <clinit> is existing behavior
+              self.code shouldBe "$self"
```
ml86, line 574: Code of self should be `self` not `$self` as this is not even valid php.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
```diff
@@ -507,4 +529,128 @@ class TypeDeclTests extends PhpCode2CpgFixture {
...
+        |  const BAZZ = 'bazz';
+        |  public static function bar() {}
+        |};
+        |""".stripMargin)
+
+    "Create .<class> type decl" in {
+      inside(cpg.typeDecl.name("Test0.php:<global>@anon-class-0.<class>").l) {
+        case anonTypeDecl :: Nil =>
+          anonTypeDecl.fullName shouldBe "Test0.php:<global>@anon-class-0.<class>"
+          anonTypeDecl.modifier.map(_.modifierType).contains(ModifierTypes.STATIC) shouldBe true
```
ml86, line 606: Again no static modifier, like above.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -49,11 +49,16 @@ trait AstCreatorHelper(disableFileContent: Boolean)(implicit withSchemaValidatio
...
   }
 
-  protected def composeMethodFullName(methodName: String, isStatic: Boolean): String = {
+  protected def composeMethodFullName(methodName: String, isStatic: Boolean, appendClass: Boolean = false): String = {
     if (methodName == NamespaceTraversal.globalNamespaceName) {
       globalNamespace.fullName
     } else {
-      val className       = getTypeDeclPrefix
+      val className = if (appendClass) {
+        getTypeDeclPrefix.map(name => s"${name}.<class>")
```
ml86, line 57: Use a const variable for `<class>` and rename it to `<metaclass>`.

DavidBakerEffendi:
> hello, I have a question for this pr. what is the benefit of moving static members into meta typedecl. It seems like it would increase workload for downstream static analysis

The idea of this is to model the behaviour of how the language handles inheritance and access to simplify downstream analysis, and try to avoid baking in language-specific logic beyond AST creator. Much of the downstream analysis in open-source is not object-sensitive, so this logic also helps separate possible overtainting.

A motivating example for PHP (where this also exists in Python, JS, Ruby, etc.), we can have something like
```php
class Foo {
  const CONSTANT = 'constant value';
}

Foo::CONSTANT; // static access, expected
$foo = new Foo();
$foo->CONSTANT; // dynamic access, less expected
```

In the above, `CONSTANT` is meant to be a singleton and live on this "meta class", but the fact that all instances can access it suggests that each instance has a pointer to the meta class. This is true when looking at how the Zend engine constructs classes, but this internal behaviour is not available to us in PHP, but we model it here.

In languages like Ruby, the difference between `extends` and `includes` also decides from which class (singleton, or object) methods and constants are inherited, hence the separation. In PHP, it's fortunately a bit simpler, but the above example hopefully explains why we have this separation.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -66,7 +116,13 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this:
         typeDeclMember.astParentType(NodeTypes.TYPE_DECL)
       }
 
+      val metaTypeDeclMember = NewMember()
+        .name(s"${className}.<class>")
+        .code(s"${className}.<class>")
+        .dynamicTypeHintFullName(Seq(s"${classFullName}.<class>"))
+
```
ml86, line 123: There seem to be no tests for this. Also make sure to that this is tested in the CS data flow tests.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -46,15 +78,33 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this:
...
 
-    val bodyStmts      = astsForClassLikeBody(stmt, stmt.stmts, stmt.hasConstructor)
+    val bodyStmts      = astsForClassLikeBody(stmt, dynamicStmts, stmt.hasConstructor)
     val modifiers      = stmt.modifiers.map(modifierNode(stmt, _)).map(Ast(_))
     val annotationAsts = stmt.attributeGroups.flatMap(astForAttributeGroup)
 
     scope.popScope()
+
+    scope.surroundingAstLabel.foreach(metaTypeDeclTemp.astParentType(_))
+    scope.surroundingScopeFullName.foreach(metaTypeDeclTemp.astParentFullName(_))
```
ml86, line 102: Is the assumption here that `surroundingAstLabel` and `surroundingScopeFullName` always return non empty?
I guess it is because otherwise `astParentType` and `astParentFullName` are not set. 
Please state this in the code be using `.get` instead of `.foreach`.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -46,15 +78,33 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this:
       alias = None
     )
 
+    val metaTypeDeclTemp = typeDeclNode(
```
ml86, line 81: This does not seem to be a temp variable because it ends up in the resulting AST. Please rename.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -131,22 +194,68 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this:
...
+      List.empty
+    }
+
     scope.popScope()
 
-    Ast(typeDecl).withChildren(modifiers).withChildren(bodyStmts).withChildren(annotationAsts)
+    val classTypeDeclAst = Ast(typeDecl).withChildren(modifiers).withChildren(bodyStmts).withChildren(annotationAsts)
+
+    // Ignore the metaTypeDecl when we are in the <global> method / type
+    if (!name.name.equals(NamespaceTraversal.globalNamespaceName)) {
```
ml86, line 225: Why?

ml86 (changes requested):

d1tto:
> > hello, I have a question for this pr. what is the benefit of moving static members into meta typedecl. It seems like it would increase workload for downstream static analysis
> 
> The idea of this is to model the behaviour of how the language handles inheritance and access to simplify downstream analysis, and try to avoid baking in language-specific logic beyond AST creator. Much of the downstream analysis in open-source is not object-sensitive, so this logic also helps separate possible overtainting.
> 
> A motivating example for PHP (where this also exists in Python, JS, Ruby, etc.), we can have something like
> 
> ```
> class Foo {
>   const CONSTANT = 'constant value';
> }
> 
> Foo::CONSTANT; // static access, expected
> $foo = new Foo();
> $foo->CONSTANT; // dynamic access, less expected
> ```
> 
> In the above, `CONSTANT` is meant to be a singleton and live on this "meta class", but the fact that all instances can access it suggests that each instance has a pointer to the meta class. This is true when looking at how the Zend engine constructs classes, but this internal behaviour is not available to us in PHP, but we model it here.
> 
> In languages like Ruby, the difference between `extends` and `includes` also decides from which class (singleton, or object) methods and constants are inherited, hence the separation. In PHP, it's fortunately a bit simpler, but the above example hopefully explains why we have this separation.

Thanks for your explain. I got it.

ml86 (approved):

PR #5504, opened 2025-05-21
ExternalCommand: cleanup, `verifySuccess` convenience method

mpollmeier:
some changes as I was using it


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommand.scala
```diff
@@ -1,3 +1,130 @@
...
+object ExternalCommand {
+  private val logger = LoggerFactory.getLogger(getClass)
+
+  case class ExternalCommandResult(exitCode: Int, stdOut: Seq[String], stdErr: Seq[String], commandInfo: String) {
+
+    /** convenience method: verify that the result is a success, throws an exception otherwise */
+    def verifySuccess(): Unit =
+      toTry.get
+
+    def successOption: Option[Seq[String]] = exitCode match {
```
maltek, line 25:
```suggestion
    /** Lines of standard output, if successful. */
    def successOption: Option[Seq[String]] = exitCode match {
```

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommand.scala
```diff
@@ -1,3 +1,130 @@
...
+
+      logger.debug(s"command finished successfully")
+      val stdOut = IOUtils.readLinesInFile(stdOutFile.toPath)
+      val stdErr = stdErrFile.map(_.toPath).map(IOUtils.readLinesInFile).getOrElse(Seq.empty)
+      ExternalCommandResult(process.exitValue(), stdOut, stdErr, commandInfo)
+    } catch {
+      case NonFatal(exception) =>
+        val stdOut = IOUtils.readLinesInFile(stdOutFile.toPath)
+        val stdErr = stdErrFile.map(f => IOUtils.readLinesInFile(f.toPath)).getOrElse(Seq.empty) :+ exception.getMessage
+        logger.warn(s"command did not finish successfully. Input was: $commandInfo")
```
max-leuthaeuser, line 98: Just a small remark here:
We frequently use `ExternalCommand.run` to check if a (compatible) executable FOOastgen binary is in the path by just trying to execute it (various frontends do so). Iff not, we call the one from the Joern distribution.
This results in a WARN log entry every time. 
I am not sure whats the best option here but maybe we could use an additional parameter that disables that warning in case you kind of expect the execution to fail (most people do not have the astgen binary in their PATH but some do)?

maltek (approved):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommand.scala
mpollmeier: Good point. I guess when I added that code (prior to this PR) I thought it's ok because users can configure their log levels accordingly. Rethinking, I guess it's better to not log here at all, and leave the handling of the result to the caller. Will change.

PR #5505, opened 2025-05-21
[c2cpg] Fix variadic param handling

max-leuthaeuser:


max-leuthaeuser:
The PR is not needed; the old behavoir was correct.

https://en.cppreference.com/w/cpp/language/variadic_arguments

>The comma preceding the ellipsis can be omitted. (deprecated in C++26) 
```C++
// the function declared as follows:
int printx(const char* fmt, ...);
int printx(const char* fmt...); // same as above, but deprecated since C++26
```

max-leuthaeuser:
Needs to be done differently.

PR #5509, opened 2025-05-22
[php] update logic as to when to add <metaclass> to static call

AndreiDreyer:
Fixed a bug where the `<metaclass>` was being added to a call name twice. This was happening when you had a static call in a static function:
```php
class Foo {
  public static function foo() {
    self::bar();
  }
 
  private static function bar()
}
```

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/utils/Scope.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/utils/Scope.scala
@@ -104,6 +104,9 @@ class Scope(summary: Map[String, Seq[SymbolSummary]] = Map.empty)(implicit nextC
     methods
   }
 
+  def isSurroundedByMetaclassTypeDecl: Boolean =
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/utils/Scope.scala
```diff
@@ -104,6 +104,9 @@ class Scope(summary: Map[String, Seq[SymbolSummary]] = Map.empty)(implicit nextC
     methods
   }
 
+  def isSurroundedByMetaclassTypeDecl: Boolean =
+    stack.map(_.scopeNode.node).collectFirst { case td: NewTypeDecl if td.name.endsWith("<metaclass>") => td }.isDefined
```
maltek, line 108: I'm not fully immersed in the implementation details of the scope classes. But I suspect this could go wrong with nested classes. (E.g. an anonymous class defined within a static method.) Should it be sth like this?

```suggestion
    stack.map(_.scopeNode.node).collectFirst { case td: NewTypeDecl => td }.exists(_ .name.endsWith(MetaTypeDeclExtension))
```

maltek (approved):

ml86 (approved):

PR #5512, opened 2025-05-23
[php] Normalize method name delimiter to `.`

andreimid:
* Normalized all method delimiters to `.`, instead of `->` and `::`
* Updated `PhpTypeRecovery` to now check for `<metaclass>` to determine static calls. Previously we had `classA<metaclass>::static_method-><returnValue>` and it would split on the `->`, but since we can no longer account for that, we now split on `<metaclass>` for static calls and create the method full name from the parts.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
@@ -49,16 +49,12 @@ trait AstCreatorHelper(disableFileContent: Boolean)(implicit withSchemaValidatio
     identifierNode(originNode, name, s"$$$name", typeFullName)
   }
 
-  protected def composeMethodFullName(
-    methodName: String,
-    isStatic: Boolean,
-    appendMetaTypeDeclExt: Boolean = false
-  ): String = {
+  protected def composeMethodFullName(methodName: String, appendMetaTypeDeclExt: Boolean = false): String = {
     if (methodName == NamespaceTraversal.globalNamespaceName) {
       globalNamespace.fullName
     } else {
       val className = if (appendMetaTypeDeclExt) {
-        getTypeDeclPrefix.map(name => s"${name}$MetaTypeDeclExtension")
+        getTypeDeclPrefix.map(name => s"$name$MetaTypeDeclExtension")
       } else {
         getTypeDeclPrefix
       }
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/passes/PhpTypeRecoveryPassTests.scala
```diff
@@ -370,12 +370,12 @@ class PhpTypeRecoveryPassTests extends PhpCode2CpgFixture() {
...
     "be properly resolved when called with $this" in {
       val List(fooCall) = cpg.method("bar").ast.isCall.take(1).l
-      fooCall.methodFullName shouldBe "ClassA->foo"
+      fooCall.methodFullName shouldBe "ClassA.foo"
     }
 
     "be properly resolved when called through class with known type" in {
-      val List(fooCall) = cpg.method("baz").ast.isCall.filter(_.code == "$a->foo()").take(1).l
-      fooCall.methodFullName shouldBe "ClassA->foo"
+      val List(fooCall) = cpg.method("baz").ast.isCall.filter(_.code == "$a.foo()").take(1).l
```
ml86, line 377: Why would you also change the code property? Please leave that as it was to reflect the actual code.

ml86 (changes requested):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/MemberTests.scala
```diff
@@ -146,7 +146,7 @@ class MemberTests extends PhpCode2CpgFixture {
       |""".stripMargin)
 
     inside(cpg.call.nameExact(Operators.fieldAccess).l) { case List(fieldAccess) =>
-      fieldAccess.code shouldBe "Foo::X"
+      fieldAccess.code shouldBe "Foo.X"
```
ml86, line 149: ... still a modified code field.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
```diff
@@ -449,12 +449,12 @@ class TypeDeclTests extends PhpCode2CpgFixture {
     "generate __construct calls" in {
       inside(cpg.call.name("__construct").l) {
         case constructAnonClass0 :: constructAnonClass1 :: Nil =>
-          constructAnonClass0.code shouldBe "Test0.php:<global>@anon-class-0->__construct(10)"
+          constructAnonClass0.code shouldBe "Test0.php:<global>@anon-class-0.__construct(10)"
```
ml86, line 452: Can you use a `.` instead of the `@` please. Or is there any reference in the PHP world that anonymous classes are `@` prefixed?

ml86 (changes requested):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
AndreiDreyer: Changes made.

There is nothing specific about it being an `@`, I was just following the naming pattern on the name changes to go lambdas on codescience.

Issue #5297, opened 2025-02-06
[Bug] "SOE getting binding table" exception when enabling withKeepTypeArguments


jbellis:
My code is
```
  private def createNewCpg(): Cpg = {
    println(s"Creating code graph at $outputPath")
    val config = Config()
      .withInputPath(absolutePath.toString)
      .withOutputPath(outputPath.toString)
      .withEnableTypeRecovery(true)
//      .withKeepTypeArguments(true)
    JavaSrc2Cpg().createCpg(config).get
  }
```

If I uncomment `withKeepTypeArguments(true)`, I get thousands of errors like this when loading the DataStax Cassandra repo (https://github.com/datastax/cassandra/).  I am not sure what features of this codebase trigger it but I do know that I have run `withKeepTypeArguments` successfully on much smaller codebases.  Happy to help narrow it down if you can point me in the right direction.

```
java.lang.RuntimeException: SOE getting binding table for org.apache.cassandra.db.lifecycle.LogRecord$Type
        at io.joern.javasrc2cpg.util.BindingTable$.createBindingTable$$anonfun$2(BindingTable.scala:68) ~[brokk-0.1-SNAPSHOT-jar-with-dependencies.jar:?]
        at scala.runtime.function.JProcedure1.apply(JProcedure1.java:15) ~[brokk-0.1-SNAPSHOT-jar-with-dependencies.jar:?]
        at scala.runtime.function.JProcedure1.apply(JProcedure1.java:10) ~[brokk-0.1-SNAPSHOT-jar-with-dependencies.jar:?]
        at scala.collection.immutable.List.foreach(List.scala:333) ~[brokk-0.1-SNAPSHOT-jar-with-dependencies.jar:?]
        at io.joern.javasrc2cpg.util.BindingTable$.createBindingTable(BindingTable.scala:72) ~[brokk-0.1-SNAPSHOT-jar-with-dependencies.jar:?]
        at io.joern.javasrc2cpg.astcreation.AstCreator.getMultiBindingTable$$anonfun$1(AstCreator.scala:300) ~[brokk-0.1-SNAPSHOT-jar-with-dependencies.jar:?]
        at scala.collection.mutable.HashMap.getOrElseUpdate(HashMap.scala:469) ~[brokk-0.1-SNAPSHOT-jar-with-dependencies.jar:?]
[...]

PR #5516, opened 2025-05-27
[javasrc2cpg] Fix crash when attempting to resolve type params

johannescoetzee:
This fixes an `UnsupportedOperationException` when attempting to resolve type params under some conditions. I found this in some larger projects and haven't found a small reproducing example yet, which is why this doesn't include a unit test. I did re-check the large projects with the fix and didn't get the error anymore though.


max-leuthaeuser:
Is this somehow releated to https://github.com/joernio/joern/issues/5297?

max-leuthaeuser (approved):

PR #5523, opened 2025-06-02
[php] Fix Field Identifier Representation

DavidBakerEffendi:
At current, many field access calls have `Identifier` RHS nodes, whereas a `FieldIdentifier` node would be the correct node.

Breaks up the diff of https://github.com/joernio/joern/pull/5493

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
@@ -116,9 +116,9 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) { 
     val callRoot  = callNode(call, code, name, fullName, dispatchType, Some(signature), Some(Defines.Any))
 
     val receiverAst = (targetAst, nameAst) match {
-      case (Some(target), Some(n)) =>
+      case (Some(target), Some(_)) =>
         val fieldAccess     = operatorCallNode(call, codePrefix, Operators.fieldAccess, None)
-        val fieldIdentifier = Ast(fieldIdentifierNode(call, n.rootCodeOrEmpty.stripPrefix("$"), n.rootCodeOrEmpty))
+        val fieldIdentifier = Ast(fieldIdentifierNode(call, name.stripPrefix("$"), name))
         Option(callAst(fieldAccess, target :: fieldIdentifier :: Nil))
       case (Some(target), None) => Option(target)
       case (None, Some(n))      => Option(n)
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -489,26 +490,35 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
       if (expr.isStatic)
-        StaticMethodDelimiter
+        "::"
       else if (expr.isNullsafe)
-        s"?$InstanceMethodDelimiter"
+        "?->"
       else
-        InstanceMethodDelimiter
-    }
+        "->"
```
ml86: Why do you stop using the defines?

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -117,8 +117,9 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
 
     val receiverAst = (targetAst, nameAst) match {
       case (Some(target), Some(n)) =>
-        val fieldAccess = operatorCallNode(call, codePrefix, Operators.fieldAccess, None)
-        Option(callAst(fieldAccess, target :: n :: Nil))
+        val fieldAccess     = operatorCallNode(call, codePrefix, Operators.fieldAccess, None)
+        val fieldIdentifier = Ast(fieldIdentifierNode(call, n.rootCodeOrEmpty.stripPrefix("$"), n.rootCodeOrEmpty))
```
ml86: Why would you work on the mangled code from another node just to mangle it back?
Please just work with the original data from the PHP AST.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -489,26 +490,35 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
 
-    val fieldAst = expr.name match {
-      case name: PhpNameExpr => Ast(fieldIdentifierNode(expr, name.name, name.name))
-      case other             => astForExpr(other)
+    def fieldNodeAndName(nameExpr: PhpExpr): (PhpExpr, String) = nameExpr match {
+      case name: PhpNameExpr => name -> name.name
+      case variable: PhpVariable =>
+        val (expr, name) = fieldNodeAndName(variable.value)
+        (expr, s"$$$name")
+      case other => other -> ""
```
ml86: I deem it bad style to use two different ways of creating a tuple in two consecutive lines. 
I suggest just using `(..., ...)` unless you are in a mapping scenario.

d1tto:
For PHP, it seems more reasonable to use an `Identifier` to handle variable variables, as it can distinguish them from regular property access and conveniently track their data flow. For example, in `$x->$y`, if I want to track the data flow of $y to determine its possible value range, using an `Identifier` would ensure the dataflow engine calculates its data flow, whereas a FieldIdentifier would not.

PR #5529, opened 2025-06-11
[kotlin2cpg] Populate offset and offset fields

johannescoetzee:
This PR populates the offset fields for all node types, but only tests methods and type decls since these are the only node types for which this is currently used.


max-leuthaeuser:
Afair we had some issues with strange unicode characters throwing off the offset calculations.
There are tests for that, e.g., for jssrc2cpg somewhere. Could you add similar tests here?

PR #5530, opened 2025-06-12
Change .location, NewLocation, and Location to lazily compute location information

ml86:
This PR is a copy of https://github.com/joernio/joern/pull/5482 which
got closed because Michael Flanders deleted the repository from which
he made the pull request.

- Adds Loc class as replacement for LocationCreator and the location node step. Loc is a replacement that is meant to compute and store less up-front compared to LocationCreator and location. IIUC, scala lazy values also take up space, so I eyeballed the operations that might take more steps to compute and made those lazy and those that are just lookups on the underlying node and made these methods.
- Adds a duplicate of one of the c2cpg test files to use Loc rather than replacing the other tests for now
- Add deprecation warnings to the previous location code


johannescoetzee (approved):
Approving on behalf of @ml86. From a Slack message from him:
`Consider the joern PR approved. Since i opened it, you need to approve it`

PR #5531, opened 2025-06-12
[c2cpg] Adjust INLINE call fullnames to our usual format

max-leuthaeuser:


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/MacroHandler.scala
```diff
@@ -99,23 +96,66 @@ trait MacroHandler { this: AstCreator =>
...
+    val signature        = s"$typeFullName(${argAsts.size})"
+    val fileName_        = fileName(macroDef)
+    val macroDefFullName = s"$fileName_:$macroDefName:$signature"
+
+    val callNode_ =
+      callNode(node, code, macroDefName, macroDefFullName, DispatchTypes.INLINED, Some(signature), Some(typeFullName))
+
+    // we stub the corresponding method manually because we want different line/column information
+    // (call information from the actual AST node but the method gets the IASTPreprocessorMacroDefinition information)
+    createMacroMethodStub(macroDefName, macroDefFullName, signature, fileName_, argAsts.size, macroDef)
```
maltek: This seems to be happening at the use-site of a macro. I don't think we want to create a new method stub every time a macro is used.

max-leuthaeuser: Totally messed this up. Will fix.

PR #5533, opened 2025-06-16
WIP: [testing] Translated distro test shell scripts to Python for testing on all OS

max-leuthaeuser:


.github/workflows/release.yml
```diff
@@ -36,8 +36,13 @@ jobs:
         run: rm -rf /home/runner/.cargo
         if: runner.os == 'Linux'
       - run: sbt scalafmtCheck test
-      - run: ./testDistro.sh
-      - run: |
+      - name: Test distribution
```
mpollmeier, line 39: Maybe worth skipping the tests here, given that they ran on PR and master already...

max-leuthaeuser: Yes, definitely 👍 
Did not want to change the behavior of the other workflows too much.
Feel free to change that in a PR following this one here (some workflows could need some cleaning anyway).

PR #5536, opened 2025-06-18
[testing] Added schema-extender test to testDistro.py

max-leuthaeuser:
Also:
- refactored `testDistro.py` for less code duplication and gracefull `subprocess.run` error handling
- enabled `console/src/test/scala/io/joern/console/PluginManagerTests.scala` on Windows again. There was a file created releated to the schema-extender stuff but that was never used appearently.
- removed copies of `sbt` and `scalac` scripts that were never used


testDistro.py
```diff
@@ -10,16 +10,20 @@
...
+def executable_name(executable, from_path = False):
     # Determine script name for Windows vs. Unix
     if os.name == 'nt':
         return f"{executable}.bat"
     else:
-        return f"./{executable}"
+        if not from_path:
+            return f"./{executable}"
+        else:
+            return executable
```
mpollmeier, line 21: hmm, I'd rather not have that `from_path` handling and be explicit on the call side, i.e. pass `./foo` if it should be the local one...

mpollmeier: but, either way is fine with me, just an idea

max-leuthaeuser: But that would not work for Windows. I do not want to switch OS specific executable names at the call site.

mpollmeier: true..

PR #5540, opened 2025-06-20
[swiftsrc2cpg][c2cpg] Initial round of refactorings

max-leuthaeuser:
~~And set block children order instead of argumentIndex.~~

Removed setting explicit order by relying on `io.joern.x2cpg.Ast#setOrderWhereNotSet`

Just some initial cleanups before starting with PRs for new features.

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
@@ -28,20 +28,6 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
     Ast(unknownNode(node, code(node)))
   }
 
-  protected def setOrder(asts: Seq[Ast]): Unit = {
-    var order = 1
-    asts.foreach { a =>
-      a.root match {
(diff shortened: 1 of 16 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -39,6 +28,20 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
+    asts.foreach { a =>
+      a.root match {
+        case Some(x: AstNodeNew) =>
+          x.order = order
+          order = order + 1
+        case None => // do nothing
+        case _ =>
+          order = order + 1
+      }
+    }
```
mpollmeier:
```suggestion
    asts.map(_.root).foreach {
      case Some(x: AstNodeNew) =>
        x.order = order
        order = order + 1
      case None => // do nothing
      case _ =>
        order = order + 1
    }
```

max-leuthaeuser: Actually, there is no need to set order explicitly at all.
This is done by: `io.joern.x2cpg.Ast#setOrderWhereNotSet`

I removed that everywhere.

PR #5548, opened 2025-06-24
refactor and fix ExternalCommand

mpollmeier:
* working dir: defaults to `.` anyway, no need to specify it everywhere
* do not falsely claim that we run from `System.getProperty("user.dir")`
* add and use `logIfFailed`
* various small style issues

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommandResult.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommandResult.scala
@@ -40,6 +40,7 @@ case class ExternalCommandResult(
   def successOption: Option[Seq[String]] =
     toTry.toOption
 
+  /** Lines of stdout, if successful. Otherwise an exception with message=stderr. */
   def toTry: Try[Seq[String]] = {
     if (successful) Success(stdOut)
     else Failure(new RuntimeException(stdErr.mkString("\n")))


max-leuthaeuser (approved):

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommandResult.scala
```diff
@@ -0,0 +1,47 @@
...
+  stdErr: Seq[String],
+  input: String,
+  additionalContext: Option[String]
+) {
+
+  def successful: Boolean =
+    exitCode == 0
+
+  /** Lines of standard output (if successful), or else a combination of stdout and stderr, plus some context.
+    */
```
maltek, line 18: this comment seems wrong?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommandResult.scala
```diff
@@ -0,0 +1,47 @@
...
+  def verifySuccess(): this.type = {
+    toTry.get
+    this
+  }
+
+  /** Lines of standard output, if successful. */
+  def successOption: Option[Seq[String]] =
+    toTry.toOption
+
+  def toTry: Try[Seq[String]] = {
```
maltek, line 44:
```suggestion
  /** Lines of stdout, if successful. Otherwise an exception with message=stderr. */
  def toTry: Try[Seq[String]] = {
```

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommandResult.scala
mpollmeier: yes, thanks, i went through a few iterations and the compiler didn't help me with that one :)

PR #5554, opened 2025-06-27
Implement more robust method finding logic in LazyLocation

johannescoetzee:
This should fix a match error causing crashes later.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Location.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Location.scala
@@ -104,7 +104,18 @@ class LazyLocation(storedNode: StoredNode, filenameOverride: Option[String]) ext
 
   private lazy val method: Method = storedNode match {
     case cfgNode: CfgNode => cfgNode.method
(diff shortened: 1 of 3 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/Location.scala
```diff
@@ -104,7 +104,18 @@ class LazyLocation(storedNode: StoredNode, filenameOverride: Option[String]) ext
 
   private lazy val method: Method = storedNode match {
     case cfgNode: CfgNode => cfgNode.method
-    case local: Local     => local.method.head
+    case _ => findParentMethod(storedNode).getOrElse {
+      throw IllegalStateException(s"Node not contained in a method $storedNode")
```
maltek: We have nodes like `Member` or `DomNode` which just don't exist within a method. So this should be `.getOrElse(defaultString)` like with the other values.

maltek (changes requested):

johannescoetzee:
@maltek There are still some issues with the unit tests, so I'll fix those and let you know when they're ready

PR #5556, opened 2025-06-30
[dot] Make max code char limit configurable

max-leuthaeuser:
Same mechanic as within `io.joern.x2cpg.AstNodeBuilder` but for the `.dot` representation.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/dotgenerator/DotSerializer.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/dotgenerator/DotSerializer.scala
@@ -15,7 +15,7 @@ object DotSerializer {
 
   private val DefaultCharLimit: Int = 50
   // maximum length of code fields in number of characters
-  private lazy val MaxCharLimit: Int =
+  private lazy val maxCharLimit: Int =
     sys.env.get("JOERN_MAX_DOT_CODE_LENGTH").flatMap(_.toIntOption).getOrElse(DefaultCharLimit)
 
   case class Graph(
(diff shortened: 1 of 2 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/dotgenerator/DotSerializer.scala
```diff
@@ -13,7 +13,10 @@ import scala.language.postfixOps
 
 object DotSerializer {
 
-  private val CharLimit = 50
+  private val DefaultCharLimit: Int = 50
+  // maximum length of code fields in number of characters
+  private lazy val MaxCharLimit: Int =
```
mpollmeier:
```suggestion
  private lazy val maxCharLimit: Int =
```

Uppercase naming convention is for constants only: https://docs.scala-lang.org/style/naming-conventions.html#constants-values-and-variables

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/dotgenerator/DotSerializer.scala
```diff
@@ -65,22 +68,22 @@ object DotSerializer {
     sb.append(s"""digraph "$name" {  \n""")
   }
 
-  private def limit(str: String): String = StringUtils.abbreviate(str, CharLimit)
+  private def limit(str: String): String = StringUtils.abbreviate(str, MaxCharLimit)
```
mpollmeier:
```suggestion
  private def limit(str: String): String = StringUtils.abbreviate(str, maxCharLimit)
```

PR #5557, opened 2025-06-30
Swift ConfigFilePass and MetaDataPass

itsacoderepo:
Adding initial ConfigFilePass to swift, renaming SwiftMetaDataPass to MetaDataPass


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/MetaDataPass.scala
max-leuthaeuser, line 1: SwiftMetaDataPass is not deleted?

itsacoderepo: ah yes, good catch.

max-leuthaeuser (approved):

PR #5560, opened 2025-07-02
[ruby] Create stable lambda names

AndreiDreyer:
* Create stable lambda names using surrounding structural infromation instead of simply having a global counter.


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/datastructures/RubyScope.scala
```diff
@@ -191,6 +196,74 @@ class RubyScope(summary: RubyProgramSummary, projectRoot: Option[String])
...
+  def getNewClassTmp: String = {
+    val surroundingFullName = this.surroundingScopeFullName.getOrElse(RubyDefines.Main)
+
+    stack
+      .collectFirst {
+        case ScopeElement(scopeNode: NamespaceScope, _) => s"$surroundingFullName.${scopeNode.getNextClassTmp}"
+        case ScopeElement(node: TypeLikeScope, _)       => s"$surroundingFullName.${node.getNextClassTmp}"
+        case ScopeElement(node: MethodLikeScope, _)     => s"$surroundingFullName.${node.getNextClassTmp}"
+      }
+      .getOrElse(s"${surroundingFullName}.${this.getNextClassTmp}")
```
ml86: Why do we need this fallback handling here and for all the other kinds of temp variables?

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/ClassTests.scala
```diff
@@ -363,9 +363,9 @@ class ClassTests extends RubyCode2CpgFixture {
         |""".stripMargin)
 
     "generate a type decl with the associated members" in {
-      inside(cpg.typeDecl.nameExact("<anon-class-0>").l) {
+      inside(cpg.typeDecl.nameExact("Test0.rb:<main>.<anon-class-0>").l) {
```
ml86: In other tests you use `$Main` instead of `<main>` please do so here as well as in the rest of this file.

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
```diff
@@ -1165,4 +1165,31 @@ class MethodTests extends RubyCode2CpgFixture {
       case xs => fail(s"Expected one call, got ${xs.code.mkString(",")}")
     }
   }
+
+  "tmp-vars in different methods should have same names" in {
```
ml86: Bad test description, "should both start from zero" is better.

ml86 (changes requested):
I am also missing at least a single test where the tmp/anon counters are greater then zero.

ml86 (approved):

PR #5563, opened 2025-07-03
[c2cpg] Consistent handling of unparsable C++20 coroutine keywords

max-leuthaeuser:


maltek (commented):
I don't know if hacking around parser deficiencies like this is really worth it. It seems likely to fail in all kinds of fun new weird ways every time we come across another new edge-case.

maltek:
That thing where we create a fake method and then put its contents somewhere in the AST of a real method is new to me. And to me that approach seems like it's just asking for trouble.

max-leuthaeuser:
Guess you are right.
I changed it back to the generation of Unknown nodes without any further handling of broken sub parse trees.

PR #5565, opened 2025-07-04
[php] Rework `PhpScopeElement`

AndreiDreyer:
Refactored the `PhpScopeElement` to have separate `Scope` classes (similar to `rubysrc2cpg`) instead of trying to do it all through one class with differing `apply` methods. This makes adding a new field to a single `Scope` type (like `MethodScope`) much simpler, and no longer requires a new `Option` parameter on the base class just to cater for a single scope type.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/utils/Scope.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/utils/Scope.scala
@@ -1,12 +1,11 @@
 package io.joern.php2cpg.utils
 
-import flatgraph.DiffGraphBuilder
 import io.joern.php2cpg.astcreation.AstCreator.NameConstants
 import io.joern.php2cpg.parser.Domain.{PhpExpr, PhpNode}
 import io.joern.php2cpg.parser.Domain.MetaTypeDeclExtension
 import io.joern.php2cpg.passes.SymbolSummaryPass.*
 import io.joern.x2cpg.Ast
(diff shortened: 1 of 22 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/utils/Scope.scala
```diff
@@ -30,109 +35,113 @@ class Scope(summary: Map[String, Seq[SymbolSummary]] = Map.empty)(implicit nextC
...
-      case Some(node) =>
-        s"${this.surroundingScopeFullName.getOrElse("<global>")}.${node.scopeNode.getNextClassTmp}"
-      case None =>
+      case Some(_ @ScopeElement(namespace: NamespaceScope, _)) =>
+        s"${this.surroundingScopeFullName.getOrElse("<global>")}.${namespace.getNextClassTmp}"
+      case Some(_ @ScopeElement(typeScope: TypeScope, _)) =>
+        s"${this.surroundingScopeFullName.getOrElse("<global>")}.${typeScope.getNextClassTmp}"
+      case Some(_ @ScopeElement(methodScope: MethodScope, _)) =>
+        s"${this.surroundingScopeFullName.getOrElse("<global>")}.${methodScope.getNextClassTmp}"
+      case _ =>
```
ml86: Here and all other places where you use `_@XXX()` the `_@` is unnecessary. Please remove it.

ml86 (approved):

PR #5566, opened 2025-07-06
Delombok resource limits

jbellis:
- Add -Xmx1G to spawned JVMs [instead of default 1/4 system ram]
- Use a fixed-size thread pool to avoid "compensation" threads that the default FJP spawns when it detects blocking (like for a subprocess)

This takes it from spawning 100s of delombok jvms for a large delombok effort to the expected jvm-per-core.


jbellis:
cc @maltek

johannescoetzee (approved):
This looks good to me and local testing didn't reveal any issues. Thanks for the contribution!

PR #5568, opened 2025-07-07
[swiftsrc2cpg] Clean fullnames from generics and remove ImportsPass

max-leuthaeuser:
* clean all fullnames from generics and linebreaks (that was totally missing) and also use them for identifier typefullnames if possible
* remove ImportsPass (there are no import calls in Swift, just import statements which are already handled)


maltek:
Anything type-related for Swift should probably go through @bbrehm, not me. I don't really know what the plan is there.

PR #5571, opened 2025-07-08
[php] Capture variables used in arrow closure expressions

andreimid:
Added capturing for variables used in arrow closure expressions. Variables used in arrow closures that are not parameters passed to the arrow closure automatically capture the value of the variable outside the arrow closure

```php
$foo = "foo";
fn () => $foo; // This captures the value of $foo outside the scope of the arrow closure
```

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
@@ -256,6 +256,28 @@ class ClosureTests extends PhpCode2CpgFixture {
     }
   }
 
+  "multiple recursive arrow functions" should {
+    val cpg = code(
+      """<?php
+        |$a = 5
+        |fn () => fn () => $a;
+        |fn () => fn () => $a;
+        |""".stripMargin,
+      "foo.php"
+    )
+
+    "correctly number lambdas" in {
+      inside(cpg.method.isLambda.l) {
+        case firstOuterInnerLambda :: firstOuterLambda :: secondOuterInnerLambda :: secondOuterLambda :: Nil =>
+          firstOuterInnerLambda.name shouldBe "foo.php:<global>.<lambda>0.<lambda>0"
+          firstOuterLambda.name shouldBe "foo.php:<global>.<lambda>0"
+          secondOuterInnerLambda.name shouldBe "foo.php:<global>.<lambda>1.<lambda>0"
+          secondOuterLambda.name shouldBe "foo.php:<global>.<lambda>1"
+        case xs => fail(s"Expected four lambdas, instead got ${xs.name.mkString(", ")}")
+      }
+    }
+  }
+
   "recursive arrow functions with captured variables" should {
     val cpg = code(
       """<?php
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
```diff
@@ -127,14 +127,64 @@ class ClosureTests extends PhpCode2CpgFixture {
...
+        |$x = fn ($value) => $value + $captured + $captured + $captured;
         |""".stripMargin,
       fileName = "foo.php"
     )
 
     "not leave orphan identifiers" in {
       cpg.identifier.filter(node => Try(node.astParent.toList).isFailure).toList shouldBe Nil
     }
+
+    "local variable $captured exists" in {
```
ml86, line 140: More tests needed:
1. Capturing of multiple parameters/locals into an arrow function
2. Recursive capturing of arrow function in arrow function

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -155,16 +159,60 @@ trait AstCreatorHelper(disableFileContent: Boolean)(implicit withSchemaValidatio
...
+        val methodName = scope.surroundingScopeFullName.get // should throw error if this does not exist
+        val methodRef  = scope.getSurroundingArrowClosureMethodRef
+
+        val closureBindingId = s"$methodName:$name"
+        val localNode_ =
+          localNode(expr, name, local.code, local.typeFullName, closureBindingId = Option(closureBindingId))
+
+        createClosureBinding(methodRef, closureBindingId, local, localNode_)
+
+        localNode_
```
ml86: There is a trait in the graph schema called `LocalLike` which you could use to handle the parameter and local cases together but that is not a must. What is a must is that removal of this ugly code duplication.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -155,16 +159,60 @@ trait AstCreatorHelper(disableFileContent: Boolean)(implicit withSchemaValidatio
...
   }
 
+  private def createClosureBinding(
+    methodRef: Option[NewMethodRef],
+    closureBindingId: String,
+    existingNode: (NewLocal | NewMethodParameterIn),
+    newLocalNode: NewLocal
+  ): Unit = {
+    val closureBindingNode =
+      NewClosureBinding().closureBindingId(closureBindingId).evaluationStrategy(EvaluationStrategies.BY_SHARING)
```
ml86: There already two other places in the php2cpg frontend which create a closure binding from scratch ( `NewClosureBinding()`. Is it possible to get some sensible code reuse with those?

ml86 (changes requested):

ml86 (approved):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
```diff
@@ -128,14 +128,262 @@ class ClosureTests extends PhpCode2CpgFixture {
...
+        |""".stripMargin,
+      "foo.php"
+    )
+
+    "create a local variable in innermost lambda" in {
+      // Confirming only 1 local is created for each captured variable
+      cpg.method.name("foo.php:<global>.<lambda>0.<lambda>1").local.name("a").size shouldBe 1
+      cpg.method.name("foo.php:<global>.<lambda>0.<lambda>1").local.name("b").size shouldBe 1
+
+      val localNode = cpg.method.name("foo.php:<global>.<lambda>0.<lambda>1").local.name("a").head
```
ml86: Naming of the lambda function suggest that there is shared suffix counter. I would have expected a counter per scope and that every lambda has its own scope. Please use separate counter per scope so that the lambda name is `foo.php:<global>.<lambda>0.<lambda>0`

TNSelahle: Yep, that's correct. I've confirmed this and implemented a fix. Just testing now and will ping when the PR is ready for your review again.

PR #5584, opened 2025-07-11
[swiftsrc2cpg] Several improvements

max-leuthaeuser:
* dictionary, tuple, and list lowering
 * proper typeRefNode in cast expressions
 * no dynamicTypeHintFullName anymore; just possibleTypes if applicable
 * added support for missing addressOf calls
 * fixed wrong Operators.plus calls; it's Operators.addition
 * fixed wrong Operators.minus calls; it's Operators.subtraction

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
@@ -2,38 +2,101 @@ package io.joern.swiftsrc2cpg.astcreation
 
 import io.joern.swiftsrc2cpg.parser.SwiftNodeSyntax.*
 import io.joern.swiftsrc2cpg.passes.GlobalBuiltins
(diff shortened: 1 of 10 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
```diff
@@ -2,38 +2,101 @@ package io.joern.swiftsrc2cpg.astcreation
...
+
+          val slicedElements = elements.slice(0, MaxInitializers).toList
+
+          val propertiesAsts = slicedElements.map {
+            case dictElement: DictionaryElementSyntax =>
+              val (lhsNode, rhsAst) = {
+                val key = dictElement.key
+                val fieldName = key match {
+                  case decl: (DeclReferenceExprSyntax | StringLiteralExprSyntax) =>
+                    stripQuotes(code(decl))
```
maltek: just use an index access here instead of a field access - that allows arbitrary expressions as index, and we don't need such hacks in order to satisfy the constraints of a `FieldIdentifier`

maltek (approved):

PR #5585, opened 2025-07-11
use `loneElement` instead of `onlyChecked`

mpollmeier:


maltek (approved):
didn't you just add a new message argument, to make errors from missing edges easier to understand?

mpollmeier:
while writing the hints, i noticed that something like `AstNode.astParentOption` would be useful and I added that... would you agree?

maltek:
> while writing the hints, i noticed that something like AstNode.astParentOption would be useful and I added that... would you agree?

it is technically more correct - there are roots that just don't have a parent. IME it's rare to accidentally run into that, except with frontend bugs where crashing loudly is likely the best course of action. Usually you know where you want to go in the AST and stop long before you reach a `File` node. But overall it's probably good to have it there.

mpollmeier:
kotlin2cpg is already fixed in https://github.com/joernio/joern/pull/5587 - will await the merge and rebase then

PR #5589, opened 2025-07-14
move walkUpAst and walkUpContains into companion object of CfgNodeMethods

mpollmeier:
* these don't work on the node of the value class, so it's rather
  confusing
* their code was copied without changes in codescience, this way we can
  reuse it

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CfgNodeMethods.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CfgNodeMethods.scala
@@ -123,14 +123,14 @@ class CfgNodeMethods(val node: CfgNode) extends AnyVal with NodeExtension {
 
 object CfgNodeMethods {
 
-  /** Attention: this only works for some special CfgNodes that are guaranteed to always be direct children of a method
-    * \- that's why it's private!
+  /** Attention: this only works for some special CfgNodes that are guaranteed to always be direct children of a
+    * method... that's why it's private!
     */
   private def walkUpAst(node: CfgNode): Method =
     node.astParent.asInstanceOf[Method]
 
-  /** Attention: this only works for some special CfgNodes that are guaranteed to always be direct children of a method
-    * \- that's why it's private!
+  /** Attention: this only works for some special CfgNodes that are guaranteed to always be direct children of a
+    * method... that's why it's private!
     */
   private def walkUpContains(node: StoredNode): Method = {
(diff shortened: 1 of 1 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CfgNodeMethods.scala
```diff
@@ -118,10 +119,14 @@ class CfgNodeMethods(val node: CfgNode) extends AnyVal with NodeExtension {
     node.lineNumber.map(_.toLong.toHexString)
   }
 
-  private def walkUpAst(node: CfgNode): Method =
+}
+
+object CfgNodeMethods {
+
+  def walkUpAst(node: CfgNode): Method =
```
maltek: This should not be a public method, because it does not work for arbitrary cfg nodes. Only for some special few ones that are guaranteed to always be direct children of a method.

I was thinking we could change the parameter type to a union type - but that doesn't work because the `CallRepr` that's not a `Call` from line 107 only exists downstream in codescience, not here in joern. 

So I'd say let's just keep it private and duplicated.

mpollmeier: ok thanks, will close this one and add a comment instead

mpollmeier: or actually, let's still move it to the object..

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/language/nodemethods/CfgNodeMethods.scala
```diff
@@ -118,10 +119,20 @@ class CfgNodeMethods(val node: CfgNode) extends AnyVal with NodeExtension {
...
+
+  /** Attention: this only works for some special CfgNodes that are guaranteed to always be direct children of a
+    * method... that's why it's private!
+    */
   private def walkUpAst(node: CfgNode): Method =
     node.astParent.asInstanceOf[Method]
 
-  private def walkUpContains(node: StoredNode): Method =
+  /** Attention: this only works for some special CfgNodes that are guaranteed to always be direct children of a
+    * method... that's why it's private!
```
maltek:
```suggestion
  /** Attention: this only works for those `CfgNodes` that have `CONTAINS` edges (see `ContainsEdgePass`)... that's why it's private! 
```

maltek (approved):

PR #5592, opened 2025-07-14
[php2cpg] Add method bindings to class methods to fix dataflow

johannescoetzee:
This PR adds method bindings to php2cpg class methods to fix dataflow through dynamic calls. This is a very bare-bones implementation which only adds the bindings for methods defined directly in the class and ignores inheritance since, in general, we won't have any information about ancestor methods at this point in AST creation. We would need either a pass after AST creation, or changes in the back-end logic, to handle inherited methods.


johannescoetzee:
I've updated the PR with the change to use the empty string for php signatures

PR #5594, opened 2025-07-16
[swiftsrc2cpg] Fix for typeRefs, fullNames, trailing closures, and more

max-leuthaeuser:
- no more broken typrefs from type declarations (some slipped through in the last PR)
- fullnames use dots instead of colons to separated types/method names now
- no more method nodes from trailing closures under calls (method refs as expected now)
- handle accessor declarations 
- fullname uniqueness pass
- add .xib config files


maltek (approved):
- signatures will need a larger revamp, but I will talk to you about that outside this PR
- I just learned from some of those tests that `<` and `>` are legal in method names in Swift, so all our synthetic names using those can theoretically clash with real names. I don't think I really want those to be different than in all the other frontends though.

PR #5601, opened 2025-07-22
Use indexAccess operator for php field access calls with arbitrary expressions

johannescoetzee:
This PR changes the representation for php property accesses with arbitrary expressions on the RHS, for example:
```
$obj->{foo()}
```

Previously, these were represented as field access calls, but this causes a crash in the back-end since it breaks the assumption that a field access calls have an identifier or field identifier as the second argument. Instead, these are now represented as index access calls, per the suggestion from @maltek

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
@@ -517,7 +517,6 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) { 
 
     val fieldAst = fieldName match {
       case None =>
-        logger.debug(s"Unable to determine field identifier node from ${fieldNode.getClass} (parent node $expr)")
         astForExpr(fieldNode)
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -513,22 +513,33 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
-    val fieldAst = fieldNodeAndName(expr.name) match {
-      case (other, None) =>
-        logger.warn(s"Unable to determine field identifier node from ${other.getClass} (parent node $expr)")
-        astForExpr(other)
-      case (expr, Some(name)) => Ast(fieldIdentifierNode(expr, name.stripPrefix("$"), name))
+    val (fieldNode, fieldName) = fieldNodeAndName(expr.name)
+
+    val fieldAst = fieldName match {
+      case None =>
+        logger.debug(s"Unable to determine field identifier node from ${fieldNode.getClass} (parent node $expr)")
```
maltek: I'd say we can remove this log - now that we lower it correctly

PR #5607, opened 2025-07-25
[php] Add `ListExpr` support in foreach loop

AndreiDreyer:
* Added `PhpListExpr` deconstruction support in foreach loop
* Removed `astForListExp` as it is no longer used

```php
$data = [[1, 2]]
foreach ($data as list($a, $b)) {
  echo $a; // prints 1
  echo $b; // prints 2
}
```


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -706,29 +705,8 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
-     * and the case:
-     *   foreach ($arr as list($a, $b)) { ... }
-     * can be lowered as above for each $arr[i];
-     *
-     * The below is just a placeholder to prevent crashes while figuring out the cleanest way to
-     * implement the above lowering or to think of a better way to do it.
-     */
-
+    // This is only used in `AstForControlStructureesCreator::astForForeachStatement:assignItemTargetAst` so that the
+    // code field can be populated.
```
ml86: I do not like this solution. Creating a hole AST just to take the code property from it is mental and computational overhead. Please try to find a better solution.

PR #5610, opened 2025-07-28
Populate static receiver field for javasrc2cpg

johannescoetzee:


joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/CallTests.scala
```diff
@@ -131,7 +134,47 @@ class NewCallTests extends JavaSrcCode2CpgFixture {
...
         |""".stripMargin)
 
-    cpg.call.name("foo").methodFullName.l shouldBe List("foo.Foo.foo:java.lang.String()")
+    inside(cpg.call.name("foo").l) { case List(fooCall) =>
+      fooCall.methodFullName shouldBe "foo.Foo.foo:java.lang.String()"
+      fooCall.staticReceiver shouldBe Some("foo.Foo")
+    }
+  }
+
+  "calls to inherited static methods" should {
```
ml86, line 143: Please also add a test where the static method is called on an instance and not directly via the class name.

ml86 (changes requested):

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/CallTests.scala
johannescoetzee: Will do! I didn't know that was a thing in Java as well.

ml86 (approved):

johannescoetzee:
We've decided not to populate the STATIC_RECEIVER field for javasrc, but to use information from JavaParser to set the methodFullName instead

PR #5612, opened 2025-07-30
fix json serializer: define only for nodes, not for all Products

mpollmeier:
prior to this change:
```
List((1,2,3)).toJson
// [{"head":1,"next":{"head":2,"next":{"head":3,"next":{}}}}]
```

now:
```
List((1,2,3)).toJson
// [1,2,3]
```

It's a regression that was introduced in https://github.com/joernio/joern/pull/5530
I have no idea why Michael Flanders made this change.
This is also adding a test so we catch it early next time round.


ml86 (approved):

mpollmeier:
I now understand why he changed the serializer to Product - the newly introduced Location types are not AbstractNodes...
I'm cleaning it up, updating the tests etc.

ml86 (approved):

PR #5614, opened 2025-07-31
update cpg, adapt to change of ambiguous callIn/Out method

maltek:


mpollmeier (approved):
wouldn't it make sense to give it a nicer step name, e.g. `method`? i.e. define it on the `stepNameOut` in the cpg schema.

PR #5617, opened 2025-08-07
Fix maven dependency fetching on Windows

johannescoetzee:
Without this option set, the `mvn` command can't be found even if it is in the PATH. With this option set, dependency fetching breaks on Linux, so we need to do the OS check.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/MavenDependencies.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/MavenDependencies.scala
@@ -47,8 +47,9 @@ object MavenDependencies {
   }
 
   private[dependency] def get(projectDir: Path): Option[collection.Seq[String]] = {
+    val isWindows = System.getProperty("os.name").toLowerCase.startsWith("windows")
     val lines = ExternalCommand
-      .run(command = fetchCommandWithOpts, workingDir = Option(projectDir))
+      .run(command = fetchCommandWithOpts, workingDir = Option(projectDir), isShellCommand = isWindows)
       .logIfFailed()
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/MavenDependencies.scala
```diff
@@ -47,8 +47,9 @@ object MavenDependencies {
   }
 
   private[dependency] def get(projectDir: Path): Option[collection.Seq[String]] = {
+    val isWindows = System.getProperty("os.name").toLowerCase.startsWith("windows")
```
max-leuthaeuser: You may want to use `import scala.util.Properties.isWin` and use it like e.g., in https://github.com/joernio/joern/blob/master/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommand.scala#L64

max-leuthaeuser (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/MavenDependencies.scala
mpollmeier: yes, what max said

mpollmeier (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/MavenDependencies.scala
johannescoetzee: Thanks! I was looking for something like that, but missed it

PR #5621, opened 2025-08-12
refactor x2cpg configuration to use dependent method types

maltek:
This removes the whole `withInheritedFields()` foot gun that each frontend config type had to adhere to. Instead they have a simple case class member variable of type `X2CpgConfig.GenericConfig` now, and a `withGenericConfig()` method to get a copy of that config with a different value for that member.

Many of the x2cpg traits no longer have a generic type argument for the frontend config type. Instead, `X2CpgFrontend` defines an abstract type member that each frontend can set to the config type it uses. (I think that's a dependent type?) This allows us to remove a good chunk (not all) of type casts and boilerplate from each frontend config definition.

This also changes all the config types to be immutable, instead of some `with*` methods mutating (as in the builder pattern) and some returning copies like case classes tend to do. We could go further there and remove most frontend-specific `with*` methods in favor of the regular `copy(option=...)`, but this PR does not go so far since the `with*` helpers are still convenient for the shared options and it looks a bit weird to mix the two styles.

`X2CpgMain` also defines a default implementation for `run()` now - that was mostly the same for each frontend before.


max-leuthaeuser (approved):
Looking good! Thanks 👍 

Fingers crossed the new generic run method handles absolut/relative path conversions like each frontend used to do individually. I checked and I hope I did not miss anything.

PR #5629, opened 2025-08-25
[php2cpg] Change method and call representation to fix dataflow through lambdas

johannescoetzee:
This PR adds support for closed source dataflow tracking through callable variable invocations, for example
```
$x = fn() => 42;
$x();
```

This is done by creating a callable type decl for the lambda function, in this example, with an `__invoke` method defined on it. The call is then represented as an explicit call to this `__invoke` method, so `$x->__invoke` in this case. 

This call representation handles invocations on variables instantiated from regular php callables by default, since those would have the `__invoke` method defined as well.

The last case is for first class callables as described in https://www.php.net/manual/en/functions.first_class_callable_syntax.php.

To support any method being potentially callable, a type decl is created for each method with the same name and fullname as the method. This typeDecl then has the `__invoke` method defined on it.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
@@ -491,6 +491,12 @@ class ClosureTests extends PhpCode2CpgFixture {
         lambdaMethod.fullName shouldBe "foo.<lambda>0"
         lambdaMethod.astParentType shouldBe "TYPE_DECL"
         lambdaMethod.astParentFullName shouldBe "Test0.php:<global>"
+
+        inside(lambdaMethod.parameter.l) { case List(xParam) =>
+          xParam.name shouldBe "x"
+          xParam.code shouldBe "$x"
+          xParam.lineNumber shouldBe Some(3)
+        }
       }
     }
 


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -154,11 +168,45 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
...
+      val paramCopies = parameters.map(paramAst => (paramAst, paramAst.root)).collect {
+        case (paramAst, Some(root: AstNodeNew)) => paramAst.subTreeCopy(root)
+      }
+      createAndPushTypeDeclForMethod(decl, method, isAnonymousDecl, paramCopies)
+    }
 
     methodAst
   }
 
+  private def createAndPushTypeDeclForMethod(
```
johannescoetzee, line 188: This still needs a comment explaining why this is needed.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
```diff
@@ -83,9 +84,10 @@ class ClosureTests extends PhpCode2CpgFixture {
       closureMethod.fullName shouldBe expectedName
       closureMethod.signature shouldBe ""
       closureMethod.code shouldBe s"function $expectedName($$value) use($$use1, &$$use2)"
-      closureMethod.parameter.size shouldBe 1
+      closureMethod.parameter.size shouldBe 2
 
-      inside(closureMethod.parameter.l) { case List(valueParam) =>
+      inside(closureMethod.parameter.l) { case List(thisParam, valueParam) =>
+        thisParam.name shouldBe "this"
```
ml86: Does PHP really capture "this" into lambdas? I ran a small example
```
<?php
$x = function() {
	print($this);
};
$x();
```

And got the error: `Fatal error: Uncaught Error: Using $this when not in object context in /home/user/scripts/code.php:3`

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
```diff
@@ -446,4 +449,63 @@ class ClosureTests extends PhpCode2CpgFixture {
...
+
+          inside(invokeMethod.parameter.l) { case List(thisParameter, xParameter) =>
+            thisParameter.name shouldBe "this"
+
+            xParameter.name shouldBe "x"
+          }
+        }
+      }
+    }
+  }
```
ml86, line 521: We seem to have "foo.<lambda>0" method and "foo.<lambda>0.__invoke" method. Is one just invoking the other? I cant remember we agreed to have such a forwarder method.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
```diff
@@ -76,11 +75,28 @@ class TypeDeclTests extends PhpCode2CpgFixture {
       }
     }
 
+    "have the correct bindings for the synthetic foo method class" in {
+      inside(cpg.typeDecl("foo").bindsOut.sortBy(_.name).l) { case List(invokeBinding) =>
+        invokeBinding.name shouldBe "__invoke"
+        invokeBinding.methodFullName shouldBe "<empty>"
```
ml86, line 81: Since `<empty` is the default value, you do not have to explicitly set it. Here and in the other places.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
johannescoetzee: The `$this` parameter refers to the instance of the closure, not `$this` captured from the outer scope, for example:
```
class AnonymousClass {
  function anonymousMethod($this, $value) {...}
}

$x = new AnonymousClass();
$x.__invoke($value); <-- $x is passed to $this
```

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
johannescoetzee: `foo.0.__invoke` refers to `foo.0` via a REF edge. It's currently done this way since we would require both for regular functions/methods, but for anonymous functions and lambdas I think we only need the `__invoke`. I don't think this case is something we thought about or discussed specifically.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
johannescoetzee: I added the test for this here since the testcase above did change from having it set to not set. I think having this test to keep track of changes is useful

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
ml86: Yes, I your example for a class and class method this makes sense. But it seems not to be the case for lambda methods.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
ml86: How can `foo.0.__invoke` refer to `foo.0` via REF edge? It is not support have METHOD -> REF_EDGE -> METHOD in the CPG.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/TypeDeclTests.scala
ml86: Ok, if you put it on purpose thats fine.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
johannescoetzee: We use the same lowering for lambda methods and anonymous functions, so in both cases we create the Anonymous/LambdaClass which the `$this` ends up referring to. In the double-method representation (where we have the method with and without the invoke) we could potentially leave this out of the closureMethod, but with the `__invoke` I'd expect this param to be necessary for the dataflow engine. If only the receiver argument is required without needing a corresponding parameter, then this could be left out.

I could also rename it to something more descriptive to avoid confusion between this and `$this` from the enclosing scope.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
johannescoetzee: This is true. I was confusing it with the binding. The `foo.0.__invoke` method isn't actually required as long as the correct binding exists, so I'll remove it.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
johannescoetzee: For anyone reading this later, we had a call to discuss this point. The methods created from lambdas and anonymous functions should not have the `$this` parameter. At the callsite of `$x.__invoke`, for example, `$x` should be given as the receiver of the `__invoke` call, but not as an argument with index 0.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ClosureTests.scala
```diff
@@ -446,4 +447,70 @@ class ClosureTests extends PhpCode2CpgFixture {
       }
     }
   }
+
+  "a simple lambda definition and call" should {
```
ml86, line 451: Missing a test for the lambda parameters here or does that exists somewhere else?

johannescoetzee: There is one for a testcase earlier in the file, but I'll add another here to make it easier to spot for anyone using the tests as documentation (which I've done myself)

PR #5630, opened 2025-08-26
[sbt] Auto-reload sbt when application.conf changes

TNSelahle:
Auto-reload `sbt` when `application.conf` changes.

[Screencast from 2025-08-26 08-40-31.webm](https://github.com/user-attachments/assets/fdb01fee-9a8e-4038-889a-60b5b7e71d75)


TNSelahle:
This works for all frontends that use an astgen with the version specified in `application.conf`.

ml86 (approved):

PR #5632, opened 2025-08-29
[swiftsrc2cpg] Implement type and declaration fullname extraction from the Swift compiler

max-leuthaeuser:
Introduces support for retrieving full Swift compiler type and declaration fullname information via new CLI options.
 - refactored AST creation to use type/fullname mapping for more accurate type/fullname resolution
 - added utility helpers for type and fullname handling
 - updated tests and dependencies to support these changes
 - bumped astgen version
 - added Google Gson for large JSON file processing (from swiftc dump ast invocations)

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
@@ -191,7 +191,7 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
     }
   }
 
-  private val ReturnTypeMatcher = """\(.*\)->(.+)""".r
+  private val ReturnTypeMatcher = """\(.*\)(->|:)(.+)""".r
 
   protected def methodInfoForFunctionDeclLike(node: FunctionDeclLike): MethodInfo = {
     val name = calcMethodName(node)
@@ -199,12 +199,12 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
       case Some(fullNameWithSignature) =>
         val (fullName, signature) = methodInfoFromFullNameWithSignature(fullNameWithSignature)
         val returnType = node match {
-          case init: InitializerDeclSyntax =>
-            ReturnTypeMatcher.findFirstMatchIn(signature).map(_.group(1)).getOrElse(astParentInfo()._2)
-          case deinit: DeinitializerDeclSyntax =>
+          case _: DeinitializerDeclSyntax =>
             Defines.Void
+          case _: InitializerDeclSyntax =>
+            ReturnTypeMatcher.findFirstMatchIn(signature).map(_.group(2)).getOrElse(astParentInfo()._2)
           case _ =>
-            fullnameProvider.typeFullname(node).getOrElse(Defines.Any)
+            ReturnTypeMatcher.findFirstMatchIn(signature).map(_.group(2)).getOrElse(Defines.Any)
         }
         registerType(returnType)
         MethodInfo(name, fullName, signature, returnType)


joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/utils/FullnameProviderTests.scala
```diff
@@ -0,0 +1,101 @@
...
+import scala.jdk.CollectionConverters.*
+
+class FullnameProviderTests extends AnyFlatSpec with Matchers {
+
+  private class MockFullNameProvider(typeMap: SwiftFileLocalTypeMapping) extends FullnameProvider(typeMap) {
+    override def typeFullname(range: (Int, Int), nodeKind: String): Option[String] =
+      super.typeFullname(range, nodeKind)
+
+    override def declFullname(range: (Int, Int), nodeKind: String): Option[String] =
+      super.declFullname(range, nodeKind)
```
ml86, line 16: Why do we need those overrides?

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/utils/FullnameProviderTests.scala
```diff
@@ -0,0 +1,101 @@
+package io.joern.swiftsrc2cpg.utils
+
+import io.joern.swiftsrc2cpg.utils.SwiftTypesProvider.{ResolvedTypeInfo, SwiftFileLocalTypeMapping}
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import scala.collection.mutable
+import scala.jdk.CollectionConverters.*
+
+class FullnameProviderTests extends AnyFlatSpec with Matchers {
```
ml86, line 9: Looks like this class only tests the easy cases but not the more complicate and not intuitive ones like the recursive case and the `+-1` handling.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/utils/SwiftCompilerFullnameTests.scala
```diff
@@ -0,0 +1,77 @@
...
+      helloMethod.fullName shouldBe "SwiftTest.Main.hello:()->()"
+      val List(helloCall) = cpg.call.nameExact("hello").l
+      helloCall.methodFullName shouldBe "SwiftTest.Main.hello:()->()"
+
+      helloMethod.fullName shouldBe helloCall.methodFullName
+
+      val List(mainTypeDecl) = cpg.typeDecl.nameExact("Main").l
+      mainTypeDecl.fullName shouldBe "SwiftTest.Main"
+      val List(mainConstructor) = mainTypeDecl.ast.isMethod.isConstructor.l
+      mainConstructor.fullName shouldBe "SwiftTest.Main.init:()->SwiftTest.Main"
```
ml86, line 35: Two questions:
1. Is the current format where the parameters and return value are written as `(<parms>) -> <ret>` directly coming from the compiler or is that already mangled? I am wondering how much effort it would be do mangle it into the full name format we use for java
2. Does the Swift AST contain an allocator call or does it only have a constructor call. It looks here like there is only a constructor call because the constructor returns an instance of type Main. I am again thinking in the direction of using a representation which is closer to the other languages.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/utils/FullnameProviderTests.scala
max-leuthaeuser: The actual implementation is a protected function (being a helper function for `def typeFullname(node: SwiftNode): Option[String]` in `FullNameProvider`.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/utils/FullnameProviderTests.scala
max-leuthaeuser: Right, will add these.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/utils/SwiftCompilerFullnameTests.scala
max-leuthaeuser: 1.) thats the exact signature coming from the compiler after demangling the name. I already checked re-formatting it to the usual format but that would require parsing the fullname/signature for more complex signatures like e.g., nested stuff (`Foo.foo(lambda:(Int)->String)->Bar`). I think its much simpler to stick to what Swift itself uses.

max-leuthaeuser: 2.) It only has the constructor call.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/utils/FullnameProviderTests.scala
ml86: Ah yes, right.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/utils/SwiftCompilerFullnameTests.scala
ml86: Regarding 1: Ok, then lets stick to this format. I guess this will show us the places where people did not adhere to the "do not parse the full name" rule.

Regarding 2: I think we should at the minimum rewrite the constructors to have the object instance passed in as param 0. Otherwise it would be a relative pain in the ass for @maltek to handle the constructor policies differently. Adding the `newObject = <operator>.allocator` could also be added but is not really necessary.

max-leuthaeuser: > I think we should at the minimum rewrite the constructors to have the object instance passed in as param 0

~~Yes, will do.~~
Done.

ml86 (approved):

PR #5635, opened 2025-08-29
[csharpsrc2cpg] Use temporary CPG location in tests

max-leuthaeuser:
Hopefully this fixes the csharpsrc2cpg test issues we see lately, e.g., https://github.com/joernio/joern/actions/runs/17240626125/job/48916564889


joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/testfixtures/CSharpCode2CpgFixture.scala
```diff
@@ -83,7 +84,10 @@ trait CSharpFrontend extends LanguageFrontend {
       .getOrElse(Config().withSchemaValidation(ValidationMode.Enabled))
 
   override def execute(sourceCodeFile: File): Cpg = {
-    new CSharpSrc2Cpg().createCpg(defaultConfig.withInputPath(sourceCodeFile.getAbsolutePath)).get
+    val cpgOutFile = FileUtil.newTemporaryFile(suffix = "cpg.bin")
+    FileUtil.deleteOnExit(cpgOutFile)
```
mpollmeier, line 88:
```suggestion
    // n.b. this will only delete the temp file if the jvm exits normally, but that's gotta be good enough for now..
    FileUtil.deleteOnExit(cpgOutFile)
```

max-leuthaeuser: Yeah, same for all other frontend test fixtures that uses this approach.
The files being temporary files should not result in any issues anyway.

PR #5639, opened 2025-09-03
Remove withErrorsToConsole from X2CpgFrontend.run().

ml86:
This resulted in exceptions that are not caught by the frontend specific
code to be logged twice which can be very confusing.
The other log location is in X2CpgMain.main() which i deem more
appropriate.


maltek (approved):
maybe add it back for the server mode (line 170 or https://github.com/joernio/joern/blob/54ac4775b5bc822c699a4ec870266d7e42a8931b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/server/FrontendHTTPServer.scala#L109), we don't seem to log the stack trace anywhere in that case

also `withErrorsToConsole` is unused after this change

PR #5640, opened 2025-09-03
markus/kotlinfix

ml86:
- **Stop iterating over map. Use List instead.**
- **Remove unnecessary field from GradleDepsInitScript case class.**
- **Handle FileAlreadyExistsException during dependency extraction.**

Fix for https://shiftleftinc.atlassian.net/browse/SEN-4407

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/DependencyResolver.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/DependencyResolver.scala
@@ -88,7 +88,7 @@ object DependencyResolver {
     GradleDependencies.get(projectDir, maybeProjectNameOverride, maybeConfigurationOverride) match {
       case dependencies if dependencies.exists(_._2.nonEmpty) =>
         val allDependencies = dependencies.flatMap(_._2)
-        Option(allDependencies.distinctBy(path => path.split(java.io.File.separatorChar).last))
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/DependencyResolver.scala
```diff
@@ -86,8 +86,9 @@ object DependencyResolver {
     val maybeProjectNameOverride   = params.forGradle.get(GradleConfigKeys.ProjectName)
     val maybeConfigurationOverride = params.forGradle.get(GradleConfigKeys.ConfigurationName)
     GradleDependencies.get(projectDir, maybeProjectNameOverride, maybeConfigurationOverride) match {
-      case dependenciesMap if dependenciesMap.values.exists(_.nonEmpty) =>
-        Option(dependenciesMap.values.flatten.toList.distinctBy(path => path.split(java.io.File.separatorChar).last))
+      case dependencies if dependencies.exists(_._2.nonEmpty) =>
+        val allDependencies = dependencies.flatMap(_._2)
+        Option(allDependencies.distinctBy(path => path.split(java.io.File.separatorChar).last))
```
maltek: there are weird platforms which have more than one valid separator char (on Windows both `/` and `\` work interchangeably), so

```suggestion
        Option(allDependencies.distinctBy(path => new java.io.File(path).getName))
```

PR #5641, opened 2025-09-04
Use SourceFiles.filterFile logic to also filter config files

johannescoetzee:
This PR introduces filtering for config files using the same exclude mechanism we currently use for `SourceFiles`. It's implemented here as an `isExcludedFile` method which can be overridden per-frontend if we want to (either to disable this completely or to use a custom filter), but is enabled by default here (which I'm happy to change if need be).

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/testfixtures/Py2CpgTestContext.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/testfixtures/Py2CpgTestContext.scala
@@ -25,7 +25,7 @@ class Py2CpgTestContext {
 
   private val codeAndFile       = mutable.ArrayBuffer.empty[Py2Cpg.InputPair]
   private var buildResult       = Option.empty[Cpg]
-  private val absTestFilePath   = "<absoluteTestPath>/"
+  private val absTestFilePath   = "absoluteTestPath/"
   private var enableFileContent = true
 
   def withEnabledFileContent(value: Boolean): Py2CpgTestContext = {
--- a/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/testfixtures/Py2CpgTestContext.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/testfixtures/Py2CpgTestContext.scala
@@ -1,6 +1,6 @@
 package io.joern.pysrc2cpg.testfixtures
 
-import io.joern.pysrc2cpg.Py2Cpg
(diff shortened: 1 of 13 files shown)


maltek (approved):

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ConfigFileCreationPass.scala
```diff
@@ -5,7 +5,7 @@ import io.shiftleft.codepropertygraph.generated.Cpg
 
 import java.nio.file.Path
 
-class ConfigFileCreationPass(cpg: Cpg) extends XConfigFileCreationPass(cpg) {
+class ConfigFileCreationPass(cpg: Cpg, config: Config) extends XConfigFileCreationPass(cpg, config = config) {
```
max-leuthaeuser, line 9: Thats just missing its import.

johannescoetzee: Hmm, I thought I compiled/tested everything, but missed this somehow. Should be fixed now

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/testfixtures/Py2CpgTestContext.scala
```diff
@@ -45,16 +45,14 @@ class Py2CpgTestContext {
   }
 
   def buildCpg: Cpg = {
+    val config = Py2CpgOnFileSystemConfig()
+      .withInputPath(absTestFilePath)
```
max-leuthaeuser, line 52: Almost missed that: please also add `.withOutputPath` or we run into the same problem as with csharpsrc2cpg in Windows: https://github.com/joernio/joern/pull/5635

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XConfigFileCreationPass.scala
```diff
@@ -71,6 +73,18 @@ abstract class XConfigFileCreationPass(cpg: Cpg, private val rootDir: Option[Str
...
+    val isAcceptedFile = SourceFiles.filterFile(
+      filePath.toAbsolutePath.toString,
+      rootPath,
+      Option(config.defaultIgnoredFilesRegex),
+      Option(config.ignoredFilesRegex),
+      Option(config.ignoredFiles)
+    )
+
+    !isAcceptedFile
+  }
```
ml86: I would not invert the logic and stick to the `isAccepted` logic of `filterFile`. Just easier to comprehend. But i can live with the current version as well.

ml86 (approved):

PR #5644, opened 2025-09-08
[swiftsrc2cpg] Refactor Swift type mapping to use ConcurrentHashMap

max-leuthaeuser:
Even if the compute function looks more complicated now this ensures that multiple concurrent updates to the inner Map are not lost and computed at most once (TrieMap actually uses a retry mechanic which results in additional performance overhead as we call the swift-demangle binary externally).

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
@@ -2,8 +2,8 @@ package io.joern.swiftsrc2cpg.utils
 
 import io.joern.swiftsrc2cpg.Config
 import io.joern.swiftsrc2cpg.astcreation.AstCreatorHelper
+import io.joern.swiftsrc2cpg.utils.ExternalCommand.*
 import io.joern.x2cpg.utils.Environment
-import ExternalCommand.*
 import io.shiftleft.semanticcpg.utils.FileUtil.PathExt
 import io.shiftleft.utils.IOUtils
 import org.slf4j.LoggerFactory
@@ -12,11 +12,11 @@ import versionsort.VersionHelper
 import java.io.{BufferedReader, InputStreamReader, StringReader}
 import java.nio.file.*
 import java.nio.file.attribute.BasicFileAttributes
-import scala.jdk.CollectionConverters.*
 import scala.collection.parallel.CollectionConverters.ImmutableSeqIsParallelizable
 import scala.collection.parallel.ExecutionContextTaskSupport
 import scala.collection.{immutable, mutable}
 import scala.concurrent.ExecutionContext
+import scala.jdk.CollectionConverters.*
 import scala.util.matching.Regex
 import scala.util.{Failure, Try, Using}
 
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
```diff
@@ -88,8 +88,23 @@ object SwiftTypesProvider {
     */
   case class ResolvedTypeInfo(typeFullname: Option[String], declFullname: Option[String], nodeKind: String)
 
-  type SwiftFileLocalTypeMapping = mutable.HashMap[(Int, Int), mutable.HashSet[ResolvedTypeInfo]]
-  type SwiftTypeMapping          = TrieMap[String, SwiftFileLocalTypeMapping]
+  /**   - [[java.util.concurrent.ConcurrentHashMap]] provides atomic compute and computeIfAbsent operations used here to
+    *     create-or-update entries and mutate the inner HashSet safely per key, without extra locks or retry loops.
+    *   - TrieMap lacks equivalent APIs with the same per-key atomicity semantics; achieving the same would require
+    *     CAS/retries or external synchronization, hurting clarity and throughput.
+    *   - ConcurrentHashMap is highly optimized and scales better under contention; in practice it outperforms
```
mpollmeier, line 95: if you happen to have a link to your research or benchmarking results at hand, would be good to include here, for the future reader...

mpollmeier (approved):

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
max-leuthaeuser: Sure!

A small benchmark can be done with https://github.com/migueldeicaza/SwiftGodot.
Before this PR on my machine: ~70 sec.
After: ~55 sec.
(done a few rounds, varies from run to run due to GC but thats expected)

- Raw throughput (single-thread or low contention): ConcurrentHashMap is typically faster. There are reports that note TrieMap’s absolute performance is slightly below JDK 8+ ConcurrentHashMap for simple get/put [1,2]. Even so TrieMap is designed for highly scalable concurrent insert/remove and shows good scaling characteristics in published benchmarks [2] we do not need this as concurrency is bound to the maximum number of processors available (work-stealing thread pool).
- Iteration/snapshots: TrieMap supports O(1) snapshots and consistent iteration; ConcurrentHashMap  provides weakly consistent iterators (no snapshot). Snapshots are not needed at all here.
- Progress guarantees: TrieMap is non-blocking/lock-free; ConcurrentHashMap is not lock-free. Under heavy contention this can positively affect tail latencies for updates but we do not need this. We want the exact opposite. If a key (node range) is not found calculating the entry should be blocking and done exactly once (because the swift-demangle binary is called).
- Memory/cache behavior: TrieMap is described as memory-efficient and cache-aware; ConcurrentHashMap structure evolved (tree bins since JDK 8) and is also well-optimized. Actual memory/usefulness is workload-dependent. I have not seen any significant differences for our use-case here, so thats not too important.

[1] https://stackoverflow.com/questions/29499381/what-is-a-triemap-and-what-is-its-advantages-disadvantages-compared-to-a-hashmap
[2] https://www.researchgate.net/publication/221643801_Concurrent_Tries_with_Efficient_Non-Blocking_Snapshots

mpollmeier: I actually meant in the code, but this is ok too :)

PR #5656, opened 2025-09-22
Fix bug in issue 5654 keyword final

jjzdxmd:
I create a issue#5654 in joern, and i fix the bug mentioned in it about the missing modifier "final"


maltek (approved):
thanks for the contribution!

PR #5658, opened 2025-09-24
[kotlin2cpg] Handle null pointer graceful and log.

ml86:
Handle a null pointer return from getDispatchReceiverParameter
gracefully so that not the hole file is dropped. When this happens we
log an error with some debug information. Hopefully enough to fix this
if the customer resubmits.

Issue which caused this: https://shiftleftinc.atlassian.net/browse/SEN-4407

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
@@ -240,9 +240,14 @@ class AstCreator(
     }
   }
 
-  private def logDebugWithTestAndStackTrace(message: String): Unit = {
+  protected def logDebugWithTestAndStackTrace(message: String): Unit = {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -240,11 +240,16 @@ class AstCreator(
     }
   }
 
-  private def logDebugWithTestAndStackTrace(message: String): Unit = {
+  protected def logDebugWithTestAndStackTrace(message: String): Unit = {
     val declString = debugScope.headOption.map(_.getText).getOrElse("Declaration scope empty")
     logger.debug(message + "\nIn declaration:\n" + declString + "\nStack trace to declaration:" + getStackTrace)
   }
 
+  protected def logWarnWithTestAndStackTrace(message: String): Unit = {
```
mpollmeier, line 248: code reuse anyone? 😏

mpollmeier (approved):

PR #5664, opened 2025-10-01
[c2cpg] Handle StackOverflowError

max-leuthaeuser:
A single problematic file should not crash the whole process. That is not a real fix but will gain is more insights what is going on if we encounter a file causing CDT to crash.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
@@ -294,9 +294,10 @@ trait AstForStatementsCreator { this: AstCreator =>
     val columnNumber = column(statement)
     // We only handle un-parsable macros here for now
     val isFromMacroExpansion = statement.getProblem.getNodeLocations.exists(_.isInstanceOf[IASTMacroExpansionLocation])
+    val code                 = statement.getRawSignature
+    val file                 = Paths.get(statement.getContainingFilename)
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -294,9 +294,10 @@ trait AstForStatementsCreator { this: AstCreator =>
     val columnNumber = column(statement)
     // We only handle un-parsable macros here for now
     val isFromMacroExpansion = statement.getProblem.getNodeLocations.exists(_.isInstanceOf[IASTMacroExpansionLocation])
+    val code                 = statement.getRawSignature
+    val file                 = Paths.get(statement.getContainingFilename)
     val asts = if (isFromMacroExpansion) {
-      new CdtParser(config, headerFileFinder, None, global)
-        .parse(statement.getRawSignature, Paths.get(statement.getContainingFilename)) match
+      new CdtParser(config, headerFileFinder, None, global).parse(code, file) match
```
mpollmeier: I'd really prefer having braces over significant whitespace... and that's how we do it normally...

mpollmeier (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
max-leuthaeuser: Intellij auto-complete bullshit - will fix.

PR #5665, opened 2025-10-02
improve DefaultSemantics definition for strncat

qkaiser:
From a signature perspective, `strncpy` and `strncat` are equivalent. strncat was missing the `(3, 3)` pair indicating that length is not sanitized.


maltek (approved):
looks good, thank you

PR #5671, opened 2025-10-08
[kotlin2cpg] Fix NoSuchElementException in SAM interface handling.

ml86:
When only partial type information is available, the expected type of a
lambda or function expression may not be available. This scenario was
not correctly handled which led to the NoSuchElementException.

Fix for https://shiftleftinc.atlassian.net/browse/SEN-4472

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForFunctionsCreator.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForFunctionsCreator.scala
@@ -528,11 +528,11 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) {
             call.getResolvedCallAtom
           }
 
-        resolvedCallAtom.map { callAtom =>
+        resolvedCallAtom.flatMap { callAtom =>
           callAtom.getCandidateDescriptor match {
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForFunctionsCreator.scala
```diff
@@ -528,11 +528,11 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) {
             call.getResolvedCallAtom
           }
 
-        resolvedCallAtom.map { callAtom =>
+        resolvedCallAtom.flatMap { callAtom =>
           callAtom.getCandidateDescriptor match {
             case samConstructorDesc: SamConstructorDescriptor =>
               // Lambda is wrapped e.g. `SomeInterface { obj -> obj }`
-              samConstructorDesc.getBaseDescriptorForSynthetic
+              Some(samConstructorDesc.getBaseDescriptorForSynthetic)
```
mpollmeier, line 535: and you're 100% sure `getBaseDescriptorForSynthetic` will never return `null`?

ml86: Yes, the return value of getBaseDescriptorForSynthetic is `T` and not `T?`.

PR #5682, opened 2025-10-20
[kotlin2cpg] Structured ClosureBindingId

TNSelahle:
Use structured closureBindingIds instead of UUIDs.

Fixes https://github.com/ShiftLeftSecurity/codescience/issues/8508

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/LambdaTests.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/LambdaTests.scala
@@ -25,7 +25,7 @@ class LambdaTests extends KotlinCode2CpgFixture(withOssDataflow = false, withDef
     "should contain a CLOSURE_BINDING node for the captured parameter with the correct props set" in {
       val List(cb) = cpg.closureBinding.l
       cb.evaluationStrategy shouldBe EvaluationStrategies.BY_REFERENCE
-      cb.closureBindingId shouldBe Some("f1.<lambda>0")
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/LambdaTests.scala
```diff
@@ -68,7 +79,23 @@ class LambdaTests extends KotlinCode2CpgFixture(withOssDataflow = false, withDef
...
+      cb.evaluationStrategy shouldBe EvaluationStrategies.BY_REFERENCE
+      cb.closureBindingId shouldBe Some("simple.pkg.Bar.foo.<lambda>0.x")
+
+      cb._refOut.size shouldBe 1
+    }
+
+    "should contain a CLOSURE_BINDING node for captured `this` with the correct props set" in {
+      val List(cb) = cpg.closureBinding.filter(_._methodParameterInViaRefOut.name.l == List("this")).l
+      cb.evaluationStrategy shouldBe EvaluationStrategies.BY_REFERENCE
+      cb.closureBindingId shouldBe Some("simple.pkg.Bar.foo.<lambda>0.this")
```
ml86, line 98: `x` and `this` should not be captured into the lambda function if they are not used.

TNSelahle: Ah, right. The existing implementation captures them, and so I added in the missing test cases without thinking about whether it's correct :D 

I'll see how I can ensure that they aren't captured.

PR #5685, opened 2025-10-22
[swiftsrc2cpg] Proper closure call handling

max-leuthaeuser:


joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/ClosureWithCompilerTests.scala
```diff
@@ -0,0 +1,258 @@
...
+          |	   print("Hey there, \(name).")
+          |	 }
+          |	 greetUser("Alex")
+          |
+          |	 let findSquare = { (num: Int) -> (Int) in
+          |    let square = num * num
+          |	   return square
+          |	 }
+          |	 let findSquareResult = findSquare(5)
+          |}""".stripMargin
```
ml86, line 49: Please split up the different examples.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/ClosureWithCompilerTests.scala
```diff
@@ -0,0 +1,258 @@
...
+
+      val cpg = codeWithSwiftSetup(testCode)
+
+      // compare ((Swift.String,Swift.String)->Swift.Bool)
+      val List(compareLocal) = cpg.local.nameExact("compare").l
+      compareLocal.typeFullName shouldBe "Swift.Function"
```
ml86: We cannot just use a generic `Swift.Function` as type as this would conflate more functions into a single hierarchy then necessary, resulting in higher computational effort during e.g. call resolution. So we should use separate types for all the different function signatures. Same goes for all other places where `Swift.Function` is used.

max-leuthaeuser: I just did the same thing as javasrc2cpg did where everything is `java.util.function.Function`.
Should I change it to incorparate the signature for differentations, e.g., here in this example `Swift.Function:(Swift.String,Swift.String)->Swift.Bool`?

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/ClosureWithCompilerTests.scala
max-leuthaeuser: I did not do that on purpose because every Swift compilation for such a snippet (+ its boilerplate project code) takes ~30 seconds on my Windows system. We test 5 closures here so that means 5 times 30 seconds which does not help when running the tests. On Mac and Linux it is no issue at all (~5 seconds).

ml86: Ok, in that case keep it like this but add comments indicating why this is not split up.

max-leuthaeuser: Added the comment. 👍

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/ClosureWithCompilerTests.scala
max-leuthaeuser: Signatures are now added to `Swift.Function` with surrounding `<...>`.

PR #5692, opened 2025-11-05
[swiftsrc2cpg] Call and type decl member/access rework

max-leuthaeuser:
This PR includes the following changes:
- no creation of locals from assignments to type decl members
- assignments to type decl members are now correctly attached to the constructor block
- no creation of members from methods/functions under type decls
- proper creation of call base for dynamic calls
- proper creation of implicit self access to members from identifiers

There is no way to determine whether a call is a call to a static method/function at the moment. The SwiftParser does not provide that information. I will extend the utils handling compiler information for this purpose. This would go beyond the scope of this PR (its already large enough). Will do a follow up PR.

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
@@ -64,8 +64,10 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
               val lhsAst = astForNode(dictElement.key)
               val rhsAst = astForNode(dictElement.value)
 
-              val lhsTmpNode            = Ast(identifierNode(dictElement, tmpName))
-              val lhsIndexAccessCallAst = createIndexAccessCallAst(dictElement, lhsTmpNode, lhsAst)
+              val lhsTmpNode = identifierNode(dictElement, tmpName)
+              scope.addVariableReference(tmpName, lhsTmpNode, Defines.Any, EvaluationStrategies.BY_REFERENCE)
+
+              val lhsIndexAccessCallAst = createIndexAccessCallAst(dictElement, Ast(lhsTmpNode), lhsAst)
 
               createAssignmentCallAst(
                 dictElement,
@@ -77,6 +79,7 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
           }
 
           val tmpNode = identifierNode(node, tmpName)
+          scope.addVariableReference(tmpName, tmpNode, Defines.Any, EvaluationStrategies.BY_REFERENCE)
 
           scope.popScope()
           localAstParentStack.pop()
@@ -231,7 +234,7 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
     callAst(callNode_, args, Option(baseAst))
   }
 
-  private def constructorInvocationBlockAst(expr: FunctionCallExprSyntax): Ast = {
+  private def astForConstructorInvocation(expr: FunctionCallExprSyntax): Ast = {
     // get call is safe as this function is guarded by isRefToConstructor
     val tpe = fullnameProvider.typeFullname(expr).get
     registerType(tpe)
@@ -241,9 +244,12 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
     scope.pushNewBlockScope(blockNode_)
 
     val tmpNodeName  = scopeLocalUniqueName("tmp")
-    val tmpNode      = identifierNode(expr, tmpNodeName, tmpNodeName, tpe)
-    val localTmpNode = localNode(expr, tmpNodeName, tmpNodeName, tpe)
+    val localTmpNode = localNode(expr, tmpNodeName, tmpNodeName, tpe).order(0)
+    diffGraph.addEdge(blockNode_, localTmpNode, EdgeTypes.AST)
     scope.addVariable(tmpNodeName, localTmpNode, tpe, VariableScopeManager.ScopeType.BlockScope)
+
+    val tmpNode = identifierNode(expr, tmpNodeName, tmpNodeName, tpe)
+    scope.addVariableReference(tmpNodeName, tmpNode, tpe, EvaluationStrategies.BY_SHARING)
 
     val allocOp          = Operators.alloc
     val allocCallNode    = callNode(expr, allocOp, allocOp, allocOp, DispatchTypes.STATIC_DISPATCH)
@@ -274,10 +280,9 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
 
     val retNode = identifierNode(expr, tmpNodeName, tmpNodeName, tpe)
     scope.addVariableReference(tmpNodeName, retNode, tpe, EvaluationStrategies.BY_SHARING)
-    val retAst = Ast(retNode)
 
     scope.popScope()
-    Ast(blockNode_).withChildren(Seq(assignmentAst, constructorCallAst, retAst))
+    Ast(blockNode_).withChildren(Seq(assignmentAst, constructorCallAst, Ast(retNode)))
   }
 
   private def astForFunctionCallExprSyntax(node: FunctionCallExprSyntax): Ast = {
@@ -299,7 +304,7 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
           val memberCode = code(m.declName)
           handleCallNodeArgs(node, astForNode(m.base.get), memberCode)
         case other if isRefToConstructor(node, other) =>
-          constructorInvocationBlockAst(node)
+          astForConstructorInvocation(node)
         case other if isRefToClosure(node, other) =>
           astForClosureCall(node)
         case declReferenceExprSyntax: DeclReferenceExprSyntax if code(declReferenceExprSyntax) != "self" =>


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -164,16 +164,23 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
 
   protected def astForIdentifier(node: SwiftNode): Ast = {
     val identifierName = code(node)
-    val identNode      = identifierNode(node, identifierName)
-    val variableOption = scope.lookupVariable(identifierName)
-    val tpe = variableOption match {
-      case Some((_, variableTypeName)) if variableTypeName != Defines.Any => variableTypeName
-      case None if identNode.typeFullName != Defines.Any                  => identNode.typeFullName
-      case _                                                              => Defines.Any
+    dynamicInstanceTypeStack.headOption match {
```
maltek: From some quick testing, local variables can shadow member variables. So I think we have to weave in the member variables in the right level in the regular scope lookup.

Unlike in Java/Scala, inner types don't have access to an instance (or its members) of its outer types - but these members still shadow locals/globals from an even larger scope, and the compiler rejects their use. So in the end, we don't need to worry about getting anything right if there is a member in an outer type - it's impossible.


```swift
let x = "a"
class Outer {

    let x = "b"
    var inner = Inner()

    class Inner {

        var x = "c"

        func foo() -> String {
            let x = "d"
            return x
        }

    }
}
print(Outer().inner.foo())
```

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstNodeBuilder.scala
```diff
@@ -150,18 +153,19 @@ trait AstNodeBuilder(implicit withSchemaValidation: ValidationMode) { this: AstC
         t
       }
       .getOrElse(name match {
-        case "this" | "self" | "Self" => typeHintForThisExpression().headOption.getOrElse(Defines.Any)
+        case "this" | "self" | "Self" => typeHintForSelfExpression().headOption.getOrElse(Defines.Any)
```
maltek: I don't think `this` has any special meaning in Swift?

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/CallTests.scala
```diff
@@ -0,0 +1,75 @@
...
+      fooCall.signature shouldBe "()->()"
+      val List(fooCallReceiver) = fooCall.receiver.isIdentifier.l
+      fooCallReceiver.name shouldBe "self"
+      fooCallReceiver.typeFullName shouldBe "SwiftTest.Foo"
+
+      val List(barCall) = cpg.call.nameExact("bar").l
+      barCall.methodFullName shouldBe "SwiftTest.Foo.bar:()->Swift.String"
+      barCall.signature shouldBe "()->Swift.String"
+      val List(barCallReceiverCall) = barCall.receiver.fieldAccess.l
+      barCallReceiverCall.typeFullName shouldBe "Swift.Function<()->Swift.String>"
```
maltek: This uses our 'calling conventions' for the receiver for dynamically typed languages, when it should use the one for statically typed languages. In one case it's just `self`, in the other a field access `self.bar`

@ml86 will know more about the details, but I think the root cause for the different convention is the question whether we can statically set up bindings for the `TypeDecl` or not.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
maltek: this is still wrong for closures that access a member of `self` - that's why I mentioned weaving in the member definitions into the scope.

E.g. this here prints `b`:

```swift
let x = "a"
class MyType {
  var x = "b"
  func foo() -> String {
    func closure() -> String {
      return x
    }
    return closure()
  }
}
print(MyType().foo())
```

max-leuthaeuser: Yes right, missed that. It's part of the scope now (VariableScopeManager).

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -164,16 +164,39 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
     val identifierName = code(node)
-    val identNode      = identifierNode(node, identifierName)
-    val variableOption = scope.lookupVariable(identifierName)
-    val tpe = variableOption match {
-      case Some((_, variableTypeName)) if variableTypeName != Defines.Any => variableTypeName
-      case None if identNode.typeFullName != Defines.Any                  => identNode.typeFullName
-      case _                                                              => Defines.Any
+    // scoping rules for self access:
+    // 1) if identifier is in local method scope it shadows everything else:
+    if (scope.variableIsInMethodScope(identifierName)) {
```
maltek: afaics, the handling of case 1 and 3 is identical

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/SimpleAstCreationPassTest.scala
```diff
@@ -201,6 +201,230 @@ class SimpleAstCreationPassTest extends AstSwiftSrc2CpgSuite {
...
+      val List(bAccess) = cpg.call.nameExact("handleF").argument.fieldAccess.l
+      bAccess.code shouldBe "self.f"
+      inside(bAccess.argument.l) { case List(id: Identifier, fieldName: FieldIdentifier) =>
+        id.name shouldBe "self"
+        id.typeFullName shouldBe "Test0.swift:<global>.Foo"
+        fieldName.canonicalName shouldBe "f"
+      }
+    }
+
+    "have correct structure for implicit self access from within nested functions" in {
```
maltek, line 279: in the closure tests (plural) could you also assert that the closure bindings are set up correctly?

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -164,16 +164,26 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
-      case _                                                              => Defines.Any
+    if (scope.variableIsInTypeDeclScope(identifierName)) {
+      // we found it as member of the surrounding type decl
+      // (Swift does not allow to access any member / function of the outer class instance)
+      val tpe      = typeForSelfExpression()
+      val selfNode = identifierNode(node, "self", "self", tpe)
+      scope.addVariableReference("self", selfNode, selfNode.typeFullName, EvaluationStrategies.BY_REFERENCE)
+      fieldAccessAst(node, node, Ast(selfNode), s"self.$identifierName", identifierName, tpe)
+    } else {
+      // otherwise it must come from the outer scope somewhere
```
maltek:
```suggestion
      // otherwise it must come from a variable (potentially captured from an outer scope)
```

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/CallTests.scala
```diff
@@ -0,0 +1,75 @@
...
+          |}
+          |""".stripMargin
+      val cpg = codeWithSwiftSetup(testCode)
+
+      val List(fooCall) = cpg.call.nameExact("foo").l
+      fooCall.methodFullName shouldBe "SwiftTest.Foo.foo:()->()"
+      fooCall.signature shouldBe "()->()"
+      val List(fooCallReceiver) = fooCall.receiver.isIdentifier.l
+      fooCallReceiver.name shouldBe "self"
+      fooCallReceiver.typeFullName shouldBe "SwiftTest.Foo"
```
ml86, line 77: Dispatch type should be tested here.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/CallTests.scala
ml86: I agree with @maltek this representation is wrong unless there is a very specific reason why we cant use the same call representation as for other statically typed languages. In the representation you would have just `this` as receiver and `bar` as name. The name property test seems to be missing here anyway.

max-leuthaeuser: Has been fixed, guess you reviewed an older commit on this PR

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/CallTests.scala
```diff
@@ -0,0 +1,132 @@
...
+          |  func bar() -> String { return "" }
+          |  func main() {
+          |    foo()
+          |    self.bar()
+          |  }
+          |}
+          |""".stripMargin
+      val cpg = codeWithSwiftSetup(testCode)
+
+      val List(fooCall) = cpg.call.nameExact("foo").l
```
ml86, line 70: Please also check the name property of call nodes

max-leuthaeuser: But its right there? `cpg.call.nameExact("foo").l`

I added tests for calls to functions from extensions and protocols.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/CallTests.scala
```diff
@@ -111,6 +113,148 @@ class CallTests extends SwiftCompilerSrc2CpgSuite {
...
+      fooCallReceiver.typeFullName shouldBe "SwiftTest.Foo"
+
+      val List(barCall) = cpg.call.nameExact("bar").l
+      barCall.methodFullName shouldBe "SwiftTest.Foo.bar:()->()"
+      barCall.signature shouldBe "()->()"
+      barCall.dispatchType shouldBe DispatchTypes.DYNAMIC_DISPATCH
+
+      val List(barCallReceiverCall) = barCall.receiver.isIdentifier.l
+      barCallReceiverCall.name shouldBe "self"
+      barCallReceiverCall.typeFullName shouldBe "SwiftTest.Foo"
```
ml86: Extension methods are statically dispatched in Swift.

max-leuthaeuser: We are talking about Extensions (https://docs.swift.org/swift-book/documentation/the-swift-programming-language/extensions/) here. Are you sure?
E.g.,
```
extension Int {
    mutating func square() {
        self = self * self
    }
}
var someInt = 3
someInt.square()
```

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/CallTests.scala
ml86: Ups, overlooked that.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/CallTests.scala
ml86: Yes those extensions. Mutability of `self` is unrelated.

ml86: Why did you cite the `mutating` example? Maybe we have a miss understating here.

max-leuthaeuser: That was just an example, did not pick it on purpose.

Ok, then we have to model these extension calls as static calls. Currently there is no way to detect this as the swift-parser AST does not carry this information. I need to extend the compiler info queries to be able to detect static calls reliably.
This is out of the scope here (this PR is already large). I will change these tests to `pendingUntilFixed`.

ml86: Ok, doing it as a followup is fine. It is already a large PR and it improves the current situation a lot in any case.

PR #5695, opened 2025-11-12
Add a validator pass

bbrehm:
This is an initial validator pass for CPGs, post-frontend.

This PR activates the validator for unit-tests in some frontends. The validator can be set to "warn-only" or to throw an exception on encountering issues.

The actual validator code is intentionally kept very stupid and simple. This is because we will only learn what we want of it by using it. Less stupid code exists (I wrote that before integrating into joern), but we should start with this until we reach a point where unit-tests (and then sptests) successfully pass the graph validation.

From the integration, you see that c2cpg, csharpsrc2cpg, gosrc2cpg, jssrc2cpg, rubysrc2cpg, kotlin2cpg all cause validation errors in the unit tests.

The frontends javasrc2cpg and pysrc2cpg currently pass.

I don't know yet about swiftsrc2cpg, using this PR to find out (local build troubles).

I will add the other frontends in due time.

In order to better debug, consider setting a breakpoint at the noted line of code; and consider "arming" the validator by setting `throwOnError=true`.

### DIFF
--- a/console/src/main/scala/io/joern/console/BridgeBase.scala
+++ b/console/src/main/scala/io/joern/console/BridgeBase.scala
@@ -79,7 +79,7 @@ trait BridgeBase extends InteractiveShell with ScriptExecution with PluginHandli
         .valueName("'import Int.MaxValue'")
         .unbounded()
         .optional()
-        .action((x, c) => c.copy(runBefore = c.runBefore :+ x))
+        .action((x, c) => c.copy(runBefore = c.runBefore ++ x.linesIterator.toIndexedSeq))
         .text("given code will be executed on startup - this may be passed multiple times")
 
       opt[String]("runAfter")
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForFunctionsCreator.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForFunctionsCreator.scala
@@ -328,7 +328,7 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
       .possibleTypes(possibleTypes)
     diffGraph.addEdge(getParentTypeDecl, memberNode_, EdgeTypes.AST)
     addModifier(functionNode, func.json)
-    Ast(functionNode)
+    Ast()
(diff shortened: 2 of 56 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/validation/validation.scala
```diff
@@ -0,0 +1,78 @@
...
+        case m: nodes.TypeDecl =>
+          if (cpg.typeDecl.fullNameExact(m.fullName).size != 1) {
+            registerViolation(
+              "fullname uniqueness typedecl",
+              s"Fullname ${m}  - ${m.fullName} should occur at most oncee"
+            )
+          }
+        case _ =>
+      }
+    }
```
ml86, line 125: Put the 5 different tests into individual methods.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/validation/validation.scala
```diff
@@ -0,0 +1,78 @@
...
+    val tmp = getViolationsCompressed()
```
ml86: Better name than `tmp` please. I suggest `violations`.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/validation/validation.scala
```diff
@@ -0,0 +1,78 @@
...
+import io.shiftleft.semanticcpg.language.*
+import org.slf4j.{Logger, LoggerFactory}
+
+class ValidationError(msg: String) extends RuntimeException(msg) {}
+
+abstract class AbstractValidator(cpg: Cpg) extends CpgPass(cpg) {
+  val violations: mutable.ArrayBuffer[(String, String)] = mutable.ArrayBuffer[(String, String)]()
+  def registerViolation(key: String, msg: String): Unit = {
+    // when debugging, add the breakpoint here
+    violations.addOne((key, msg))
```
ml86: Use enum as key.

bbrehm: The point here is to allow people to quick-and-dirty add more validation tests; and once we have a good test package, then we can optimize that to run fast.

PR #5698, opened 2025-11-13
[php2cpg] fix invalid offset for non utf-8 encoded files

TNSelahle:
Fix invalid offset for non-UTF-8 encoded files.

Fixes https://github.com/ShiftLeftSecurity/codescience/issues/8500

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/OffsetTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/OffsetTests.scala
@@ -17,7 +17,9 @@ class OffsetTests extends PhpCode2CpgFixture {
         |// ääú
         |
         |print("hello");
-        |""".stripMargin, "test.php", StandardCharsets.ISO_8859_1
+        |""".stripMargin,
+      "test.php",
+      StandardCharsets.ISO_8859_1
     ).withConfig(contentEnabled)
 
     "have correct offsets" in {
@@ -47,7 +49,9 @@ class OffsetTests extends PhpCode2CpgFixture {
         |// ääú
         |
         |print("hello");
-        |""".stripMargin, "test.php", StandardCharsets.UTF_8
+        |""".stripMargin,
+      "test.php",
+      StandardCharsets.UTF_8
     ).withConfig(contentEnabled)
 
     "have correct offsets" in {
--- a/joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/utils/TestCodeWriter.scala
+++ b/joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/utils/TestCodeWriter.scala
@@ -29,7 +29,7 @@ trait TestCodeWriter {
     codeFileNamePairs.append(Code(code, Option(fileName), None))
     this
   }
-  
+
   def moreCode(code: String, fileName: String, charset: Charset): this.type = {
     codeFileNamePairs.append(Code(code, Option(fileName), Option(charset)))
     this


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -45,9 +31,9 @@ trait AstCreatorHelper(disableFileContent: Boolean)(implicit withSchemaValidatio
   override protected def offset(phpNode: PhpNode): Option[(Int, Int)] = {
     Option.when(!disableFileContent) {
       val startPos =
-        new String(fileContent.get.getBytes.slice(0, phpNode.attributes.startFilePos), StandardCharsets.UTF_8).length
+        new String(fileContentBytes.slice(0, phpNode.attributes.startFilePos), fileCharset).length
       val endPos =
-        new String(fileContent.get.getBytes.slice(0, phpNode.attributes.endFilePos), StandardCharsets.UTF_8).length
+        new String(fileContentBytes.slice(0, phpNode.attributes.endFilePos), fileCharset).length
```
ml86, line 36: Please use this `String` constructor to avoid the extra slice operation:
```
public String​(byte[] bytes,
              int offset,
              int length,
              [String](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html) charsetName)
```

joern-cli/frontends/php2cpg/build.sbt
```diff
@@ -15,11 +15,12 @@ dependsOn(
...
-  "com.lihaoyi"       %% "ujson"               % Versions.upickle,
-  "io.shiftleft"      %% "codepropertygraph"   % Versions.cpg,
-  "com.github.sh4869" %% "semver-parser-scala" % Versions.semverParser,
-  "org.scalatest"     %% "scalatest"           % Versions.scalatest % Test
+  "com.lihaoyi"       %% "upickle"                % Versions.upickle,
+  "com.lihaoyi"       %% "ujson"                  % Versions.upickle,
+  "io.shiftleft"      %% "codepropertygraph"      % Versions.cpg,
+  "com.github.sh4869" %% "semver-parser-scala"    % Versions.semverParser,
+  "org.scalatest"     %% "scalatest"              % Versions.scalatest % Test,
+  "com.github.albfernandez" % "juniversalchardet" % Versions.juniversalchardet
```
ml86, line 23: There is an open [ticket](https://github.com/albfernandez/juniversalchardet/issues/29) on github which claims that this library does not have support for ISO8859-1. Why does this still seem to work?

ml86 (changes requested):

joern-cli/frontends/php2cpg/build.sbt
TNSelahle: Hmm, that's actually true. I played around more with it and found that it detects the code as `Windows-1255`. I also gave it more characters from the ISO-8859-1 encoding to see if it'll guess differently but it still guesses `Windows-1255`. The full list of detectable encodings is on https://github.com/albfernandez/juniversalchardet/blob/b43c5b5e9b4519cfb7ae7702a305572212f7a11b/README.md. 

Since most files are in UTF-8, we could opt to first try decoding as UTF-8 then if there are any errors, resort to guessing using the encoding as a best-effort, using the library. What's your thoughts?

ml86: Its unclear to me how we would detect errors for the first attempt of loading as UTF-8. I guess using juniversalchardet is fine for now as it solves the problem at hand and does not seem to cause problems in the standard UTF-8 input case.

ml86 (approved):

joern-cli/frontends/php2cpg/build.sbt
TNSelahle: When decoding, you can have it throw an exception if there were decoding issues. The current setup just replaces the problematic characters. 

I'll merge this one for now though.

PR #5700, opened 2025-11-19
[console] fix line splitting in --runBefore

maltek:
before

    ./joern --runBefore '//> using dep org.apache.commons:commons-exec:1.5.0
    //> using dep io.methvin:directory-watcher:0.19.1'

would fail because the version number for commons-exec would not end at "1.5.0", but instead also include the whole second line.


mpollmeier (approved):
I was first afraid this would break other multiline 

```
./joern --runBefore 'def foo = bar
def bar = 40'
```

but that's not the case - so 👍🏻

PR #5707, opened 2025-11-27
[swiftsrc2cpg] Extension handling rework

max-leuthaeuser:
- Extension methods are now processed separately, and extension inheritance information is collected and passed through a new global structure. The AST generation for extensions now creates function nodes with a `<extension>` tag in their fullNames for each function in the extension and binds them appropriately. No TypeDecl is generated for the extension itself.
- The `inheritsFrom` logic was updated to work for both regular type declarations and extensions, and inheritance relationships for extensions are now registered globally for later processing in a separate pass.
- The logic for determining enclosing TypeDecl fullNames was improved in several places, ensuring that references to `self` and initializer return types are correct.
- Updated the frontend pass pipeline to include new passes for extension calls and inheritance, and reordered some passes for correctness (TypeDecl and Type generation for used types).

Needs a fix at: https://github.com/ShiftLeftSecurity/codescience/blob/c8d155ffd34ccb6d1f8296a5722a9edeaa8a5c30/securityprofile/src/main/scala/io/shiftleft/securityprofile/passes/finalcustom/swiftsrc/generic/NSURLConnCertCheckPass.scala#L36
Will do that in a PR at CS.

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/CallTests.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/CallTests.scala
@@ -208,7 +208,7 @@ class CallTests extends SwiftCompilerSrc2CpgSuite {
       val cpg = codeWithSwiftSetup(testCode)
 
       val List(negateCall) = cpg.call.nameExact("negate").l
-      negateCall.methodFullName shouldBe "Swift.Swift.SignedNumeric<extension>.negate:()->()"
+      negateCall.methodFullName shouldBe "Swift.SignedNumeric<extension>.negate:()->()"
     }
 
     "be correct for calls to generic functions from extensions with compiler support" in {
@@ -233,8 +233,8 @@ class CallTests extends SwiftCompilerSrc2CpgSuite {
       val cpg = codeWithSwiftSetup(testCode)
 
       val List(barMethodInt, barMethodDouble) = cpg.method.nameExact("bar").l
-      barMethodInt.fullName shouldBe "SwiftTest.SwiftTest.Foo<AwhereA==Swift.Int><extension>.bar:()->()"
-      barMethodDouble.fullName shouldBe "SwiftTest.SwiftTest.Foo<AwhereA==Swift.Double><extension>.bar:()->()"
+      barMethodInt.fullName shouldBe "SwiftTest.Foo<AwhereA==Swift.Int><extension>.bar:()->()"
+      barMethodDouble.fullName shouldBe "SwiftTest.Foo<AwhereA==Swift.Double><extension>.bar:()->()"
 
       val List(intBarCall, doubleBarCall) = cpg.call.nameExact("bar").l
       intBarCall.methodFullName shouldBe barMethodInt.fullName
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
@@ -573,6 +573,8 @@ case class SwiftTypesProvider(config: Config, parsedSwiftInvocations: Seq[Seq[St
     */
   private val ExtensionNameRegex: Regex = """^\(extension\sin\s([^)]+)\):(.*)""".r
 
+  private val ExtensionInSignatureRegex: Regex = """\(extension in ([^)]+)\):""".r
+
   /** Calculates a demangled declaration fullName from a mangled Swift declaration name.
     *
     * This method demangles the Swift declaration name, applies special transformations for member names and extensions,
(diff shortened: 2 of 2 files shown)


joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/CallTests.scala
```diff
@@ -170,32 +167,66 @@ class CallTests extends SwiftCompilerSrc2CpgSuite {
...
-        barCall.methodFullName shouldBe "SwiftTest.Foo.bar:()->()"
-        barCall.signature shouldBe "()->()"
-        barCall.dispatchType shouldBe DispatchTypes.STATIC_DISPATCH // Should be STATIC_DISPATCH for extension methods, but it is not
+      val List(fooCall) = cpg.call.nameExact("foo").l
+      fooCall.methodFullName shouldBe fooMethod.fullName
+      fooCall.signature shouldBe "()->()"
+      fooCall.dispatchType shouldBe DispatchTypes.STATIC_DISPATCH
+      val List(fooCallReceiver) = fooCall.receiver.isIdentifier.l
+      fooCallReceiver.name shouldBe "self"
+      fooCallReceiver.typeFullName shouldBe "SwiftTest.Foo"
```
ml86: `self` is should not be given as receiver. It should only be an argument. Receiver relation in the CPG is only valid for dynamically dispatched calls.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/CallTests.scala
```diff
@@ -170,32 +167,66 @@ class CallTests extends SwiftCompilerSrc2CpgSuite {
...
-        barCallReceiverCall.typeFullName shouldBe "SwiftTest.Foo"
-      }
+      val List(barCall) = cpg.call.nameExact("bar").l
+      barCall.methodFullName shouldBe barMethod.fullName
+      barCall.signature shouldBe "()->()"
+      barCall.dispatchType shouldBe DispatchTypes.STATIC_DISPATCH
+
+      val List(barCallReceiverCall) = barCall.receiver.isIdentifier.l
+      barCallReceiverCall.name shouldBe "self"
+      barCallReceiverCall.typeFullName shouldBe "SwiftTest.Foo"
```
ml86: Same as above.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/ExtensionMultiModuleTests.scala
```diff
@@ -0,0 +1,64 @@
...
+      .moreCode(moduleBCode, "SwiftTest/Sources/ModuleB/FooExt.swift")
+
+  "ExtensionMultiModuleTests" should {
+
+    "create correct extension method fullNames" in {
+      val List(fooTypeDecl) = cpg.typeDecl.nameExact("Foo").l
+      fooTypeDecl.fullName shouldBe "ModuleA.Foo"
+
+      val List(foo) = cpg.method.nameExact("foo").l
+      foo.fullName shouldBe "ModuleA.Foo<extension>.foo:()->()"
```
ml86, line 52: Why does `foo` not have the full name `ModuleA.ModuleA.Foo<extension>.foo:()->()`?
Seems like you use a different naming pattern for extension in the same module as the extended type. I think a uniform naming schema is better or am I missing some reason other than the shorter full name?

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/inheritance/ClassExtensionTests.scala
```diff
@@ -80,10 +77,20 @@ class ClassExtensionTests extends SwiftSrc2CpgSuite {
...
+
+    /** TODO: Re-enable once extension methods are properly accessible via EXTENSION_BLOCK
+      * typeDeclFoo.boundMethod.fullName.l shouldBe List(
+      * "Test0.swift:<global>.Foo.init:(paramA:Swift.String,paramB:Swift.Int)->Test0.swift:<global>.Foo",
+      * "Test0.swift:<global>.Foo.someFunc:()->ANY", "Test0.swift:<global>.Foo.someMethod:()->ANY",
+      * "Test0.swift:<global>.Foo.square:()->ANY", "Test0.swift:<global>.Foo.<clinit>:()->Test0.swift:<global>.Foo",
+      * "Test0.swift:<global>.Foo<extension>.someOtherFunc:()->ANY" )
+      */
+
+    typeDeclFoo.inheritsFromTypeFullName.sorted.l shouldBe List("AnotherProtocol", "Bar", "SomeProtocol")
```
ml86, line 90: These are not full names. We need to talk about your type full name concept. As those contain file names they do not seem appropriate for linking.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ExtensionCallPass.scala
```diff
@@ -0,0 +1,49 @@
...
+  * it:
+  *   - updates the call's dispatch type to STATIC_DISPATCH
+  *   - replaces the call's methodFullName with the mapped extension method full name
+  *
+  * @param cpg
+  *   the code property graph to operate on
+  * @param extensionFullNameMapping
+  *   mapping from compiler generated methodFullName -> CPG extension methodFullName (for uniqueness)
+  */
+class ExtensionCallPass(cpg: Cpg, extensionFullNameMapping: Map[String, String])
```
ml86, line 24: I think the approach of fixing up the call sites later has reached its end because as mentioned about the receiver edges are wrong for static calls. Of course we could now remove them here but a pre-pass which collects the necessary information to directly generate the correct calls seems more appropriate.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/inheritance/ClassExtensionTests.scala
max-leuthaeuser: Will be handled in a follow-up PR.

PR #5712, opened 2025-12-02
go 1.25

mpollmeier:
context: go 1.18 isn't available for download any longer, and that blocks other PRs:
    ```
    Setup go version spec 1.18
    Attempting to download 1.18...
    matching 1.18...
    Not found in manifest.  Falling back to download directly from Go
    Install from dist
    Acquiring go1.18.10 from https://storage.googleapis.com/golang/go1.18.10.darwin-arm64.tar.gz
    Error: Failed to download version 1.18: Error: Unexpected HTTP response: 403
    ```
https://github.com/joernio/joern/actions/runs/19851681566/job/56879725873?pr=5711

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -21,7 +21,7 @@ jobs:
         if: matrix.os == 'macos-latest'
         run: brew install php
       - name: Set up Go
-        uses: actions/setup-go@v5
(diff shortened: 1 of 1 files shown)


max-leuthaeuser (approved):
If that works: needs the same update in all other workflows too.

PR #5715, opened 2025-12-03
[php2cpg] fix: binding table fix

TNSelahle:
Removing unnecessary entries from the binding table. 

Fixes https://github.com/ShiftLeftSecurity/codescience/issues/8573

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForTypesCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForTypesCreator.scala
@@ -1,6 +1,7 @@
 package io.joern.php2cpg.astcreation
 
 import io.joern.php2cpg.astcreation.AstCreator.{NameConstants, TypeConstants}
+import io.joern.php2cpg.parser.Domain
 import io.joern.php2cpg.parser.Domain.*
 import io.joern.php2cpg.utils.TypeScope
 import io.joern.x2cpg.{Ast, Defines, ValidationMode}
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -334,11 +334,12 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this:
 
     val bindings = typeDecl match {
       case Some(typeDecl) =>
-        (anonymousMethodAsts ++ otherBodyStmts).flatMap(_.root).collect { case method: NewMethod =>
-          val bindingNode = NewBinding().name(method.name).signature("")
-          diffGraph.addNode(bindingNode)
-          diffGraph.addEdge(typeDecl, bindingNode, EdgeTypes.BINDS)
-          diffGraph.addEdge(bindingNode, method, EdgeTypes.REF)
+        otherBodyStmts.flatMap(_.root).collect {
+          case method: NewMethod if method.name != "<global>" =>
```
ml86: Having a the `<global>` method in the bodyStmts of a class already seems wrong. Please fix that instead of filtering this method out here.

ml86 (changes requested):

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForTypesCreator.scala
TNSelahle: @ml86 this seems to be due to how the parser represents this. 
image

There is a `PhpClassLikeStmt` for `<global>`, and in that, a `PhpMethodDecl` for `<global>`

ml86: Concluded in zoom that we want to go ahead with the current status as changing it would be a too big refactoring for now.

PR #5716, opened 2025-12-03
[php2cpg] feat: represent yield as return statement

TNSelahle:
Represent `yield`-like statements as return statements. This is to aid dataflow tracking.

Fixes https://github.com/ShiftLeftSecurity/codescience/issues/8572

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
@@ -1064,7 +1064,7 @@ class ControlStructureTests extends PhpCode2CpgFixture {
           inside(kv.astChildren.l) { case (literal: Literal) :: (identifier: Identifier) :: Nil =>
             literal.code shouldBe "1"
             literal.typeFullName shouldBe "int"
-            
+
             identifier.name shouldBe "x"
           }
         }


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -770,9 +770,9 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
         s"yield ${maybeKey.map(_.rootCodeOrEmpty).getOrElse("")}${maybeVal.map(_.rootCodeOrEmpty).getOrElse("")}".trim
     }
 
-    val yieldNode = controlStructureNode(expr, ControlStructureTypes.YIELD, code)
+    val node = returnNode(expr, code)
 
-    Ast(yieldNode)
+    Ast(node)
       .withChildren(maybeKey.toList)
       .withChildren(maybeVal.toList)
```
TNSelahle: @ml86 since one can yield a key/value pair in PHP, I'm not sure if this is the best way to represent it.

ml86: That is invalid format. A return node can only have one AST child. I suggest you represent this as a pair. If PHP has a builtin pair type the frontend should already have means to represent this in the CPG and you might be able to reuse that functionality with only minor adjustments.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
```diff
@@ -1056,18 +1052,14 @@ class ControlStructureTests extends PhpCode2CpgFixture {
...
-          key.code shouldBe "1"
-          key.lineNumber shouldBe Some(3)
-
-          value.name shouldBe "x"
-          value.code shouldBe "$x"
-          value.lineNumber shouldBe Some(3)
+        inside(returnNode.astChildren.l) { case List(kv: Call) =>
+          kv.code shouldBe "1 => $x"
+          kv.lineNumber shouldBe Some(3)
+          kv.methodFullName shouldBe "<operator>.doubleArrow"
```
TNSelahle, line 1062: Also check the children of this doubleArrow

PR #5719, opened 2025-12-08
[pysrc2cpg] Avoid tmp variable creation in for non complex receivers.

ml86:
So far the python frontend emitted tmp variables in the context of call
site lowering if the `<x>` in a `<x>.y()` call was anything other than
an identifier/name. Now we also avoid emitting tmp variables if `<x>` is
a member access chain. E.g. `x.y.z()`. This leads to less clutter and
better alias handling.


maltek (approved):
Maybe rename `mayHaveSideEffects` to `assumeNoSideEffects` or so? In principle, python does support computed getters/setters that can have side effects. To be clear, I'm not saying we should revisit our long-standing assumption that such getters/setters actually behave like a simple field access. Just that we should spell this assumption out more clearly here.

PR #5722, opened 2025-12-15
[swiftsrc2cpg] Enhance inheritance handling for Swift types and xcodebuild support

max-leuthaeuser:
We use proper fullNames now (if we have the compiler support).

Also:
- removed unused dependency from test snippets (to make them compile with both the 6.1.x and 6.2.x swift toolchain without the need to specify the SDK path).
- cleaned / refactored GsonTypeInfoReader (handles both 6.1.x and 6.2.x inheritance JSON specs).
- adjusted tests
- handle xcodebuild output correctly (for https://github.com/ShiftLeftSecurity/security-team/blob/master/vuln_iosapp/)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/GsonTypeInfoReader.scala
```diff
@@ -235,6 +235,10 @@ object GsonTypeInfoReader {
       .orElse(safePropertyValue(obj, "interface_type"))
   }
 
+  private def isInBuildFolder(filename: String): Boolean = {
+    filename.contains("/.build/") || filename.contains("\\.build\\") || filename.contains("/Build/")
```
maltek: at least on mac/Windows those comparisons should be case-insensitive

ml86 (approved):

PR #5724, opened 2025-12-18
Require JDK13+ due to the use of `FileSystems.newFileSystem(Path)`

gacevedo:
The method `FileSystems.newFileSystem(Path)` was added in Java 13. Updated the Java version check in `build.sbt` to expect `13` instead of `11`, and added a new assertion in `project/meta-build.sbt` to fail earlier if JDK13 or higher is not in use.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -70,7 +70,7 @@ ThisBuild / compile / javacOptions ++= Seq(
 ThisBuild / scalacOptions ++= Seq(
   "-deprecation", // Emit warning and location for usages of deprecated APIs.
   "--release",
-  "13",
+  "11",
   "-Wshadow:type-parameter-shadow"
 )
 


build.sbt
```diff
@@ -59,18 +59,18 @@ ThisBuild / libraryDependencies ++= Seq(
 ThisBuild / compile / javacOptions ++= Seq(
   "-g", // debug symbols
   "-Xlint",
-  "--release=11"
+  "--release=13"
```
mpollmeier:
```suggestion
  "--release=11"
```

mpollmeier (changes requested):
this changes two things:
1) require jdk13 for building joern - that's fine and good (since it reflects reality).

2) increase the class version to jdk13 in the compiled artifact - that's really not required (the mentioned file is in `project`, i.e. the sbt project, not the main build. So let's not do that unless really necessary. Hence the suggested change.

build.sbt
```diff
@@ -61,16 +61,16 @@ ThisBuild / compile / javacOptions ++= Seq(
...
-  assert(javaVersion.toInt >= 11, s"this build requires JDK11+ - you're using $javaVersion")
+  assert(javaVersion.toInt >= 13, s"this build requires JDK13+ - you're using $javaVersion")
   Nil
 }
 
 ThisBuild / scalacOptions ++= Seq(
   "-deprecation", // Emit warning and location for usages of deprecated APIs.
   "--release",
-  "11",
+  "13",
```
gacevedo: @mpollmeier How about this? Also needs to go back to 11?

maltek (approved):

Issue #5726, opened 2025-12-22
[Bug][javasrc2cpg] Code got from TypeDecl is different from actual code


ventusfortis:
I'm trying to get the code property from a typeDecl, and the result is different from the actual code.

The code I loaded:

```java
package com.example.cvedemo.config;

import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;
import org.springframework.security.authorization.AuthorizationDecision;
import org.springframework.security.config.annotation.method.configuration.EnableMethodSecurity;
import org.springframework.security.config.annotation.web.builders.HttpSecurity;
import org.springframework.security.core.userdetails.User;
import org.springframework.security.core.userdetails.UserDetails;
import org.springframework.security.crypto.factory.PasswordEncoderFactories;
import org.springframework.security.crypto.password.PasswordEncoder;
import org.springframework.security.provisioning.InMemoryUserDetailsManager;
import org.springframework.security.web.SecurityFilterChain;

@Configuration
@EnableMethodSecurity
public class SecurityConfig {

    @Bean
    public InMemoryUserDetailsManager users() {
        PasswordEncoder encoder = PasswordEncoderFactories.createDelegatingPasswordEncoder();
        UserDetails admin = User.withUsername("admin")
                .password(encoder.encode("adminpass"))
                .roles("ADMIN")
                .build();

        UserDetails user = User.withUsername("user")
                .password(encoder.encode("userpass"))
                .roles("USER")
                .build();

        return new InMemoryUserDetailsManager(admin, user);
    }

    @Bean
    public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
        http
            .authorizeHttpRequests(auth -> auth
                .anyRequest().authenticated()
            )
            .httpBasic()
            .and()
            .csrf().disable(); // упрощение для примера
        return http.build();
    }
}
```

The query I used and the result:

```bash
joern> cpg.annotation.method.typeDecl.code.l
val res25: List[String] = List("public class GenericService")
```

So the signature is different, there's no abstract keyword and it's not generic.

PR #5728, opened 2025-12-23
Fix TypeDecl code to include abstract/final and generics #5726

ventusfortis:
Include abstract/final modifiers and generic parameters in TypeDecl.code for Java; add a regression test. Fixes incorrect rendering described in #5726.


maltek (commented):
Thank you!

I think it would be better to loop over all modifiers of the type, instead of listing them here one by one. Otherwise we will always be missing some of them. I.e. do something like `typ.getModifiers.asScala.map(_.asString).mkString("", " ", " ")` (I haven't tested that snippet though.)

Issue #5699, opened 2025-11-14
[Bug][javasrc2cpg] Trying to get a location node from an annotation returns an error


ventusfortis:
When trying to get location node from an annotation returns an error:
```scala
joern> cpg.annotation.fullName(".*SpringBootApplication.*").location.l
java.util.NoSuchElementException: next on empty iterator
  at scala.collection.Iterator$$anon$19.next(Iterator.scala:973)
  at scala.collection.Iterator$$anon$19.next(Iterator.scala:971)
  at scala.collection.Iterator$$anon$6.next(Iterator.scala:492)
  at flatgraph.traversal.GenericSteps$.head$extension(Language.scala:98)
  at io.shiftleft.semanticcpg.language.nodemethods.CfgNodeMethods$.method$extension(CfgNodeMethods.scala:109)
  at io.shiftleft.semanticcpg.language.LazyLocation.method$lzyINIT1(Location.scala:117)
  at io.shiftleft.semanticcpg.language.LazyLocation.method(Location.scala:117)
  at io.shiftleft.semanticcpg.language.LazyLocation.filename(Location.scala:99)
  at io.shiftleft.semanticcpg.language.LazyLocation.productElements$lzyINIT1(Location.scala:143)
  at io.shiftleft.semanticcpg.language.LazyLocation.productElements(Location.scala:140)
  at io.shiftleft.semanticcpg.language.LazyLocation.productArity(Location.scala:156)
  at replpp.shaded.pprint.Walker.treeify$$anonfun$1(Walker.scala:100)
  at scala.Option.getOrElse(Option.scala:201)
  at replpp.shaded.pprint.Walker.treeify(Walker.scala:128)
  at replpp.shaded.pprint.Walker.treeify$$anonfun$1$$anonfun$3(Walker.scala:85)
  at scala.collection.Iterator$$anon$9.next(Iterator.scala:584)
  at replpp.shaded.pprint.Renderer.rec(Renderer.scala:42)
  at replpp.PPrinter$$anon$2.rec(PPrinter.scala:48)
  at replpp.PPrinter$$anon$1.tokenize(PPrinter.scala:51)
  at replpp.shaded.pprint.PPrinter.apply(PPrinter.scala:121)
  at replpp.PPrinter$.apply(PPrinter.scala:22)
  at replpp.PPrinter.apply(PPrinter.scala)
[...]

PR #5729, opened 2025-12-23
Fix error when trying to get class annotation #5699

ventusfortis:
Replace unsafe head with headOption.orNull in CfgNodeMethods.method for annotations, so annotations not inside a method no longer throw NoSuchElementException, as described in #5699


maltek (commented):
Thank you for the contribution!

You deleted a new line there, could you please fix that?

ml86:
@maltek No, just because there is already bad code in there, it does not mean that we will accept more. This kind of code will make our lifes hell in the future.

Issue #5651, opened 2025-09-15
[Bug] Location node is empty for imports in JS and python (at least)


ventusfortis:
The location node got from imports (`cpg.imports.location.l`) is empty (by that I mean that all its fields are empty) in:
- JS files (I also tried `cpg.imports.file.l` to get the filename)

<img width="560" height="506" alt="Image" src="https://github.com/user-attachments/assets/2236ee06-78aa-4a87-8e59-b235f8dcc1d4" />

<img width="767" height="73" alt="Image" src="https://github.com/user-attachments/assets/b9b7dbf0-b449-4ce6-930e-575f67e2ab04" />

- Python files

<img width="735" height="502" alt="Image" src="https://github.com/user-attachments/assets/f22c6913-6def-4606-a89f-477e5be62c60" />

<img width="777" height="80" alt="Image" src="https://github.com/user-attachments/assets/7ff7d2ce-def2-4832-8740-ca3501066094" />

It is not empty when loading Java files. but for these two languages it doesn't work.

PR #5730, opened 2025-12-24
[pysrc2cpg] Enrich Python import nodes with location metadata and exact code in the code property (#5651)

ventusfortis:
This commit adds a runOnPart method for python imports, which enriches import nodes with missing data about location (#5651  and also fixes the code property to be exact in output

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/frontendspecific/pysrc2cpg/PythonTypeRecovery.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/frontendspecific/pysrc2cpg/PythonTypeRecovery.scala
@@ -56,10 +56,7 @@ private class RecoverForPythonFile(cpg: Cpg, cu: File, builder: DiffGraphBuilder
           // If the imported member is called, seed the call type directly.
           val fullName = createCallFromIdentifierTypeFullName(basePath, memberName)
           i.call.inAst.collectAll[Method].headOption.foreach { method =>
-            cpg.call
-              .nameExact(entityName)
-              .filter(_.inAst.collectAll[Method].fullNameExact(method.fullName).nonEmpty)
-              .foreach(call => symbolTable.append(call, Set(fullName)))
+            method.ast.isCall.nameExact(entityName).foreach(call => symbolTable.append(call, Set(fullName)))
           }
         case UnknownMethod(fullName, alias, receiver, _) =>
           symbolTable.put(CallAlias(alias, receiver), fullName)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/frontendspecific/pysrc2cpg/ImportsPass.scala
```diff
@@ -23,6 +25,25 @@ class ImportsPass(cpg: Cpg) extends XImportsPass(cpg) {
     }
   }
 
+  override def runOnPart(diffGraph: DiffGraphBuilder, part: (Call, Assignment)): Unit = {
```
maltek: please change the base implementation of `XImportsPass.runOnPart` (and/or `Imports.createImportNodeAndLink` which it calls) so that your improvement applies for all frontends making use of it, not just for python.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/frontendspecific/pysrc2cpg/ImportsPass.scala
```diff
@@ -40,4 +61,14 @@ class ImportsPass(cpg: Cpg) extends XImportsPass(cpg) {
     }
   }
 
+  private def importCodeFromCall(call: Call): String = {
```
maltek: This is a very welcome change, but this logic doesn't belong in the imports pass.

I believe the call for the `import` statement is emitted here, so that's where the `code` field should be set to a more appropriate value: 
https://github.com/joernio/joern/blob/de6cfa2d60ca39e67c21a492e4a8369a4e13d911/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitorHelpers.scala#L122

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/frontendspecific/pysrc2cpg/ImportsPass.scala
ventusfortis: Updated XImportsPass.runOnPart to populate Import metadata (code/order/line/column + AST edge) so the same enrichment is applied to all frontends using XImportsPass
https://github.com/joernio/joern/pull/5730/commits/c33d8a4548cf0e89ee974e925b82784c50b52010

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/frontendspecific/pysrc2cpg/ImportsPass.scala
ventusfortis: Moved import code rendering to the AST creation site by passing the import AST node into createCall() in PythonAstVisitorHelpers, so call.code is derived from the original syntax at emission time rather than inside the imports pass.
Updated ImportsPassTests and relevant DataFlowTests expectations to match the new import code strings, now it expects the exact code rather then the intermidiate lowering
https://github.com/joernio/joern/pull/5730/commits/7f5215a72b3b0f241dd64725ee6982aa5e962fe3

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/passes/frontend/XImportsPass.scala
```diff
@@ -22,9 +23,22 @@ abstract class XImportsPass(cpg: Cpg) extends ForkJoinParallelCpgPass[(Call, Ass
...
+      .lineNumber(call.lineNumber)
+      .columnNumber(call.columnNumber)
+
+    call.inAst.collectAll[nodes.Method].headOption.foreach { m =>
+      diffGraph.addEdge(m, importNode, EdgeTypes.AST)
+    }
   }
 
+  /** Hook for frontends to customize how the import code string is rendered (defaults to lowering in call.code). */
+  protected def importCode(call: Call): String = call.code
```
maltek: with the other changes, this is no longer overridden for python or anywhere else. So just remove this method and use `call.code` directly.

maltek (commented):
Looks good! We're almost there.

To get past the code style checker, run `sbt scalafmtAll` and commit the changed code.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/frontendspecific/pysrc2cpg/PythonTypeRecovery.scala
```diff
@@ -53,6 +53,14 @@ private class RecoverForPythonFile(cpg: Cpg, cu: File, builder: DiffGraphBuilder
             .filterNot(_ == Defines.Any)
             .toSet
           symbolTable.put(LocalVar(entityName), memberTypes)
+          // If the imported member is called, seed the call type directly.
+          val fullName = createCallFromIdentifierTypeFullName(basePath, memberName)
+          i.call.inAst.collectAll[Method].headOption.foreach { method =>
+            cpg.call
+              .nameExact(entityName)
+              .filter(_.inAst.collectAll[Method].fullNameExact(method.fullName).nonEmpty)
```
maltek: just a nit-pick: it's more efficient to walk the graph starting from the method here, i.e. `method.ast.isCall.nameExact(entityName)`

PR #5735, opened 2026-01-12
[swiftsrc2cpg] Fix handling of static extension methods

max-leuthaeuser:
Update extension call handling to include static methods (and exclude operator methods) and add tests for these static extensions.


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ExtensionCallPass.scala
```diff
@@ -30,7 +30,7 @@ class ExtensionCallPass(cpg: Cpg, extensionFullNameMapping: Map[String, String])
   override def generateParts(): Array[Call] =
     cpg.call
       .methodFullNameNot(x2cpg.Defines.DynamicCallUnknownFullName)
-      .dispatchType(DispatchTypes.DYNAMIC_DISPATCH)
+      .filterNot(_.methodFullName.startsWith("<operator>"))
```
maltek: we unfortunately have some operators that have a plural `<operators>` name, so
```suggestion
      .filterNot(_.methodFullName.startsWith("<operator"))
```

PR #5736, opened 2026-01-12
[x2cpg] Add retryable assertion scalatest util with timeout and exponential backoff support

max-leuthaeuser:
I have seen a lot of 403s and timeouts while downloading stuff during tests for frontends (especially during wacky Cloudflare things in the last weeks). Some frontend tests rely on downloading dependencies (e.g., kotlin2cpg for resolving types, csharpsrc2cpg) during tests and fail hard if some resource is affected.

We could gradually fix/adjust the tests affected in these frontends over time.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/utils/RetryableAssertion.scala
+++ b/joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/utils/RetryableAssertion.scala
@@ -1,22 +1,31 @@
 package io.joern.x2cpg.utils
 
-import org.scalatest.Assertion
-
 import scala.annotation.tailrec
 import scala.concurrent.duration.*
 import scala.util.{Failure, Success, Try}
 
-/** Utility providing retrying assertions with timeout and optional exponential backoff.
+/** Provides a test-focused retry helper for flaky or asynchronous assertions.
   *
-  * Use `eventually` to execute an assertion block repeatedly until it succeeds or a configured timeout / retry limit is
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/utils/RetryableAssertion.scala
```diff
@@ -0,0 +1,120 @@
...
+    *   The assertion block to execute
+    * @param config
+    *   Retry configuration
+    * @return
+    *   The assertion result if successful
+    * @throws AssertionError
+    *   if all retries are exhausted or timeout is reached
+    */
+  def eventually[T](assertion: => T)(implicit config: RetryConfig = RetryConfig()): T = {
+    val startTime = System.currentTimeMillis()
```
maltek: `System.nanoTime` is better for purposes where one only cares about how things took because it's a monotonic clock (it doesn't jump even with NTP adjustments or suspend/resume cycles). The official javadoc doesn't explicitly guarantee that, but different from `currentTimeMillis` the opaque return value is designed to allow a monotonic clock and de facto it also is implemented as one.

maltek:
I think I'd add a generic type parameter then to constrain the exception types which trigger a retry. For better-behaved tests we can be more specific then. For the others, we can still just use `Exception` or `Throwable`. (Though ideally we'd fix them to not swallow the network exceptions.)

With that, 👍 from me.

maltek (approved):

PR #5739, opened 2026-01-13
[kotlin2cpg] Lambda capture fix (#8544)

SuperUserDone:
Resolves: https://github.com/ShiftLeftSecurity/codescience/issues/8544

This commit solves an issue where lambdas incorrectly capture variables that aren't used within the lambda. As an example, for the given code, the method parameter x and Bar.this are captured into the lambda. Only baz should be captured into the lambda.

```kotlin
package simple.pkg

class Bar {
    fun foo(x: String) {
        val baz: String = "BAZ"
        1.let { println(baz) }
    }
}
```

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForFunctionsCreator.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForFunctionsCreator.scala
@@ -15,6 +15,7 @@ import io.shiftleft.codepropertygraph.generated.Properties
 import io.shiftleft.semanticcpg.language.types.structure.NamespaceTraversal
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.descriptors.DeclarationDescriptor
 import org.jetbrains.kotlin.descriptors.FunctionDescriptor
 import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.descriptors.ParameterDescriptor
@@ -29,7 +30,7 @@ import org.jetbrains.kotlin.resolve.source.KotlinSourceElement
 import scala.collection.mutable
 import scala.jdk.CollectionConverters.*
 
-trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) {
+trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) extends KotlinAstVisitorHelpers {
   this: AstCreator =>
 
   import AstCreator.ClosureBindingDef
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForFunctionsCreator.scala
ml86, line 433: Here the unfiltered locals get added to the scope. I am no expert in the kt2cpg frontend either but this does not look right.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForFunctionsCreator.scala
```diff
@@ -488,14 +492,80 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+  private def astForLambdaBlock(
+    expr: KtBlockExpression,
+    argIdxMaybe: Option[Int],
+    argNameMaybe: Option[String],
+    pushToScope: Boolean = true,
+    localsForCaptures: List[NewLocal] = List(),
+    implicitReturnAroundLastStatement: Boolean = false,
+    preStatements: Option[Seq[Ast]] = None
+  ): (Ast, List[NewLocal]) = {
+    val typeFullName = registerType(exprTypeFullName(expr).getOrElse(TypeConstants.Any))
```
ml86: Majority of this methods body is copied from `astForBlock` please make a common method for the shared code.

SuperUserDone: Made a common method for the shared code.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForFunctionsCreator.scala
SuperUserDone: Will get back to you on this. AFIK this should not be an issue, as locals are not directly added from the scope to the cpg, but would need to dive deeper to confirm.

SuperUserDone:
I have changed the logic, as the previous approach did not properly handle nested lambdas. Now the logic traverses the CompilerAPI AST directly to find captures instead of trying to first create the CPG then trying to deduce closure bindings from that. 

This PR now depends on https://github.com/joernio/joern/pull/5839

SuperUserDone:
I have added assertions to an existing test. Let me know if you think that would be sufficient?
https://github.com/joernio/joern/pull/5739/changes#diff-f49ca26417025a34de759762bf498683a5d2a6aaa1a4463d46e4d6328ccda0d5R591-R626

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/LambdaTests.scala
```diff
@@ -102,20 +94,28 @@ class LambdaTests extends KotlinCode2CpgFixture(withOssDataflow = false, withDef
...
+        |     val baz: String = "BAZ"
+        |     val unused: String = "UNUSED"
+        |     1.let { println(baz) }
+        |   }
+        |}
+        |""".stripMargin)
 
-      cb._refOut.size shouldBe 1
+    "should only capture the used local" in {
+      cpg.methodRef.outE.collectAll[Capture].size shouldBe 1
```
ml86: Wherever possible avoid going to the edge level in tests and use steps instead. Here `_closureBindingViaCaptureOut`.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/LambdaTests.scala
```diff
@@ -102,20 +94,28 @@ class LambdaTests extends KotlinCode2CpgFixture(withOssDataflow = false, withDef
...
+      cpg.closureBinding._localViaRefOut.name.toSet shouldBe Set("baz")
+      cpg.closureBinding._methodParameterInViaRefOut.name.toSet shouldBe empty
```
ml86: Instead of starting with new `cpg....` queries it would be better to make a variable for the method reference and then expand from there. Like it is now you do not test that the closure bindings are associated with the right methodRef.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForFunctionsCreator.scala
```diff
@@ -395,12 +452,52 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+      .map(referencedCaptureNamesIn(_, declaredParamNames))
+      .getOrElse(Set.empty)
+
     val closureBindingEntriesForCaptured = scope
       .pushClosureScope(lambdaMethodNode)
       .collect {
         case node: NewMethodParameterIn => NodeContext(node, node.name, node.typeFullName)
         case node: NewLocal             => NodeContext(node, node.name, node.typeFullName)
       }
+      .filter(capturedNodeContext => hasDestructuringParam || referencedNames.contains(capturedNodeContext.name))
```
ml86: Why are all `capturedNodeContext` taken if there is even a single destructuring parameter?
Are there tests for this?

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForFunctionsCreator.scala
```diff
@@ -440,9 +537,12 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) {
...
         parameters.zipWithIndex.foreach { (paramDesc, idx) =>
           val param = paramDesc.getSource.asInstanceOf[KotlinSourceElement].getPsi.asInstanceOf[KtParameter]
-          paramAsts.append(astForParameter(param, valueParamStartIndex + idx))
           if (param.getDestructuringDeclaration != null) {
-            destructedParamAsts.appendAll(astsForDestructuring(param))
+            val generatedName = s"${Constants.DestructedParamNamePrefix}${destructedParamKeyPool.next}"
+            paramAsts.append(astForParameter(param, valueParamStartIndex + idx, Some(generatedName)))
+            destructedParamAsts.appendAll(astsForDestructuring(param, Some(generatedName)))
+          } else {
+            paramAsts.append(astForParameter(param, valueParamStartIndex + idx))
```
ml86, line 605: Seems to be an unrelated fix for the naming for destructuring parameters. Cant find a corresponding change in the tests. What am I missing?

SuperUserDone: The generated destructuring parameter name is passed into `astsForDestructuring(...)` so the synthetic it initializer `(tmp = it)` can be wired to the real lambda parameter node instead of a standalone local, which restores correct closure/dataflow edges (at least for the OSS dataflow engine).
That was needed because, after tightening capture to used variables, the old destructuring path lost the parameter-to-destructured-value link and caused the lambda dataflow test to go empty, so this change preserves the expected flow.

I'll add a test to test the parameter name generation. The Kotlin2cpg frontend does have a lot of missing tests for edge cases like this. I'll make a note for future work to also check that tests are sufficient to test edge cases like this.

The test that caught this issue was this:

https://github.com/joernio/joern/blob/bac40363833e6ec71801e2d37f1e326f51d7f71a/joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/dataflow/LambdaTests.scala

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForFunctionsCreator.scala
SuperUserDone: Thinking a bit more about it, this is a bit of a hack. The idea is that a lambda with at least one destructuring parameter bypasses usage filtering and captures all visible outer locals/params as a compatibility workaround for destructuring. The OSS dataflow engine seems very particular about these sorts of things. I'll try to think of a better method to solve this issue. 

Ideally true usage-based capturing would be possible, but unfortunately this heuristic based method is the best that I found possible with my current understanding of the kotlin compiler API and internals. I'll do some more reading into the internals and reading more into how other frontends handle this for a potential cleaner fix.

SuperUserDone: This test covers this case indirectly, but I'll add a test to cover this behavouir directly.

https://github.com/joernio/joern/blob/bac40363833e6ec71801e2d37f1e326f51d7f71a/joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/dataflow/LambdaTests.scala

SuperUserDone: I have removed this hack and replaced it with a proper filtering solution.

PR #5746, opened 2026-01-14
Upgraded sealed traits in Model.scala to scala3

SuperUserDone:
Not sure if the following cases should be converted too, inside the same Model.scala file.

- Named defines shared methods and is used as a trait bound
- JavaTypeSignature and ReferenceTypeSignature form a nested inheritance hierarchy, possible, but don't I think idiomatic scala 3 enums would improve the code/usage of these classes.
- ResolvedType includes a non-case class (ResolvedTypeDecl) with mutable state, which isn't suitable for enum conversion. Would require a large amount of changes by separating the types out completely, or changing usage to make use of immutable objects.


ml86:
Please leave the other cases as they are.

PR #5750, opened 2026-01-14
Upgraded scala 2 enums to scala 3 in javasrc2cpg Scope

SuperUserDone:


ml86:
`ScopeVariable` does not behave much like an Enum those we will not merge this.

PR #5751, opened 2026-01-15
[swiftsrc2cpg] Handle computed property functions

max-leuthaeuser:
- support property accessors in type decls and extensions
- fixed member/property call methodFullName mapping
- Unified ExtensionMemberPass, ExtensionCallPass, and ExtensionInheritancePass into a single pass (ExtensionsPass)

### DIFF
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/DefaultSemantics.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/DefaultSemantics.scala
@@ -1,6 +1,7 @@
 package io.joern.dataflowengineoss
 
-import io.joern.dataflowengineoss.semanticsloader.{FlowSemantic, PassThroughMapping, FullNameSemantics}
+import io.joern.dataflowengineoss.semanticsloader.{FlowSemantic, FullNameSemantics}
+import io.joern.dataflowengineoss.semanticsloader.FlowPath.{FlowMapping, PassThroughMapping}
 import io.shiftleft.codepropertygraph.generated.Operators
 
 import scala.annotation.unused
--- a/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/nodemethods/ExpressionMethods.scala
+++ b/dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/language/nodemethods/ExpressionMethods.scala
@@ -1,6 +1,7 @@
 package io.joern.dataflowengineoss.language.nodemethods
 
 import io.joern.dataflowengineoss.semanticsloader.*
+import io.joern.dataflowengineoss.semanticsloader.FlowPath.{FlowMapping, PassThroughMapping}
 import io.shiftleft.codepropertygraph.generated.nodes.{Call, Expression, Method}
 import io.shiftleft.semanticcpg.language.*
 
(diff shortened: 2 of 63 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ExtensionsPass.scala
```diff
@@ -0,0 +1,183 @@
...
+    *     - `NAME` and `METHOD_FULL_NAME` to the resolved getter
+    *     - `DISPATCH_TYPE` to `DYNAMIC_DISPATCH`
+    *     - ensures a `RECEIVER` edge from the access to the base expression and fixes argument indices/orders
+    */
+  private def handleMemberPropertyCalls(diffGraph: DiffGraphBuilder): Unit = {
+    val fieldAccesses = cpg.fieldAccess
+    fieldAccesses.foreach { fieldAccess =>
+      for {
+        base         <- fieldAccess.arguments(1).headOption
+        baseFullName <- base.flatMap(typeFullNameOf)
```
maltek: I think `flatMap` only works here because of an implicit conversion of a plain value to an iterator
```suggestion
        baseFullName <- typeFullNameOf(base)
```

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ExtensionsPass.scala
```diff
@@ -0,0 +1,183 @@
...
+        memberName   <- fieldAccess.arguments(2).headOption.map(_.code)
```
maltek: if a matching property is found, this field identifier here should be removed from the graph

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ExtensionsPass.scala
```diff
@@ -0,0 +1,183 @@
...
+          * the `inheritsFrom` property and create `INHERITS_FROM` edges to the correct `typ` nodes.
+          *
+          * The value is `lazy` to avoid performing lookups when the `for` loop below does not find a `TypeDecl` for
+          * `typeDeclFullName`.
+          */
+        lazy val inheritsFullNamesResolved = inheritsFullNames.map { name =>
+          cpg.typeDecl.nameExact(name).headOption match {
+            case Some(td) => td.fullName
+            case None     => name
+          }
```
maltek: if there are multiple different types with this same simple name, we just take the first one (in database order). The conservative approach here would be to only rewrite like this if there is a unique match, i.e. 
`cpg.typeDecl.nameExact(name).fullName.loneElementOption`.

Do you think this would be too strict?

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ExtensionsPass.scala
```diff
@@ -0,0 +1,183 @@
...
+          typeDecl <- cpg.typeDecl.fullNameExact(typeDeclFullName)
+        } setInherits(diffGraph, typeDecl, inheritsFullNamesResolved)
+      }
+    }
+  }
+
+  /** Merge the provided inherited type fullNames with existing ones and set them on the type declaration. */
+  private def setInherits(diffGraph: DiffGraphBuilder, typeDecl: TypeDecl, inheritFullNames: Set[String]): Unit = {
+    val existingInherits = typeDecl.inheritsFromTypeFullName
+    val allInherits      = (existingInherits ++ inheritFullNames).sorted
```
maltek: this should also do a `.distinct`, otherwise we get duplicates here due to the `None     => name` fallback

PR #5752, opened 2026-01-15
Deprecate FileUtil.deleteOnExit in cpgcreation

SuperUserDone:


console/src/main/scala/io/joern/console/Console.scala
```diff
@@ -39,25 +39,27 @@ class Console[T <: Project](loader: WorkspaceLoader[T], baseDir: Path = FileUtil
...
-      } match {
-        case Success(_) =>
-          // We never handle the actual result anywhere.
-          // Hence, we just pass a success message.
-          Success(s"Running viewer for '$tmpFile' finished.")
-        case Failure(exc) =>
-          System.err.println("Executing image viewer failed. Is it installed? ")
-          System.err.println(exc)
-          Failure(exc)
+      // and gets removed immediately after running this viewer instance synchronously via .run().
```
maltek: the new code has the same problem that the copy was supposed to prevent - the file is immediately deleted, potentially before the image viewer had a chance to actually read it.

On Unix, we could proably get away with deleting the file after a fixed delay that's generally long enough for even a very slow image viewer to have read the file. On Windows, that'd just fail though if the image viewer keeps an open file handle.

Overall, I think this may be the exception where delete on exit is acceptable?

SuperUserDone: Have restored original functionality for that function in latest commit

PR #5761, opened 2026-01-16
Replace .orNull call

max-leuthaeuser:


max-leuthaeuser:
That only shifts the issue to `CfgNodeMethods.method` (or higher up the call chain). All these methods are not null-safe. 
Also: having no parent method always indicates a frontend issue so crashing early is good.

ml86:
As discussed on Zoom, we can remove this special handling this the frontends have already been fixed.

ml86:
I thought that is already in and released...

PR #5764, opened 2026-01-19
Deprecate delete onExitTest in test fixtures

SuperUserDone:
Depends on #5759 for the default output path change.


joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/testfixtures/CSharpCode2CpgFixture.scala
```diff
@@ -86,9 +86,8 @@ trait CSharpFrontend extends LanguageFrontend {
 
   override def execute(sourceCodeFile: File): Cpg = {
     val cpgOutFile = FileUtil.newTemporaryFile(suffix = "cpg.bin")
```
ml86: Also remove now unused `cpgOutFile`.

PR #5771, opened 2026-01-21
Deprecate FileUtil.deleteOnExit in kotlin2cpg tests

SuperUserDone:
Note: CompilerAPI.scala still uses raw deleteOnExit() for resource dependencies. This is a more complex case requiring refactoring of the compiler configuration lifecycle. Not sure it would be practical.


ml86 (approved):
Lets leave the CompilerAPI.scala as is. We might revisit it at the end again.

PR #5774, opened 2026-01-21
Deprecate FileUtil.deleteOnExit in jimple2cpg tests

SuperUserDone:
Note: There is still an instance in `JimpleCodeToCpgFixture.scala`. This would require some larger changes to remove.


ml86 (approved):
Like with the others we will leave the more complex one for now as it is.

PR #5775, opened 2026-01-21
[kotlin2cpg] Add support for KtCallableReferenceExpression

SuperUserDone:


joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,157 @@
...
+        |fun isOdd(x: Int) = x % 2 != 0
+        |
+        |fun main() {
+        |  val numbers = listOf(1, 2, 3)
+        |  numbers.filter(::isOdd)
+        |}
+        |""".stripMargin)
+
+    inside(cpg.call.name("filter").argument.l) { case List(receiver, methodRef: MethodRef) =>
+      methodRef.methodFullName shouldBe "<unresolvedNamespace>.isOdd:boolean(int)"
```
ml86: Does `isOdd` really have `<unresolvedNamespace>.isOdd:boolean(int)` as method full name? If yes that is already a problem, if not the methodRef should use the same method full name as the referenced method.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,157 @@
...
+        |  val reference = numbers::forEach
+        |}
+        |""".stripMargin)
+
+    val methodRefs = cpg.methodRef.code("numbers::forEach").l
+    methodRefs.size shouldBe 1
+    val methodRef = methodRefs.head
+
+    // The exact type will depend on Kotlin stdlib resolution
+    methodRef.methodFullName should include("forEach")
```
ml86: `forEach` is not a full name. I would have expected something like e.g. `kotlin.collection.forEach(...)`.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,157 @@
...
+        |  fun doNothing(c: (Int) -> Unit) {}
+        |
+        |  fun foo() {
+        |    doNothing(Bar::bar)
+        |  }
+        |}
+        |""".stripMargin)
 
+    inside(cpg.call.name("doNothing").argument.l) { case List(thisArg: Identifier, methodRef: MethodRef) =>
+      thisArg.name shouldBe "this"
```
ml86: Testing for `this` argument does not really belong to the thing we want to check here. Please remove it.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,157 @@
...
+
+      methodRef.methodFullName shouldBe "com.test.Bar.bar:void(int)"
+      methodRef.typeFullName shouldBe "com.test.Bar"
```
ml86: This is wrong. You seem to have looked at Javasrc2cpg, but that is/was sadly not a good example and Johannes is fixing it there as we speak.

You basically have to create a new synthetic TYPE_DECL and a corresponding TYPE which is then referenced here in typeFullName. This new TYPE_DECL then needs bindings that reflect the implemented interface.

I suggest you and @johannescoetzee talk tomorrow so that he can tell you the details.
One pointer where most of this is already implemented and tested are the Lambda function tests in Javasrc2cpg [here](https://github.com/joernio/joern/blob/f5f66d2fff2a7ec001f2f889f786921eac4a048d/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/LambdaTests.scala#L8)

ml86 (changes requested):

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: AFAIK, the forEach FQN might change depending on the Kotlin stdlib version. I used the `should include("forEach")`  check instead of an exact comparison to accommodate. Should I change it to be exact for the latest version?

ml86: I find that very surprising. What are those different full names?
Even if they are different we should have a fixed one in our test scenario right?

SuperUserDone: 1.0 >= Kotlin < 1.3 kotlin.collections.List.forEach
Kotlin >= 1.3 kotlin.collections.Iterable.forEach

Should be `kotlin.collections.Iterable` though for the test yes.

SuperUserDone:
The latest set of changes are major overhaul of the methodology of this PR. This now closely follows what the kotlin compiler does.

For unbound references to a global function `::globalFunction`, it works similarly to the lambda method @johannescoetzee and I worked through for the javasrc2cpg case. It creates a `MethodRef` node, along with a dummy `Type`, `TypeDecl` and  bindings for the method being overridden in the SAM Interface (Eg invoke for a function). If the SAM Interface is generic, an type-erased binding is also created.

For bound references to a companion object `Type::function` and to a regular method `instance::method`, a new `Type` and `TypeDecl` is created. The `TypeDecl` inherits from `kotlin.jvm.internals.CallableReference`. This class provides an `this.receiver` object. A constructor and implementation of the SAM Method is lowered into the CPG. The TypeDecl contains bindings for the method with concrete types, and if applicable a type-erased binding is also generated.

This does not work around kotlin compiler bug https://youtrack.jetbrains.com/issue/KT-54316, so the behavior is identical to the Kotlin compiler in the documented case. This only affects the companion case for assigned references.

This affects cases like this. This code won't compile because the Utils::validate ref has a broken signature of (Utils, Int) -> Boolean.
```kotlin
class Utils {
    companion object {
        fun validate(x: Int): Boolean = x > 0
    }
}

val ref: (Int) -> Boolean = Utils::validate
```

This case is fine tough and types are resolved correctly

```kotlin
class Utils {
    companion object {
        fun validate(x: Int): Boolean = x > 0
    }
}

fun bar(ref: (Int) -> Boolean) {}

bar(Utils::validate)
```

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+      params.head.name shouldBe "this"
+      params(1).typeFullName shouldBe "int"
+      params(2).typeFullName shouldBe "java.lang.String"
+    }
+
+    "forward all parameters to the referenced method" in {
+      val processCall   = invokeMethod.ast.isCall.name("process").head
+      val forwardedArgs = processCall.argument.isIdentifier.l
+      forwardedArgs.size shouldBe 2
+      forwardedArgs.map(_.typeFullName).toSet shouldBe Set("int", "java.lang.String")
```
johannescoetzee: With the way this is written, it's possible to miss non-identifier arguments or to have the correct arguments in the wrong order without the test catching it. It's a nitpick, but in my opinion it is better to test explicitly, so something like
```
processCall.argument.map(_.typeFullName).toList shouldBe List(<something>, "int", "java.lang.String")
```

or 
```
inside(processCall.argument.l) { case List(<something>, intArg: Identifier, stringArg: Identifier) =>
  intArg.typeFullName shouldBe "int"
  ...
}
```

It's more cumbersome to write and can be harder to read, but is also occasionally useful for catching weird bugs

johannescoetzee: Please check that REF edges from the `process` arguments to the `invoke` parameters are being created correctly as well (in this method, but also wherever an `Identifier` node is added to the CPG). A lot of dataflow issues in the past were caused by missing REF edges (or REF edges referring to the wrong local/param) and testing it in the frontend AST tests is much faster than figuring it out through dataflow debugging

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+    }
+
+    "use DYNAMIC_DISPATCH with cast and field access to receiver" in {
```
johannescoetzee, line 82: Is it necessary to add a cast call here? I know the Kotlin compiler uses the cast, but is there anything stopping us from just using `com.test.Handler` as the type for the `receiver` field?

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+      val processCall = invokeMethod.ast.isCall.name("process").head
+      processCall.dispatchType shouldBe "DYNAMIC_DISPATCH"
+
+      // The receiver should be a cast operation
+      val castCall = processCall.receiver.isCall.name(Operators.cast).head
+      castCall.methodFullName shouldBe Operators.cast
+      castCall.typeFullName shouldBe "ANY"
```
johannescoetzee: If we do need the cast, the `typeFullName` should be the type being cast to (in this case `com.test.Handler`)

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+      castCall.dispatchType shouldBe "STATIC_DISPATCH"
+
+      // Cast should have field access as first argument
```
johannescoetzee: Are you sure the field access should be the first argument? `javasrc2cpg` does it the other way around

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+        .head
+        .typeFullName shouldBe "kotlin.jvm.internal.CallableReference"
+
+      // Cast should have TypeRef as second argument
+      val typeRef = castCall.argument.argumentIndex(2).isTypeRef.head
+      typeRef.typeFullName shouldBe "com.test.Handler"
+    }
+
+    "create a constructor call for the synthetic type with receiver as parameter" in {
+      val ctorCalls = cpg.call.nameExact("<init>").methodFullName(".*Function2Impl.*<init>.*").l
```
johannescoetzee:
```suggestion
      val ctorCalls = cpg.call.methodFullName(".*Function2Impl.*<init>.*").l
```
should return the same, unless there's a type decl with `<init>` in the name. You could also use `methodFullNameExact` with the exact method full name to avoid any conflicts.

johannescoetzee: This is perhaps a stylistic thing, but I also prefer to write these in a way that gives more information about where the call occurs, for example `cpg.method("test").call...`. How far to take this depends on the complexity of the code in question though and here just finding the `<init>` directly is probably fine

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+      ctorCalls.size shouldBe 1
+
+      val ctorCall = ctorCalls.head
+      ctorCall.methodFullName shouldBe "com.test.Handler.process$kotlin.jvm.functions.Function2Impl.<init>:void(com.test.Handler)"
+      ctorCall.typeFullName shouldBe "void"
+      ctorCall.signature shouldBe "void(com.test.Handler)"
+
+      val args = ctorCall.argument.l
+
+      val receiverArg = args(1).asInstanceOf[Identifier]
```
johannescoetzee, line 104: Another nitpick: `args.argumentIndex(1)` is better since the argument edges aren't guaranteed to be added in the correct order. In most cases they will be (for example any call created with `callAst`), but if arguments are added/removed/modified later it could mess up the order. 

The suggestion in the the previous comment where I mentioned doing something like
```
inside(call.argument.l) { case List(...) => ... }
```
runs into the same issue and using `call.argument.sortBy(_.argumentIndex).l` would be safer as well. I'm not consistent about how paranoid to be about this, but have found some bugs where the difference mattered.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+      receiverArg.typeFullName shouldBe "com.test.Handler"
+    }
+
+    "have PUBLIC and VIRTUAL modifiers on invoke method" in {
+      val modifiers = invokeMethod.modifier.l
+      modifiers.size shouldBe 2
+      modifiers.map(_.modifierType).toSet shouldBe Set("PUBLIC", "VIRTUAL")
+    }
+
+    "have a constructor method with CONSTRUCTOR modifier" in {
```
johannescoetzee, line 115: Please extend this test (or add a separate one) showing that the field assignment is done correctly in the constructor

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+      val nestedTypeDecls = cpg.typeDecl.fullName(".*Function2Impl.*invoke.*").l
+      nestedTypeDecls.size shouldBe 1
+
+      val nestedTypeDecl = nestedTypeDecls.head
+      nestedTypeDecl.name shouldBe "invoke"
+      nestedTypeDecl.fullName shouldBe "com.test.Handler.process$kotlin.jvm.functions.Function2Impl.invoke:boolean(int,java.lang.String)"
+      nestedTypeDecl.astParentType shouldBe "TYPE_DECL"
+      nestedTypeDecl.astParentFullName shouldBe "com.test.Handler.process$kotlin.jvm.functions.Function2Impl"
+      nestedTypeDecl.inheritsFromTypeFullName should contain("kotlin.Function")
+    }
```
johannescoetzee: Why is this necessary? Could you not just change the typeFullName of `com.test.Handler.process$kotlin.jvm.functions.Function2Impl` to `com.test.Handler.process$kotlin.jvm.functions.Function2Impl.invoke:boolean(int,java.lang.String)` and only use that?

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+          |   val handler = Handler()
+          |   val ref: (String?) -> Boolean = handler::process
+          |}
+          |""".stripMargin)
+
+      val syntheticTypeDecl = cpg.typeDecl.fullName(".*Function1Impl.*").head
+      val invokeMethod      = syntheticTypeDecl.method.name("invoke").head
+
+      val param = invokeMethod.parameter.index(1).head
+      param.typeFullName should include("String")
```
johannescoetzee: Please check the full type full name (and same for similar tfn/signature checks below)

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+      val transformerImpl = cpg.typeDecl.fullName(".*TransformerImpl.*").head
+      val constructors    = transformerImpl.method.name("<init>").l
+      constructors.size shouldBe 1
+
+      val constructor = constructors.head
+      constructor.modifier.map(_.modifierType).head shouldBe "CONSTRUCTOR"
+    }
+  }
+
+  "SAM implementation edge cases" should {
```
johannescoetzee, line 373: For all of these, please check at least the following:
1. The `invoke` method is created with the correct mfn/signature and parameters
2. The constructor for the impl type is created with the same
3. The constructor call is created with correct signature and arguments
4. The invoke call is created with the correct signature and arguments

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+          |}
+          |""".stripMargin)
+
+      val syntheticTypeDecl = cpg.typeDecl.fullName(".*Function1Impl.*").head
+      val invokeMethod      = syntheticTypeDecl.method.name("invoke").head
+
+      invokeMethod.signature shouldBe "boolean(int)"
+      val checkCall = invokeMethod.ast.isCall.name("check").head
+      checkCall.methodFullName shouldBe "Validator.check:boolean(int)"
+    }
```
johannescoetzee: What is this testing that isn't already covered by other tests? The body of the method being referred to shouldn't affect the method ref lowering

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+          |}
+          |""".stripMargin)
+
+      val syntheticTypeDecl = cpg.typeDecl.fullName(".*Function1Impl.*").head
+      val invokeMethod      = syntheticTypeDecl.method.name("invoke").head
+      val thisParam         = invokeMethod.parameter.index(0).head
+
+      thisParam.dynamicTypeHintFullName should not be empty
+      thisParam.dynamicTypeHintFullName.head should include("Function1Impl")
+    }
```
johannescoetzee: This could be combined with the similar test above

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,552 @@
...
+      val syntheticTypeDecl = cpg.typeDecl.fullName(".*Function0Impl.*").head
+      val constructor       = syntheticTypeDecl.method.name("<init>").head
+      val thisParam         = constructor.parameter.index(0).head
+
+      thisParam.dynamicTypeHintFullName should not be empty
+      thisParam.dynamicTypeHintFullName.head should include("Function0Impl")
+    }
+  }
+
+  "enforce uniqueness for duplicate SAM implementations" should {
```
johannescoetzee, line 594: Please add another uniqueness test where the refs appear in separate files

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: Have adjusted some of the tests to use this pattern where I thought it made sense. In most of the tests also added checks for the ref edges to the identifier nodes. There was a bug that I caught through that and resolved.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: Have removed the cast call. Everything I tested works as expected with the cast removed.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: Cast has been removed

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: Have made sure to add sorting and use argumentIndex where apropriate

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: Added checks for this

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: Have added some checks for these points. Checking is not exhaustive yet for these cases but should be much better

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: This case slipped in from some playing around I did. Initially it was meant to check functions that have implicit return types and how far you can push it, but the Kotlin compiler API handles it properly so is not needed.

I have removed the case.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: Have done so.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: have combined the nullable parameter and return cases

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: The ctor calls the ctor of the superclass (CallableReference) I have added a check for that call.

johannescoetzee: I'm not sure if that would work (not saying it won't, just that I don't know if it will). Could you please add a dataflow test for something like the below to check it (I'm mixing java and kotlin syntax, but I hope the intent is clear):

```
class Foo {
  String s;

  public Foo(String s) {
    this.s = s;
  }

  public String foo() {
    return s;
  }
}

void main() {
  val fooInstance = new Foo(source())
  val ref = fooInstance::foo
  sink(ref.invoke())
}
```

SuperUserDone: Hmm yes, it seems like it does not track into the definition of the `CallableReference` Class constructor defined in the kotlin stdlib.  Planning to work around this by removing  the `CallableReference` Class from the inheritance hierarchy, and directly assigning to `this.receiver` in the generated ctor instead. Does that sound good?

SuperUserDone: I have made this adjustment. Dataflow tracking now works as expected

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -501,9 +564,27 @@ class AstCreator(
...
+      val argumentTypeMaybe = calledFuncDesc
+        .flatMap { funcDesc =>
+          val params = funcDesc.getValueParameters.asScala.toList
+          if (idx - 1 < params.size) {
+            Some(params(idx - 1))
+          } else {
+            None
+          }
+        }
+        .flatMap(paramDesc => Some(paramDesc.getType))
```
ml86: Extra flatMap seems superfluous. Just do the `.getType` in the above flatMap.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -534,6 +615,247 @@ class AstCreator(
...
+          Seq((typeDecl, genericBinding, EdgeTypes.BINDS), (genericBinding, methodNode, EdgeTypes.REF))
+        )
+      )
+    }
+  }
+
+  protected def createUnboundSamTypeDecl(samInfo: UnboundSamInfo): Ast = {
+    val samTypeDecl = typeDeclNode(
+      samInfo.expr,
+      samInfo.samImplClass.split('.').last,
```
ml86, line 656: Is this quaranteed to always have a '.'?

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -534,6 +615,247 @@ class AstCreator(
...
+      samInfo.samImplClass,
+      relativizedPath,
+      samInfo.inheritsFrom.map(registerType),
+      None
+    )
+
+    val samMethodFullName = s"${samInfo.methodRefName}:${samInfo.signature}"
+
+    val methodRefBinding =
```
ml86: Reused `createBindingInfo`

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -534,6 +615,247 @@ class AstCreator(
...
+      bindingInfoQueue.prepend(BindingInfo(genericBinding, Seq((samTypeDecl, genericBinding, EdgeTypes.BINDS))))
+    }
+
+    Ast(samTypeDecl)
+  }
+
+  protected def createBoundSamTypeDecl(samInfo: BoundSamInfo): Ast = {
+    val samTypeDecl = typeDeclNode(
+      samInfo.expr,
+      samInfo.samImplClass.split('.').last,
```
ml86: Garanteed to have `.`?

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -534,6 +615,247 @@ class AstCreator(
...
+    methodName: String,
+    signature: String,
+    methodFullName: String,
+    typeDecl: NewTypeDecl,
+    methodNode: NewMethod,
+    genericSignature: Option[String]
+  ): Unit = {
+    val binding = bindingNode(methodName, signature, methodFullName)
+    bindingInfoQueue.prepend(
+      BindingInfo(binding, Seq((typeDecl, binding, EdgeTypes.BINDS), (binding, methodNode, EdgeTypes.REF)))
```
ml86: The edge meta data for the REF edge is superfluous as long as it points to the same method as indicated by methodFullName. The backend will create the edge.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -534,6 +615,247 @@ class AstCreator(
...
+    )
+
+    genericSignature.foreach { genericSig =>
+      val genericBinding = bindingNode(methodName, genericSig, methodFullName)
+      bindingInfoQueue.prepend(
+        BindingInfo(
+          genericBinding,
+          Seq((typeDecl, genericBinding, EdgeTypes.BINDS), (genericBinding, methodNode, EdgeTypes.REF))
```
ml86: Again superfluous REF edge.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -534,6 +615,247 @@ class AstCreator(
...
+    val receiverAssignmentNode =
+      operatorCallNode(samInfo.expr, s"this.receiver = $receiverParamName", Operators.assignment, None)
+    val receiverAssignmentAst = callAst(receiverAssignmentNode, Seq(receiverFieldAccessAst, receiverArgAst))
+
+    val ctorBlock =
+      blockNode(samInfo.expr, s"this.receiver = $receiverParamName", registerType(TypeConstants.Void))
+    val ctorReturn    = methodReturnNode(samInfo.expr, TypeConstants.Void)
+    val ctorModifiers = Seq(modifierNode(samInfo.expr, ModifierTypes.CONSTRUCTOR))
+
+    createBindingInfo("<init>", s"void(${receiverType})", ctorFullName, samTypeDecl, ctorNode, None)
```
ml86: Constructor is not supposed to be in the binding table since it is never dynamically dispatched.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -534,6 +615,247 @@ class AstCreator(
...
+      samInfo.expr,
+      samInfo.samImplClass.split('.').last,
+      samInfo.samImplClass,
+      relativizedPath,
+      samInfo.inheritsFrom.map(registerType),
+      None
+    )
+
+    val samMethodFullName = s"${samInfo.samImplClass}.${samInfo.samMethodName}:${samInfo.samMethodSig}"
+    val samMethodNode = methodNode(
```
ml86, line 677: Factor out SAM method creation into separate method.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -534,6 +615,247 @@ class AstCreator(
...
+      samMethodFullName,
+      samTypeDecl,
+      samMethodNode,
+      samInfo.samGenericMethodSig
+    )
+
+    val receiverType = samInfo.receiverTypeFullName
+    val ctorFullName = s"${samInfo.samImplClass}.<init>:void(${receiverType})"
+    val ctorNode =
+      methodNode(samInfo.expr, "<init>", "<init>", ctorFullName, Some(s"void(${receiverType})"), relativizedPath)
```
ml86, line 756: Factor out constructor method creation into separate method.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -387,9 +439,20 @@ class AstCreator(
     }
     val lambdaTypeDecls =
       lambdaBindingInfoQueue.flatMap(_.edgeMeta.collect { case (node: NewTypeDecl, _, _) => Ast(node) })
+
+    val samImplTypeDecls = samImplInfoQueue.values.flatMap { samInfo =>
```
ml86, line 444: This is an iteration over a HashMap. We do not want to do that because the iteration order is not guaranteed which results in semantical identical but structurally different CPGs.  We want to avoid that as much as possible. So if you must iterate over this either use a map which allows for insertion order iteration or sort it before iteration.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -534,6 +615,247 @@ class AstCreator(
...
+      Some(samInfo.signature),
+      Some(registerType(returnType))
+    )
+
+    val callArgumentAsts = samInfo.samMethodParams.zipWithIndex.zip(valueParameterAsts).map {
+      case (((paramName, paramType), idx), paramAst) =>
+        val identNode =
+          identifierNode(samInfo.expr, paramName, paramName, registerType(paramType)).argumentIndex(idx + 1)
+        paramAst.root
+          .collect { case p: NewMethodParameterIn =>
```
ml86: Avoid single letter variable names.

ml86: Best to have `valueParameterAsts` be not yet `Ast` wrapped `NewMethodParameterIn` nodes. That way you can avoid the `collect`.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -27,11 +28,60 @@ import scala.annotation.tailrec
...
+
+  case class BoundSamInfo(
+    samImplClass: String,
+    methodRefName: String,
+    signature: String,
+    inheritsFrom: List[String],
+    samMethodName: String,
+    samMethodSig: String,
+    samGenericMethodSig: Option[String],
+    receiverParam: (Ast, String),
```
ml86: The second tuple value seems never to be used.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -534,6 +615,247 @@ class AstCreator(
...
+
+    val receiverType = samInfo.receiverTypeFullName
+    val ctorFullName = s"${samInfo.samImplClass}.<init>:void(${receiverType})"
+    val ctorNode =
+      methodNode(samInfo.expr, "<init>", "<init>", ctorFullName, Some(s"void(${receiverType})"), relativizedPath)
+    val ctorThisParam =
+      parameterInNode(samInfo.expr, "this", "this", 0, false, EvaluationStrategies.BY_SHARING, samInfo.samImplClass)
+        .dynamicTypeHintFullName(IndexedSeq(samInfo.samImplClass))
+    val receiverParamName =
+      samInfo.receiverParam._1.root.collect { case expr: ExpressionNew => expr.code }.getOrElse("receiver")
```
ml86: If `expr` is e.g. a call expression this will result in strange looking parameter names. I suggest you just always name the parameter `receiver`.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -534,6 +615,247 @@ class AstCreator(
...
+  }
+
+  protected def createSamImplTypeDecl(samInfo: SamImplInfo): Ast = {
+    samInfo match {
+      case unbound: UnboundSamInfo => createUnboundSamTypeDecl(unbound)
+      case bound: BoundSamInfo     => createBoundSamTypeDecl(bound)
+    }
+  }
+
+  private def extractReturnTypeFromSignature(signature: String): String = {
```
ml86: The method signature you pass into this function was at some point composed from a return type and the parameter types and now you parse it in order to extract the return type again. Please avoid that and propagated the not yet combined return and parameter types.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
SuperUserDone: No, its not, but does `String.split(..)` not contain the original string as the only element of the array if so? Am I missing something?

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
SuperUserDone: Good to know. Was one of the pain points as it was not possible to get a ref to the method node here, hence the duplication here: https://github.com/joernio/joern/pull/5775#discussion_r2882928331

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
SuperUserDone: Will swap to using `SeqMap`

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
ml86: You are right. For whatever reason I was thinking that `split` returns an empty array if the character is not contained in the string.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
ml86: Same as above. Please ignore.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -769,31 +745,24 @@ class AstCreator(
...
       parameterInNode(samInfo.expr, "this", "this", 0, false, EvaluationStrategies.BY_SHARING, samInfo.samImplClass)
         .dynamicTypeHintFullName(IndexedSeq(samInfo.samImplClass))
-    val receiverParamName =
-      samInfo.receiverParam._1.root.collect { case expr: ExpressionNew => expr.code }.getOrElse("receiver")
+
     val ctorReceiverParam = parameterInNode(
       samInfo.expr,
-      receiverParamName,
-      receiverParamName,
+      "receiver",
```
ml86: Use constant for "receiver" string.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForExpressionsCreator.scala
```diff
@@ -760,58 +762,67 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
-      abstractMethods.headOption
+    val samMethodFromExprType = exprType.flatMap { kotlinType =>
+      Option(kotlinType.getConstructor.getDeclarationDescriptor)
+        .collect { case classDescriptor: ClassDescriptor => classDescriptor }
+        .flatMap { classDescriptor =>
+          classDescriptor.getUnsubstitutedMemberScope
+            .getContributedDescriptors(DescriptorKindFilter.FUNCTIONS, _ => true)
+            .asScala
+            .collectFirst {
+              case functionDescriptor: FunctionDescriptor if functionDescriptor.getModality.toString == "ABSTRACT" =>
```
ml86: I suspect that we can avoid using `.toString` and instead check the modality directly?

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,692 @@
...
+        |}
+        |
+        |fun test() {
+        |   val handler = Handler()
+        |   val ref: (Int, String) -> Boolean = handler::process
+        |}
+        |""".stripMargin)
+
+    val syntheticTypeDecl = cpg.typeDecl.fullName(".*Function2Impl.*").head
+    val invokeMethod      = syntheticTypeDecl.method.name("invoke").head
```
ml86: All code outside of an actual concrete test that references `cpg` must be evaluated lazily because it will otherwise trigger the CPG generation even if none of theses tests are selected. The `cpg` value itself is an exception and does not need to be lazily evaluated because the CPG is only created open first us of `cpg`. For details look into the `KotlinCode2CpgFixture` and its super classes.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,692 @@
...
+      processCall.dispatchType shouldBe "DYNAMIC_DISPATCH"
+
+      // Should have fieldAccess
+      val receiverAccess = processCall.receiver.isCall.name(Operators.fieldAccess).head
+      receiverAccess.typeFullName shouldBe "com.test.Handler"
+      receiverAccess.ast.isFieldIdentifier.canonicalName.head shouldBe "receiver"
+      receiverAccess.ast.isIdentifier
+        .name("this")
+        .head
+        .typeFullName shouldBe "kotlin.jvm.internal.CallableReference"
```
ml86: Static type of `this` should be the name of the generated wrapper class.

In order to test the field identifier and the `this` identifier better use `receiverAccess.argument(2).asInstanceOf[FieldIdentifier]...` and `receiverAccess.argument(1).asInstanceOf[Identifier]...`. That way you encode the expected positions and check that these are really direct AST children of the call instead of being somewhere in the AST below it.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,692 @@
...
+    }
+
+    "create a constructor call for the synthetic type with receiver as parameter" in {
+      val ctorCalls = cpg.method("test").call.methodFullName(".*Function2Impl.*<init>.*").l
+      ctorCalls.size shouldBe 1
+
+      val ctorCall = ctorCalls.head
+      ctorCall.methodFullName shouldBe "com.test.Handler.process$kotlin.jvm.functions.Function2Impl.invoke:boolean(int,java.lang.String).<init>:void(com.test.Handler)"
```
ml86, line 98: Fullname seems longer than necessary. Seems like `.invoke:boolean(int,java.langString)` can be omitted without the danger of a naming collisions.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,692 @@
...
+
+      val ctorModifiers = constructor.modifier.l
+      ctorModifiers.size shouldBe 1
+      ctorModifiers.head.modifierType shouldBe "CONSTRUCTOR"
+    }
+
+    "have constructor body that assigns receiver field" in {
+      val constructor = syntheticTypeDecl.method.name("<init>").head
+
+      constructor.ast.isCall.name("<init>").methodFullName(".*CallableReference.*").l shouldBe List.empty
```
ml86: What is this supposed to test?

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,692 @@
...
+        |    }
+        |}
+        |
+        |fun test() {
+        |   val ref: (Int) -> Boolean = Utils::validate
+        |}
+        |""".stripMargin)
+
+    val syntheticTypeDecl = cpg.typeDecl.fullName(".*Function1Impl.*").head
+    val invokeMethod      = syntheticTypeDecl.method.name("invoke").head
```
ml86: Again make these lazily evaluated.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,692 @@
...
+
+    "use STATIC_DISPATCH without receiver access" in {
```
ml86, line 191: Validation of static call target is missing. Check `call.methodFullName`.

ml86: I would also like to see validation that the companion object is given as argument 0 to the `validate` call.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,692 @@
...
+    }
+
+    "create a binding from the synthetic type to the original function" in {
+      val bindings = cpg.all
+        .collectAll[Binding]
+        .nameExact("invoke")
+        .l
+        .filter(_.bindingTypeDecl.fullName.contains("globalFunction"))
+        .sortBy(_.methodFullName)
+      bindings.size shouldBe 2
```
ml86, line 289: Better to select the type decl you are interested in and than use `.methodBinding`.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
```diff
@@ -1,30 +1,692 @@
...
+  MethodParameterIn,
+  MethodRef
+}
+import io.shiftleft.codepropertygraph.generated.Operators
 import io.shiftleft.semanticcpg.language.*
 
 class CallableReferenceTests extends KotlinCode2CpgFixture(withOssDataflow = false) {
 
-  "CPG for code with simple callback usage" should {
+  "SAM implementation for bound instance method reference" should {
```
ml86, line 11: A binding table test for the generated type decl seems to be missing.

ml86 (changes requested):

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: Sorry, was a leftover from https://github.com/joernio/joern/pull/5775#discussion_r2862853370. Will remove.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: If `com.test.Handler.process` is overloaded, it might cause a collision, so at the very least the `.invoke` can be omitted, but I think the signature should stay.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForExpressionsCreator.scala
SuperUserDone: That is correct. Will change that. I cannot remember why I wrote it like this.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: Yes, only for the unbound case currently exist. Will add for the bound case.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
SuperUserDone: Will also explicitly set the argumentIndex in the implementation code to ensure this.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/CallableReferenceTests.scala
ml86: You are right, somehow we need the signature of `process` in the method full name.
Questions is whether we keep it like it is or if we do:
`com.test.Handler.process:boolean(int,java.lang.String)$kotlin.jvm.functions.Function2Impl.invoke.<init>:void(com.test.Handler)`. 
We can discus that in our next meeting.

PR #5776, opened 2026-01-22
[kotlin2cpg] Fix LHS of array access not treated as an expression

SuperUserDone:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForExpressionsCreator.scala
```diff
@@ -656,14 +656,13 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
   ): Ast = {
     val arrayExpr     = expression.getArrayExpression
     val typeFullName  = registerType(exprTypeFullName(expression).getOrElse(TypeConstants.Any))
-    val identifier    = identifierNode(arrayExpr, arrayExpr.getText, arrayExpr.getText, typeFullName)
-    val identifierAst = astWithRefEdgeMaybe(arrayExpr.getText, identifier)
+    val identifierAst = astsForExpression(arrayExpr, None)
```
ml86: Variable name should also be updated. Something like `baseExpression` seems appropriate.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/ArrayAccessExprsTests.scala
```diff
@@ -70,4 +70,24 @@ class ArrayAccessExprsTests extends KotlinCode2CpgFixture(withOssDataflow = true
...
+        |}
+        |""".stripMargin)
+
+    "should contain a listOf Call with the correct arguments" in {
+      cpg.call.name("listOf").length shouldBe 1
+      val List(call) = cpg.call.name("listOf").l
+      call.argument.length shouldBe 3
+      call.argument(1).code shouldBe "1"
+      call.argument(2).code shouldBe "2"
+      call.argument(3).code shouldBe "3"
```
ml86, line 98: This call to `listOf` could be anywhere in the CPG.
The test would have much better form if you start by selecting the array access call, expand to argument 0 and check if that is a call to `listOf`. From there you can then check its arguments like you already do.

PR #5779, opened 2026-01-23
[swiftsrc2cpg ] Fix try expression handling / Fix setter handling

max-leuthaeuser:
- the AST creation for Swift 'try' expressions was simplified to return the expression AST directly, preserving data-flow and removing explicit try modeling
- rewriting setter calls needs to be applied after the getter rewriting finished

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ExtensionsPass.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ExtensionsPass.scala
@@ -40,7 +40,7 @@ class ExtensionsPass(
 
   /** Creates a pass that rewrites computed-property member accesses to use the corresponding setter calls. We need to
     * separate this from the main pass as setter handling needs to happen after getter handling. (For the case where a
-    * setter is called on the result of a getter.)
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ExtensionsPass.scala
```diff
@@ -35,10 +35,23 @@ class ExtensionsPass(
     handleExtensionMembers(diffGraph)
     handleExtensionCalls(diffGraph)
     handleMemberPropertyGetterCalls(diffGraph)
-    handleMemberPropertySetterCalls(diffGraph)
     handleInherits(diffGraph)
   }
 
+  /** Creates a pass that rewrites computed-property member accesses to use the corresponding setter calls. We need to
+    * separate this from the main pass as setter handling needs to happen after getter handling. (For the case where a
+    * setter is called on the result of a getter.)
```
maltek: it took me a while to parse this. Is this example correct?
```suggestion
    * setter is called on the result of a getter, i.e. `foo.bar.baz = ...`.)
```

max-leuthaeuser: Exactly.

PR #5782, opened 2026-01-26
[php2cpg] feat: support scope resolution operators

TNSelahle:
Add support for `static` and `parent` class scope resolution operators. 

Relates to https://github.com/ShiftLeftSecurity/codescience/issues/8640

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
@@ -1,11 +1,11 @@
 package io.joern.php2cpg.querying
 
 import io.joern.php2cpg.astcreation.AstCreator.NameConstants
-import io.joern.php2cpg.testfixtures.PhpCode2CpgFixture
 import io.joern.php2cpg.parser.Domain
+import io.joern.php2cpg.testfixtures.PhpCode2CpgFixture
 import io.joern.x2cpg.Defines
+import io.shiftleft.codepropertygraph.generated.nodes.{Call, Identifier, Literal, Method}
 import io.shiftleft.codepropertygraph.generated.{DispatchTypes, Operators}
-import io.shiftleft.codepropertygraph.generated.nodes.{Call, FieldIdentifier, Identifier, Literal}
 import io.shiftleft.semanticcpg.language.*
 
 class CallTests extends PhpCode2CpgFixture {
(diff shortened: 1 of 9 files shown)


joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
+    "contain a dynamically dispatched call" in {
+      inside(cpg.call("bar").l) { case (call: Call) :: Nil =>
+        call.methodFullName shouldBe "Foo.bar"
+        call.code shouldBe "static::bar()"
+        call.dispatchType shouldBe DispatchTypes.DYNAMIC_DISPATCH
+      }
+    }
+
+    "have 'this' as the call receiver" in {
+      inside(cpg.call("bar").receiver.isIdentifier.l) { case (identifier: Identifier) :: Nil =>
```
johannescoetzee, line 529: Please check that the `this` argument has a `REF` edge to the `this` parameter of the enclosing method

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
+        call.dispatchType shouldBe DispatchTypes.DYNAMIC_DISPATCH
+
+        inside(call.receiver.isIdentifier.l) { case (identifier: Identifier) :: Nil =>
+          identifier.name shouldBe NameConstants.StaticReceiver
+          identifier.code shouldBe NameConstants.Static
+        }
+      }
+    }
+
+    "contain <staticReceiver> as argument 0 of static functions" in {
```
johannescoetzee:
```suggestion
    "contain <staticReceiver> as parameter 0 of static functions" in {
```

johannescoetzee: Just a nit pick

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
+
+    "contain a dynamically dispatched call" in {
+      inside(cpg.call("bar").l) { case (call: Call) :: Nil =>
+        call.methodFullName shouldBe "Foo.bar"
+        call.code shouldBe "static::bar()"
+        call.dispatchType shouldBe DispatchTypes.DYNAMIC_DISPATCH
+
+        inside(call.receiver.isIdentifier.l) { case (identifier: Identifier) :: Nil =>
+          identifier.name shouldBe NameConstants.StaticReceiver
+          identifier.code shouldBe NameConstants.Static
```
johannescoetzee: Are we going to add special support for this to the dataflow engine? I could be missing something, but it looks like the `<staticReceiver>` call receiver will be a dead-end for dataflow tracking as it is currently

johannescoetzee (changes requested):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
TNSelahle: Yeah, but the details I discussed with @ml86 around this are escaping me :D. I'm gonna chat about it in my 1-on-1 with him later today.

ml86: `<staticReceiver>` is from the data flow engines perspective just a normal parameter and `static::` calls are normal dynamic call sites. This is just a big fat misnomer from the PHP people.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
+      }
+    }
+
+    "contain <staticReceiver> as argument 0 of static functions" in {
+      inside(cpg.method.isStatic.l) { case (fooMethod: Method) :: (barMethod: Method) :: _ :: Nil =>
+        fooMethod.name shouldBe "foo"
+        fooMethod.parameter.headOption.map(_.name) shouldBe Some(NameConstants.StaticReceiver)
+
+        barMethod.name shouldBe "bar"
+        barMethod.parameter.headOption.map(_.name) shouldBe Some(NameConstants.StaticReceiver)
```
ml86: First parameter returned by `.parameter` step is not guaranteed to be the one with order 0. You have to filter for that.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
+        call.code shouldBe "self::bar()"
+        call.dispatchType shouldBe DispatchTypes.STATIC_DISPATCH
+      }
+    }
+
+    "have the correct receivers" in {
+      inside(cpg.call("bar").l) { case (call: Call) :: Nil =>
+        call.staticReceiver shouldBe Some("Foo")
+      }
+    }
```
ml86, line 564: `<staticReceiver>` parameter of `foo` needs to be argument 0 to `bar` call. At least the check for that is missing.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
```
ml86, line 602: `$this` parameter of `foo` needs to be argument 0 to `bar` call. At least the check for that is missing.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
```
ml86, line 640: `$this` parameter of `foo` needs to be argument 0 to `bar` call. At least the check for that is missing.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
+        call.code shouldBe "parent::bar()"
+        call.dispatchType shouldBe DispatchTypes.STATIC_DISPATCH
+      }
+    }
+
+    "have the correct receivers" in {
+      inside(cpg.call("bar").l) { case (call: Call) :: Nil =>
+        call.staticReceiver shouldBe Some("Base")
+      }
+    }
```
ml86, line 680: Like above the argument 0 check is missing.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
```
ml86, line 720: Like above the argument 0 check is missing.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
```
ml86, line 760: Like above the argument 0 check is missing.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
```
ml86, line 818: Like above the argument 0 check is missing.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -383,4 +383,255 @@ class CallTests extends PhpCode2CpgFixture {
...
+
+    "be a statically dispatched call to the base class" in {
+      inside(cpg.call("bar").l) { case (call: Call) :: Nil =>
+        call.methodFullName shouldBe "Base.bar"
+        call.code shouldBe "parent::bar()"
+        call.dispatchType shouldBe DispatchTypes.STATIC_DISPATCH
+      }
+    }
+
+    "have the correct receivers" in {
```
ml86: Here and in all the other tests above: Please change the test name to "have the correct staticReceiver property."

ml86 (changes requested):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
johannescoetzee: For the slightly extended example

```
<?php
class Foo {
  public static function foo($test) {
    static::bar($test);
  }

  private static function bar($value) { echo $value; }
}

Foo::foo("hello");
```

my concern is how we get the type of `<staticReceiver>` from the `Foo::foo` call. I spoke @TNSelahle and I did miss the fact that static functions now have the `<staticReceiver>` parameter 0, but I don't think we're adding a corresponding argument at the call site. We have the `STATIC_RECEIVER` field which we could use for this, but we'd need to add special backend handling for that, as far as I am aware.

ml86: In this example we use TYPE_REF Foo as argument 0 for the `Foo::foo` call.

johannescoetzee: That is missing from this PR (foo only has the literal argument). @TNSelahle could you please add a testcase for the static call arguments, along with providing the type ref for that?

PR #5784, opened 2026-01-28
[swiftsrc2cpg] Fix handling of trailing closures on static function calls

max-leuthaeuser:
Also fixed extension fullName handling.

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
@@ -660,22 +660,19 @@ case class SwiftTypesProvider(config: Config, parsedSwiftInvocations: Seq[Seq[St
     * @return
     *   The normalized fullName with extension qualifiers rewritten.
     */
-  private def replaceExtensionInSignature(fullName: String): String = {
-    @scala.annotation.tailrec
-    def loop(cur: String): String = {
-      ExtensionInSignatureRegex.findFirstMatchIn(cur) match {
-        case Some(m) =>
-          val head     = m.before
-          val name     = m.group(1)
-          val tail     = m.after.toString
-          val rest     = tail.stripPrefix(s"$name.")
-          val replaced = s"$head$name.$rest"
-          loop(replaced)
-        case None =>
-          cur
-      }
+  @scala.annotation.tailrec
+  private[utils] final def replaceExtensionInSignature(fullName: String): String = {
+    ExtensionInSignatureRegex.findFirstMatchIn(fullName) match {
+      case Some(m) =>
+        val head     = m.before
+        val name     = m.group(1)
+        val tail     = m.after.toString
+        val rest     = tail.stripPrefix(s"$name.")
+        val replaced = s"$head$name.$rest"
+        replaceExtensionInSignature(replaced)
+      case None =>
+        fullName
     }
-    loop(fullName)
   }
 
   /** Resolves a TypeInfo object to a ResolvedTypeInfo by demangling type and declaration fullNames.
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
```diff
@@ -648,18 +648,34 @@ case class SwiftTypesProvider(config: Config, parsedSwiftInvocations: Seq[Seq[St
...
-      fullName,
-      m => {
-        val head = m.before
-        val name = m.group(1)
-        val rest = m.after.toString.stripPrefix(s"$name.")
-        // Only extension methods are tagged with <extension> in their fullName.
-        // The signature only contains the module and type fullNames.
-        s"$head$name.$rest"
+    @scala.annotation.tailrec
+    def loop(cur: String): String = {
```
maltek: we don't need this `loop` function - we can just define `replaceExtensionInSignature` with `@tailrec` (we may need to declare it `final` for that to go through)

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/utils/ReplaceExtensionInSignatureTests.scala
```diff
@@ -0,0 +1,37 @@
+package io.joern.swiftsrc2cpg.utils
+
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+class ReplaceExtensionInSignatureTests extends AnyFunSuite with Matchers {
+
+  private def callReplaceExtensionInSignature(fullName: String): String = {
```
maltek: I don't like this. Let's just mark the method `private[utils]` to allow the tests to access it.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
```diff
@@ -648,18 +648,34 @@ case class SwiftTypesProvider(config: Config, parsedSwiftInvocations: Seq[Seq[St
...
+      ExtensionInSignatureRegex.findFirstMatchIn(cur) match {
```
maltek: are there any types with `)` in them that can have an extension defined? Maybe tuples or function types? Then the regex would stop too soon again, at the first `)`

max-leuthaeuser:
> I don't fully understand the extension fullnames:
> 
> in the closure tests we emit a fullname with `<extension>`:
> 
> ```
> SwiftTest.Swift.String<extension>.withPrefix:(_:Swift.String,transform:(Swift.String)->Swift.String)->Swift.String
> ```
> 
> but in the extension replacer we emit fullnames without `<extension>` like:
> 
> ```
> FooExt.Foo.bar() -> Swift.String
> ```
> 
> shouldn't they be the same?

The `<extension>` is only emitted on calls. In your example: `withPrefix` is a call in `SwiftTest.Swift.String<extension>`.
The unit test tests the a type fullName where we never tag something and replace only.

I'll update the tests to reflect that.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
max-leuthaeuser: No. Swift extension can only be declared for nominal types (e.g., struct, class, enum, protocol, and some built-in types like Optional, Array, etc.). A function type like `(Int) -> String` is not a nominal type, so you cannot write an extension on it.

For tuples it was also rejected for the time being:
https://github.com/swiftlang/swift-evolution/pull/290

maltek (approved):

PR #5786, opened 2026-01-28
[swiftsrc2cpg] Fix handling of implicit member expressions

max-leuthaeuser:
Swift's documentation refers to this as "implicit member expression" or "shorthand syntax for enumeration cases". This syntax is not limited to enums. It works with several Swift types where the compiler can infer the type from context to access static members. This is a commonly used pattern.
For enums only we could emit a Literal AST node representing the enum member. But as it works for other types as well, we emit an Unknown node here to avoid making potentially wrong assumptions.

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
@@ -8,7 +8,7 @@ import io.joern.x2cpg.datastructures.VariableScopeManager
 import io.joern.x2cpg.frontendspecific.swiftsrc2cpg.Defines
 import io.joern.x2cpg.{Ast, ValidationMode}
 import io.shiftleft.codepropertygraph.generated.*
-import io.shiftleft.codepropertygraph.generated.nodes.{ExpressionNew, NewCall, NewMethodParameterIn}
+import io.shiftleft.codepropertygraph.generated.nodes.{ExpressionNew, NewCall}
 
 import scala.annotation.unused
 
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
```diff
@@ -356,10 +356,17 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
         case other if isRefToClosure(node, other) =>
           astForClosureCall(node)
         case declReferenceExprSyntax: DeclReferenceExprSyntax if code(declReferenceExprSyntax) != "self" =>
-          val selfTpe  = fullNameOfEnclosingTypeDecl()
-          val selfNode = identifierNode(declReferenceExprSyntax, "self", "self", selfTpe)
-          scope.addVariableReference(selfNode.name, selfNode, selfTpe, EvaluationStrategies.BY_REFERENCE)
-          handleCallNodeArgs(node, Ast(selfNode), calleeCode)
+          if (
+            !scope.lookupVariable(calleeCode).exists(_._1.isInstanceOf[NewMethodParameterIn]) &&
```
maltek: this condition seems incorrect - functions as values can be the result of all kinds of other expressions, too. In the case where we have a simple name (which I guess is a `DeclReferenceExprSyntax`?), it could at least also be a local or captured variable. Or a member variable - which requires a different 'calling convention' than a member function.

max-leuthaeuser: That should have not been in this PR to begin with, sorry. I reverted that. Call handling for these cases needs to be addressed separately.

PR #5787, opened 2026-01-29
[swiftsrc2cpg] Remove call argument and parameter names

max-leuthaeuser:
Call arguments use the ARGUMENT_LABEL property now.

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForSyntaxCreator.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForSyntaxCreator.scala
@@ -258,25 +258,14 @@ trait AstForSyntaxCreator(implicit withSchemaValidation: ValidationMode) { this:
   private def astForKeyPathSubscriptComponentSyntax(node: KeyPathSubscriptComponentSyntax): Ast = notHandledYet(node)
 
   private def astForLabeledExprSyntax(node: LabeledExprSyntax): Ast = {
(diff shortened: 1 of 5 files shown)


max-leuthaeuser:
> I guess you will create a different PR then which stops using the label for the parameter names?

That was the plan but the commits are small. I just added it here.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForSyntaxCreator.scala
```diff
@@ -258,14 +257,24 @@ trait AstForSyntaxCreator(implicit withSchemaValidation: ValidationMode) { this:
...
+
+    /** In general, labeled expressions are just syntactic sugar and can be ignored when creating the AST. However, in
+      * `Package.swift` files, they are used to name the arguments for package dependencies. Hence, we handle labeled
+      * expressions in Package.swift files differently to capture the argument names for package dependencies. See:
+      * [[io.joern.swiftsrc2cpg.passes.DependenciesPass]] and its test `DependenciesPassTest` for more details.
+      *
+      * We don't have types/fullNames nor do we do anything dataflow-related there anyway, so it's safe to just add the
+      * argument names.
+      */
+    if (parserResult.filename == "Package.swift") {
```
maltek: we don't have any guarantees that the project root isn't a subdirectory of the directory that gets scanned, so strip off the directory name for this check. Also, on mac and Windows this comparison should be case-insensitive. (Unfortunately, I don't see an easy way to query this in the JVM world. It may be easiest - and more consistent - to just always use a case insensitive check.)

PR #5788, opened 2026-01-30
[javasrc2cpg] Add support for switch expressions

johannescoetzee:
This PR adds support for java switch expressions, for example
```java
int x = switch (o) {
  case String s -> 0;
  default -> 1;
}
```

as well as yield statements in case bodies. 

This represents switch expressions as `MATCH` control structures, similar to PHP, and yield statements as a `YIELD` control structure, which doesn't seem to be used by any of the joern frontends, but exists in the CPG schema already. We will probably need to add at least proper CFG support to handle this properly, but for now it isn't breaking any of our integration or sptests: https://buildbot.global.sltr.io/#/builders/82/builds/384


joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/ControlStructureTests.scala
```diff
@@ -898,4 +900,282 @@ class ControlStructureTests extends JavaSrcCode2CpgFixture {
...
+
+                defaultResult.code shouldBe "\"other\""
+                defaultResult.typeFullName shouldBe "java.lang.String"
+                defaultResult.order shouldBe 5
+            }
+          }
+        }
+      }
+    }
+  }
```
ml86: Looks to me that those two test cases are basically the same as the very first one except that the `switch` expression itself is the child of different AST nodes. Is there a reason you test that separately? I would have hoped that the hole handling of the `switch` sub AST is completely independent of its AST parent node.

johannescoetzee: Good point. I've done it this way for other nodes where it does matter (notably constructor invocations), but you're right that it shouldn't be the case here. I'll remove one of the two

ml86 (approved):

PR #5790, opened 2026-02-02
[jssrc2cpg] Fixed faulty identifiers from class method parameter decorators

max-leuthaeuser:


ml86 (approved):

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/TsDecoratorAstCreationPassTests.scala
```diff
@@ -448,6 +448,18 @@ class TsDecoratorAstCreationPassTests extends AstJsSrc2CpgSuite(".ts") {
...
+    "not have identifier used multiple times on class method parameters with multiple decorators" in {
+      val cpg = code("""
+          |export class Component {
+          |  constructor(
+          |    @Optional() @Inject(FOO) data: Data
+          |  ) {}
+          |}
+          |
+          |""".stripMargin)
+      cpg.identifier.filter(_._astIn.size != 1) shouldBe empty
```
maltek, line 461: I know there are only so many hours in the day. But it would be really nice if we could get to the point where the [post-frontend validator](https://github.com/joernio/joern/pull/5695) was actually enabled for jssrc (and other frontends) to warn us about such errors.

maltek (approved):

joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/passes/ast/TsDecoratorAstCreationPassTests.scala
max-leuthaeuser: For jssrc at least we could do that right now. I will do a PR for that.

PR #5793, opened 2026-02-02
[pysrc2cpg] Add missing tests for `for` statement.

ml86:


maltek (approved):
there is also the `for x, *y, z` syntax for unpacking some elements from the head/tail and a list for the rest that isn't tested here

PR #5795, opened 2026-02-02
[javasrc2cpg] Improve lambda capture handling

johannescoetzee:
The primary purpose of this PR is to fix lambda variable capturing in 2 cases:
1. Nested lambdas where the inner lambda uses a variable captured from the outer scope outside the outer lambda. This is done by adding locals and closure bindings for the captured variable to all of the intermediate lambdas as well. This case is probably quite rare.
2. Lambdas capturing fields from the enclosing type decl. This is likely more common and is handled by adding a `this` local to the lambda method with a closure binding to the `this` parameter of the enclosing method. 

For the sake of implementation, I added the `CaptureUseFinder` class (I'm open to suggestions about the name) which, given an AST node, returns all used names without a corresponding declaration in scope in the subtree with the given node as a root. his is used to find captured variables and add the relevant locals to the scope before processing the lambda body, which means we don't need to keep track of this in a separate structure apart from the usual scope variables lists.

Results from the `CaptureUseFinder` are cached to ensure the AST for a given file is only traversed once but, since it's only being used for lambdas at the moment, it's unlikely that this would make a big difference anyways. The plan is to refactor other capture handling to make use of the `CaptureUseFinder` as well, which would significantly simplify capture handling logic and especially constructor creation overall.

The interaction between lambda captures and anonymous class captures is not handled by this. It's still undecided how exactly we want to proceed with this, since adding (probably complex) handling for that only to throw it out if we choose a common capture representation is something we'd like to avoid.

I've also added ignored tests for 2 examples for which the AST isn't being created correctly. The first, and more significant, is for `LocalRecordDeclarations`, which aren't being added to the CPG at all (only an `Unknown` node is added). The second is for qualified `this` expressions, for example 
```
class Outer {
  String outerField;
  class Inner {
    String innerField = Outer.this.outerField;
  }
}
```

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/expressions/AstForLambdasCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/expressions/AstForLambdasCreator.scala
@@ -326,7 +326,7 @@ private[expressions] trait AstForLambdasCreator { this: AstCreator =>
     scope.pushBlockScope()
     val outerScopeVariableNames = variablesInScope.map(x => x.name -> x).toMap
 
-    val capturedVariableUses = captureUseFinder.getCaptureUses(lambdaExpr)
+    val capturedVariableUses = captureUseFinder.getUndeclaredVariables(lambdaExpr)
     var thisCaptureHandled   = false
     val closureBindingsAndLocalsForCaptures = capturedVariableUses.toList.sorted.flatMap { name =>
       val capturedResult = scope.lookupVariable(name)
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/CaptureUseFinder.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/CaptureUseFinder.scala
@@ -29,49 +29,63 @@ import scala.jdk.CollectionConverters.*
 import scala.jdk.OptionConverters.*
 import scala.util.Try
 
+/** The CaptureUseFinder does a conceptually bottom-up (practically top-down post-order reverse) traversal of the given
+  * AST and adds all variable names not defined within that AST (i.e. captured from scope surrounding this subtree).
+  * This effectively means traversing the source code in reverse starting at the end of the given structure. When it
+  * encounters an identifier, that identifier is added to the undeclaredVariables set since a corresponding declaration
+  * could not have been seen yet. Then, when a declaration is seen, the matching name is removed from undeclared
+  * variables since the corresponding NameExpr is no longer capturing anything from the already processed scope.
+  *
+  * The actual implementation is a bit muddier due to the use of mutable sets to avoid the performance cost of
+  * constantly creating and merging new sets. This means that at some pre-order points, new sets for undeclared
+  * variables need to be created to handle nodes that create new scope (blocks, methods, etc.). Once these subtrees have
+  * been fully processed, these temporary sets are merged into the main set keeping track of undeclared variables.
+  *
+  * Once a node declaring a new type has been processed, the results are cached, so a given file should only be
+  * traversed once
+  */
 class CaptureUseFinder {
 
   private val cache: mutable.Map[Node, Set[String]] = mutable.Map()
   // Not actually used, but is needed as an argument to the context constructors
   private val typeSolver = new CombinedTypeSolver()
 
-  def getCaptureUses(node: Node): Set[String] = {
+  def getUndeclaredVariables(node: Node): Set[String] = {
     cache.get(node) match {
-      case Some(captureUses) => captureUses
+      case Some(undeclaredVariables) => undeclaredVariables
 
       case None =>
-        val captureUses = mutable.Set[String]()
-        storeCaptureUses(node, captureUses)
-        captureUses.toSet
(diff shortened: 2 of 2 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/expressions/AstForLambdasCreator.scala
```diff
@@ -298,13 +323,74 @@ private[expressions] trait AstForLambdasCreator { this: AstCreator =>
     variablesInScope: Seq[ScopeVariable],
     returnType: Option[String]
   ): LambdaBody = {
+    scope.pushBlockScope()
     val outerScopeVariableNames = variablesInScope.map(x => x.name -> x).toMap
 
+    val capturedVariableUses = captureUseFinder.getCaptureUses(lambdaExpr)
+    var thisCaptureHandled   = false
+    val closureBindingsAndLocalsForCaptures = capturedVariableUses.toList.flatMap { name =>
```
ml86: This is an iteration of an unordered set. Please make the iteration order deterministic.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/CaptureUseFinder.scala
```diff
@@ -0,0 +1,225 @@
...
+      case Some(captureUses) => captureUses
+
+      case None =>
+        val captureUses = mutable.Set[String]()
+        storeCaptureUses(node, captureUses)
+        captureUses.toSet
+    }
+  }
+
+  // This method does a bottom-up traversal of the given AST and adds all variable names not defined within that AST
```
ml86: Looking at the code of the method, this does not look like bottom-up but rather top-down.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/util/CaptureUseFinderTests.scala
```diff
@@ -0,0 +1,543 @@
...
+  *   - Qualified this (Outer.this) not properly distinguished
+  *
+  * Working correctly:
+  *   - TryStmt resources scoping
+  *   - LocalRecordDeclarationStmt scoping
+  *   - Pattern matching in loop conditions
+  *   - Fields of local classes not leaked
+  *   - Instance initializers scoping
+  */
+class CaptureUseFinderTests extends JavaSrcCode2CpgFixture with Inside {
```
ml86: I would not name this class after the name of an individual implementation detail but rather something like `CapturingTests`.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/CaptureUseFinder.scala
ml86: How is the interaction between `.add` and `.remove` on `captureUses` supposed to work? Why do we need both and not just `.add`?

johannescoetzee: > Looking at the code of the method, this does not look like bottom-up but rather top-down.

Hmm, I guess this is true. The way I thought about it is bottom up, but I guess the implementation is top-down (mostly) post-order (where children are iterated over in reverse). 

My prototype was initially implemented with a proper top-down algorithm where I'd keep track of declarations in scope and only add those names that don't have a corresponding declaration (similar to what we do for the regular scope). This made caching the results difficult though, so I switched to the approach implemented here. 

The basic idea to traverse the source code in reverse, starting at the right-most leaf in the AST and traversing in reverse. If a `NameExpr` is found, it is always added to the `captureUses` set since we could not have seen a declaration for it yet (since the declaration would have to appear before it in source code and we're traversing backwards. Then, when we get to a declaration, we can remove the matching name from `captureUses` since we've seen the declaration and it is no longer capturing anything from outside the scope we've processed. This has the benefit that when we've visited all the children of a typeDecl, anonymousClass, etc. we can just add the results as they are at that point in processing to the cache.

The actual implementation is a bit muddier since I chose to pass around a mutable set to keep track of results to avoid having to constantly create and merge new sets. This means that at points where a new scope is created (blocks, method declarations, etc.) temporary sets are needed to handle that specific subtree before merging the partial results with the overall `captureUses`. That's being done manually here, but I could hide this in some scope structure or we could accept the (probably minor) performance hit for using immutable sets and go for the theoretically cleaner approach.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/util/CaptureUseFinder.scala
```diff
@@ -0,0 +1,225 @@
...
+  // (i.e. captured from scope surrounding this subtree). When it encounters an identifier, that identifier
+  // is added to the captureUses set
+  private def storeCaptureUses(node: Node, captureUses: mutable.Set[String]): Unit = {
```
ml86: I would name `captureUses` `undeclaredVariables`.

johannescoetzee: That is a much better name!

PR #5796, opened 2026-02-03
[jssrc2cpg] Remove AstJsSrc2CpgSuite and friends

max-leuthaeuser:


joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/io/TranspiledFileDetectionTests.scala
```diff
@@ -25,11 +25,17 @@ class TranspiledFileDetectionTests extends AstJsSrc2CpgSuite {
       cpg.file.name.l shouldBe List("index.ts")
     }
 
-    "skip transpiled files correctly (with other file types)" in {
+    "skip ejs transpilation if there is a transpiled file" in {
       val cpg = code("console.log('Hello World!');", "index.vue")
         .moreCode("console.log('Hello World!');", "index.ejs")
         .moreCode("console.log('Hello World!');", "index.js")
         .moreCode("", "index.js.map")
+      cpg.file.name.l shouldBe List("index.js", "index.vue")
```
max-leuthaeuser, line 33: This is now fixed - it was wrong the whole time. The tests did not catch this as the AstJsSrc2CpgSuite used the wrong paths for AstGenRunner.
We won't transpile ejs if the transpiled result is already available.

PR #5802, opened 2026-02-03
Use sizeof call for javasrc array.length

johannescoetzee:
With this PR, javasrc will also represent native `array.length` field accesses as calls to `<operators>.sizeOf`, consistent with `java2cpg`. This is done to fix an access path conflict between field accesses and index accesses on arrays leading to duplicate flows, for example 
```
for (int i = 0; i < values.length /* this field access */; i++) {
  sink(values[i]); // and this index access are conflated in a wildcard access path
}
```

This did break open-source dataflows since `values.length` in the condition is no longer a transparent field access. The fix here is to just say that `sizeOf` is effectively a `.length` field access and to add it to `MemberAccess.isFieldAccess`, but I don't know if this is the best solution.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/ArrayTests.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/ArrayTests.scala
@@ -140,4 +140,27 @@ class ArrayTests extends JavaSrcCode2CpgFixture {
       index.code shouldBe "0"
     }
   }
+
+  "array length field access" should {
+    val cpg = code("""
+        |class Foo {
+        |  public void foo(String[] xs) {
+        |    int l = xs.length;
+        |  }
+        |}""".stripMargin)
+
+    "be represented as <operators>.sizeOf calls" in {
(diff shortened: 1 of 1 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/MemberAccess.scala
```diff
@@ -32,7 +32,9 @@ object MemberAccess {
     (name == Operators.indirectFieldAccess) ||
     (name == Operators.indexAccess) ||
     (name == Operators.indirectIndexAccess) ||
-    (name == Operators.getElementPtr)
+    (name == Operators.getElementPtr) ||
+    // sizeOf is added here since it is effectively used as `array.length`
+    (name == Operators.sizeOf)
```
johannescoetzee, line 37: @fabsx00 Are you OK with this change or can you think of a better solution?

ml86 (approved):
Please wait a few more days to give @fabsx00 a chance to respond.

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/MemberAccess.scala
fabsx00: Yeah, that's ok

PR #5805, opened 2026-02-05
[x2cpg] Reject files based on configurable max. file size

max-leuthaeuser:
Max. size is `Integer.MAX_VALUE - 2L`.
Max. size for `Array[Byte]` in JVM is `Integer.MAX_VALUE`, but `String` can hold at most `Integer.MAX_VALUE - 2` bytes.

Frontend devs may configure anything up to that value or an env. var may be passed.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
@@ -17,12 +17,7 @@ object SourceFiles {
   // Max. size for Array[Byte] in JVM is Integer.MAX_VALUE, but String can hold at most Integer.MAX_VALUE - 2 bytes
   private[x2cpg] val DefaultMaxFileSizeBytes: Long = Integer.MAX_VALUE - 2L
 
-  // Human-readable representation for log messages; derived from DefaultMaxFileSizeBytes.
-  private def DefaultMaxFileSizeString: String = {
-    val gib = 1024L * 1024L * 1024L
-    if (DefaultMaxFileSizeBytes % gib == 0) s"${DefaultMaxFileSizeBytes / gib}GiB"
-    else "2GB"
-  }
+  private def DefaultMaxFileSizeString: String = s"${DefaultMaxFileSizeBytes}B"
 
   /** A failsafe implementation of a [[FileVisitor]] that continues iterating through files even if an [[IOException]]
     * occurs during traversal.


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
```diff
@@ -8,11 +8,70 @@ import java.nio.file.{FileVisitOption, FileVisitResult, FileVisitor, Files, Path
...
+            val bytes = Math.min(Math.round(bytesDouble), DefaultMaxFileSizeBytes)
+            if (bytes < DefaultMinFileSizeBytes) Some(DefaultMinFileSizeBytes) else Some(bytes)
+          }
+        }.toOption.flatten
+      case _ => None
+    }
+  }
+
+  private def maxFileSizeFromEnv(): Long = {
+    sys.env.get("MAX_FILE_SIZE") match {
```
Ferada: This looks like a very generic key. Could we prefix that with something? E.g. `QWIET_MAX_FILE_SIZE`?

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
```diff
@@ -8,11 +8,70 @@ import java.nio.file.{FileVisitOption, FileVisitResult, FileVisitor, Files, Path
...
 
   private val logger = LoggerFactory.getLogger(getClass)
 
+  // Supported formats (case-insensitive):
+  //   - "123"
+  //   - "123MB"
+  //   - "2GB"
+  //   - "1.5GB"
+  //   - "0.25MB"
+  private val FileSizePattern = "(?i)^([0-9]+(?:\\.[0-9]+)?)\\s*(MB|GB)?$".r
```
Ferada: I would suggest also accepting single letters and the kilo prefix, that is: "k", "m" and "g".

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
```diff
@@ -113,6 +117,27 @@ object SourceFiles {
...
+    *   Path to the file to check.
+    */
+  private def isTooLarge(file: String): Boolean = {
+    val path = Paths.get(file)
+    if (Files.isDirectory(path) || !Files.isRegularFile(path)) return false
+    Try {
+      val size     = Files.size(path)
+      val tooLarge = size > DefaultMaxFileSizeBytes
+      if (tooLarge) {
+        logger.warn(s"'$file' ignored (too large: ${size}B > 2GB)")
```
ml86: Use `DefaultMaxFileSizeBytes` in log message instead of `2GB`.

joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/SourceFilesTests.scala
```diff
@@ -98,5 +104,44 @@ class SourceFilesTests extends AnyWordSpec with Matchers with Inside {
...
+      FileUtil.usingTemporaryDirectory() { tmpDir =>
+        val file  = tmpDir / "a.c"
+        val bytes = Array.fill[Byte](1024)(0)
+        Files.write(file, bytes)
+
+        SourceFiles.filterFile(file = file.toString, inputPath = tmpDir.toString) shouldBe true
+      }
+    }
+
+    "allow a file whose size is exactly JVM String max. size" taggedAs NotInMacOS in {
```
ml86, line 126: I are we not running this under MacOS?

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
```diff
@@ -8,11 +8,18 @@ import java.nio.file.{FileVisitOption, FileVisitResult, FileVisitor, Files, Path
...
 
 object SourceFiles {
 
   private val logger = LoggerFactory.getLogger(getClass)
 
+  // Max. size for Array[Byte] in JVM is Integer.MAX_VALUE, but String can hold at most Integer.MAX_VALUE - 2 bytes
+  private[x2cpg] val DefaultMaxFileSizeBytes = Integer.MAX_VALUE - 2L
+
+  // it actually is Integer.MAX_VALUE - 2 bytes, just for readability in log messages
+  private val DefaultMaxFileSizeString = "2GB"
```
ml86: Combine this into one define.

joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/SourceFilesTests.scala
max-leuthaeuser: Because for whatever reason the macOS runners here at GH crash out with a OOM exception when trying to allocate the arrays there. It works locally on our Macs.

max-leuthaeuser: Found the issue: https://github.com/actions/runner-images/issues/11899

ml86 (approved):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala
```diff
@@ -8,11 +8,22 @@ import java.nio.file.{FileVisitOption, FileVisitResult, FileVisitor, Files, Path
...
   private val logger = LoggerFactory.getLogger(getClass)
 
+  // Max. size for Array[Byte] in JVM is Integer.MAX_VALUE, but String can hold at most Integer.MAX_VALUE - 2 bytes
+  private[x2cpg] val DefaultMaxFileSizeBytes: Long = Integer.MAX_VALUE - 2L
+
+  // Human-readable representation for log messages; derived from DefaultMaxFileSizeBytes.
+  private def DefaultMaxFileSizeString: String = {
+    val gib = 1024L * 1024L * 1024L
+    if (DefaultMaxFileSizeBytes % gib == 0) s"${DefaultMaxFileSizeBytes / gib}GiB"
+    else "2GB"
```
ml86: No hardcoded things here. Otherwise this might not match the actual value! Print it in non human readable if you have to.

PR #5806, opened 2026-02-05
[kotlin2cpg] Add missing type-erased binding for generic class methods

SuperUserDone:
This is a dependency for #5775 

This PR solves an issue where the dataflow engine is unable to find the correct implementation of methods overridden in a generic superclass. This was because no binding for the type-erased signature was created for the class type declaration.

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForDeclarationsCreator.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForDeclarationsCreator.scala
@@ -27,6 +27,14 @@ import org.jetbrains.kotlin.psi.*
 import scala.collection.mutable
 import scala.jdk.CollectionConverters.*
 import scala.util.Random
+import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.descriptors.FunctionDescriptor
+import org.jetbrains.kotlin.descriptors.{CallableMemberDescriptor, SimpleFunctionDescriptor}
+import org.jetbrains.kotlin.types.SimpleType
+import scala.jdk.CollectionConverters.*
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForDeclarationsCreator.scala
```diff
@@ -262,6 +273,85 @@ trait AstForDeclarationsCreator(implicit withSchemaValidation: ValidationMode) {
...
+        }
+      case _ =>
+        nameRenderer.typeFullName(kotlinType)
+    }
+  }
+
+  /** Creates an erased signature for a function descriptor, replacing type parameters with their upper bounds.
+    */
+  private def createErasedSignature(functionDesc: FunctionDescriptor): Option[String] = {
+    val extRecvType = Option(functionDesc.getExtensionReceiverParameter) match {
```
ml86, line 332: I am surprised that`getExtensionReceiverParameter` is what we want here. Just judging by the name `getDispatchReceiverParameter` sounds like the one we would want to use?

SuperUserDone: As far as I could gather, the extension receiver is what we need because it's part of the function's actual type signature (`String.() -> Int` becomes `Int(String)` in JVM land). The dispatch receiver is the containing class for member functions, which isn't part of the callable type.  (meaning `this` is not explicitly declared as part of the sig)

Using dispatch receiver would add the declaring class to member function signatures when it shouldn't be there. Extension receiver captures the explicit receiver parameter that becomes part of the erased signature.

Please let me know if I am misunderstanding something.

ml86: Ok, thanks for the explanation. Sounds very much like you picked the right one.

PR #5808, opened 2026-02-05
[javasrc2cpg] Fix ref edge creation for captured variable argument to this call in constructor

johannescoetzee:
When local classes (also anonymous and nested classes) capture variables, those are represented as fields in the local class which are initialised in the constructor (with corresponding parameters added to the constructors). They are only initialised in constructors which do not contain a `this` constructor invocation themselves, however. If a constructor invokes another, then the parameter value for the first constructor is used as an argument for the second. 

Due to the order in which these were created and the information kept in scope, the `REF` edges for the argument to the `this` call were not pointing at the constuctor parameter, but rather at the original captured node in a different class. This PR fixes this issue by storing the necessary information in the scope as well.

Constructor method and invocation creation with captures is already spaghetti and this PR makes it worse, but it does fix the issue for now. The medium-term plan is to use the `CaptureUseFinder` introduced in https://github.com/joernio/joern/pull/5795 to pre-compute all of the information needed to handle these, avoiding the need to split creation into multiple steps which is the main cause of pasta

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/declarations/AstForMethodsCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/declarations/AstForMethodsCreator.scala
@@ -635,7 +635,7 @@ private[declarations] trait AstForMethodsCreator { this: AstCreator =>
           }
         Nil
       } else {
-paramsForCaptures.map(assignmentForCapture(partialConstructor.originNode, _, partialConstructor.thisNode))
+        paramsForCaptures.map(assignmentForCapture(partialConstructor.originNode, _, partialConstructor.thisNode))
       }
 
     val bodyAst =


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/declarations/AstForMethodsCreator.scala
```diff
@@ -624,9 +624,17 @@ private[declarations] trait AstForMethodsCreator { this: AstCreator =>
...
+        partialConstructor.bodyStatementAsts.headOption
+          .flatMap(_.root)
+          .collect { case initCall: NewCall => initCall }
+          .foreach { initCall =>
+            scope.enclosingTypeDecl.foreach(_.registerThisCallCaptureParams(initCall, paramsForCaptures))
+          }
         Nil
-      else
+      } else
         paramsForCaptures.map(assignmentForCapture(partialConstructor.originNode, _, partialConstructor.thisNode))
```
maltek: it may be a personal preference, but if one part of an `if/else` has braces I much prefer for the other part to have them as well
```suggestion
      } else {
paramsForCaptures.map(assignmentForCapture(partialConstructor.originNode, _, partialConstructor.thisNode))
      }

```

PR #5810, opened 2026-02-05
[swiftsrc2cpg] Fixed handling for offsets for unicode characters

max-leuthaeuser:


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -240,9 +240,9 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
   protected def methodInfoForFunctionDeclLike(node: FunctionDeclLike): MethodInfo = {
     val name = calcMethodName(node)
 
-    val nodeRange = node.json("range")
+    val nodeRange = node.asInstanceOf[SwiftNode].json("range").obj.copy()
```
maltek, line 246: does this get mutated, and must therefore be copied? Or is this a debug left-over?

max-leuthaeuser: Yes, it gets mutated. I'll put in an explanation.

PR #5811, opened 2026-02-09
[swiftsrc2cpg] Added support for Objective-C constructor calls

max-leuthaeuser:


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ObjcCallFullNamePass.scala
```diff
@@ -0,0 +1,37 @@
...
+    for {
+      call <- cpg.call.isStatic.nameExact("init").methodFullName(".+\\.init:\\(.*\\)->.+")
+      if call.typeFullName != PropertyDefaults.TypeFullName && AstCreatorHelper.isObjcCall(call.methodFullName)
+      constructorMethod <- cpg.typeDecl
+        .fullNameExact(call.typeFullName)
+        .method
+        .isConstructor
+        .filter(m =>
+          m.fullName.startsWith(s"${call.typeFullName}.init:") && m.fullName.endsWith(s")->${call.typeFullName}")
+        )
```
maltek: Constructors can be overloaded, so we need a check for the signature here.

Or, if signatures are unknown with objective c then we should be conservative and use `loneElementOption` to ensure we don't pick the wrong constructor.

PR #5812, opened 2026-02-09
[php2cpg] fix parser out-of-memory issue

TNSelahle:
Fixed sptest out-of-memory issue caused by the `php.ini` file needed by the parser being deleted before it is used. It contains a config to remove the memory limit, hence the out-of-memory issue we're now seeing.

sptests running successfully :+1: 

Fixes https://github.com/ShiftLeftSecurity/codescience/issues/8712

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
@@ -218,20 +218,13 @@ object PhpParser {
 
         f(parserOption)
       case _ =>
-        Try {
-          FileUtil.usingTemporaryFile(suffix = "-php.ini") { tmpIni =>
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/parser/PhpParser.scala
```diff
@@ -212,16 +212,27 @@ object PhpParser {
...
+            Files.writeString(tmpIni, iniContents)
+            val parserOption = getParser(config, tmpIni.absolutePathAsString)
+
+            f(parserOption)
+          }
+        } match {
+          case Success(result) => result
+          case Failure(exception) =>
+            logger.error(s"Failed to set up temporary php.ini: ${exception.getMessage}")
+            f(None)
```
maltek: I don't think we want to run `f` again, when `f` failed. And if something as fundamental as writing to a temporary file fails, I don't think we need a fallback code path for that either. So I would just remove the `Try` here.

PR #5815, opened 2026-02-11
[swiftsrc2cpg] Remove self param from closures

max-leuthaeuser:


joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/ClosureTests.scala
```diff
@@ -109,6 +109,38 @@ class ClosureTests extends SwiftSrc2CpgSuite {
       p2.typeFullName shouldBe "MyType"
     }
 
+    "testClosure6" in {
+      val cpg = code("""
+          |class A {
+          |  var x = 1
+          |  func foo(_ a: ANY, _ b: ANY, _ c: ANY, _ d: ANY, _ x: ANY) {}
+          |  func method() {
```
maltek: I don't think this is intentionally testing access of undefined variables, right?

```suggestion
          |  func method(_ a: ANY, _ b: ANY, _ c: ANY, _ d: ANY, _ x: ANY) {
```

PR #5816, opened 2026-02-11
[semanticcpg] check for `order` collisions between CFG nodes 

xavierpinho:
Adds a check to ensure that CFG nodes have no `order` collisions, as CfgCreatorPass processes its child nodes according to their `order` (and having collisions makes it ambiguous which statement is processed first.)

To prevent warnings on benign collisions, such as between LOCAL and CALL, we whitelist Declaration nodes.


max-leuthaeuser:
> Though it's probably not a big deal to emit them with unique orders either.

Did that here: https://github.com/joernio/joern/pull/5818

PR #5822, opened 2026-02-13
[swiftsrc2cpg] Handle Self access in static context

max-leuthaeuser:


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/VariableScopeManager.scala
```diff
@@ -372,6 +376,17 @@ class VariableScopeManager {
     }
   }
 
+  def isInStaticMethodScope: Boolean = {
```
maltek: Swift allows defining classes within a method-local scope. For those, the current logic is wrong. The good news is that Swift doesn't allow any access to `self` or `Self` from the outer scope within a nested/local class. There is no capturing at all for those, unlike with lambdas.

So for Swift we don't want to walk the whole stack and check for the existence of any `static` method. Instead, we want to walk up until we find the first non-lambda and simply check the static-ness of that method.


For reference, I've peen playing with variations of these snippets:

```swift
class Foo {
  static let msg2 = "world"
  static func foo() {
    class Bar {
      let msg1 = "hello"
      func bar() {
        print(msg1)
        print(Foo.msg2)
      }
    }
    let numbers = [1]

    numbers.forEach { num in
      print(Self.msg2)
      Bar().bar()
    }
  }
}
Foo.foo()
```

```swift
class Foo {
  let msg2 = "world"
  func foo() {
    let numbers = [1]

    numbers.forEach { num in
      Bar.fooInstance = self
      Bar.bar()
    }

    class Bar {
      static let msg1 = "hello"
      static var fooInstance: Foo? = nil

      static func bar() {
        print(msg1)
        print(fooInstance!.msg2)
      }
    }
  }
}
Foo().foo()
```

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/VariableScopeManager.scala
```diff
@@ -372,6 +377,23 @@ class VariableScopeManager {
...
```
maltek: this logic is very swift-specific - I'd move it out of x2cpg.

PR #5825, opened 2026-02-16
[kotlin2cpg] Avoid duplicate METHOD_REF argument edges

SuperUserDone:
Resolves part of: https://github.com/ShiftLeftSecurity/codescience/issues/8701

In a specific qualified-call shape, the same METHOD_REF was attached twice:
- once as the receiver argument, and
- once again via generic argument-edge emission.

This change prevents duplicate incoming ARGUMENT edges on a METHOD_REF in the qualified-call case where a single lambda method reference is treated as the call receiver.

Note: Currently lambda CPG emission in kotlin2cpg has some issues. This PR just fixes existing behavior.

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForExpressionsCreator.scala
@@ -344,14 +344,13 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
         },
         argIdx
       ).argumentName(argNameMaybe)
-    val useSingleMethodRefArgAsReceiver =
-      argAsts.sizeIs == 1 && argAsts.head.root.get.isInstanceOf[NewMethodRef]
-    val receiverNode =
-      if (useSingleMethodRefArgAsReceiver) argAsts.head.root.get
-      else receiverAst.root.get
-    val argumentNodes =
-      if (useSingleMethodRefArgAsReceiver) Seq.empty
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForExpressionsCreator.scala
```diff
@@ -344,15 +344,20 @@ trait AstForExpressionsCreator(implicit withSchemaValidation: ValidationMode) {
...
+      if (useSingleMethodRefArgAsReceiver) Seq.empty
+      else argAsts.map(_.root.get)
 
     Ast(node)
       .withChild(receiverAst)
       .withArgEdge(node, receiverNode)
       .withChildren(argAsts)
-      .withArgEdges(node, argAsts.map(_.root.get))
+      .withArgEdges(node, argumentNodes)
       .withReceiverEdge(node, receiverNode)
```
johannescoetzee: This could perhaps be done cleaner with a `match`, e.g.

```
argAsts match {
  case List(receiverAst) if receiverAst.root.exists(_.isInstanceOf[NewMethodRef]) =>
    callAst(node, args=Nil, receiver=receiverAst)
  case _ =>
    callAst(node, argAsts)
```

I know you didn't write the original code for this, but this is a good opportunity to clean up some old code

SuperUserDone: Have done the best I could. Using the callAst function breaks type resolution as it changes some argumentIndex values, but I have switched to using a match to clean up the code a bit instead of using the if/else block.

PR #5826, opened 2026-02-16
[kotlin2cpg] Make fake-global fullnames file-unique

SuperUserDone:
Resolves Partially: https://github.com/ShiftLeftSecurity/codescience/issues/8701

Multiple Kotlin files in the same package generated the same synthetic fake-global full names, causing collisions.

This change makes synthetic file-scope fake-global TYPE_DECL/METHOD full names file-unique instead of package-unique. Fake-global full names now include file identity (relativizedPath), so each file-level wrapper is uniquely named.

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
@@ -197,7 +197,7 @@ class Kotlin2Cpg extends X2CpgFrontend with UsesService {
     config: Config,
     filesWithJavaExtension: List[String],
     kotlinAstCreatorTypes: List[String]
-  ): Unit = {
+  ): List[String] = {
     if (config.includeJavaSourceFiles && filesWithJavaExtension.nonEmpty) {
       val javaAstCreator = JavaSrcInterop.astCreationPass(config.inputPath, filesWithJavaExtension, cpg)
       javaAstCreator.createAndApply()
@@ -206,9 +206,9 @@ class Kotlin2Cpg extends X2CpgFrontend with UsesService {
       javaAstCreator.sourceParser.cleanupDelombokOutput()
       javaAstCreator.clearJavaParserCaches()
 
-      TypeNodePass
-        .withRegisteredTypes((javaAstCreatorTypes.toSet -- kotlinAstCreatorTypes.toSet).toList, cpg)
-        .createAndApply()
+      javaAstCreatorTypes
+    } else {
+      List.empty
     }
   }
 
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstCreator.scala
```diff
@@ -366,14 +366,25 @@ class AstCreator(
       }
     methodAstParentStack.push(node)
 
-    val name     = NamespaceTraversal.globalNamespaceName
-    val fullName = node.fullName
+    val name                       = NamespaceTraversal.globalNamespaceName
+    val fullName                   = node.fullName
+    val fakeGlobalTypeDeclFullName = s"$fullName:$relativizedPath"
```
johannescoetzee: Small nit: in other frontends that use the file path as part of the full name, the path is the prefix (the equivalent of`$relativizedPath:$fullName`). I think it would be better to change this to stick to that convention.

joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/validation/DefaultImportsTests.scala
```diff
@@ -4,6 +4,14 @@ import io.joern.kotlin2cpg.testfixtures.KotlinCode2CpgFixture
 import io.shiftleft.semanticcpg.language.*
 
 class DefaultImportsTests extends KotlinCode2CpgFixture(withOssDataflow = false) {
+  // TODO HACK: windows paths are not supported in SourceFiles.toRelativePath, so we need to filter out files with windows paths in SourceFilesPicker.shouldFilter.
```
johannescoetzee: I'm worried this hack could hide a legitimate issue (either now or a regression in future). I had a quick look at `SourceFiles.toRelativePath` and didn't see anything obvious that would cause an issue. What was actually going wrong to cause the Windows failures?

SuperUserDone: Prefix matching fails to make the paths relative, so they remain absolute.

I have not fully triaged the issue yet, but I suspect one of these reasons:
- Paths on windows are case-insensitive
- shortened 8.3 path format

On Linux/Unix:
```
HashSet(ANY, void, kotlin.Function, <global>:Test0.kt, java.lang.String[], kotlin.Pair, int, java.lang.String, java.util.Map, main:void(java.lang.String[]), java.lang.Object, kotlin)
```

On windows:
```
HashSet("ANY", "void", "kotlin.Function", "java.lang.String[]", "int", "main:void(java.lang.String[])", "<global>:C:/Users/RUNNER~1/AppData/Local/Temp/x2cpgTestTmpDir7040576959906842737/Test0.kt", "java.lang.Object", "kotlin")
```

johannescoetzee: Hmm, this will be an issue for filenames in the CPG in general then and is something we definitely want fixed. I could have a look, or you could do it if you want, but I think we should just fix it instead of merging this hack

SuperUserDone: Cool, will have a look at it either later today or a bit later this week

johannescoetzee: Cool, thanks!

SuperUserDone:
I have pulled the #5829 changes from master into here, now the windows hack is no longer needed.

@johannescoetzee Please take a look again and let me know if something else needs addressing.

PR #5827, opened 2026-02-16
[kotlin2cpg] Filter generated bin Kotlin sources

SuperUserDone:
Resolves part of: https://github.com/ShiftLeftSecurity/codescience/issues/8701

Some projects include generated/compiled Kotlin copies in `bin` alongside canonical `src/main/...` sources.

This change filters generated Kotlin sources under bin from frontend input selection to avoid analyzing duplicate source copies.


SuperUserDone:
I have done some more investigation, and the generation of the files in ./bin seem rather random, and even for the same project depending on kotlin/gradle/jvm/dependency versions. For now I'll be closing this PR, and from there try to determine the root of the generation of these files. If I can determine the root of why the files in ./bin are generated, I'll reopen with findings. They do not seem to appear while running SPTests on the buildbot.

PR #5829, opened 2026-02-17
[x2cpg] Support Windows Case-Insensitive And 8.3 Style Paths

SuperUserDone:
This PR is draft for now until I've had some more discussions on if this is the way we want to do this.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/SourceFilesTests.scala
+++ b/joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/SourceFilesTests.scala
@@ -29,13 +29,7 @@ class SourceFilesTests extends AnyWordSpec with Matchers with Inside {
 
   private object UnixOnly
       extends Tag(
-        if (!scala.util.Properties.isWin && !scala.util.Properties.isMac) ""
-        else classOf[Ignore].getName
-      )
-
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/x2cpg/src/test/scala/io/joern/x2cpg/SourceFilesTests.scala
```diff
@@ -21,6 +21,24 @@ class SourceFilesTests extends AnyWordSpec with Matchers with Inside {
...
+        if (!scala.util.Properties.isWin && !scala.util.Properties.isMac) ""
+        else classOf[Ignore].getName
+      )
+
+  private object MacOnly
+      extends Tag(
+        if (scala.util.Properties.isMac) ""
+        else classOf[Ignore].getName
+      )
+
```
johannescoetzee: It looks like this is currently unused, so it should either be removed or mac tests added, in my opinion. Since case-sensitivity seems to be the only difference between Mac and at least major Linux filesystems, and since case-sensitivity for either of these is not consistent and is very configuration-dependent, I reckon we could just leave it out and modify `UnixOnly` to include Mac

johannescoetzee (approved):
Looks good to me. I think the switch to `Path.startsWith` is a big improvement since case-sensitivity is a nightmare to deal with

PR #5832, opened 2026-02-19
[x2cpg] Add ArtifactFetcher for fetching artifacts for unit tests

johannescoetzee:
This PR adds support for artifact fetching described in https://github.com/ShiftLeftSecurity/codescience/issues/8694

Some notes:
- The default cache directory is `~/.shiftleft/test-artifacts`, but this can be overridden with the `X2CPG_TEST_ARTIFACT_CACHE` environment variable
- It supports reading the first set of credentials from `~/.sbt/.credentials` by default. The credentials file path can be overridden with `X2CPG_CREDENTIALS_FILE`, or alternatively the username and password can be specified via the `X2CPG_SBT_CREDENTIALS_{USER,PASS}` environment variables.
- There's no test for fetching an artifact requiring credentials to download here, but I did test it manually and will add a test fetching an artifact from our registry to codescience.
- The `ArtifactFetcher` is defined in x2cpg main (not test), mostly since codescience doesn't already have joern as a test dependency. We could either keep it as is (maybe a legitimate non-test use-case exists for this as well?), or move it to test and add a test dependency on joern to the CS modules that need it.
- Support for this will need to be added to individual language fixtures. I've implemented support for inference jars in the javasrc2cpg fixture as an example, but this would need to be duplicated in CS.
- This adds a `NetworkTest` tag which we could exclude from test runs if these ever cause issues due to flaky internet access or anything like that.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ArtifactFetcher.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ArtifactFetcher.scala
@@ -58,38 +58,32 @@ class ArtifactFetcher(val cacheDir: Path) {
     val entryDir = cacheDir.resolve(urlHash)
     val target   = entryDir.resolve(fileName)
 
-    val future   = new CompletableFuture[Option[Path]]()
-    val existing = inFlight.putIfAbsent(artifact.url, future)
-
-    if (existing != null) {
-      // existing != null means another thread is already downloading, so wait for the result
-      try {
-        existing.get(DownloadTimeoutMinutes, TimeUnit.MINUTES)
-      } catch {
-        case e: Exception =>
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ArtifactFetcher.scala
```diff
@@ -0,0 +1,190 @@
...
+    val urlHash = HashUtil.sha256(artifact.url.getBytes("UTF-8"))
+
+    val fileName = filenameFromUrl(artifact.url)
+    val entryDir = cacheDir.resolve(urlHash)
+    val target   = entryDir.resolve(fileName)
+
+    val future   = new CompletableFuture[Option[Path]]()
+    val existing = inFlight.putIfAbsent(artifact.url, future)
+
+    if (existing != null) {
```
mpollmeier: please be consistent with braces (e.g. the else block doesn't have them)

also: readability and correctness would improve if you used Try here

PR #5833, opened 2026-02-20
[kotlin2cpg] Emit control structure for throw statement

SuperUserDone:
Resolves: https://github.com/ShiftLeftSecurity/codescience/issues/8707


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForStatementsCreator.scala
```diff
@@ -522,6 +523,17 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) {
     Ast(node)
   }
 
+  def astForThrowExpression(
+    expr: KtThrowExpression,
+    argIdxMaybe: Option[Int],
+    argNameMaybe: Option[String],
+    annotations: Seq[KtAnnotationEntry] = Seq()
+  ): Ast = {
+    val thrownValue = astsForExpression(expr.getThrownExpression, None)
```
ml86, line 532: Please explicitly set ORDER to 1 as this is the expectation of the CfgCreator and also test for this ORDER value.

PR #5844, opened 2026-02-25
[frontends] Enhance AstGenRunner to validate executable binaries and provide error handling / exit

max-leuthaeuser:
jssrc2cpg, swiftsrc2cpg, gosrc2cpg, and csharpsrc2cpg use standalone binaries to parse. We now exit with error code 1 and an appropriate error message if the binary is not available / not executable to prevent generating (nearly) empty CPGs silently.

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/jar4import/UsesService.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/jar4import/UsesService.scala
@@ -22,13 +22,16 @@ trait UsesService { this: Kotlin2Cpg =>
     } catch {
       case _: MalformedURLException =>
         println(s"The specified jar4import service url parameter `$serviceUrl` is not a valid URL. Exiting.")
-        scala.sys.exit(1)
+        System.exit(1)
+        None
       case _: java.net.ConnectException =>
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/utils/AstGenRunner.scala
```diff
@@ -163,8 +163,17 @@ object AstGenRunner {
...
+        val localPath = s"$executableDir/$executableName"
+        if (io.joern.x2cpg.astgen.AstGenRunner.isExecutableFile(localPath)) {
+          localPath
+        } else {
+          logger.error(s"""Local astgen binary not found at '$localPath' or is not executable!
+               |Please make sure to have a compatible astgen version installed and available
+               |on this system or set the environment variable ASTGEN_BIN to the full path of an executable astgen binary.
+               |""".stripMargin)
+          System.exit(1)
+          localPath // this will never be reached but is needed to satisfy the return type
```
maltek:
```suggestion
          scala.sys.exit(1)
```

`scala.sys.exit()` solves this, returning the `Nothing` type

mpollmeier (commented):
thank you 💯!

i just double checked: isExecutableFile also handles symlinks, so looks good!

PR #5846, opened 2026-02-26
[swiftsrc2cpg] Fix subscript handling with nested accessor methods

max-leuthaeuser:
- does not emit a separate METHOD node for subscript functions if they contain nested accessors anymore
- resolves closure binding issues there (identifiers ref to subscript function parameters correctly now)

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
@@ -356,8 +356,7 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
     fullNameSubscriptPrefix: String = ""
   ): MethodInfo = {
     val accessorName = code(node.accessorSpecifier)
-    val namePrefix = if (variableName.nonEmpty) { s"$variableName." }
-    else { "" }
+    val namePrefix   = if (variableName.nonEmpty) s"$variableName." else ""
     val name = accessorName match {
       case "set" => s"${namePrefix}setter"
       case "get" => s"${namePrefix}getter"
@@ -374,14 +373,9 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
         }
         MethodInfo(name, fullName, signature, tpe)
       case None =>
-        val (methodName, methodFullName) = calcNameAndFullName(name)
+        val (methodName, methodFullName) = calcNameAndFullName(name, fullNameSubscriptPrefix)
         registerType(tpe)
-        val methodFullNameWithSubscriptPrefix = if (fullNameSubscriptPrefix.nonEmpty) {
-          methodFullName.replaceFirst(s"\\.$methodName", s".$fullNameSubscriptPrefix.$methodName")
-        } else {
-          methodFullName
-        }
-        MethodInfo(methodName, methodFullNameWithSubscriptPrefix, tpe, tpe)
+        MethodInfo(methodName, methodFullName, tpe, tpe)
     }
   }
 
@@ -483,10 +477,14 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
     }
   }
 
-  private def calcNameAndFullName(name: String): (String, String) = {
+  private def calcNameAndFullName(name: String, fullNameSubscriptPrefix: String = ""): (String, String) = {
     val fullNamePrefix = s"${parserResult.filename}:${scope.computeScopePath}"
-    val fullName       = s"$fullNamePrefix.$name"
-    (name, fullName)
+    val methodFullNameWithSubscriptPrefix = if (fullNameSubscriptPrefix.nonEmpty) {
+      s"$fullNamePrefix.$fullNameSubscriptPrefix.$name"
+    } else {
+      s"$fullNamePrefix.$name"
+    }
+    (name, methodFullNameWithSubscriptPrefix)
   }
 
   private def calcMethodName(func: SwiftNode): String = {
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -347,18 +338,30 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
-  protected def methodInfoForAccessorDecl(node: AccessorDeclSyntax, variableName: String, tpe: String): MethodInfo = {
+  protected def methodInfoForAccessorDecl(
+    node: AccessorDeclSyntax,
+    variableName: String,
+    tpe: String,
+    fullNameSubscriptPrefix: String = ""
+  ): MethodInfo = {
     val accessorName = code(node.accessorSpecifier)
+    val namePrefix = if (variableName.nonEmpty) { s"$variableName." }
+    else { "" }
```
maltek: This style of `if/else` that scalafmt sometimes spits out where a `else` in expression context doesn't get any indentation just doesn't parse in my head. I seem to see the `else` first, and then search around where I've missed the `if`.

I always manually edit them like this (which scalamft won't undo):
```suggestion
    val namePrefix =
      if (variableName.nonEmpty) { s"$variableName." }
      else { "" }
```

or this (which might be undone by `scalafmt` if the line is too long):
```suggestion
    val namePrefix = if (variableName.nonEmpty) s"$variableName."  else ""
```

It's still just a minor style thing so do as you wish with that.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -373,7 +376,12 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
       case None =>
         val (methodName, methodFullName) = calcNameAndFullName(name)
         registerType(tpe)
-        MethodInfo(methodName, methodFullName, tpe, tpe)
+        val methodFullNameWithSubscriptPrefix = if (fullNameSubscriptPrefix.nonEmpty) {
+          methodFullName.replaceFirst(s"\\.$methodName", s".$fullNameSubscriptPrefix.$methodName")
```
maltek: could we build the fullname correctly in `calcNameAndFullName` to begin with? These kinds of string replacements have a tendency to break in weird edge cases.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/ExtensionTests.scala
```diff
@@ -91,65 +90,92 @@ class ExtensionTests extends SwiftSrc2CpgSuite {
...
+          |      return path.reduce(self) { $0[sub: $1] }
+          |    }
+          |  }
+          |
+          |  public subscript(path: JSONSubscriptType...) -> JSON {
+          |    get {
+          |      return self[path]
+          |    }
+          |  }
+          |}""".stripMargin
```
maltek, line 151: it would be good to also have a test with a read-only subscript without the explicit `get` block (3rd code sample at https://docs.swift.org/swift-book/documentation/the-swift-programming-language/subscripts/#Subscript-Syntax )

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/ExtensionTests.scala
```diff
@@ -91,65 +90,92 @@ class ExtensionTests extends SwiftSrc2CpgSuite {
...
-      val List(getRef1)                  = subscriptMethod1.ast.isMethodRef.code("get").l
-      val List(closureBindingPath1)      = getRef1.captureOut.l
-      closureBindingPath1.closureBindingId shouldBe Option(
-        "Ext1.swift:<global>.Foo.subscript:(path:Swift.Array)->JSON:get:ANY:path"
-      )
-      closureBindingPath1.refOut.head shouldBe pathParam1
-      closureBindingPath1.evaluationStrategy shouldBe EvaluationStrategies.BY_REFERENCE
-
-      inside(cpg.call.codeExact("self[path]").argument(2).l) { case List(id: Identifier) =>
+      inside(subscriptFooGet2.ast.isCall.codeExact("self[path]").argument(2).l) { case List(id: Identifier) =>
```
maltek, line 199: from the argument index 2 I gather this is still lowered as a index access, not as a call to the overloaded subscript method? That's fine for now, but another todo :)

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForDeclSyntaxCreator.scala
```diff
@@ -595,15 +595,55 @@ trait AstForDeclSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
...
+    }
+  }
+
+  private def astForAccessorInSubscript(
+    node: SubscriptDeclSyntax,
+    handleAccessor: (AccessorDeclSyntax, String, String, String, Seq[SwiftNode]) => Unit
+  ): Ast = {
+    val methodInfo                                       = methodInfoForFunctionDeclLike(node)
+    val MethodInfo(methodName, _, signature, returnType) = methodInfo
+    val subscriptSignaturePrefix                         = s"$methodName:${signature.replace(s"->$returnType", "")}"
```
maltek: similar concern here about the string replacement. Here, I'm pretty sure it would break when some function pointers are involved. (Even if I don't know why someone would implement a subscript that takes functions as arguments.)

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/ExtensionTests.scala
max-leuthaeuser: Yes, that's for another PR.

PR #5848, opened 2026-02-26
[SBT] Use more robust coursier download API

max-leuthaeuser:
I have seen a lot of 502s and 504s during `ensureIsAvailable` calls when trying to download resources, e.g., from GH during project initialization (astgen downloads).

The Coursier stuff claims to be more robust in handling transient HTTP failures so lets give that a try.

If that works, we could apply that to all projects currently using plain old `sbt.IO.transfer`.

### DIFF
--- a/project/meta-build.sbt
+++ b/project/meta-build.sbt
@@ -7,12 +7,15 @@ initialize := {
 }
 
 libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


max-leuthaeuser:
Still ends up with `download error: Caught java.nio.channels.OverlappingFileLockException ...`. I guess this does not work. Will do that differently in another PR.

PR #5852, opened 2026-03-03
[swiftsrc2cpg] Enhance subscript call handling with getter and setter methods in ExtensionsPass

max-leuthaeuser:


joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/ExtensionTests.scala
maltek, line 215: shouldn't this be converted to a getter call now, with different argument indizes?

max-leuthaeuser: We have no way to figure this out if no fullnames from the compiler are available (call site).

maltek: OK, I see. That file contains tests without compiler support. Makes sense that it wouldn't work in that scenario.

PR #5854, opened 2026-03-04
[kotlin2cpg] Use unique namespace fullNames for imports and for packages

SuperUserDone:
This solves the policylinker issue, but not sure that imports are being represented correctly. From what I can tell javasrc does not create namespace blocks for imports.

https://github.com/joernio/joern/blob/master/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/AstCreator.scala#L236-L259


joern-cli/frontends/kotlin2cpg/src/test/scala/io/joern/kotlin2cpg/querying/NamespaceBlockTests.scala
ml86: Name property of the namespace block for `PackageFoo` is wrong. The name property shall contain the fully qualified name `com.test.PackageFoo`.

ml86:
You are right, the imports are not represented correctly in the sense that an import statement should not result in the generation of a namespace block for the imported entity.
Please also address this in this PR.

ml86 (approved):

PR #5858, opened 2026-03-05
[swiftsrc2cpg] Added ClosureBindingsPass to manage bindings for closure functions on Swift.Function

max-leuthaeuser:
This includes the fully typed Function TypeDecls

Alternative to: https://github.com/joernio/joern/pull/5857

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
@@ -433,7 +433,7 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
       callName,
       callMethodFullname,
       DispatchTypes.DYNAMIC_DISPATCH,
-      Option(signature),
+      Option(Defines.ErasedSignature),
       Option(tpe)
     )
     callAst(callNode_, args, Option(baseAst))
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstNodeBuilder.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstNodeBuilder.scala
@@ -187,7 +187,7 @@ trait AstNodeBuilder(implicit withSchemaValidation: ValidationMode) { this: AstC
     val (inherits, bindingName, signature) = if (node.isInstanceOf[ClosureExprSyntax]) {
       val inheritsFunctionFullName = s"${Defines.Function}<${methodNode.signature}>"
       registerType(inheritsFunctionFullName)
-      (Seq(inheritsFunctionFullName), Defines.ClosureApplyMethodName, "")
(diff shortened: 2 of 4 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ClosureBindingsPass.scala
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+      cpg.typeDecl.fullNameExact(closureMethodFullName).loneElement
+    }
+  }
+
+  override def run(diffGraph: DiffGraphBuilder): Unit = {
+    for {
+      closureMethod            <- cpg.method.filter(_.name.startsWith(io.joern.x2cpg.Defines.ClosurePrefix))
+      closureMethodeTypeDecl   <- cpg.typeDecl.fullNameExact(closureMethod.fullName).loneElementOption
+      inheritsFromTypeFullName <- closureMethodeTypeDecl.inheritsFromTypeFullName.loneElementOption
```
maltek: I have not tested it, but I think all edges should already be available for this to work:
```suggestion
      closureMethodTypeDecl    <- cpg.method.isLambda.bindingTypeDecl
      inheritsFromTypeFullName <- closureMethodTypeDecl.inheritsFromTypeFullName.loneElementOption
```

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/SwiftTypeNodePass.scala
```diff
@@ -14,6 +15,11 @@ object SwiftTypeNodePass {
         typeName match {
           case name if name.endsWith(NamespaceTraversal.globalNamespaceName) =>
             NamespaceTraversal.globalNamespaceName
+          case name if name.startsWith(s"${Defines.Function}<") && name.endsWith(">") =>
```
maltek: I don't know if it's too fancy - this can also be written as
```suggestion
          case s"${Defines.Function}<$_>" =>
```

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/SwiftTypeNodePass.scala
```diff
@@ -14,6 +15,11 @@ object SwiftTypeNodePass {
...
+            "Function"
+          case name if name.contains(s"${io.joern.x2cpg.Defines.ClosurePrefix}") =>
+            val index = name.indexOf(io.joern.x2cpg.Defines.ClosurePrefix)
```
maltek: lambdas can be nested, so we need to get the last index. But then, I think lambdas can also contain named functions. So we also have to check that the characters after the prefix up to the `:` are all digits.

Sucks that this interface requires parsing of fullnames.

maltek: do we actually need this case at all? If we rearrange the default case like this, I think it should just work for lambdas. Including nested lambdas and named functions inside those.

`typeName.takeWhile(_ != ':').split('.').lastOption.getOrElse(typeName)`

PR #4892, opened 2024-09-02
[x2cpg] Added server mode

max-leuthaeuser:

PR #5860, opened 2026-03-06
remove jlhttp dependency (GPLv2 license)

mpollmeier:
Having a GPLv2 license in the dependencies limits us downstream
and really isn't required for a simple http server. This replaces
it with HttpServer from the JDK.

jlhttp was initially introduced via https://github.com/joernio/joern/pull/4892/


max-leuthaeuser (commented):
Did you test that with e.g., the CS integrations tests?
Looking at the failed tests here: looks like this is not thread-safe anymore?
`server` and `frontendHTTPHandler` were both objects.

max-leuthaeuser (approved):

PR #5861, opened 2026-03-06
ArtifactFetcher: fix for systems with multiple partitions

mpollmeier:
if the repo and tmp are on different partitions, atomic move isn't supported

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ArtifactFetcher.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ArtifactFetcher.scala
@@ -136,13 +136,7 @@ class ArtifactFetcher(val cacheDir: Path) {
               throw new RuntimeException(s"Hash mismatch: expected ${artifact.sha256}, got $downloadedHash")
             }
 
-            try {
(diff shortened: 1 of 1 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/ArtifactFetcher.scala
```diff
@@ -136,7 +136,13 @@ class ArtifactFetcher(val cacheDir: Path) {
...
             }
 
-            Files.move(tmpFile, target, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING)
+            try {
+              Files.move(tmpFile, target, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING)
+            } catch {
+              case _: java.nio.file.AtomicMoveNotSupportedException =>
+                // if the repo and tmp are on different partitions, atomic move isn't supported
+                Files.move(tmpFile, target, StandardCopyOption.REPLACE_EXISTING)
+            }
```
maltek: might as well just do the move without the atomic flag from the beginning. The only thing it does is raise an error instead of falling back to a non-atomic move.
```suggestion
            Files.move(tmpFile, target, StandardCopyOption.REPLACE_EXISTING)
```

PR #5864, opened 2026-03-06
[swiftsrc2cpg] Simplify astForIdentifier and fix executableDir for Bloop

max-leuthaeuser:
Replace nested if/else in `astForIdentifier` (swiftsrc2cpg) with a `match` expression and extract the repeated self-access node building into a private `fieldAccessAstForSelf` helper.

Replace brittle string-based `executableDir` logic in both jssrc2cpg and swiftsrc2cpg AstGenRunners with a call to the shared `ExternalCommand.executableDir` utility.

Add Bloop support to `ExternalCommand.executableDir`: detect the `.bloop/<project>/` path prefix, parse the project's Bloop config JSON to find the real project directory, and resolve `bin/` from there, enabling astgen resolution when running under Metals/Cursor.
This also allows to use Metals LSP (which uses Bloop instead of sbt under the hood) in any IDE.

### DIFF
--- a/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommand.scala
+++ b/semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommand.scala
@@ -106,10 +106,8 @@ object ExternalCommand {
         while (dir.toString.contains("target"))
           dir = dir.getParent
         dir
-      } else if (packagePathAbsolute.toString.contains(s"${File.separator}.bloop${File.separator}")) {
-        resolveBloopProjectDir(packagePathAbsolute).getOrElse(Paths.get("."))
       } else {
-        Paths.get(".")
+        resolveBloopProjectDir(packagePathAbsolute).getOrElse(Paths.get("."))
       }
 
     fixedDir.resolve("bin/").toAbsolutePath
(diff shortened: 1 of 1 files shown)


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommand.scala
```diff
@@ -106,11 +106,39 @@ object ExternalCommand {
...
+      val afterBloop    = pathStr.substring(bloopIdx + bloopMarker.length)
+      val sepIdx        = afterBloop.indexOf(File.separatorChar)
+      val projectName   = if (sepIdx >= 0) afterBloop.substring(0, sepIdx) else afterBloop
+      val configFile    = workspaceRoot.resolve(".bloop").resolve(s"$projectName.json")
+
+      val content    = IOUtils.readLinesInFile(configFile).mkString("\n")
+      val dirPattern = """"directory"\s*:\s*"([^"]+)"""".r
+      dirPattern.findFirstMatchIn(content).map(m => Paths.get(m.group(1)))
+    } catch {
+      case NonFatal(_) => None
```
mpollmeier: better log a warning here?

semanticcpg/src/main/scala/io/shiftleft/semanticcpg/utils/ExternalCommand.scala
```diff
@@ -106,11 +106,42 @@ object ExternalCommand {
...
     fixedDir.resolve("bin/").toAbsolutePath
   }
 
+  private val BloopDirPattern = """"directory"\s*:\s*"([^"]+)"""".r
+
+  /** Resolves the actual project directory when running under Bloop. Bloop stores compiled classes under
+    * `.bloop/<projectName>/...` at the workspace root, so we read the project's Bloop config JSON
+    * (`.bloop/<projectName>.json`) to find the real project directory.
+    */
+  private def resolveBloopProjectDir(packagePath: Path): Option[Path] = {
```
maltek, line 122: I would rewrite the first half of this as a loop calling `Path.getName(i)` until `Path.of(".bloop")` is found to rely a little less on string manipulation. But I don't see anything that's actually wrong with your code as it is.

PR #5867, opened 2026-03-09
[javasrc2cpg] Add java 25 support

johannescoetzee:
This PR upgrades JavaParser and adds effectively complete java 25 support. It includes a few changes:
* a minor change in pattern handling due to the new `PatternExpr` hierarchy which includes unnamed/match-all patterns. These are just ignored in our lowering. 
* The JdkJarTypeSolver is renamed to JarTypeSolver and made somewhat more flexible to avoid a bug introduced in JavaParser `v3.28.0`. We now use this for both inference jars and jdk jars/jmods.
* It adds an implementation for the `tryToSolveTypeInModule` for all of our custom type solvers.

I had to modify one of our unit tests for this due to a change in JavaParser. An improvement to method resolution was made, but this improvement only works for methods where all the types in the signature are resolvable, and throws an `UnsolvedSymbolException` when not. This means that results will be worse than before for scans without dependency information, but since the results for those are already effectively unusable, I think it's fine to incorporate this as-is.

This also doesn't include support for the to-be-added `isModuleImport` field for import nodes, but that should not affect dataflow tracking.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
@@ -144,7 +144,7 @@ class AstCreationPass(config: Config, cpg: Cpg, sourcesOverride: Option[List[Str
     }
 
     combinedTypeSolver.addNonCachingTypeSolver(
-      JarTypeSolver.fromJdkPath(jdkPath, config.cacheJdkTypeSolver, enableVerboseTypeLogging)
+      JarTypeSolver.fromPath(jdkPath, config.cacheJdkTypeSolver, enableVerboseTypeLogging)
     )
 
     val sourceTypeSolver =
@@ -163,9 +163,7 @@ class AstCreationPass(config: Config, cpg: Cpg, sourcesOverride: Option[List[Str
     }
     (jarsList ++ dependencies)
       .foreach { path =>
-        Try(
-          JarTypeSolver.fromJdkPath(path, useCache = true, enableVerboseTypeLogging = enableVerboseTypeLogging)
-        ) match {
+        Try(JarTypeSolver.fromPath(path, useCache = true, enableVerboseTypeLogging = enableVerboseTypeLogging)) match {
           case Success(jarTypeSolver) =>
             combinedTypeSolver.addNonCachingTypeSolver(jarTypeSolver)
             if (enableVerboseTypeLogging) {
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/expressions/AstForPatternExpressionsCreator.scala
```diff
@@ -279,24 +279,25 @@ trait AstForPatternExpressionsCreator { this: AstCreator =>
       .map(_.getDeclaredFields.asScala.map(_.getName).toList)
       .getOrElse(patternList.map(_ => Defines.UnknownField))
 
-    patternList.zip(fieldNames).map { case (childPatternExpr, fieldName) =>
-      val childTypeFullName = getPatternTypeFullName(childPatternExpr) match {
-        case typeFullName if isResolvedTypeFullName(typeFullName) => Option(typeFullName)
-        case _                                                    => None
-      }
+    patternList.zip(fieldNames).collect { case (pat: PatternExpr, fieldName) => (pat, fieldName) }.map {
```
ml86, line 292: What about the other kinds of `ComponentPatternExpr`? Inline comment to explain why the are not of interest would make sense.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/TypeDeclTests.scala
```diff
@@ -2,13 +2,47 @@ package io.joern.javasrc2cpg.querying
...
+      }
+    }
+
+    "contain a member node for s" in {
+      inside(cpg.member.name("s").l) { case List(sMember) =>
+        sMember.name shouldBe "s"
+        sMember.typeFullName shouldBe "java.lang.String"
+
+        sMember.astParent.isInstanceOf[TypeDecl] shouldBe true
+        sMember.astParent.asInstanceOf[TypeDecl].fullName shouldBe "$COMPACT_CLASS"
```
ml86, line 32: So we handle package declaration in compact classes correctly? We only seem to be handling the special case of being in the root package or are compact classes only allowed in root package namespace?

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/expressions/AstForPatternExpressionsCreator.scala
johannescoetzee: The PatternExpr type hierarchy is now

```
ComponentPatternExpr
|-- MatchAllPatternExpr (doesn't have a type)
|-- PatternExpr (ABC for patterns with types)
     |-- TypePatternExpr
     |-- RecordPatternExpr
```

For the purposes of our lowering, we only care about `MatchAllPatternExpr` when getting the field names corresponding to the patterns in the list (the `zip` here), and after that we ignore them. I'll add a comment explaining this in the source.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/TypeDeclTests.scala
johannescoetzee: We do handle it correctly, but I didn't think to add a test for it. I've added one now

PR #5868, opened 2026-03-09
Fixed compilation warnings

max-leuthaeuser:


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/compiler/CompilerAPI.scala
```diff
@@ -27,7 +27,7 @@ object CompilerAPI {
     messageCollector: MessageCollector
   ): KotlinCoreEnvironment = {
     val config = new CompilerConfiguration()
-    config.put(CLIConfigurationKeys.MESSAGE_COLLECTOR_KEY, messageCollector)
```
max-leuthaeuser, line 30: This was deprecated. They just moved the enum.

mpollmeier (approved):

PR #5869, opened 2026-03-09
frontend http server: attempt 1 to fix OOM issue

mpollmeier:
We're seeing occasional OOM errors, which is likely due to the recent
change in the FrontendHttpServer (930caed). This is an attempt to fix
it. Idea being:

* null out the http server so it can get gc-collected
* shutdown the executorservice


maltek:
If the switch to the different HTTP server is the problem, then I suspect it's something related to leaked classpaths again. We enable forked tests in this repo, so I'd be a little surprised. But still, `sbt` really wants to manage classpaths on its own, and the initial PR already showed some signs that the Sun server doesn't play too nicely with that.

Let's see though what the CI says 🤞

PR #5873, opened 2026-03-10
chore: enable strict format testing

TNSelahle:
Enabling strict format testing.

Closes https://github.com/ShiftLeftSecurity/codescience/issues/8705

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForFunctionsCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForFunctionsCreator.scala
@@ -179,8 +179,21 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
     val attributeAsts = decl.attributeGroups.flatMap(astForAttributeGroup)
     val methodBody    = blockAst(methodBodyNode, methodBodyStmts)
 
-    scope.popScope()
-    val methodAst = methodAstWithAnnotations(method, parameters, methodBody, methodReturn, modifiers, attributeAsts)
+    val scope_ = scope.popScope()
+    val additionalBodyChildren = scope_ match {
+      case Some(ms: MethodScope) =>
+        ms.additionalBodyChildren.map(Ast(_)).toList
+      case _ =>
+        List()
+    }
+    val methodAst = methodAstWithAnnotations(
+      method,
+      parameters,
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForFunctionsCreator.scala
```diff
@@ -179,8 +179,21 @@ trait AstForFunctionsCreator(implicit withSchemaValidation: ValidationMode) { th
...
+    val additionalBodyChildren = scope_ match {
+      case Some(ms: MethodScope) =>
+        ms.additionalBodyChildren.map(Ast(_)).toList
+      case _ =>
+        List()
+    }
+    val methodAst = methodAstWithAnnotations(
+      method,
+      parameters,
+      methodBody.withChildren(additionalBodyChildren),
```
ml86, line 192: METHOD_REF need to be at the position in the AST where the lambda/method is defined and this looks to me as if it is just concatenated to the back of the method body.
I am also missing a test case or the adjustment of an existing test which checks for the position of the method ref.

TNSelahle: This is specifically for normal functions. Anonymous functions are handled differently. There is a guard around the code that populates the scope with the methodRef. 

https://github.com/joernio/joern/blob/2baeab5578b978303571765b83e91b407fc1cdcb/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala#L183

It was a bit of a mission trying to rework this to be more uniform so decided to just have a solution that works with what's there. 

Actually, my bad on not explaining in the description why I made this particular change. The newly added duplicate order validation check fails as it finds 2 methodRefs which are AST siblings having the same order (-1). This was occurring in this case:

```php
<?php
$a = 10;
$b = 20;
if (true) {
  function foo() {
    global $a;
    global $b;
  }
} else {
  function foo() {
    global $a;
    global $b;
  }
}
```
The `methodRef` nodes for both functions are children of the global method body and because they were being added directly to the diffgraph, had the same order of -1.

ml86: Ok, missed it, in that case it is fine.

PR #5887, opened 2026-03-16
[php2cpg] feat: add multi array dimension assignment support

TNSelahle:
- Add multi array dimension assignment support
- Fix tmp var numberings not being unique when a tmp variable is generated while in a block scope

Closes https://github.com/ShiftLeftSecurity/codescience/issues/8691

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
@@ -104,7 +104,7 @@ class ArrayTests extends PhpCode2CpgFixture {
       block.code shouldBe "$xs[][2][] = $val"
       block.lineNumber shouldBe Some(3)
 
-      inside(block.astChildren.isLocal.l) { case List(tmp0, tmp1, tmp2, tmp3, tmp4, xs) =>
+      inside(block.astChildren.isLocal.l) { case List(tmp0, tmp1, tmp2, tmp3, xs) =>
         tmp0.name shouldBe "foo@tmp-0"
         tmp0.code shouldBe "$foo@tmp-0"
         tmp1.name shouldBe "foo@tmp-1"
@@ -113,33 +113,34 @@ class ArrayTests extends PhpCode2CpgFixture {
         tmp2.code shouldBe "$foo@tmp-2"
         tmp3.name shouldBe "foo@tmp-3"
         tmp3.code shouldBe "$foo@tmp-3"
-        tmp4.name shouldBe "foo@tmp-4"
-        tmp4.code shouldBe "$foo@tmp-4"
         xs.name shouldBe "xs"
         xs.code shouldBe "$xs"
       }
 
       inside(block.astChildren.not(_.isLocal).l) {
-        case List(assignOne: Call, assignTwo: Call, assignThree: Call, arrayPushCall: Call, retIden: Identifier) =>
+        case List(assignOne: Call, assignTwo: Call, arrayPushCall: Call, retIden: Identifier) =>
           assignOne.name shouldBe Operators.assignment
-          assignOne.code shouldBe "$foo@tmp-0 = $val"
+          assignOne.code shouldBe "$foo@tmp-0 = array($val)"
           assignOne.order shouldBe 1
 
           assignTwo.name shouldBe Operators.assignment
-          assignTwo.code shouldBe "$foo@tmp-1 = array($foo@tmp-0)"
+          assignTwo.code shouldBe "$foo@tmp-2 = array(2 => $foo@tmp-0)"
           assignTwo.order shouldBe 2
 
-          assignThree.name shouldBe Operators.assignment
-          assignThree.code shouldBe "$foo@tmp-3 = array(2 => $foo@tmp-1)"
-          assignThree.order shouldBe 3
-
           arrayPushCall.name shouldBe "array_push"
-          arrayPushCall.code shouldBe "$xs[] = $foo@tmp-3"
-          arrayPushCall.order shouldBe 4
+          arrayPushCall.code shouldBe "$xs[] = $foo@tmp-2"
+          arrayPushCall.order shouldBe 3
+          inside(arrayPushCall.argument.l) { case List(lhsIdentifier: Identifier, rhsIdentifier: Identifier) =>
+            lhsIdentifier.name shouldBe "xs"
+            lhsIdentifier.code shouldBe "$xs"
 
-          retIden.name shouldBe "foo@tmp-0"
-          retIden.code shouldBe "$foo@tmp-0"
-          retIden.order shouldBe 5
+            rhsIdentifier.name shouldBe "foo@tmp-2"
+            rhsIdentifier.code shouldBe "$foo@tmp-2"
+          }
+
+          retIden.name shouldBe "val"
+          retIden.code shouldBe "$val"
+          retIden.order shouldBe 4
       }
     }
   }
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
```diff
@@ -65,6 +65,108 @@ class ArrayTests extends PhpCode2CpgFixture {
...
+          assignOne.name shouldBe Operators.assignment
+          assignOne.code shouldBe "$foo@tmp-0 = $val"
+          assignOne.order shouldBe 1
+
+          assignTwo.name shouldBe Operators.assignment
+          assignTwo.code shouldBe "$foo@tmp-1 = array($foo@tmp-0)"
+          assignTwo.order shouldBe 2
+
+          assignThree.name shouldBe Operators.assignment
+          assignThree.code shouldBe "$foo@tmp-3 = array(2 => $foo@tmp-1)"
```
johannescoetzee: This lowering is fine in the case where `$xs` does not already exist, but I think it could be problematic if `$xs[1][2]` is an array that was already tainted. This step plus the next would then overwrite the tainted array.

The lowering in general looks to be the other way around from what we discussed, but I also wasn't involved in later discussions. Did the plan change after we last spoke?

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
```diff
@@ -65,6 +65,108 @@ class ArrayTests extends PhpCode2CpgFixture {
...
+          retIden.code shouldBe "$foo@tmp-0"
+          retIden.order shouldBe 5
+      }
+    }
+  }
+
+  "assignments using the multi array dimension fetch syntax with last dimension should be rewritten as multiple assignments" in {
+    val cpg = code("""<?php
+        |function foo($val) {
+        |  $xs[1][2][] = $val;
```
johannescoetzee, line 206: A test for something like `$xs[1][][2]` not ending with a dimensionless index access is also missing.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
```diff
@@ -1345,10 +1345,10 @@ class ControlStructureTests extends PhpCode2CpgFixture {
         lhsIdent.code shouldBe s"$$${variableName}"
 
         rhsIdxAccess.methodFullName shouldBe Operators.indexAccess
-        rhsIdxAccess.code shouldBe s"$$foo@tmp-0[${arrayIndex}]" // $foo@tmp-0 is the name of the variable holding the current value
+        rhsIdxAccess.code shouldBe s"$$foo@tmp-1[${arrayIndex}]" // $foo@tmp-1 is the name of the variable holding the current value
```
johannescoetzee, line 1348: I'm guessing these changing are due to the duplicate ASTs being generated for the code? I do think we should find another option there. Duplicating work generating the AST isn't ideal, especially if it's only for the code field

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
TNSelahle: Hmm, do you mean in that case, the previous taint information would be lost?

Regarding the lowering, the plan initially didn't change but when I started with the implementation, it was a bit more intuitive implementing it going top-to-bottom instead of bottom-to-top.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
TNSelahle: These are changing due to my updates to the `getNewVarTmp` method in `Scope`. The method previously used the counter from the top scope if it was a method-, class- or namespace- scope. If the top scope was a block scope, a new counter would be used and but the tmp var name would still be dependent on the enclosing method, type, or namespace name. So then it was possible to get a name that wasn't unique.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
ml86: The lowering looks good to me. If an array is tainted with a wildcard access path the assignment `$xs[1][2]=...` does not remove the taint.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
ml86: Please separate the temp variable name fix into a separate PR.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
johannescoetzee: For `$xs[][2][] = $val;` this would be fine, but for `$xs[1][2][] = $val` we run into an issue. As it is lowered here, the first couple of assignments effectively reduce to
```
$xs[1] = array(2 => $val)
```
which means if we had something like `$xs[1] = array(1 => "bad data")`, that first element would be removed after the lowered assignment above.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
```diff
@@ -65,6 +65,108 @@ class ArrayTests extends PhpCode2CpgFixture {
...
+          assignThree.code shouldBe "$foo@tmp-3 = array(2 => $foo@tmp-1)"
+          assignThree.order shouldBe 3
+
+          arrayPushCall.name shouldBe "array_push"
+          arrayPushCall.code shouldBe "$xs[] = $foo@tmp-3"
+          arrayPushCall.order shouldBe 4
+
+          retIden.name shouldBe "foo@tmp-0"
+          retIden.code shouldBe "$foo@tmp-0"
+          retIden.order shouldBe 5
```
johannescoetzee: This is also not correct. This is saying that `$val` appears as the last child in the method body, but that should not be the case

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
johannescoetzee: Edit to add the two specific lines in the lowering I'm concerned about:

```
assignTwo.code shouldBe "$foo@tmp-1 = array($foo@tmp-0)" // effectively = array(val)
assignFour.code shouldBe "$xs[1] = $foo@tmp-3"           // $foo@tmp-3 == array(2 => foo@tmp-1)
```

To give a concrete example, it's the difference between 

```php
<?php

$xs = array(array(1 => "bad"));

$xs[0][2] = "good";

var_dump($xs);

// output
array(1) {
  [0]=>
  array(2) {
    [1]=>
    string(3) "bad"
    [2]=>
    string(4) "good"
  }
}
```

and 

```php
<?php

$xs = array(array(1 => "bad"));

$xs[0] = array(2 => "good");

var_dump($xs);

// output
array(1) {
  [0]=>
  array(1) {
    [2]=>
    string(4) "good"
  }
}
```

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
TNSelahle: The whole assignment expression is represented by the BLOCK node and so I've made the `$val` the last child of that. I think you may have confused that block as being the method block.

johannescoetzee: You're right, I did miss one of the `astChildren`s in the query.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
johannescoetzee: I've been thinking some more and I think I didn't really understand the implications of using wildcard access paths for taint. If we're just using wildcards, then 
```
xs[0][2] = ...
```
and 
```
xs[0] = array(2 => ...)
```
should be equivalent since from the DF engine's perspective, it's unknown what's being overwritten in any case.

TNSelahle: Very good catch! I missed this. This poses a problem anytime an index in the chain is pointing to an array. Whoops.

ml86: Ok, now I see your point @johannescoetzee. I think we can relatively easy fix this buy just emitting a normal access operator chain for every up to the first `[]` from where on we know that no old taint can be overwritten.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
```diff
@@ -65,6 +93,172 @@ class ArrayTests extends PhpCode2CpgFixture {
...
+        tmp4.name shouldBe "foo@tmp-4"
+        tmp4.code shouldBe "$foo@tmp-4"
+        xs.name shouldBe "xs"
+        xs.code shouldBe "$xs"
+      }
+
+      inside(block.astChildren.not(_.isLocal).l) {
+        case List(assignOne: Call, assignTwo: Call, assignThree: Call, arrayPushCall: Call, retIden: Identifier) =>
+          assignOne.name shouldBe Operators.assignment
+          assignOne.code shouldBe "$foo@tmp-0 = $val"
```
ml86: Lets avoid creating a temp variable also for the right hand side of the assignment. We already create one for each `array` invocation which makes the lowering complicated enough.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
```diff
@@ -65,6 +93,172 @@ class ArrayTests extends PhpCode2CpgFixture {
...
+
+          assignThree.name shouldBe Operators.assignment
+          assignThree.code shouldBe "$foo@tmp-3 = array(2 => $foo@tmp-1)"
+          assignThree.order shouldBe 3
+
+          arrayPushCall.name shouldBe "array_push"
+          arrayPushCall.code shouldBe "$xs[] = $foo@tmp-3"
+          arrayPushCall.order shouldBe 4
+
+          retIden.name shouldBe "foo@tmp-0"
```
ml86: Use the right hande side of assignment in this case `$val` as last identifier of the block.

ml86 (changes requested):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ArrayTests.scala
TNSelahle: The Ast for the right-hand side of the assignment is generated somewhere within the recursive method `traverseArrayDimExpr`. I didn't want to over-complicated the implementation. I initially called `astForExpr` for the RHS just before creating the Ast for the block but then since the RHS expression gets passed into `traverseArrayDimExpr`, `astForExpr` will be called again on the RHS expression. 

I assign it to a tmp variable since calling `astForExpr` twice on the same `PhpVariable` isn't as bad as doing it on whatever complexion expression might be on the RHS

ml86 (approved):

ml86 (approved):

PR #5910, opened 2026-03-30
[pysrc2cpg] Fix **kwargs handling, walrus operator and comprehension test coverage

allsmog:
Fixes Python **kwargs handling and adds test coverage for untested features.

- Fix **kwargs unpacking to preserve dict argument in CPG for taint tracking (was silently dropped)
- Use `<keyword_dict>` val as synthetic argument name instead of string literal
- Remove resolved TODO comments (`// TODO test` on namedExpr, `// TODO implement` on kwargs)
- Add walrus operator (assignment expression) tests verifying semantics in conditions
- Add comprehensive comprehension tests (list, set, dict, generator)

## Test plan
- [x] New kwargs tests in CallCpgTests (both direct and member calls)
- [x] New walrus operator tests in AssignCpgTests
- [x] New ComprehensionCpgTests
- [x] All existing pysrc2cpg tests pass

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
@@ -2184,8 +2184,8 @@ class PythonAstVisitor(
 object PythonAstVisitor {
   private val logger = LoggerFactory.getLogger(getClass)
 
-  val typingPrefix      = "typing."
-  val metaClassSuffix   = "<meta>"
+  val typingPrefix       = "typing."
+  val metaClassSuffix    = "<meta>"
   val keywordDictArgName = "<keyword_dict>"
 
   val noLineAndColumn = LineAndColumn(-1, -1, -1, -1, -1, -1)
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
@@ -1,11 +1,11 @@
 package io.joern.pysrc2cpg
 
-import PythonAstVisitor.{logger, metaClassSuffix, noLineAndColumn}
+import PythonAstVisitor.{keywordDictArgName, logger, metaClassSuffix, noLineAndColumn}
 import io.joern.pysrc2cpg.memop.*
 import io.joern.pysrc2cpg.memop.MemoryOperation.{Del, Load, Store}
 import io.joern.x2cpg.frontendspecific.pysrc2cpg.Constants.builtinPrefix
 import io.joern.pythonparser.{AstPrinter, ast}
-import io.joern.pythonparser.ast.{Arguments, MatchAs, MatchClass, MatchMapping, MatchOr, MatchSequence, MatchSingleton, MatchStar, MatchValue, iast, iexpr, ipattern, istmt}
+import io.joern.pythonparser.ast.{Arguments, MatchAs, iast, iexpr, istmt}
 import io.joern.x2cpg.frontendspecific.pysrc2cpg.Constants
 import io.joern.x2cpg.{AstCreatorBase, ValidationMode}
 import io.shiftleft.codepropertygraph.generated.*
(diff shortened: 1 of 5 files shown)


joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/AssignCpgTests.scala
```diff
@@ -203,6 +203,33 @@ class AssignCpgTests extends PySrc2CpgFixture with Matchers {
     }
   }
 
+  "walrus operator (named expression)" should {
```
ml86: Not `(named expression)` but rather `(assignment expression)`

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/CallCpgTests.scala
```diff
@@ -194,6 +194,36 @@ class CallCpgTests extends PySrc2CpgFixture(withOssDataflow = false) {
     }
   }
 
+  "call with **kwargs unpacking" should {
+    lazy val cpg = code("""func(a, **my_dict)""".stripMargin, "test.py")
```
ml86: Remove `lazy`. The cpgs returned by `code` are always created lazily.
There is also no need to explicitly specify the filename of the test code.
There are multiple other instances of this. Fix those as well.

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/MatchCpgTests.scala
```diff
@@ -40,13 +40,13 @@ class MatchCpgTests extends PySrc2CpgFixture() {
           jumpTargetFirstCase.code shouldBe "case [a, b] if 3 == 4"
 
           blockFirstCase.label shouldBe NodeTypes.BLOCK
-          blockFirstCase.code shouldBe "print(1)"
+          blockFirstCase.astChildren.code.l should contain allOf ("a", "b", "print(1)")
```
ml86: Having the case values `a`, `b` as individual statements in the first case block does not seem like a good representation. It does not encode the semantic and I fail to see in which scenarios that helps. Please explain.

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
```diff
@@ -1848,13 +1865,14 @@ class PythonAstVisitor(
...
       if (keyword.arg.isDefined) {
-        Some((keyword.arg.get, convert(keyword.value)))
+        (keyword.arg.get, convert(keyword.value))
       } else {
         // keyword.arg == None. This is the case for func(**dict) style arguments.
-        // TODO implement handling for this case.
-        None
+        // We use a synthetic argument name to preserve the unpacked dict as an argument
+        // in the CPG so that data flow tracking can follow through it.
+        ("**", convert(keyword.value))
```
ml86: Use a `val` instead of string constant and instead of `**` use `<keyword_dict>`.

ml86 (changes requested):

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/MatchCpgTests.scala
allsmog: You're right, the pattern variables as bare statements don't encode the destructuring semantics. Reverted this change; a proper representation would need dedicated AST constructs for match patterns (similar to how javasrc2cpg handles pattern matching with Local nodes and assignment operations). Will address in a separate PR if there's interest.

ml86 (approved):

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/MatchCpgTests.scala
ml86: We intend to introduce direct support in the CPG to represent this. So I fear providing a lowered representation now would not be worth the effort.

PR #5922, opened 2026-04-08
[jssrc2cpg] Optimize performance for AST node processing

max-leuthaeuser:
Replace several JVM-costly patterns in the per-node hot path:
- Reflection-based BabelAst.fromString with pre-built Map lookup
- Try/exception-based JSON access with ujson Option APIs
- SortedMap position lookups with array-based binary search
- Redundant createBabelNodeInfo calls with lightweight nodeTypeOf
- Unnecessary .arr.toList intermediate allocations
- Regex replaceAll with literal replace in EjsPreprocessor
- Option-allocating HashMap patterns with direct getOrElse/update


ml86:
I have to ask you to split this up. I cant review it like this properly. At the very minimum I need individual commits.

max-leuthaeuser:
They are individual commits now.

ml86 (approved):

PR #5923, opened 2026-04-08
[javasrc2cpg] Add support for getting type information from modules jimage archive

johannescoetzee:
One of the sources of type information used by javasrc2cpg is java itself. While we originally used the `ReflectionTypeSolver` included in JavaParser for this, we decided at some point to start using the JDK jars/jmods instead to decouple the java version used for type inference from that used for running javasrc2cpg. In the original implementation of that, I didn't consider the `modules` jimage archive containing the necessary runtime libs built by `jlink`, however.

This PR introduces support for that as well, which means we can get type information both from the JDK and the bundled JRE (with the jmods taking preference over the `modules` archive)

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
@@ -44,7 +44,7 @@ class AstCreationPass(config: Config, cpg: Cpg, sourcesOverride: Option[List[Str
 
   def usedTypes(): Set[String] = _usedTypes
 
-  val (sourceParser, symbolSolver) = initParserAndUtils(config)
+  val (sourceParser, symbolSolver, combinedTypeSolver) = initParserAndUtils(config)
 
   override def createAccumulator(): AstCreationPass.Accumulator = AstCreationPass.Accumulator()
 
@@ -92,11 +92,16 @@ class AstCreationPass(config: Config, cpg: Cpg, sourcesOverride: Option[List[Str
     JavaParserFacade.clearInstances()
   }
 
-  private def initParserAndUtils(config: Config): (SourceParser, JavaSymbolSolver) = {
-    val dependencies = getDependencyList(config.inputPath)
-    val sourceParser = SourceParser(config, sourcesOverride)
-    val symbolSolver = createSymbolSolver(config, dependencies, sourceParser)
-    (sourceParser, symbolSolver)
+  /** Close type solver resources (open JAR files, jrt: file systems). Should be called after all AST creation is done
+    * and [[clearJavaParserCaches]] has been called.
+    */
+  def closeTypeSolvers(): Unit = combinedTypeSolver.close()
+
+  private def initParserAndUtils(config: Config): (SourceParser, JavaSymbolSolver, SimpleCombinedTypeSolver) = {
+    val dependencies                   = getDependencyList(config.inputPath)
+    val sourceParser                   = SourceParser(config, sourcesOverride)
+    val (symbolSolver, combinedSolver) = createSymbolSolver(config, dependencies, sourceParser)
+    (sourceParser, symbolSolver, combinedSolver)
   }
 
   private def getDependencyList(inputPath: String): List[String] = {
@@ -131,7 +136,7 @@ class AstCreationPass(config: Config, cpg: Cpg, sourcesOverride: Option[List[Str
     config: Config,
     dependencies: List[String],
     sourceParser: SourceParser
-  ): JavaSymbolSolver = {
+  ): (JavaSymbolSolver, SimpleCombinedTypeSolver) = {
     val verboseDebugLoggingEnvVarValue = Option(System.getenv(JavaSrcEnvVar.EnableVerboseTypeLogging.name))
     val enableVerboseTypeLogging       = verboseDebugLoggingEnvVarValue.isDefined || config.enableVerboseTypeLogging
     if (enableVerboseTypeLogging) {
@@ -191,7 +196,7 @@ class AstCreationPass(config: Config, cpg: Cpg, sourcesOverride: Option[List[Str
         }
       }
 
-    symbolSolver
+    (symbolSolver, combinedTypeSolver)
   }
 
   private def recursiveJarsFromPath(path: String): List[String] = {
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
@@ -114,7 +114,7 @@ object JrtRuntimeImageClassPath {
   /** Max directory depth (in terms of file depth from the search root) when searching under findRuntimeImage. Since
     * `lib/modules` sits 2 levels below a java.home, a maxDepth of 6 allows java.home directories up to 4 levels deep.
     */
-  val DefaultRuntimeImageSearchMaxDepth: Int = 4
+  val DefaultRuntimeImageSearchMaxDepth: Int = 10
 
   /** Find the shallowest `<java.home>/lib/modules` layout under searchRoot. maxDepth limits how deep the `modules` file
     * itself can be. Does not follow symbolic links (avoids symlink loops, consistent with jar discovery).
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
@@ -29,11 +29,7 @@ class JrtRuntimeImageClassPath(javaHome: Path) extends ClassPath with AutoClosea
       logger.warn(s"jrt: file system has no /modules root for java.home=$javaHome")
       (Map.empty[String, Path], Map.empty[String, List[String]])
     } else {
-      val exports = buildExports(modulesRoot)
-
-      val index = buildIndex(modulesRoot)
-
-      (index, exports)
+      buildExportsAndIndex(modulesRoot)
     }
   }
 
@@ -42,52 +38,51 @@ class JrtRuntimeImageClassPath(javaHome: Path) extends ClassPath with AutoClosea
   /** Module name to exported package names (for JarTypeSolver.tryToSolveTypeInModule). */
   val moduleExportsMap: Map[String, List[String]] = moduleExports
 
-  private def buildExports(modulesRoot: Path): Map[String, List[String]] = {
+  private def buildExportsAndIndex(modulesRoot: Path): (Map[String, Path], Map[String, List[String]]) = {
     val exportsBuilder = Map.newBuilder[String, List[String]]
-    Using.resource(Files.walk(modulesRoot)) { recursiveFiles =>
-      recursiveFiles.iterator().asScala.foreach { moduleDir =>
-        if (Files.isDirectory(moduleDir)) {
-          val moduleInfo = moduleDir.resolve("module-info.class")
-          if (Files.isRegularFile(moduleInfo)) {
-            Try(Using.resource(Files.newInputStream(moduleInfo)) { in =>
-              val (moduleName, exportPkgs) = JrtRuntimeImageClassPath.readModuleExports(in)
-              exportsBuilder += moduleName -> exportPkgs
-            }).recover(e => logger.debug(s"Could not read module descriptor at $moduleInfo", e))
-          }
-        }
-      }
-    }
-    exportsBuilder.result()
-  }
-
-  private def buildIndex(modulesRoot: Path): Map[String, Path] = {
-    val indexMutable = mutable.Map.empty[String, Path]
+    val indexMutable   = mutable.Map.empty[String, Path]
     var duplicateCount = 0
+
     Using.resource(Files.walk(modulesRoot)) { recursiveFiles =>
       recursiveFiles.iterator().asScala.foreach { path =>
         if (Files.isRegularFile(path) && path.getFileName.toString.endsWith(".class")) {
-          if (path.getFileName.toString != "module-info.class") {
-            Try(
-              Using.resource(Files.newInputStream(path))(BytecodeIndexedClassPath.readClassNameFrom)
-            ).toOption match {
-              case Some(className) =>
-                if (!indexMutable.contains(className)) {
-                  indexMutable(className) = path
-                } else {
-                  duplicateCount += 1
-                }
-              case None =>
-                logger.debug(s"Could not read class name from $path")
+          Try(Using.resource(Files.newInputStream(path)) { recursiveFiles =>
+            if (path.getFileName.toString == "module-info.class") {
+              handleModuleInfo(recursiveFiles, exportsBuilder)
+            } else {
+              if (handleClassFile(recursiveFiles, path, indexMutable)) {
+                duplicateCount += 1
+              }
             }
-          }
+          }).recover { case e => logger.debug(s"Could not read class files at $path", e) }
         }
       }
     }
+
     if (duplicateCount > 0) {
       logger.debug(s"JRT index: $duplicateCount duplicate class names across modules (first module wins)")
     }
 
-    indexMutable.toMap
+    (indexMutable.toMap, exportsBuilder.result())
+  }
+
+  private def handleModuleInfo(
+    in: InputStream,
+    exportsBuilder: mutable.Builder[(String, List[String]), Map[String, List[String]]]
+  ): Unit = {
+    val (moduleName, exportPkgs) = JrtRuntimeImageClassPath.readModuleExports(in)
+    exportsBuilder += moduleName -> exportPkgs
+  }
+
+  /** Returns true if the class name was a duplicate. */
+  private def handleClassFile(in: InputStream, path: Path, indexMutable: mutable.Map[String, Path]): Boolean = {
+    val className = BytecodeIndexedClassPath.readClassNameFrom(in)
+    if (!indexMutable.contains(className)) {
+      indexMutable(className) = path
+      false
+    } else {
+      true
+    }
   }
 
   override def find(classname: String): URL = {
@@ -125,10 +120,7 @@ object JrtRuntimeImageClassPath {
   /** Find the shallowest `<java.home>/lib/modules` layout under searchRoot. maxDepth limits how deep the `modules` file
     * itself can be. Does not follow symbolic links (avoids symlink loops, consistent with jar discovery).
     */
-  def findRuntimeImage(
-    searchRoot: Path,
-    maxDepth: Int = DefaultRuntimeImageRootSearchMaxDepth
-  ): Option[Path] = {
+  def findRuntimeImage(searchRoot: Path, maxDepth: Int = DefaultRuntimeImageRootSearchMaxDepth): Option[Path] = {
     if (!Files.exists(searchRoot)) {
       None
     } else {
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
@@ -18,7 +18,7 @@ import scala.util.{Failure, Success, Try, Using}
   * The FileSystem is kept open for the lifetime of this instance (same lifecycle as BytecodeIndexedClassPath holding a
   * JarFile).
   */
-class JrtRuntimeImageClassPath(javaHome: Path) extends ClassPath {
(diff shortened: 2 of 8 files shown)


max-leuthaeuser:
**A few issues:**

1. Resource leak: `jrtFileSystem` is never closed

`JrtRuntimeImageClassPath` opens a `FileSystem` at line 25 but never closes it. `javassist.ClassPath` has no `close()` method, so there's no lifecycle hook. While `BytecodeIndexedClassPath` has a similar issue with `JarFile`, a `FileSystem` is a heavier resource. If this class is instantiated more than once (e.g., when `useCache = false`), leaked file systems may accumulate.

Suggestion: Consider implementing `AutoCloseable` / `Closeable` on `JrtRuntimeImageClassPath` and closing the filesystem when the solver is no longer needed. At minimum, add a comment acknowledging the intentional leak (matching the `BytecodeIndexedClassPath` pattern if that's the accepted convention).

2. `DefaultRuntimeImageSearchMaxDepth = 48` is very deep

A max depth of 48 is very generous. A typical JDK layout has `lib/modules` at depth 2 from `java.home`. Walking 48 levels deep under the input path (which could be `/usr` or `/opt`) would traverse a large tree. This could cause significant startup latency.

Suggestion: Consider reducing to something like 8-12, or better yet, check a few well-known relative paths (e.g., `./lib/modules`, `./jre/lib/modules`) before falling back to a bounded walk.

3. No test for `determineJarPathsAllowEmpty` returning empty + no `lib/modules` present

The error path (no jars AND no runtime image) in `fromPath` throws `IllegalArgumentException`. This is tested implicitly by the existing test for `determineJarPaths`, but the new error message has changed. A dedicated test would prevent regressions.

**Minor things:**

1. `addRuntimeImage` is public but arguably should be package-private

`JarTypeSolverBuilder.addRuntimeImage` is called internally from `JarTypeSolver.fromPath` and from tests. Consider making it `private[typesolvers]` to avoid widening the API surface unnecessarily.

2. Duplicate class resolution is "last path wins"; consider "first module wins"

In `buildIndexAndExports` (line 72), when the same class name exists in multiple modules, the last one encountered wins. The JDK module system normally resolves split packages to the first readable module. This is unlikely to matter in practice (no two modules should export the same class), but worth a comment or at least logging which module won.

3. Consider early return / short-circuit in `findRuntimeImage`

Since you prefer the shallowest match and JDK images typically have `lib/modules` near the root, you could short-circuit the walk after finding a candidate at depth smaller than 3. This would prevent scanning deep subtrees for common cases.

4. Test creates a `JrtRuntimeImageClassPath` instance per test

The test class creates a fresh `JrtRuntimeImageClassPath(javaHome)` in each test that uses `assumeRuntimeImage()`. Since indexing the full JDK image is expensive, consider sharing one instance across tests (e.g., via a lazy val or `BeforeAndAfterAll`).

johannescoetzee:
I've pushed changes that I think address all of these except for 

> Consider early return / short-circuit in findRuntimeImage

I've lowered the default search depth to 4, so the search tree should be very small and easily handle-able

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JarTypeSolver.scala
```diff
@@ -95,16 +99,24 @@ class JarTypeSolverBuilder(enableVerboseTypeLogging: Boolean) {
...
+  private[typesolvers] def buildShared: JarTypeSolver =
     new JarTypeSolver(classPool, knownPackagePrefixes.toSet, exportedPackages.toMap)
-  }
 
   private def addPathToClassPool(archivePath: String): Try[BytecodeIndexedClassPath] = {
     if (archivePath.isJarPath || archivePath.isJmodPath) {
       Try {
         val classPath = new BytecodeIndexedClassPath(archivePath)
         classPool.appendClassPath(classPath)
+        ownedCloseables += classPath
```
ml86, line 119: Why use `ownedClosables` when it contains the same things then `classPool`?

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JarTypeSolver.scala
```diff
@@ -171,30 +200,58 @@ object JarTypeSolver {
...
+        builder.withJars(jarPaths)
+      } else {
+        JrtRuntimeImageClassPath.findRuntimeImage(Paths.get(inputPath)) match {
+          case Some(layout) =>
+            logger.info(
+              s"JDK type solver: using runtime image at ${layout.javaHome} (modules file: ${layout.modulesImageFile}; search root: $inputPath)"
+            )
+            builder
+              .addRuntimeImage(layout.javaHome)
+              .fold(
```
ml86: While `fold` works `getOrElse` use more appropriate.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
```diff
@@ -0,0 +1,156 @@
...
+/** ClassPath over the JDK modular runtime image (JEP 220, lib/modules jimage), exposed via the jrt file system. Used
+  * when the configured JDK path has no jar/jmod archives to index (e.g. minimal jlink images). When a full JDK is
+  * present, BytecodeIndexedClassPath on jmods and other archives is preferred instead (see JarTypeSolver.fromPath).
+  *
+  * The FileSystem is kept open for the lifetime of this instance (same lifecycle as BytecodeIndexedClassPath holding a
+  * JarFile).
+  */
+class JrtRuntimeImageClassPath(javaHome: Path) extends ClassPath with AutoCloseable {
+
+  private val logger = LoggerFactory.getLogger(this.getClass)
```
ml86: We put loggers in the companion object.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
```diff
@@ -0,0 +1,156 @@
...
+
+  private val jrtFileSystem = JrtRuntimeImageClassPath.openJrtFileSystem(javaHome)
```
ml86, line 26: Why is this a class member when it is only used in `buildIndexAndExports` which itself is also only called once. Moving it into `buildIndexAndExports` would have the benefit that `JrtRuntimeImageClassPath` does not need to auto closable.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
```diff
@@ -0,0 +1,156 @@
...
+  val moduleExportsMap: Map[String, List[String]] = moduleExports
+
+  private def buildIndexAndExports(): (Map[String, Path], Map[String, List[String]]) = {
+    val modulesRoot = jrtFileSystem.getPath("/modules")
+    if (!Files.exists(modulesRoot)) {
+      logger.warn(s"jrt: file system has no /modules root for java.home=$javaHome")
+      (Map.empty, Map.empty)
+    } else {
+      val exportsBuilder = Map.newBuilder[String, List[String]]
+      Using.resource(Files.list(modulesRoot)) { stream =>
```
ml86: `stream`: Using the type for naming is not good here.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
```diff
@@ -0,0 +1,156 @@
...
+        stream.iterator().asScala.foreach { moduleDir =>
```
ml86: Since you only filter afterwards `moduleDir` name is not correct. I suggest filtering before if you want to keep the name.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
```diff
@@ -0,0 +1,156 @@
...
+          if (Files.isDirectory(moduleDir)) {
+            val moduleInfo = moduleDir.resolve("module-info.class")
+            if (Files.isRegularFile(moduleInfo)) {
+              Try(Using.resource(Files.newInputStream(moduleInfo)) { in =>
+                val (moduleName, exportPkgs) = JrtRuntimeImageClassPath.readModuleExports(in)
+                exportsBuilder += moduleName -> exportPkgs
+              }) match {
+                case Failure(e) => logger.debug(s"Could not read module descriptor at $moduleInfo", e)
```
ml86: No single letter naming.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
```diff
@@ -0,0 +1,156 @@
...
+                case Success(_) =>
+              }
+            }
+          }
+        }
+      }
+      val exports = exportsBuilder.result()
+
+      val indexMutable   = mutable.Map.empty[String, Path]
```
ml86: Different pattern for map creation in the same method. Above you use `Map.newBuilder` here not. Why?

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
```diff
@@ -0,0 +1,156 @@
...
+  /** Module name to exported package names (for JarTypeSolver.tryToSolveTypeInModule). */
+  val moduleExportsMap: Map[String, List[String]] = moduleExports
+
+  private def buildIndexAndExports(): (Map[String, Path], Map[String, List[String]]) = {
+    val modulesRoot = jrtFileSystem.getPath("/modules")
+    if (!Files.exists(modulesRoot)) {
+      logger.warn(s"jrt: file system has no /modules root for java.home=$javaHome")
+      (Map.empty, Map.empty)
+    } else {
+      val exportsBuilder = Map.newBuilder[String, List[String]]
```
ml86: Split content of else block into two methods.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
```diff
@@ -0,0 +1,156 @@
...
+              javaHome <- Option(lib.getParent)
+            } yield RuntimeImageLayout(javaHome, p)
+          }
+          .minByOption(layout => root.relativize(layout.javaHome).getNameCount)
+      }
+    }
+  }
+
+  /** Directory layout root for a JDK / jlink image (java.home); see findRuntimeImage for the modules file path. */
+  def findRuntimeImageRoot(searchRoot: Path, maxDepth: Int = DefaultRuntimeImageSearchMaxDepth): Option[Path] =
```
ml86: Only used in testing. Move to test code if really needed.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
```diff
@@ -0,0 +1,156 @@
...
+      val root = searchRoot.toAbsolutePath.normalize()
+      Using.resource(Files.walk(root, maxDepth)) { stream =>
+        stream
+          .iterator()
+          .asScala
+          .filter(p => Files.isRegularFile(p) && !Files.isSymbolicLink(p) && p.getFileName.toString == "modules")
+          .flatMap { p =>
+            for {
+              lib <- Option(p.getParent)
+              if lib.getFileName != null && lib.getFileName.toString == "lib" && !Files.isSymbolicLink(lib)
```
ml86: Missing `(` `)` around condition.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
```diff
@@ -0,0 +1,156 @@
...
+    classNameToPath.get(classname).map(p => Files.newInputStream(p)).orNull
+  }
+
+  override def close(): Unit = jrtFileSystem.close()
+}
+
+object JrtRuntimeImageClassPath {
+
+  /** Result of a single filesystem walk: the jrt java.home directory and the exact lib/modules jimage path. */
+  final case class RuntimeImageLayout(javaHome: Path, modulesImageFile: Path)
```
ml86, line 109: `modulesImageFile` is only used for logging and testing. But since the relative location to `javaHome` is always fix, this variable is unnecessary.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
```diff
@@ -0,0 +1,156 @@
...
+          .asScala
+          .filter(p => Files.isRegularFile(p) && !Files.isSymbolicLink(p) && p.getFileName.toString == "modules")
+          .flatMap { p =>
+            for {
+              lib <- Option(p.getParent)
+              if lib.getFileName != null && lib.getFileName.toString == "lib" && !Files.isSymbolicLink(lib)
+              javaHome <- Option(lib.getParent)
+            } yield RuntimeImageLayout(javaHome, p)
+          }
+          .minByOption(layout => root.relativize(layout.javaHome).getNameCount)
```
ml86: Instead of taking the minimum by some random metric we should always just take the first and warn if we found more than one.

ml86 (changes requested):
Did not yet read the test but I think for now I will stop.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JarTypeSolver.scala
johannescoetzee: `ownedCloseables` is used to when closing the classPaths and `ClassPool` doesn't expose added class paths or provide a method to close them, so we need to keep track of them separately.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JarTypeSolver.scala
johannescoetzee: I've refactored this to use a match instead (change will be visible once I've pushed). With `getOrElse`, we lose the exception that caused the issue initially

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
johannescoetzee: The paths built/indexed by `buildIndexAndExports` are paths in this jrtfs, so even though it's not used explicitly, the filesystem still needs to be open to read the classes from it in the other methods.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JarTypeSolver.scala
ml86: Then you could use `.recover(...)`.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
```diff
@@ -92,11 +92,16 @@ class AstCreationPass(config: Config, cpg: Cpg, sourcesOverride: Option[List[Str
...
+  /** Close type solver resources (open JAR files, jrt: file systems). Should be called after all AST creation is done
+    * and [[clearJavaParserCaches]] has been called.
+    */
+  def closeTypeSolvers(): Unit = combinedTypeSolver.close()
+
+  private def initParserAndUtils(config: Config): (SourceParser, JavaSymbolSolver, SimpleCombinedTypeSolver) = {
+    val dependencies                   = getDependencyList(config.inputPath)
+    val sourceParser                   = SourceParser(config, sourcesOverride)
+    val (symbolSolver, combinedSolver) = createSymbolSolver(config, dependencies, sourceParser)
+    (sourceParser, symbolSolver, combinedSolver)
```
ml86, line 104: Alternatively you could derived from `JavaTypeSolver` and handle the resources there this would require less adjustments if for some reason further resources then `combinedSolver` need to be released in the feature. But i think that is unlikely and will leave it to you.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/noncaching/JarTypeSolver.scala
```diff
@@ -171,30 +200,54 @@ object JarTypeSolver {
...
+          case Some(imageRootPath) =>
+            logger.info(s"JDK type solver: using runtime image at $imageRootPath; search root: $inputPath)")
+            builder.addRuntimeImage(imageRootPath) match {
+              case Success(_) => builder
+              case Failure(exception) =>
+                throw new IllegalArgumentException(
+                  s"Could not load JDK runtime image (jrt:) for image root path=$imageRootPath",
+                  exception
+                )
+            }
```
ml86: `.recover { exception => throw... }.get` would be cleaner.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPath.scala
ml86: Makes sense, please put a comment in the code explaining this.

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPathTests.scala
```diff
@@ -0,0 +1,90 @@
...
+    Try(cp.close())
+    super.afterAll()
+  }
+
+  private def assumeRuntimeImage(): Unit = {
+    assume(
+      JrtRuntimeImageClassPath.findRuntimeImage(javaHome).isDefined,
+      s"Skipping: no lib/modules at $javaHome (not a modular JDK layout)"
+    )
+  }
```
ml86: This would basically always skip the tests guarded by `assumeRuntimeImage`. Please make use of our cached resource downloading to always have a JRE at hand.

ml86 (changes requested):

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/passes/AstCreationPass.scala
johannescoetzee: The only somewhat-likely case I can think of is if we add type solvers for our own custom type format, but of course it's possible something else comes up as well. I don't know what that would look like though, so I'm leaning more towards refactoring and handling that if the time comes. I don't think the added complexity of a general resource handling solution for this is worth it yet

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/typesolvers/JrtRuntimeImageClassPathTests.scala
johannescoetzee: As far as I understand, any standard java runtime environment must include `lib/modules`, so unless we start doing weird custom packaging, the `java.home` system property should point at a valid JRE (which the standard JDK distributions also include). I've just removed this check since it shouldn't ever fail (so the tests will now fail if the modules file cannot be found). If the unit tests fail (although I don't expect they would) or this is still something you're concerned about (assuming the unit tests are green) I will add the JRE download though

ml86: Ok, sounds like we can then cover the relevant test cases also with any modern JDK which we can assume to be running the tests. In that case I am fine with going ahead without a dedicated JRE.

ml86 (approved):

PR #5924, opened 2026-04-09
[swiftsrc2cpg] Several performance fixes

max-leuthaeuser:
- Fix O(n²) scope path construction in `computeScopePath` using `StringBuilder`
- Eliminate O(N²) scan and fix data race in `extractFileLocalTypeMap`
- Pre-compile `SwiftCompilerArgSplit` regex pattern
- Add reverse index for `typeDeclFullNameForMember` lookup

With the update to swiftastgen code gen we also got:
- Cache `SwiftNode.toString` to avoid repeated reflection overhead 
- Replace generated AST node wrappers with direct pattern matching (~2800 lines removed)
- Replace O(n) sequential if-chain in `createSwiftNode` with `HashMap`
  dispatch for both node types and token kinds
- Replace repeated `.arr.toList.find()` child accessors with a cached
  `_childrenMap` (`Map[String, Value]`) built once per node, eliminating
  linear scans and intermediate List allocations on every property access
- Replace Try-based range accessors (`startOffset`, `endLine`, etc.) with
  safe `obj.get` lookups, avoiding exception construction and fillInStackTrace overhead when fields are absent


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/AstCreationPass.scala
```diff
@@ -25,7 +27,14 @@ class AstCreationPass(cpg: Cpg, astGenRunnerResult: AstGenRunnerResult, config:
...
 
-  private var typeMap = SwiftTypesProvider(config).map(_.retrieveMappings()).getOrElse(Map.empty)
+  private val typeMap: java.util.concurrent.ConcurrentHashMap[String, SwiftFileLocalTypeMapping] = {
+    val m      = new java.util.concurrent.ConcurrentHashMap[String, SwiftFileLocalTypeMapping]()
+    val source = SwiftTypesProvider(config).map(_.retrieveMappings()).getOrElse(Map.empty)
+    source.foreach { case (filename, mapping) =>
+      m.put(filename.replace("\\", "/"), mapping)
+    }
+    m
+  }
```
ml86: Unrelated to the change but why do we have to have this contention point in the first place?

max-leuthaeuser: The type map is built project wide via SwiftTypesProvider (indexed by filename) but accessed from individual threads per file. To reduce heap pressure we remove the file specific entry once the work is done.
I fixed that particular place with the latest commit to avoid double map conversion.

ml86 (approved):

PR #5928, opened 2026-04-13
[swiftsrc2cpg] Fix resource leaks, reduce lock contention, and batch swift-demangle calls 

max-leuthaeuser:
- Fix Process handle leak in `processFromCommand` (was `inputStreamFromCommand`): return `Process` instead of `InputStream` so callers can manage the process lifecycle; register it with `Using.Manager` in `retrieveMappings()` to ensure `destroyForcibly().waitFor()` on exit
- move `resolve(typeInfo)` which triggers `demangling` via external process out of nested `ConcurrentHashMap.compute()` lambdas in `mappingFromJson()`, so bin-level locks are only held for the cheap map insertion
- Reduce GC pressure from GSON tree materialization: in `GsonTypeInfoReader.parseObject()`, only retain child objects/arrays on the parent `JsonObject` for fields actually accessed by `extractTypeInfo` and its helpers; other fields are still recursed for side effects but not retained
- Drain remaining `stdout` in `findInStdOut` after `Iterator.find` returns, before calling `waitFor()`, to prevent pipe-buffer deadlock when a subprocess produces more output than the OS pipe buffer
- Batch swift-demangle calls into a single process invocation. This reduces the runtime even for smaller projects when using types from swiftc by 50% (e.g., on the Swift Vapor project)!

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/ExternalCommand.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/ExternalCommand.scala
@@ -3,7 +3,7 @@ package io.joern.swiftsrc2cpg.utils
 import io.shiftleft.semanticcpg.utils.ExternalCommandResult
 import org.slf4j.LoggerFactory
 
-import java.io.{BufferedReader, InputStream, InputStreamReader}
+import java.io.{BufferedReader, InputStreamReader}
 import java.nio.file.Paths
 import scala.util.{Failure, Success, Try, Using}
 
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/ExternalCommand.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/ExternalCommand.scala
@@ -65,10 +65,9 @@ object ExternalCommand {
       val process = pb.start()
       val input   = use(new InputStreamReader(process.getInputStream))
       val reader  = use(new BufferedReader(input))
-      val result = Iterator
-        .continually(reader.readLine())
-        .takeWhile(_ != null)
-        .find(find)
+      val iter    = Iterator.continually(reader.readLine()).takeWhile(_ != null)
+      val result  = iter.find(find)
+      while (iter.hasNext) iter.next()
       if (process.waitFor() == 0) { result }
       else None
     } match {
@@ -81,23 +80,23 @@ object ExternalCommand {
     }
   }
 
-  /** Executes a command and returns its output as an InputStream.
+  /** Executes a command and returns the started [[Process]].
     *
-    * This method builds and starts a process with the given command, configuring it to merge stderr into stdout and
-    * setting the working directory.
+    * Callers are responsible for destroying the process when done (e.g. via `destroyForcibly().waitFor()`). The
+    * process's stdout and stderr are merged into a single stream accessible via `process.getInputStream`.
     *
     * @param command
     *   The command to execute
     * @param workingDir
     *   The directory to execute the command in
     * @return
-    *   An InputStream containing the output of the Swift compiler process
+    *   The started Process whose merged stdout/stderr can be read via `getInputStream`
     */
-  def inputStreamFromCommand(command: Seq[String], workingDir: String): InputStream = {
+  def processFromCommand(command: Seq[String], workingDir: String): Process = {
     val builder = new ProcessBuilder(command*)
     builder.directory(Paths.get(workingDir).toFile)
     builder.redirectErrorStream(true)
-    builder.start().getInputStream
+    builder.start()
   }
 
 }
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/ExternalCommand.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/ExternalCommand.scala
@@ -3,7 +3,7 @@ package io.joern.swiftsrc2cpg.utils
 import io.shiftleft.semanticcpg.utils.ExternalCommandResult
 import org.slf4j.LoggerFactory
 
-import java.io.{BufferedReader, InputStreamReader}
+import java.io.{BufferedReader, InputStreamReader, Writer}
 import java.nio.file.Paths
 import scala.util.{Failure, Success, Try, Using}
 
@@ -67,7 +67,7 @@ object ExternalCommand {
       val reader  = use(new BufferedReader(input))
       val iter    = Iterator.continually(reader.readLine()).takeWhile(_ != null)
       val result  = iter.find(find)
-      while (iter.hasNext) iter.next()
+      reader.transferTo(Writer.nullWriter)
       if (process.waitFor() == 0) { result }
       else None
     } match {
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
```diff
@@ -540,10 +548,72 @@ case class SwiftTypesProvider(config: Config, parsedSwiftInvocations: Seq[Seq[St
...
+    try {
+      val process = pb.start()
+      try {
+        Using.resource(new BufferedWriter(new OutputStreamWriter(process.getOutputStream))) { writer =>
+          stdinNames.foreach { name =>
+            writer.write(name)
+            writer.newLine()
+          }
+        }
+        Using.resource(new BufferedReader(new InputStreamReader(process.getInputStream))) { reader =>
```
maltek: This potentiallly dead-locks - the demangler will block on stdout when the pipe buffer is full. And then we will block on writing to the stdin of the demangler. We need either a file to buffer things, or different threads for reading and writing the two pipes. (Or async IO.)

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/ExternalCommand.scala
```diff
@@ -65,10 +65,9 @@ object ExternalCommand {
       val process = pb.start()
       val input   = use(new InputStreamReader(process.getInputStream))
       val reader  = use(new BufferedReader(input))
-      val result = Iterator
-        .continually(reader.readLine())
-        .takeWhile(_ != null)
-        .find(find)
+      val iter    = Iterator.continually(reader.readLine()).takeWhile(_ != null)
+      val result  = iter.find(find)
+      while (iter.hasNext) iter.next()
```
maltek: pointless micro-optimization:
```suggestion
      import java.io.Writer
      reader.transferTo(Writer.nullWriter);
```

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
```diff
@@ -711,37 +831,15 @@ case class SwiftTypesProvider(config: Config, parsedSwiftInvocations: Seq[Seq[St
...
   }
 
   /** Retrieves Swift type mappings by executing Swift compiler commands and processing output.
     *
-    * This method executes all parsed Swift compiler invocations in parallel, processes the JSON output to extract type
-    * information, and builds a comprehensive type mapping. Progress and errors are logged appropriately.
+    * This method works in three phases:
+    *   1. Execute all Swift compiler invocations in parallel and collect raw TypeInfo objects
+    *   1. Batch-demangle all unique mangled names through a single swift-demangle process
+    *   1. Resolve TypeInfo objects (using cached demangle results) and build the type mapping
```
maltek, line 856:
```suggestion
    *   2. Batch-demangle all unique mangled names through a single swift-demangle process
    *   3. Resolve TypeInfo objects (using cached demangle results) and build the type mapping
```

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
```diff
@@ -540,10 +548,72 @@ case class SwiftTypesProvider(config: Config, parsedSwiftInvocations: Seq[Seq[St
...
+            }
+          }
+        }
+        process.waitFor()
+      } finally {
+        process.destroyForcibly()
+      }
+    } catch {
+      case e: Exception =>
+        logger.debug("Batch demangling failed, falling back to per-name demangling", e)
```
maltek, line 627: I don't see the code that does this per-name demangling?

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
```diff
@@ -751,23 +849,37 @@ case class SwiftTypesProvider(config: Config, parsedSwiftInvocations: Seq[Seq[St
...
-          val reader = use(new InputStreamReader(inputStreamFromCommand(invocationCommand, config.inputPath)))
+          val process = processFromCommand(invocationCommand, config.inputPath)
+          use(new AutoCloseable { def close(): Unit = { process.destroyForcibly().waitFor(); () } })
+          val reader = use(new InputStreamReader(process.getInputStream))
           val stdOut = use(new BufferedReader(reader))
-          ParallelLineProcessor.processLinesParallel(stdOut, pool, _.startsWith("{"))(jsonString =>
-            mappingFromJson(jsonString, mapping)
-          )
+          ParallelLineProcessor.processLinesParallel(stdOut, pool, _.startsWith("{")) { jsonString =>
+            collectTypeInfoFromJson(jsonString).foreach(allTypeInfos.add)
```
maltek, line 878: It would be much more efficient to collect intermediate results on each thread, and only merge these once. Many threads pushing to the same list will lead to contention. That is, if the work in `collectTypeInfoFromJson(jsonString)` is cheap. If it is expensive, then it's not worth optimizing.

max-leuthaeuser: collectTypeInfoFromJson is the heavy work there, so thats fine.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
max-leuthaeuser: https://github.com/joernio/joern/blob/7ec9852e4038771a24a7bfb0eeb3ca8111a74bef/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala#L708

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
max-leuthaeuser: No, see: https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html#other-formatting-notes

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
max-leuthaeuser: Yes, different threads it is. Done this.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
maltek: Does it actually reject or incorrectly parse a list with increasing numbers? I can see the point of always using `1.` in lists that are going to change often: it avoids churn in diffs whenever something in the middle is added/removed. But in a case like this where this isn't a concern, it just hurts readability in all the contexts (like the github viewer) where the comment isn't rendered intelligently.

maltek (approved):

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/SwiftTypesProvider.scala
max-leuthaeuser: Yes, scalafmt removes the indentation and everything becomes one block. If you manually add line breaks scaladoc rendering fails rendering the numbered list completely.

maltek: I see 🤮

PR #5930, opened 2026-04-14
[workflows] Update Github actions

max-leuthaeuser:
@mpollmeier
We need to find and implement replacements for:

- https://github.com/actions/create-release, and
- https://github.com/actions/upload-release-asset

Both were archived by the owner on Mar 4, 2021.

Once Github pulls the plug on Node.js 20 they will stop working.

```
Node.js 20 actions are deprecated.
The following actions are running on Node.js 20 and may not work as expected:
actions/checkout@v4, actions/create-release@v1, actions/setup-java@v4, actions/upload-release-asset@v1.
Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026.
Node.js 20 will be removed from the runner on September 16th, 2026.
Please check if updated versions of these actions are available that support Node.js 24.
```

A good candidate would be: https://github.com/softprops/action-gh-release
It allows to do releases and file uploads so we could replace both outdated plugins at once.

There is no update for SwiftyLab/setup-swift yet. I will keep an eye open for that.

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -17,7 +17,7 @@ jobs:
       contents: write
       packages: write
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
       - name: Set up Docker Buildx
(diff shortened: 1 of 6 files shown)


mpollmeier (approved):
https://github.com/softprops/action-gh-release sounds good, and we're using it in a few places already: https://github.com/search?q=action-gh-release+owner%3Ajoernio+&type=code

PR #5939, opened 2026-04-20
[javasrc2cpg] Support new control structure edges

TNSelahle:
- Add support for new control structure edges.

Relates to https://github.com/ShiftLeftSecurity/codescience/issues/8799

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/statements/AstForSimpleStatementsCreator.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/statements/AstForSimpleStatementsCreator.scala
@@ -178,7 +178,7 @@ trait AstForSimpleStatementsCreator { this: AstCreator =>
   private[statements] def astsForIf(stmt: IfStmt): Seq[Ast] = {
 
     val conditionAst =
-      astsForExpression(stmt.getCondition, ExpectedType.Boolean).headOption.toList
+      astsForExpression(stmt.getCondition, ExpectedType.Boolean).headOption
 
     val patternLocals = scope.enclosingMethod.get.getAndClearUnaddedPatternLocals().map(Ast(_))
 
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/statements/AstForSimpleStatementsCreator.scala
```diff
@@ -215,22 +215,21 @@ trait AstForSimpleStatementsCreator { this: AstCreator =>
...
-      .withChildren(elseAst.toList)
-
-    val astWithConditionEdge = conditionAst.flatMap(_.root.toList) match {
-      case r :: Nil =>
-        ast.withConditionEdge(ifNode, r)
-      case _ =>
-        ast
+
+    val astWithChildren = controlStructureAst(ifNode, conditionAst, thenAst :: elseAst.toList)
+    val astWithTrueBody = thenAst.root match {
```
johannescoetzee: If we're going to start using custom edges for different control structures, then I don't know if `controlStructureAst` makes sense as a helper anymore since that leads to the situation below where we need to remember to add the relevant edges manually. I think it would be better to have helpers like `ifAst`, `forAst`, etc. that take care of edge creation for us (similar to how `callAst` creates argument edges). 

Is this something that you think could generalise across languages or do we use inconsistent orderings/structures?

joern-cli/frontends/javasrc2cpg/src/test/scala/io/joern/javasrc2cpg/querying/ControlStructureTests.scala
```diff
@@ -748,6 +748,117 @@ class NewControlStructureTests extends JavaSrcCode2CpgFixture {
...
+          ifTwo.trueBodyOut.astChildren.code.l shouldBe List("c += 10")
+          inside(ifTwo.falseBodyOut.l) { case List(elseNode: ControlStructure) =>
+            elseNode.controlStructureType shouldBe ControlStructureTypes.ELSE
+            elseNode.astChildren.isBlock.astChildren.code.l shouldBe List("c = 10")
+          }
+      }
+    }
+  }
+
+  "`do-while` statements" should {
```
johannescoetzee, line 787: Why are tests for do loops, try statements and for loops included in this PR when it looks like only if statements are actually being updated in code?

TNSelahle: The frontend uses the methods defined in `x2cpg` to build up the asts for those statements. The methods have already been updated to emit the new edges. Just adding these tests for additional coverage.

joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/astcreation/statements/AstForSimpleStatementsCreator.scala
TNSelahle: That's a good point. From what I've seen so far, this has been pretty much the same between the 3 frontends I've touched + `kotlin2cpg` that Louis updated.

Let me take a look at it a bit and see

TNSelahle: @johannescoetzee I've made the change. Please have a look. 

I'll update the other frontends in a separate PR.

johannescoetzee (approved):
Looks good to me

PR #5941, opened 2026-04-21
[dataflowengineoss] Fix DDG visualization missing indirection operators in conditionals

### LINKED ISSUE #5935: [Bug] DDG missing connections in conditionals
**Describe the bug**
I'm trying to use the DDG to see which nodes relate to a boolean expression. I'm finding that some connections are missing, I would say that all the variables used should relate to at least one of the nodes of the expression.

**To Reproduce**
1. Copy the following code
```C
int * glob;

int main(int argc, char **argv) {
  int x = 0;
  int * xx = &x;
  while (((x) && *xx) || *glob){
    x++;
  }
  return 0;
}
```
2. Dump the DDG:
`run.dumpddg`
Image

3. See the DdgIn connections of the condition node
`cpg.controlStructure.condition.ddgIn.l`
```
val res5: List[io.shiftleft.codepropertygraph.generated.nodes.CfgNode] = List(
  Call(
    argumentIndex = 1,
    argumentName = None,
    code = "(x) && *xx",
    columnNumber = Some(value = 11),
    dispatchType = "STATIC_DISPATCH",
    dynamicTypeHintFullName = IndexedSeq(),
    lineNumber = Some(value = 6),
    methodFullName = "<operator>.logicalAnd",
    name = "<operator>.logicalAnd",
    offset = None,
    offsetEnd = None,
    order = 1,
    possibleTypes = IndexedSeq(),
    signature = "",
    staticReceiver = None,
    typeFullName = "ANY"
  ),
  Call(
    argumentIndex = 2,
    argumentName = None,
    code = "*glob",
    columnNumber = Some(value = 26),
    dispatchType = "STATIC_DISPATCH",
    dynamicTypeHintFullName = IndexedSeq(),
    lineNumber = Some(value = 6),
    methodFullName = "<operator>.indirection",
    name = "<operator>.indirection",
    offset = None,
    offsetEnd = None,
    order = 2,
    possibleTypes = IndexedSeq(),
    signature = "",
    staticReceiver = None,
    typeFullName = "ANY"
  )
)
```

4. See how in the rendered image the node `<operator>.logicalOr, 6` doesn't have any incoming connections from a `*glob` node, but in the results of the query I do see it

**Expected behavior**
Seeing the connections found by joern in the rendered graph

 - Arch linux
 - Joern 4.0.460

max-leuthaeuser:
The DDG visualization was filtering out generic member access operators (including indirection `*ptr`, field access, etc.) unconditionally, causing important data flow information to disappear from the graph. Specifically, operations like `*glob` in conditional expressions such as `while (((x) && *xx) || *glob)` were being removed even though they are critical for understanding the control flow.

Root Cause:
- `DdgGenerator` applied blanket filtering of `isGenericMemberAccessName` operators, removing them from both nodes and edges without considering their context or importance
- The `surroundingCall` method was collapsing all Expression nodes, including Call nodes representing operations, into their parent calls, creating self-loops that were then filtered out

Solution:
This PR implements selective filtering that preserves generic member access operators when they contribute meaningful information:

1. Enhanced `surroundingCall()` to explicitly preserve Call nodes while still grouping simple expressions (identifiers, literals) with their containing calls. This prevents operations from being incorrectly collapsed.

2. Added `shouldFilterMemberAccessNode()` that applies context-aware filtering:
   - Keeps operators when they are arguments to control structure conditions (if, while, for, switch). These are important for understanding control flow decisions.
   - Filters out operators in simple pass-through scenarios like basic assignments where they don't add meaningful information.

Fixes #5935


dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/dotgenerator/DdgGenerator.scala
```diff
@@ -52,8 +52,38 @@ class DdgGenerator {
...
+    *
+    * This ensures operations like `*glob` in `if (*glob)` or `while (x && *ptr)` remain visible.
+    */
+  private def shouldFilterMemberAccessNode(node: StoredNode): Boolean = {
+    node match {
+      case call: Call if isGenericMemberAccessName(call.name) =>
+        // Check if this node is an argument to a control structure condition
+        val isInCondition = call.inCall.exists { parentCall =>
+          parentCall.astParent.exists(_.isInstanceOf[ControlStructure])
+        }
```
maltek: 1. this should check for a condition edge instead of an `ControlStructure` AST parent
2. the `.inCall` is too specific to the motivating example. It just happens to work for a single `||` - but boolean operators can be nested arbitrarily.
3. there is at least one condition that doesn't use condition edges: the first argument of `Operators.conditional`

this snippet might help, though it doesn't quite match what I've asked above either:
https://github.com/ShiftLeftSecurity/codescience/blob/92f5360700f9297a7a47ddbd078d9041d05cbd37/securityprofile/src/main/scala/io/shiftleft/securityprofile/passes/finalcustom/anylang/TimingAttackPasswordCompare.scala#L87-L94

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/io/dotgenerator/DotDdgGeneratorTests.scala
```diff
@@ -48,4 +55,68 @@ class DotDdgGeneratorTests extends DataFlowCodeToCpgSuite {
...
+    "filter out indirection operators that are not in control structures" in {
+      FileUtil.usingTemporaryDirectory("dumpddg") { tmpDir =>
+        val opts         = DdgDumpOptions(tmpDir.toString)
+        val layerContext = new LayerCreatorContext(cpg2)
+        new DumpDdg(opts).run(layerContext)
+
+        val dotFile = tmpDir / "0-ddg.dot"
+        val dotContent = Using.resource(Source.fromFile(dotFile.toFile)) { source =>
+          source.getLines().mkString("\n")
+        }
```
maltek:
```suggestion
        val dotContent = java.nio.file.Files.readString(tmpDir / "0-ddg.dot")
```

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/io/dotgenerator/DotDdgGeneratorTests.scala
```diff
@@ -48,4 +55,68 @@ class DotDdgGeneratorTests extends DataFlowCodeToCpgSuite {
...
+    "include indirection operators in DDG when they are part of conditional expressions" in {
+      FileUtil.usingTemporaryDirectory("dumpddg") { tmpDir =>
+        val opts         = DdgDumpOptions(tmpDir.toString)
+        val layerContext = new LayerCreatorContext(cpg)
+        new DumpDdg(opts).run(layerContext)
+
+        val dotFile = tmpDir / "0-ddg.dot"
+        val dotContent = Using.resource(Source.fromFile(dotFile.toFile)) { source =>
+          source.getLines().mkString("\n")
+        }
```
maltek:
```suggestion
        val dotContent = java.nio.file.Files.readString(tmpDir / "0-ddg.dot")
```

max-leuthaeuser (commented):
Thanks, did that.

dataflowengineoss/src/main/scala/io/joern/dataflowengineoss/dotgenerator/DdgGenerator.scala
```diff
@@ -52,20 +52,69 @@ class DdgGenerator {
...
+    node match {
+      case expr: Expression =>
+        // Walk up parent expressions until we find a condition context
+        expr.start
+          .repeat(_.parentExpression)(_.emit)
+          .or(
+            // Check if we reach a node with an incoming CONDITION edge
+            _.in(EdgeTypes.CONDITION),
+            // Check if we're the first argument of a ternary operator
+            _.inCall.nameExact(Operators.conditional).argument.order(1).filter(_ == expr)
```
maltek: the filter doesn't work in cases where we've walked up more than one step. We have to rearrange:

```suggestion
            _.argumentIndex(1).inCall.nameExact(Operators.conditional)
```

maltek (approved):

PR #5942, opened 2026-04-21
[php2cpg] Prevent crashes on keyed foreach list destructuring

PadishahIII:
## Summary
  - avoid routing keyed `foreach (... as $key => list(...))` targets through the unimplemented `astForExpr(PhpListExpr)` path when building the `foreach` code field
  - reuse the existing list-expression code rendering for keyed `foreach` destructuring
  - add a regression test in `ControlStructureTests` for keyed `foreach` list destructuring
  - add a small file-backed integration test in `ProjectParseTests`

  ## Error
  `php2cpg` could throw a `scala.NotImplementedError` on input like:

  ```php
  foreach ($arr as $key => list($a, $b)) {
    echo $key;
    echo $a;
    echo $b;
  }
```
Original error message:
```
scala.NotImplementedError: an implementation is missing
  ...
  Caused by: scala.NotImplementedError: unexpected expression 'PhpListExpr(...)' of type class io.joern.php2cpg.parser.Domain$PhpListExpr
  at io.joern.php2cpg.astcreation.AstForExpressionsCreator.astForExpr(AstForExpressionsCreator.scala:55)
  at io.joern.php2cpg.astcreation.AstForExpressionsCreator.astForKeyValPair(AstForExpressionsCreator.scala:715)
  at io.joern.php2cpg.astcreation.AstForControlStructuresCreator.astForForeachStmt(AstForControlStructuresCreator.scala:178)

```

  ## Why this happens

  The keyed foreach code path was constructing the key => value code via astForKeyValPair(...), which calls astForExpr(...) on both sides. PhpListExpr is not implemented in astForExpr, so foreach ($arr as $key => list(...)) failed before the existing
  lowering logic for list destructuring could run.

  ## Fix

  Keep the fix narrow in astForForeachStmt by rendering PhpListExpr targets with the existing list-expression code helper instead of sending them through astForExpr(...).

  ## Test Plan

  - [x] sbt 'project php2cpg' 'scalafmt' 'Test / scalafmt'
  - [x] sbt 'project php2cpg' 'test'

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForControlStructuresCreator.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForControlStructuresCreator.scala
@@ -173,9 +173,13 @@ trait AstForControlStructuresCreator(implicit withSchemaValidation: ValidationMo
     val iterIdentifier = getTmpIdentifier(stmt, maybeTypeFullName = None, prefix = "iter_")
     val localN         = handleVariableOccurrence(stmt, iterIdentifier.name)
 
+    stmt.keyVar.collect { case PhpVariable(name: PhpNameExpr, _) =>
+      handleVariableOccurrence(stmt, name.name)
+    }
+
     // keep this just used to construct the `code` field
     val assignItemTargetString = stmt.keyVar match {
-      case Some(key) => astForKeyValPair(stmt, key, stmt.valueVar).rootCodeOrEmpty
+      case Some(key) => s"${codeForExpr(key)} => ${codeForExpr(stmt.valueVar)}"
       case None =>
         stmt.valueVar match {
           case x: PhpListExpr => createListExprCodeField(x)
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
@@ -1378,6 +1378,55 @@ class ControlStructureTests extends PhpCode2CpgFixture {
     }
   }
 
+  "foreach statements with key and list destructuring should be represented as a for" in {
+    val cpg = code("""<?php
+      |function foo($arr) {
+      |  foreach ($arr as $key => list($a, $b)) {
+      |    echo $key;
+      |    echo $a;
+      |    echo $b;
+      |  }
+      |}
+      |""".stripMargin)
+
+    val foreachStruct = inside(cpg.method.name("foo").body.astChildren.l) {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForControlStructuresCreator.scala [hunk 1]
```diff
@@ -173,14 +173,17 @@ trait AstForControlStructuresCreator(implicit withSchemaValidation: ValidationMo
...
+
     // keep this just used to construct the `code` field
     val assignItemTargetString = stmt.keyVar match {
-      case Some(key) => astForKeyValPair(stmt, key, stmt.valueVar).rootCodeOrEmpty
-      case None =>
-        stmt.valueVar match {
-          case x: PhpListExpr => createListExprCodeField(x)
-          case x              => astForExpr(x).rootCodeOrEmpty
-        }
+      case Some(key) => s"${astForExpr(key).rootCodeOrEmpty} => ${codeForForeachTarget(stmt.valueVar)}"
```
TNSelahle: Although most of this frontend does it, calling `astForExpr` purely for extracting the code field is not ideal. It's unnecessary computation and sometimes has side-effects that add temp nodes to the CPG. 

Please use the [codeForExpr](https://github.com/joernio/joern/blob/fbe5be6559680945af06d335593c1ea5c9fa2bb1/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala#L113-L115) method to generate the `code` for the key-value pair. You'll need to call this for both the key and value then generate the `code` string.

The other changes, like adding `codeForForeachTarget`, are not necessary if you go with this approach.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/io/ProjectParseTests.scala
TNSelahle, line 1: This is not necessary. Tests added in `joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala` should be sufficient enough.

TNSelahle (changes requested):
Hey @PadishahIII :). Thanks for finding and fixing the issue.

joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstForControlStructuresCreator.scala
PadishahIII: fixed

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/io/ProjectParseTests.scala
PadishahIII: removed

TNSelahle (approved):
Thanks for the fix 👍

ml86 (approved):

PR #5943, opened 2026-04-22
[rust2cpg] setup rust_ast_gen

xavierpinho:
* integrates `rust_ast_gen`
* ci: installs rust, since `rust_ast_gen` relies on `cargo` to understand the project being scanned -- added to workflows that run `sbt test` at some point.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -29,12 +29,8 @@ jobs:
           swift-version: "6.1"
       - name: Install Bundler
         run: gem install bundler -v 2.4.22
-      - name: Delete `.rustup` directory
-        run: rm -rf /home/runner/.rustup # to save disk space
-        if: runner.os == 'Linux'
-      - name: Delete `.cargo` directory # to save disk space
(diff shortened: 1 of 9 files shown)


max-leuthaeuser (approved):

.github/workflows/master.yml
```diff
@@ -29,12 +29,8 @@ jobs:
...
       - name: Install Bundler
         run: gem install bundler -v 2.4.22
-      - name: Delete `.rustup` directory
-        run: rm -rf /home/runner/.rustup # to save disk space
-        if: runner.os == 'Linux'
-      - name: Delete `.cargo` directory # to save disk space
-        run: rm -rf /home/runner/.cargo
-        if: runner.os == 'Linux'
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@stable
```
ml86, line 33: Do we really need the hole rust toolchain or just cargo? If cargo is enough, it might be better to just install that

xavierpinho: Ends up needing it in order to resolve types later

PR #5946, opened 2026-04-23
[php2cpg] fix: slice file content bytes instead of string

TNSelahle:
- `codeForExpr` slices `fileContentString` using the byte offsets. This is incorrect and leads to crashes (and potentially incorrect output) when the expression passed to `codeForExpr` has `startFilePos` and `endFilePos` exceeding the size of the string.
- The correct approach is to first slice the byte array then decode after.


ml86 (approved):

xavierpinho:
Regarding the RustAstGenRunnerTest failure: had the exact same happening with me today, and re-running the job turned out green. (Seems like it's Windows-related somehow.) I'll probably tweak the test if I can't find the root cause quickly to remove this flakiness.

TNSelahle:
@xavierpinho got the same error again just now. Rerunning yet again

xavierpinho:
Sorry about that. I just merged the patch that ignores that specific test for now. Maybe rebasing your branch would also be a safer bet.

PR #5951, opened 2026-04-28
[php2cpg] chore: cleanup unnecessary foreach nodes

TNSelahle:
There's no need to call `handleVariableOccurrence` here. It creates unnecessary temp nodes. 
This was added in https://github.com/joernio/joern/pull/5942 to avoid regressions

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
@@ -1322,9 +1322,8 @@ class ControlStructureTests extends PhpCode2CpgFixture {
       |""".stripMargin)
 
     val foreachStruct = inside(cpg.method.name("foo").body.astChildren.l) {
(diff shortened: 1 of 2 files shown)


ml86 (approved):

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/ControlStructureTests.scala
```diff
@@ -1337,6 +1336,17 @@ class ControlStructureTests extends PhpCode2CpgFixture {
         (initAsts, updateAsts, body)
     }
 
+    inside(initAsts.astChildren.l) { case List(assign: Call, kvBlock: Block) =>
+      assign.code shouldBe "$foo@iter_tmp-0 = $arr"
+
+      inside(kvBlock.astChildren.l) { case List(value: Local, key: Local, _, _) =>
+        value.code shouldBe "$val"
+        value.name shouldBe "val"
```
maltek, line 1344: @TNSelahle do I read this right that there are two different `val` locals in the same method, one at the top level and one inside the loop?

PR #5959, opened 2026-05-04
[x2cpg] Possible replacement implementation for findSupportedBuildFiles.

ml86:
This implementation does not do its own recursion and does not just take
the first ".gradle" file in each subdirectory.
Question is: Was that done on purpose? I highly doubt it but will wait
until futher clarification.

@johannescoetzee Please come back to me on this one to discuss.


max-leuthaeuser:
I guess we should make use of https://github.com/joernio/joern/blob/master/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/SourceFiles.scala here.
There are many things that can go wrong when naively traversing directories.

johannescoetzee:
I don't remember how this came to be, but agree with @max-leuthaeuser that using our `SourceFiles` utility is probably the better approach. I'll make the change and open a PR for that.

ml86:
Obsolete since we merged an alternate solution.

PR #5961, opened 2026-05-05
[linting] Added SingleLetterIdentifiers linter

max-leuthaeuser:
Only checks new code from PRs.
This way we do not have to fix the existing thousands of single letter identifiers currently sitting in the code base right now and all in once.


max-leuthaeuser:
Replaced with https://github.com/joernio/joern/pull/5962

PR #5962, opened 2026-05-05
[linting] Added SingleLetterIdentifiers linter

max-leuthaeuser:
Only run linting on actual changes.

With that, we do not have to fix thousands of single letter identifiers sitting in the existing code base right now.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -59,7 +59,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
-          fetch-depth: 0 # Fetch full history for diff comparison
+          fetch-depth: 1
+      - name: Fetch base branch with depth 1
+        run: git fetch --depth=1 origin ${{ github.event.pull_request.base.ref }}:refs/remotes/origin/${{ github.event.pull_request.base.ref }}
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yml
```diff
@@ -59,16 +59,16 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
-          fetch-depth: 1
+          fetch-depth: 0 # Fetch full history for diff comparison
```
ml86: Cant we somehow fetch `origin/${{ github.event.pull_request.base.ref }}` also with depth 1? Like this we are always fetching so much more.

max-leuthaeuser: Let me check.

max-leuthaeuser: Yes, we can do that:
1. Changed fetch-depth: 0 to fetch-depth: 1 to only fetch the PR branch HEAD
2. Added an explicit fetch step that fetches only the base branch ref with --depth=1

This way, instead of fetching the entire git history, we're only fetching:
- The PR branch HEAD commit (depth 1)
- The base branch HEAD commit (depth 1)

ml86 (approved):

PR #5970, opened 2026-05-11
[kotlin2cpg] Add fallback without dependencies and bump compiler version

ml86:
Try a second time to create the compiler binding context if the first
attempt fails. The second time we only include the standard library as
dependency.
This is mainly aimed at not failing completely for very very large
project with multiple million lines of code and not enough memory.

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
@@ -34,7 +34,7 @@ import java.util.regex.Pattern
 import scala.jdk.CollectionConverters.CollectionHasAsScala
 import scala.jdk.CollectionConverters.EnumerationHasAsScala
 import scala.jdk.StreamConverters.StreamHasToScala
-import scala.util.{Failure, Try}
+import scala.util.{Failure, Try, Using}
 import scala.util.matching.Regex
 
 object Kotlin2Cpg {
@@ -244,14 +244,17 @@ class Kotlin2Cpg extends X2CpgFrontend with UsesService {
               logger.warn("Kotlin compiler analysis failed with exception", exception)
               logger.warn("Fallback: Reconfiguring kotlin compiler environment without dependencies")
 
-              val fallbackEnvironment = CompilerAPI.makeEnvironment(
-                dirsForSourcesToCompile,
-                filesWithJavaExtension,
-                stdlibJars,
-                new ErrorLoggingMessageCollector
-              )
+              Using.resource(
+                CompilerAPI.makeEnvironment(
+                  dirsForSourcesToCompile,
+                  filesWithJavaExtension,
+                  stdlibJars,
+                  new ErrorLoggingMessageCollector
+                )
+              ) { fallbackEnvironment =>
+                createBindingContext(fallbackEnvironment)
+              }(fallbackEnvironment => Disposer.dispose(fallbackEnvironment.getProjectEnvironment.getParentDisposable))
 
-              createBindingContext(fallbackEnvironment)
             } else {
               Failure(exception)
             }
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
```diff
@@ -236,7 +237,31 @@ class Kotlin2Cpg extends X2CpgFrontend with UsesService {
 
       new MetaDataPass(cpg, Languages.KOTLIN, config.inputPath).createAndApply()
 
-      val bindingContext = createBindingContext(environment)
+      val bindingContext =
+        createBindingContext(environment)
+          .recoverWith { exception =>
+            if (dependencyContentRoots != stdlibJars) {
```
max-leuthaeuser, line 243: The condition `dependencyContentRoots != stdlibJars` may not accurately detect whether dependencies were included. If `classPathJars` is empty and `dependencyJars` is empty, then `dependencyContentRoots == stdlibJars` even though dependencies were attempted to be fetched.

Maybe it should be:
`if (dependencyJars.nonEmpty || classPathJars.nonEmpty)`

This more clearly expresses the intent: "retry without dependencies if we actually had some dependencies to remove."

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
```diff
@@ -236,7 +237,31 @@ class Kotlin2Cpg extends X2CpgFrontend with UsesService {
...
+              logger.warn("Kotlin compiler analysis failed with exception", exception)
+              logger.warn("Fallback: Reconfiguring kotlin compiler environment without dependencies")
+
+              val fallbackEnvironment = CompilerAPI.makeEnvironment(
```
max-leuthaeuser: When creating the `fallbackEnvironment`, there's no explicit disposal of the original environment. The Kotlin compiler environment holds disposable resources (see line 246: `Disposer.dispose(environment.getProjectEnvironment.getParentDisposable)`).

Document whether the failed environment needs disposal or consider wrapping in `try-finally` or using a resource manager

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
```diff
@@ -236,7 +237,31 @@ class Kotlin2Cpg extends X2CpgFrontend with UsesService {
...
+                new ErrorLoggingMessageCollector
+              )
+
+              createBindingContext(fallbackEnvironment)
+            } else {
+              Failure(exception)
+            }
+          }
+          .recover { exception =>
+            logger.error("Kotlin compiler analysis failed with exception", exception)
```
max-leuthaeuser, line 263: The fallback path logs:
`logger.warn("Kotlin compiler analysis failed with exception", exception)`

And here if the fallback also fails again:
`logger.error("Kotlin compiler analysis failed with exception", exception)`

This could be confusing in logs because the same exception appears twice with different levels.

Consider:
`logger.warn("Kotlin compiler analysis failed with exception (attempting fallback without dependencies)", exception)`

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
ml86: What does the AI try to tell us here? I only want to retry if something is different and that is encoded.

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
ml86: That is intended because the second log message can also be the only one if no fallback is attempted...

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala
ml86: Done.

gacevedo (approved):

PR #5971, opened 2026-05-12
[rust2cpg] type/method full names via rust_ast_gen.

xavierpinho:
Split into meaningful commits for code review. Hopefully the last PR of this magnitude.

In a nutshell, 
* Use type/method full names emitted by `rust_ast_gen` where available, fall back to `methodAstParentStack` otherwise (which is not enough, but let's not delay further).
* Switch full name path separator from `.` to `::` and align with the rust naming scheme.
* Add initial support for `struct` and basic `impl` block methods.
* Split tests out of `SimpleAstCreationPassTest` into dedicated test classes.
* Add basic data flow tests.


xavierpinho:
Closing this PR for now. I'll split it into a few smaller ones instead.

PR #5975, opened 2026-05-13
[swiftsrc2cpg] Add --exclude-regex support for swift-astgen v0.4.1

max-leuthaeuser:
Integrates swift-astgen v0.4.1 (released in joernio/astgen-monorepo#34), which adds `--exclude-regex` for skipping files by absolute path pattern. Same as jssrc2cpg's existing support for javascript-astgen.

Changes:
- Bump swift-astgen version to 0.4.1
- Wire up `--exclude-regex` flag in `AstGenRunner.runAstGenNative`
- Add explicit tests verifying the binary receives the flag (checks raw JSON output, not just post-filtered CPG)

### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/io/ExcludeTests.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/test/scala/io/joern/jssrc2cpg/io/ExcludeTests.scala
@@ -149,4 +149,22 @@ class ExcludeTests extends AnyWordSpec with Matchers with TableDrivenPropertyChe
 
   }
 
+  "Forwarding --exclude-regex to astgen" should {
+    "prevent the binary from emitting a JSON for an excluded file" in {
+      // Verifies that the regex actually reaches the binary: post-filtering in
+      // execute() would also drop the file from the CPG, so we inspect the raw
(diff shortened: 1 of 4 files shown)


maltek (approved):
how compatible are swift and java regex variants? If - as one of the comments says - we apply the same regex both here and in astgen, that means users must limit themselves to a subset of the regex language that has the same semantics in both engines.

max-leuthaeuser:
Yeah, valid concern. If the same exclude-regex is expected to work both in joern and in swift-astgen, then users are limited to a portability subset of regex features. For simple path filters that’s probably fine, but it should be documented somewhere. Btw: that holds for all frontends / astgens (javascript, dotnet, ruby, and swift at the moment).

A few thing can go wrong: lookbehind, named capture syntax, unicode character classes / word boundaries (e.g., `\b`, `\w`), inline flags (e.g., `(?i)`, `(?m)`, `(?s)`), backreferences.

Maybe we should add a note to the SL docs:
"Use basic regex constructs only; advanced regex features may not behave identically across all language analyzers."

Another option:
We could ensure only one part does the filtering: ideally astgen, and joern only forwards. But that's a larger refactoring. And it may still fail as the different languages implement regex differently again (and/or with slightly different semantics).

PR #5985, opened 2026-05-19
[rust2cpg] rust_ast_gen-based typeFullNames

xavierpinho:
Split of #5971 with only typeFullNames + test cleanups.

* In general, calls are STATIC_DISPATCH even when they are methods, since that's what's faithful. `dyn SomeType` would be DYNAMIC_DISPATCH, but I haven't tackled those yet. So, currently, only unresolved methods are DYNAMIC_DISPATCH.
* Some `typeFullName` are missing in this PR, viz. `typeFullNameForNameRef` and `typeFullNameForType`, but it's already a 500+ line diff.


max-leuthaeuser:
If you want to add tests for things that are currently missing / wrong you may want to use ScalaTest's `pendingUntilFixes { ... }` to document exactly what you would expect.

max-leuthaeuser (approved):

PR #5987, opened 2026-05-20
[x2cpg] Refactor for consistency and concurrency safety

max-leuthaeuser:
Did a small x2cpg cleanup pass. Each commit is independently reviewable.

- Used `Defines.Any` for `parameterIn` fallback type; drop the hardcoded `"ANY"` literal in `AstNodeBuilder.parameterInNode`.
- Made `REF`-link invariant structurally visible in `VariableScopeManager`.
- Fixed `KeyPool` field visibility and mutability.
- Encapsulated `XTypeRecoveryState` iteration counter. Replaced the `public var currentIteration` with a package-private `setIteration` method; reads stay free.
- Collapsed repeated edge handling in `Ast`: Introduced a single `typedEdgeKinds` table so `storeInDiffGraph` / `subTreeCopy` no longer repeat per-edge boilerplate, and share a `mergedWith` helper between `withChild` / `merge`. Public case class signature is unchanged, so all frontends keep working.
- Improved `FrontendHTTPServer` concurrency and add tests: Replace `var httpServer` / `var isRunning` with a single `AtomicReference`, giving at-most-once shutdown via `getAndSet(null)`. Concurrency model documented; new tests cover repeated stop, concurrent stop, stop-while-handler-in-flight, and `stopServerAfterTimeout` idle-shutdown / external-wakeup paths.

Buildbot is also [fine](https://buildbot.global.sltr.io/#/builders/82/builds/630) with it.

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/VariableScopeManager.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/VariableScopeManager.scala
@@ -39,6 +39,23 @@ object VariableScopeManager {
     case MethodScope, BlockScope, TypeDeclScope
   }
 
+  /** Result of one iteration of the link-creation loop in `createVariableReferenceLinks`.
+    *
+    * Encoding the three outcomes as an ADT makes the loop's invariant structural: every iteration that adds a REF edge
+    * either terminates the chain (`Terminate`) or names the next reference to chain from (`Capture`).
+    */
+  private enum LinkStep {
+
+    /** Add a REF edge from the current reference to `target` and stop walking the scope stack. */
+    case Terminate(target: NewNode)
+
+    /** Add a REF edge from the current reference to `target` and continue walking from `next`. */
+    case Capture(target: NewNode, next: NewNode)
+
+    /** Add no edge and continue walking the scope stack. */
+    case Skip
+  }
+
   /** Case class representing a method scope element.
     *
     * @param methodFullName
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/VariableScopeManager.scala
```diff
@@ -355,7 +355,7 @@ class VariableScopeManager {
         localOrCapturedLocalNodeOption.foreach { localOrCapturedLocalNode =>
           transferLineAndColumnInfo(currentReference, localOrCapturedLocalNode)
           diffGraph.addEdge(currentReference, localOrCapturedLocalNode, EdgeTypes.REF)
-          currentReference = nextReference
+          nextReference.foreach(currentReference = _)
```
maltek: There are code paths that don't set `done` to `true` nor change `nextReference`. If we come across one, we get multiple REF edges to `currentReference` which is clearly wrong.

Until now, we'd have gotten a REF edge to `null` which probably throws? So this is likely never reached. But then the algorithm should probably be refactored further to make this clear.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/ProgramSummary.scala
```diff
@@ -75,7 +75,10 @@ object ProgramSummary {
           namespace -> dedupedTypes
         }
         .toMap
-      assert(m.flatMap { case (name, ts) => ts.groupBy(_.name).map(_._2.size) }.forall(_ == 1))
```
maltek, line 78: We have quite a few load-bearing asserts over our code base, and long decided that compiler options that disable them are not something we will ever use.

The big difference between `assert` and `require` is that the error message for `require` puts the blame on the caller which is probably not right here.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/server/FrontendHTTPServer.scala
```diff
@@ -141,39 +160,49 @@ class FrontendHTTPServer(executor: ExecutorService, handleRequest: Array[String]
...
+      synchronized { notifyAll() }
     }
   }
 
-  /** Starts the server and returns the server's randomly chosen port
+  /** Starts the server and returns the server's randomly chosen port.
+    *
+    * `synchronized` here makes the `require(!isRunning)` check and the subsequent bind/start atomic against other
+    * `startup()` calls, so a losing concurrent caller cannot leak a started binding. It does not wait on handler
+    * progress, so it cannot deadlock against in-flight requests.
```
maltek: This synchronizes on the same monitor as the other methods, even though it protects different state.

I would prefer to synchronize on a different object. Could be the `httpServerRef`, or just a `new Object()` used only for this purpose.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/VariableScopeManager.scala
max-leuthaeuser: Yes, it would have thrown a NPE. That did not happen:

```
┌───────────────────────────────┬──────────────────────┬────────┬────────────────┐
│             Path              │       Returns        │ done?  │ nextReference? │
├───────────────────────────────┼──────────────────────┼────────┼────────────────┤
│ Found in current scope        │ Some(variableNodeId) │ ✅ set │ —              │
├───────────────────────────────┼──────────────────────┼────────┼────────────────┤
│ needsEnclosingScope           │ None                 │ —      │ —              │
├───────────────────────────────┼──────────────────────┼────────┼────────────────┤
│ MethodScope, fresh capture    │ Some(localNode)      │ —      │ ✅ set         │
├───────────────────────────────┼──────────────────────┼────────┼────────────────┤
│ MethodScope, already captured │ Some(existing)       │ ✅ set │ —              │
├───────────────────────────────┼──────────────────────┼────────┼────────────────┤
│ Other scope element           │ None                 │ —      │ —              │
└───────────────────────────────┴──────────────────────┴────────┴────────────────┘
```
But the change here shows that this invariant is distributed across nested pattern matches and is not enforced by structure. I will think about that and look for an improvement.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/server/FrontendHTTPServer.scala
```diff
@@ -141,39 +165,52 @@ class FrontendHTTPServer(executor: ExecutorService, handleRequest: Array[String]
...
-
-    isRunning = true
-
-    val port = httpServer.getAddress.getPort
+    val server = HttpServer.create(new InetSocketAddress(0), 0)
+    server.createContext("/run", new RunHandler())
+    server.setExecutor(executor)
+    // Publish before `start()` so any thread observing `isRunning` after the listener is live also sees the ref.
+    httpServerRef.set(server)
+    server.start()
```
maltek: Starting and stopping the same server instance from multiple threads seems unlikely... but this has a race with `stop()`: We must flip the order of these two operations, or else a thread stopping the server might attempt to do so before it is even started. (Or simplify the whole interaction by also locking on `startupLock` in `stop()`. In that case, I believe we could downgrade `httpServerRef` to a `volatile`.)

maltek (approved):

PR #5988, opened 2026-05-20
[rust2cpg] lower tuples, plus literal tests

xavierpinho:


joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
```diff
@@ -585,6 +585,19 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
...
   }
 
+  // TupleExpr =
+  //  Attr* '(' (Expr (',' Expr)* ','?)? ')'
+  private def visitTupleExpr(tupleExpr: TupleExpr): Ast = {
+    if (tupleExpr.expr.isEmpty) {
+      Ast(literalNode(tupleExpr, code(tupleExpr), "()"))
+    } else {
+      val typeFullName = typeFullNameForTupleExpr(tupleExpr)
+      val callNode     = operatorCallNode(tupleExpr, code(tupleExpr), "<operator>.tupleLiteral", Some(typeFullName))
```
max-leuthaeuser, line 595: Not sure about that. DF would not work if we lack proper DF semantics / policies for that operator. Afair, other frontends use some kind of `tmp` / `push` lowering?

xavierpinho: I essentially copied pysrc2cpg's lowering (and there's a `PTF("<operator>.tupleLiteral")` rule in the current DF here.)

I don't have a particular preference - I saw the `arrayInitialiazer` workaround in swift, but python's tuple literal seemed more in line with rust. Is there any question one should be looking for in order to resolve this decision?

max-leuthaeuser: Ok, its fine then (did not look into pysrc2cpg). Go ahead!

max-leuthaeuser (approved):

PR #5991, opened 2026-05-21
[swiftsrc2cpg] [jssrc2cpg] [c2cpg] Support new control structure edges 

max-leuthaeuser:
- Migrate jssrc2cpg, c2cpg, and swiftsrc2cpg to emit the new typed control-structure body edges (`TRUE_BODY`/`FALSE_BODY`, `DO_BODY`, `TRY_BODY`/`CATCH_BODY`/`FINALLY_BODY`, `FOR_INIT`/`FOR_UPDATE`/`FOR_BODY`) by routing through the shared x2cpg helpers (`ifThenElseAst`, `tryCatchAst`, `forAst`, `doWhileAst`).
- Fix `whenTrue`/`whenFalse` in semanticcpg: the legacy fallback used `_.out.collectAll[AstNode].order(...)`, but flatgraph's `.out` yields one neighbor per outgoing edge, so a body block reachable via both AST and a new typed edge was returned twice. Switched the fallback to `astChildren`, keeping legacy behavior for frontends that have not migrated.
- Add per-frontend `ControlStructureTests` exercising the new edges (if / if-elseif-else / if-without-else, do-while / repeat-while, try-catch, for).

Refs: https://github.com/ShiftLeftSecurity/codescience/issues/8799

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
@@ -242,7 +242,11 @@ trait AstForStatementsCreator { this: AstCreator =>
     val conditionAst = wrapInNullComparison(doStmt.getCondition, astForConditionExpression(doStmt.getCondition))
     val bodyAst      = nullSafeAst(doStmt.getBody)
     scope.popScope()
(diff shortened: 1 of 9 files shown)


max-leuthaeuser:
@maltek As per comments in this PR: some (jssrc) passes in CS still rely on explicit orders. I left this untouched here. Once we fixed that on CS side I will do a follow-up PR here.

ml86 (approved):

PR #5996, opened 2026-05-22
[pysrc2cpg] Use Reader interface instead of InputStream

ml86:
We do this to avoid miss matches between the character encoding
when transforming from string to input stream and from input stream
to character stream.
So far we used UTF8 when transforming from string to byte input stream
and the system encoding when going from input stream to character stream
which could cause miss matches on those rare systems which do not use
UTF8 as standard character encoding.


maltek (approved):
I guess a regression test is tricky because we'd have to mess with the JVM-global default encoding?

ml86:
Correct. I tried to create one but failed.

PR #6000, opened 2026-05-22
[x2cpg][jssrc2cpg][c2cpg][swiftsrc2cpg] Bump validation level + fixes + x2cpg adaptations

max-leuthaeuser:
Adds explicit `TRUE_BODY`, `FALSE_BODY`, `DO_BODY` edges to control structures in jssrc2cpg, c2cpg, and swiftsrc2cpg to satisfy `V2` validation.

## Changes

**x2cpg:**
- Created `switchAst` helper in x2cpg
- Added TRUE_BODY edge to `whileAst` helper

**jssrc2cpg:**
- While loops: use `whileAst` helper from x2cpg
- Switch statements: use `switchAst` helper from x2cpg
- For-in/for-of loops: Add `TRUE_BODY` edge to desugared while loop body

**c2cpg:**
- While loops: use `whileAst` helper from x2cpg
- Switch statements: use `switchAst` helper from x2cpg
- ForEach statements: Add `TRUE_BODY` edge to desugared while loop body

**swiftsrc2cpg:**
- Guard statements: Create empty block for `TRUE_BODY` when no following statements
- If expressions: Create empty blocks for `TRUE_BODY` and `CONDITION` when missing
- For loop where clauses: Create empty blocks for `TRUE_BODY` and `CONDITION` when missing
- Updated tests to check `.whenTrue.astChildren` instead of `.whenTrue`

**Validation:**
- Fixed false positive warnings for try-finally (missing `CATCH_BODY` when `FINALLY` at order 2)
- Fixed false positive warnings for multiple catch clauses (missing `FINALLY_BODY` when `CATCH` at order 3)

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
@@ -459,8 +459,6 @@ trait AstForStatementsCreator { this: AstCreator =>
     val testAst =
       createCallAst(testCallNode, base = Some(Ast(hasNextReceiverNode)), receiver = Some(Ast(hasNextReceiverNode)))
 
-    val whileLoopAst = Ast(whileLoopNode).withChild(testAst).withConditionEdge(whileLoopNode, testCallNode)
-
     // while loop variable assignment:
     val whileLoopVariableNode = identifierNode(forStmt, loopVariableName, loopVariableName, idType)
     scope.addVariableReference(loopVariableName, whileLoopVariableNode, idType, EvaluationStrategies.BY_REFERENCE)
@@ -509,13 +507,16 @@ trait AstForStatementsCreator { this: AstCreator =>
     // end surrounding block:
     scope.popScope()
 
-    val whileLoopAstWithBody = whileLoopBlockAst.root match {
-      case Some(blockRoot) => whileLoopAst.withChild(whileLoopBlockAst).withTrueBodyEdge(whileLoopNode, blockRoot)
-      case None            => whileLoopAst.withChild(whileLoopBlockAst)
-    }
+    val whileLoopAst = whileAst(
+      Option(testAst),
+      List(whileLoopBlockAst),
+      code = Option(whileLoopCode),
+      lineNumber = line(forStmt),
+      columnNumber = column(forStmt)
+    )
 
     val blockChildren =
-      List(iteratorAssignmentAst, Ast(loopVariableLocalNode), whileLoopAstWithBody)
+      List(iteratorAssignmentAst, Ast(loopVariableLocalNode), whileLoopAst)
     blockAst(blockNode, blockChildren)
   }
 
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
@@ -259,9 +259,14 @@ trait AstForStatementsCreator { this: AstCreator =>
       case _ =>
         Seq.empty
     }
-    val conditionAst = astForConditionExpression(switchStmt.getControllerExpression)
-    val stmtAsts     = nullSafeAst(switchStmt.getBody)
-    initAsts :+ controlStructureAst(switchNode, Option(conditionAst), stmtAsts)
+    val conditionAst    = astForConditionExpression(switchStmt.getControllerExpression)
+    val stmtAsts        = nullSafeAst(switchStmt.getBody)
+    val astWithChildren = controlStructureAst(switchNode, Option(conditionAst), stmtAsts)
+    val astWithEdge = stmtAsts.headOption.flatMap(_.root) match {
+      case Some(bodyRoot) => astWithChildren.withTrueBodyEdge(switchNode, bodyRoot)
+      case None           => astWithChildren
+    }
+    initAsts :+ astWithEdge
   }
 
   private def astsForCaseStatement(caseStmt: IASTCaseStatement): Seq[Ast] = {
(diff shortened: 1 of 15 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -509,22 +514,26 @@ trait AstForStatementsCreator { this: AstCreator =>
...
+
     val blockChildren =
-      List(iteratorAssignmentAst, Ast(loopVariableLocalNode), whileLoopAst.withChild(whileLoopBlockAst))
+      List(iteratorAssignmentAst, Ast(loopVariableLocalNode), whileLoopAstWithBody)
     blockAst(blockNode, blockChildren)
   }
 
   private def astForWhile(whileStmt: IASTWhileStatement): Ast = {
     val code       = s"while (${nullSafeCode(whileStmt.getCondition)})"
+    val whileNode  = controlStructureNode(whileStmt, ControlStructureTypes.WHILE, code)
```
ml86: Why do you go back from using the common x2cpg functions to composing the `while` AST from scratch? I would have expected that you adjust the common function to create the edge?
I was also wondering why, in the above cases you did not switch to using common x2cpg build functions instead of just amending the individual AST compositions?

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -509,8 +509,13 @@ trait AstForStatementsCreator { this: AstCreator =>
...
     scope.popScope()
 
+    val whileLoopAstWithBody = whileLoopBlockAst.root match {
+      case Some(blockRoot) => whileLoopAst.withChild(whileLoopBlockAst).withTrueBodyEdge(whileLoopNode, blockRoot)
+      case None            => whileLoopAst.withChild(whileLoopBlockAst)
+    }
+
     val blockChildren =
-      List(iteratorAssignmentAst, Ast(loopVariableLocalNode), whileLoopAst.withChild(whileLoopBlockAst))
+      List(iteratorAssignmentAst, Ast(loopVariableLocalNode), whileLoopAstWithBody)
```
ml86: Why is this not using the new `while` AST creation function?

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -393,8 +396,13 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
...
     localAstParentStack.pop()
 
+    val whileLoopAstWithBody = whileLoopBlockAst.root match {
+      case Some(blockRoot) => whileLoopAst.withChild(whileLoopBlockAst).withTrueBodyEdge(whileLoopNode, blockRoot)
+      case None            => whileLoopAst.withChild(whileLoopBlockAst)
+    }
+
     val blockChildren =
-      List(iteratorAssignmentAst, Ast(resultNode), Ast(loopVariableNode), whileLoopAst.withChild(whileLoopBlockAst))
+      List(iteratorAssignmentAst, Ast(resultNode), Ast(loopVariableNode), whileLoopAstWithBody)
```
ml86: Same as above. Use of common creation function.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
```diff
@@ -498,11 +498,19 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
...
-    val ifNode       = controlStructureNode(node, ControlStructureTypes.IF, code)
-    val conditionAst = astForNode(node.conditions)
-    val thenAst      = astForNode(node.body)
-    val elseAst      = node.elseBody.map(astForNode)
+    val code            = this.code(node)
+    val ifNode          = controlStructureNode(node, ControlStructureTypes.IF, code)
+    val conditionAstRaw = astForNode(node.conditions)
+    val conditionAst = conditionAstRaw.root match {
+      case Some(_) => conditionAstRaw
+      case None    => blockAst(blockNode(node.conditions), List.empty)
```
ml86, line 506: As far as I know, C/C++ `if` statements must have a condition. Why do we have this case here?

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForStmtSyntaxCreator.scala
```diff
@@ -91,9 +91,17 @@ trait AstForStmtSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
...
     node.whereClause match {
       case Some(whereClause: WhereClauseSyntax) =>
-        val ifNode  = controlStructureNode(whereClause.condition, ControlStructureTypes.IF, code(whereClause.condition))
-        val testAst = astForNode(whereClause)
-        val consequentAst = astForNode(node.body)
+        val ifNode = controlStructureNode(whereClause.condition, ControlStructureTypes.IF, code(whereClause.condition))
+        val testAstRaw = astForNode(whereClause)
+        val testAst = testAstRaw.root match {
+          case Some(_) => testAstRaw
+          case None    => blockAst(blockNode(whereClause), List.empty)
```
ml86, line 98: Same as for C: This should not be possible.

ml86 (changes requested):
There are many cases where while loop control structures are created by hand. Why?

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForStmtSyntaxCreator.scala
max-leuthaeuser: For C/C++ thats true and I will fix that.
For Swift however we may encounter things like:
`if let baz {}` where we only end up with a local for baz (and no actual identifier AST).
Or:
`if #unavailable(OSX 10.51, *) {}` which is compile-time only and also yields no AST.

ml86: `if let baz {}` is shorthand for checking if the optional baz is none empty and if it is non empty it creates a local in the true branch which is not wrapped by the Optional type. So `Optional<Int>` -> `Int`.

`if #unavailable(OSX 10.51, *) {}`: If this is compile time only, then there is no `if-else` construct left we want to represent because then we would only put the true or false branch into the CPG.
But i guess we do not get this preevaluated from the AST generator in which case we can either evaluate ourself which we likely wont be able to do because of missing information or represent it as some sort of operator call in the condition.

In any case I do not see an `if` statement without a condition.

ml86 (approved):

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForStmtSyntaxCreator.scala
max-leuthaeuser: Will do a follow-up PR for swiftsrc2cpg and clean this up.

PR #6005, opened 2026-05-26
[kotlin2cpg] Emit JUMP_ARGUMENT edges for labeled break/continue statements

SuperUserDone:
Kotlin2cpg of: https://github.com/ShiftLeftSecurity/codescience/issues/8864

### DIFF
--- a/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
+++ b/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
@@ -251,11 +251,39 @@ abstract class AstCreatorBase[Node, NodeProcessor](filename: String)(implicit wi
   def breakAst(node: Node, codeStr: String, labelName: Option[String]): Ast =
     labeledJumpAst(node, ControlStructureTypes.BREAK, codeStr, labelName)
 
+  /** Creates an AST for a break statement. When `labelNumber` is present a `Literal` child is created at order 1 and
+    * connected to the break node via a `JUMP_ARGUMENT` edge.
+    */
+  def breakAst(node: Node, codeStr: String, labelNumber: Option[Int])(implicit dummy: DummyImplicit): Ast =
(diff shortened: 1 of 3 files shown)


ml86 (changes requested):
Can you please put the code to create the AST for a break and continue in reusable form into x2cpgs `AstCreatorBase` class. That way also other frontends can benefit from it.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/AstCreatorBase.scala
```diff
@@ -257,6 +245,42 @@ abstract class AstCreatorBase[Node, NodeProcessor](filename: String)(implicit wi
...
+    * connected to the break node via a `JUMP_ARGUMENT` edge.
+    */
+  def breakAst(node: Node, codeStr: String, labelName: Option[String]): Ast =
+    labeledJumpAst(node, ControlStructureTypes.BREAK, codeStr, labelName)
+
+  /** Creates an AST for a continue statement. When `labelName` is present a `JumpLabel` child is created at order 1 and
+    * connected to the continue node via a `JUMP_ARGUMENT` edge.
+    */
+  def continueAst(node: Node, codeStr: String, labelName: Option[String]): Ast =
+    labeledJumpAst(node, ControlStructureTypes.CONTINUE, codeStr, labelName)
```
ml86: Why did you go for optional `labelName`? We currently have two kinds of break/continue statements we support. One with a label and one with an integer indicating the number of levels to jump out. The `None` case would be a third option.
I suggest you create one `breakAst` which take `String` and which take `Int` to implement those two cases. Same for `continue`.

SuperUserDone: Was under the impression that there are 3 cases:

- Label, eg `break x`
- Integer, eg `break 1`
- No argument, eg `break`

No argument would be handled by emitting no edge and no label or integer, only the control structure node.

Made it so the `break` without an argument is equivalent to `break 1`.

ml86 (approved):

PR #6006, opened 2026-05-26
[swiftsrc2cpg] Implement desugaring for optional binding syntax

max-leuthaeuser:
Add support for optional binding constructs in if/while/guard statements:

**If-let and While-let:**
- Single bindings: `if let a = foo() { ... }`
- Shorthand syntax: `if let a { ... }` (checks existing variable)
- Multiple bindings: `if let a = foo(), let b = bar() { ... }`
- Mixed cases: `if let a = foo(), let b { ... }`
- Tuple patterns: `if let (a, b) = foo() { ... }`
- Mixed simple + tuple: `if let a = foo(), let (b, c) = bar() { ... }`
- While loops: `while let item = iterator.next() { ... }`

**Guard-let:**
- Single bindings: `guard let a = foo() else { exit }`
- Shorthand syntax: `guard let a else { exit }` (checks existing variable)
- Multiple bindings: `guard let a = foo(), let b = bar() else { exit }`
- Mixed cases: `guard let a = foo(), let b else { exit }`
- Guard with other conditions: `guard let a = foo(), flag else { exit }`

_Desugaring strategy:_
- **Condition block**: Creates temp variables for simple optional bindings with initializers, evaluates expressions within nil checks with proper short-circuit semantics using && operator
- **Then/body block**: Creates properly scoped locals, unwraps temp variables into final bindings
- **Tuple patterns**: Excluded from condition desugaring, processed as-is in body block
- **Guard special handling**: Implemented in `astsForBlockElements` (AstCreatorHelper.scala) because guard's then block includes all subsequent statements in the function, not just an explicit block

_Desugaring examples:_

`if let a = foo() { body }` desugars to:
```swift
let <tmp>0;
if { (<tmp>0 = foo()) != nil } {
  let a = <tmp>0
  body
}
```

`if let a = foo(), let b = bar() { body }` desugars to:
```swift
let <tmp>0;
let <tmp>1;
if { (<tmp>0 = foo()) != nil && (<tmp>1 = bar()) != nil } {
  let a = <tmp>0
  let b = <tmp>1
  body
}
```

This ensures proper short-circuit evaluation: `bar()` is only called if `foo()` returns non-nil.

`if let a { body }` desugars to:
```swift
if a != nil {
  body
}
```

`guard let a = foo() else { exit }` desugars to:
```swift
let <tmp>0;
if { (<tmp>0 = foo()) != nil } {
  let a = <tmp>0
  // ... rest of function
} else {
  exit
}
```

`guard let a = foo(), let b else { exit }` desugars to:
```swift
let <tmp>0;
if { (<tmp>0 = foo()) != nil && b != nil } {
  let a = <tmp>0
  // ... rest of function
} else {
  exit
}
```

This does no type "unwrapping" (i.e., Optional[T] -> T).
1) by default (without types from the compiler): we do not have any types there anyway. That would only work if the user code explicitly uses type annotation which mostly just does not happen.
2) with types from the compiler: we set type fullnames for identifiers and call fullnames for calls anyway (directly from what the compiler gives us). Hence, no need to create a synthetic unwrap call. Imho, that would not bring anything to the table w.r.t. data-flow.

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
@@ -154,7 +154,7 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
       val ifNode = controlStructureNode(guardStmt, ControlStructureTypes.IF, code)
 
       // Apply optional binding desugaring for guard let
-      // First, create the block that will hold the unwrapped variables
+      // Create the block that will hold the unwrapped variables (blockNode argument is only used for location info)
       val thenBlockNode = if (elementsAfterGuard.nonEmpty) blockNode(elementsAfterGuard.head) else blockNode(guardStmt)
       scope.pushNewBlockScope(thenBlockNode)
       localAstParentStack.push(thenBlockNode)
@@ -563,12 +563,18 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
 
   /** Information about an optional binding for desugaring if-let/while-let constructs.
     *
+    * Contains the CPG-level names for optional binding desugaring. The localName is the unwrapped variable name that
+    * appears in the then/body block. The tmpName (when present) holds the optional value from the initializer and is
+    * used for the nil check in the condition block, ensuring single evaluation of the initializer expression.
+    *
     * @param localName
-    *   The name of the local variable to be created in the then/body block
+    *   CPG name for the unwrapped variable in the then/body block (e.g., "a" in `if let a = foo()`)
     * @param tmpName
-    *   Optional temp variable name (Some if has initializer, None otherwise)
+    *   CPG name for temporary holding the optional value in condition (e.g., "<tmp>0" in the nil check)
     * @param binding
-    *   The original OptionalBindingConditionSyntax node
+    *   The source-level OptionalBindingConditionSyntax node
+    * @param isWildcard
+    *   True if the pattern is a wildcard (e.g., `if let _ = foo()`), requiring generated name
     */
   protected case class BindingInfo(
     localName: String,
@@ -577,13 +583,6 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
     isWildcard: Boolean
   )
 
-  /** Collects binding information from optional binding conditions for desugaring.
-    *
-    * @param bindings
-    *   The optional binding condition nodes
-    * @return
-    *   Sequence of BindingInfo with local name, optional temp name, binding node, and wildcard flag
-    */
   protected def collectBindingInfos(bindings: Seq[OptionalBindingConditionSyntax]): Seq[BindingInfo] = {
     bindings.map { binding =>
       val (localName, isWildcard) = binding.pattern match {
@@ -595,21 +594,6 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
     }
   }
 
-  /** Analyzes conditions to determine optional binding desugaring strategy.
-    *
-    * @param conditions
-    *   The condition elements to analyze
-    * @param onAllSimple
-    *   Handler for all simple optional bindings
-    * @param onMixed
-    *   Handler for mixed simple and tuple optional bindings
-    * @param onPartial
-    *   Handler for partial desugaring (simple bindings + other conditions)
-    * @param onStandard
-    *   Handler for standard conditions (no simple bindings to desugar)
-    * @return
-    *   The result from the appropriate handler
-    */
   protected def handleOptionalBindingConditions[T](
     conditions: Iterable[ConditionElementSyntax],
     onAllSimple: Seq[OptionalBindingConditionSyntax] => T,
@@ -690,51 +674,57 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
     val hasAnyInitializer = bindingInfos.exists(_.tmpName.isDefined)
 
     if (hasAnyInitializer) {
-      // Create condition block with assignments and checks
       val condBlockNode = blockNode(node)
       scope.pushNewBlockScope(condBlockNode)
       localAstParentStack.push(condBlockNode)
 
-      val assignmentAsts = bindingInfos.flatMap { info =>
-        info.tmpName.map { tmpName =>
+      bindingInfos.foreach { info =>
+        info.tmpName.foreach { tmpName =>
           val tmpLocalNode = localNode(info.binding, tmpName, tmpName, Defines.Any).order(0)
           diffGraph.addEdge(condBlockNode, tmpLocalNode, EdgeTypes.AST)
           scope.addVariable(tmpName, tmpLocalNode, Defines.Any, VariableScopeManager.ScopeType.BlockScope)
-
-          val tmpIdentNode = identifierNode(info.binding, tmpName, tmpName, Defines.Any)
-          scope.addVariableReference(tmpName, tmpIdentNode, Defines.Any, EvaluationStrategies.BY_REFERENCE)
-          val initAst = astForNode(info.binding.initializer.get.value)
-          createAssignmentCallAst(
-            info.binding,
-            Ast(tmpIdentNode),
-            initAst,
-            s"$tmpName = ${code(info.binding.initializer.get.value)}"
-          )
         }
       }
 
-      // Create nil checks for all bindings
       val nilCheckAsts = bindingInfos.map { info =>
-        val (checkName, checkNode) = info.tmpName match {
+        info.tmpName match {
           case Some(tmpName) =>
+            val tmpIdentNode = identifierNode(info.binding, tmpName, tmpName, Defines.Any)
+            scope.addVariableReference(tmpName, tmpIdentNode, Defines.Any, EvaluationStrategies.BY_REFERENCE)
+            val initAst = astForNode(info.binding.initializer.get.value)
+            val assignAst = createAssignmentCallAst(
+              info.binding,
+              Ast(tmpIdentNode),
+              initAst,
+              s"$tmpName = ${code(info.binding.initializer.get.value)}"
+            )
+
             val tmpIdentForCheck = identifierNode(info.binding, tmpName, tmpName, Defines.Any)
             scope.addVariableReference(tmpName, tmpIdentForCheck, Defines.Any, EvaluationStrategies.BY_REFERENCE)
-            (tmpName, Ast(tmpIdentForCheck))
+            val nilNode = literalNode(info.binding, "nil", Option(Defines.Nil))
+            val checkCallNode = createStaticCallNode(
+              info.binding,
+              s"($tmpName = ${code(info.binding.initializer.get.value)}) != nil",
+              Operators.notEquals,
+              Operators.notEquals,
+              Defines.Bool
+            )
+            callAst(checkCallNode, List(assignAst, Ast(nilNode)))
+
           case None =>
-            (info.localName, astForNode(info.binding.pattern))
+            val patternAst = astForNode(info.binding.pattern)
+            val nilNode    = literalNode(info.binding, "nil", Option(Defines.Nil))
+            val checkCallNode = createStaticCallNode(
+              info.binding,
+              s"${info.localName} != nil",
+              Operators.notEquals,
+              Operators.notEquals,
+              Defines.Bool
+            )
+            callAst(checkCallNode, List(patternAst, Ast(nilNode)))
         }
-        val nilNode = literalNode(info.binding, "nil", Option(Defines.Nil))
-        val checkCallNode = createStaticCallNode(
-          info.binding,
-          s"$checkName != nil",
-          Operators.notEquals,
-          Operators.notEquals,
-          Defines.Bool
-        )
-        callAst(checkCallNode, List(checkNode, Ast(nilNode)))
       }
 
-      // Combine nil checks with additional conditions using &&
       val additionalConditionAsts = additionalConditions.map(condElem => astForNode(condElem.condition))
       val allChecks               = nilCheckAsts ++ additionalConditionAsts
       val combinedCheckAst        = combineNilChecksWithAnd(node, allChecks)
@@ -742,9 +732,8 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
       scope.popScope()
       localAstParentStack.pop()
 
-      blockAst(condBlockNode, assignmentAsts.toList :+ combinedCheckAst)
+      blockAst(condBlockNode, List(combinedCheckAst))
     } else {
-      // All bindings have no initializer: create combined != nil check without block
       val nilCheckAsts = bindingInfos.map { info =>
         val patternAst = astForNode(info.binding.pattern)
         val nilNode    = literalNode(info.binding, "nil", Option(Defines.Nil))
@@ -825,19 +814,10 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
         blockAst(bodyBlockNode, unwrapAsts.toList ++ bodyAsts)
       }
     } else {
-      // All bindings have no initializer: just use original body
       astForNode(bodyNode)
     }
   }
 
-  /** Builds unwrap assignment ASTs for optional binding info without wrapping in a body block. Used for guard let where
-    * bindings need to be in parent scope.
-    *
-    * @param bindingInfos
-    *   The binding information
-    * @return
-    *   List of assignment ASTs for unwrapping
-    */
   protected def buildUnwrapAssignments(bindingInfos: Seq[BindingInfo]): List[Ast] = {
     val bindingsWithInitializer = bindingInfos.filter(info => info.tmpName.isDefined && !info.isWildcard)
 
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
@@ -155,31 +155,38 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
 
       // Apply optional binding desugaring for guard let
       // Create the block that will hold the unwrapped variables (blockNode argument is only used for location info)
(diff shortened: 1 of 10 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -147,21 +150,54 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
 
-      val code         = this.code(guardStmt)
-      val ifNode       = controlStructureNode(guardStmt, ControlStructureTypes.IF, code)
-      val conditionAst = astForNode(guardStmt.conditions)
+      val code   = this.code(guardStmt)
+      val ifNode = controlStructureNode(guardStmt, ControlStructureTypes.IF, code)
+
+      // Apply optional binding desugaring for guard let
+      // First, create the block that will hold the unwrapped variables
+      val thenBlockNode = if (elementsAfterGuard.nonEmpty) blockNode(elementsAfterGuard.head) else blockNode(guardStmt)
```
maltek: can we make it clearer that the parameter of `blockNode` is only used for location info?

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -516,4 +552,319 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
+
+  /** Information about an optional binding for desugaring if-let/while-let constructs.
+    *
+    * @param localName
+    *   The name of the local variable to be created in the then/body block
+    * @param tmpName
+    *   Optional temp variable name (Some if has initializer, None otherwise)
+    * @param binding
+    *   The original OptionalBindingConditionSyntax node
+    */
```
maltek, line 611: I'm not really any wiser about what this class and each of its field represent after reading this comment. Are these CPG-level names or source-level names? What is the temporary variable used for?

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -516,4 +552,319 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
+    isWildcard: Boolean
+  )
+
+  /** Collects binding information from optional binding conditions for desugaring.
+    *
+    * @param bindings
+    *   The optional binding condition nodes
+    * @return
+    *   Sequence of BindingInfo with local name, optional temp name, binding node, and wildcard flag
+    */
```
maltek: this comment doesn't say anything the name and signature of the method already tell me. If there's nothing interesting to say about the method, don't make me read all that text. There are more such cases in this file, which I won't all flag individually.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -516,4 +552,319 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
+  protected def buildOptionalBindingCondition(
+    node: SwiftNode,
+    bindingInfos: Seq[BindingInfo],
+    additionalConditions: Seq[ConditionElementSyntax] = Seq.empty
+  ): Ast = {
+    val hasAnyInitializer = bindingInfos.exists(_.tmpName.isDefined)
+
+    if (hasAnyInitializer) {
+      // Create condition block with assignments and checks
```
maltek: redundant comment (and there are a bunch more in this file)

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
```diff
@@ -498,15 +498,114 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
...
+    )
+  }
+
+  /** Handles Swift optional binding (if-let) constructs.
+    *
+    * De-sugars `if let baz = foo() { body }` into:
+    *
+    * Condition: { let <tmp>0 = foo(); <tmp>0 != nil }
+    *
+    * Then block: { let baz = <tmp>0; body }
```
maltek: Do we really need the temporary here, or could we get away with assigning to `baz` immediately? Maybe by adding a surrounding block, to avoid polluting the surrounding scope with variables it doesn't contain. This would make dataflows easier to read.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
```diff
@@ -498,15 +498,114 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
...
+    *
+    * For multiple bindings `if let a = foo(), let b = bar() { body }`:
+    *
+    * Condition: { <tmp>0 = foo(); <tmp>1 = bar(); <tmp>0 != nil && <tmp>1 != nil }
+    *
+    * Then block: { a = <tmp>0; b = <tmp>1; body }
```
maltek, line 540: I highly suspect but have not verified that `bar()` only gets evaluated if `a` is not `nil`. Might be easier to get this right by lowering the condition as `(<tmp>0 = foo()) != nil && (<tmp>1 = bar()) != nil`.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
```diff
@@ -498,15 +498,114 @@ trait AstForExprSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
...
+    ifThenElseAst(ifNode, Option(conditionAst), thenAst, elseAst)
+  }
+
+  /** Handles mixed optional binding constructs with both simple and tuple patterns.
+    *
+    * De-sugars `if let a = foo(), let (b, c) = bar() { body }` into:
+    *
+    * Condition: { <tmp>0 = foo(); <tmp>0 != nil }
+    *
+    * Then block: { let a = <tmp>0; let (b, c) = bar(); body }
```
maltek, line 569: how is the tuple assignment desugared?

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForStmtSyntaxCreator.scala
```diff
@@ -561,11 +556,127 @@ trait AstForStmtSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
...
+
+  /** Handles Swift optional binding (guard-let) constructs.
+    *
+    * De-sugars `guard let x = foo() else { exit }` into:
+    *
+    * Condition: { let <tmp>0 = foo(); <tmp>0 != nil }
+    *
+    * Then block: { let x = <tmp>0 }
+    *
+    * Else block: { exit }
```
maltek: in which scope is the `x` variable defined here? It's not just in the then block I hope?

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/GuardTests.scala
```diff
@@ -208,6 +210,188 @@ class GuardTests extends SwiftSrc2CpgSuite {
...
+      thenNodes should not be empty
+
+      // Verify no new local named "optionalValue" was created in the then branch
+      val localsInThen = thenNodes.flatMap(_.ast.isLocal.nameExact("optionalValue").l)
+      localsInThen shouldBe empty
+
+      // Verify no assignment to optionalValue in the then branch
+      val assignmentsInThen        = thenNodes.flatMap(_.ast.isCall.nameExact(Operators.assignment).l)
+      val optionalValueAssignments = assignmentsInThen.filter(_.argument(1).code == "optionalValue")
+      optionalValueAssignments shouldBe empty
```
maltek: this could be a LOT simpler based on the comment at the top that this is supposed to be an empty block

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/GuardTests.scala
```diff
@@ -208,6 +210,188 @@ class GuardTests extends SwiftSrc2CpgSuite {
...
+      val List(andCheck) = condBlock.astChildren.isCall.nameExact(Operators.logicalAnd).l
+      val arguments      = andCheck.argument.l
+      arguments should have size 2
+
+      // One should be the nil check, the other should be the flag identifier
+      val nilChecks =
+        arguments.collect { case c if c.isCall => c }.flatMap(_.ast.isCall.nameExact(Operators.notEquals).l)
+      val flags = arguments.collect { case i if i.isIdentifier => i }.flatMap(_.ast.isIdentifier.nameExact("flag").l)
+
+      nilChecks.code.l should contain(s"$tmpName != nil")
```
maltek: this could be a single small `inside` match

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/GuardTopLevelTests.scala
```diff
@@ -18,12 +18,21 @@ class GuardTopLevelTests extends SwiftSrc2CpgSuite {
       val List(globalBlock) = cpg.method.nameExact("<global>").block.l
       val List(localA)      = globalBlock.local.nameExact("a").l
       localA.typeFullName shouldBe "Swift.Int"
-      val List(localB) = globalBlock.local.nameExact("b").l
-      localB.typeFullName shouldBe "ANY"
-      val assigns = cpg.call.nameExact(Operators.assignment).code.l
-      assigns shouldBe List("let a: Int? = 1", "let b = a")
+
+      // After desugaring, b is in the guard's then block, not the global block
```
maltek, line 29: that's incorrect. `b` has method scope here.

max-leuthaeuser: The whole guard statement is de-sugared like that (was already that way):

```
someFunc {
  <some code 1>
  guard <condition> else { <else code> }
  <some code 2>
}

to:

someFunc {
  <some code 1>
  if <condition> {
    <some code 2>
  } else {
    <else code>
  }
}
```

So in our case b is only valid in the then scope.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
max-leuthaeuser: But we need it for the comparison and the assignment. We can't evaluate it twice. Or do I miss something?

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
max-leuthaeuser: Damn, that's true. I need to change that.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForStmtSyntaxCreator.scala
max-leuthaeuser: See my other comment.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
maltek: The idea was to use the new variable for the comparison just like we now use the temporary. I.e. turn `if let baz = foo() { body }` into one of:


```
{ // new block
  let baz = foo();
  if (baz != nil) {
    body
  }
  // else can't access baz, so it doesn't matter that we wrote to it
}
```

```
{ // new block
  let baz;
  if ((baz = foo()) != nil) {
    body
  }
  // else can't access baz, so it doesn't matter that we wrote to it
}
```

But I've actually realized now that shadowing breaks this again:

```
let nn = 42;
func square(n: Int?) -> Int {
    if let nn = n {
      return nn * nn // local 'nn'
    } else {
      return nn // global 'nn'
    }
}
```

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForExprSyntaxCreator.scala
max-leuthaeuser: This is not done with full coverage there. That needs an additional PR.
There is special tuple de-sugaring, but only in condition/pattern-matching contexts like `if case`, `guard case`, and `switch`, where tuple subjects get de-sugared into tmp variables plus .0, .1, etc. accesses. That de-sugaring is in `astsForBindingTuplePattern` / `astsForBindingTupleExpr`. 

For optional tuple pattern bindings here we can reuse it mostly (store in tmp and do the `nil` comparison; locals and .0, .1, etc. accesses in then block).

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/GuardTopLevelTests.scala
maltek: OK, I see. Could you extend the test to show this? Just in case this is ever changed, we want the `local` to move together with the trailing code following the `guard`.

maltek (commented):
I'm not happy with some of the tests. They are so fuzzy. E.g. `AvailabilityQueryTests` is using `contains` and `not be empty` all over the place instead of complete matches that'd tell us exactly what's there and what isn't there. And some of the queries in there are also structured similarly imprecise - e.g. we know that `nilCheck` has an assignment argument, but we don't know if it has any other arguments.

(It's not just this test though, some of the others are similar.)

max-leuthaeuser:
Ok, I'll overhaul them (tbh they were bad already before this PR).

max-leuthaeuser:
> I'm not happy with some of the tests. They are so fuzzy. E.g. `AvailabilityQueryTests` is using `contains` and `not be empty` all over the place 

Are we talking about the same file? I can not find any "not be empty" assertion there.
There are a few "shouldBe empty" assertions but that's expected (testing for empty true/false blocks).

~~Do not yet re-review. I am still bug-hunting a local creation issue.~~ Got it.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/GuardTests.scala
```diff
@@ -37,16 +36,26 @@ class GuardTests extends SwiftSrc2CpgSuite {
...
-      guardIf.code should startWith("guard i % 2 == 0")
       guardIf.controlStructureType shouldBe ControlStructureTypes.IF
-      guardIf.condition.code.l shouldBe List("i % 2 == 0")
-      guardIf.whenTrue.astChildren.code.l shouldBe List("print(i)", "i = i + 1")
-      whileBlock.astChildren.isControlStructure.whenFalse.astChildren.code.l shouldBe List("i = i + 1", "continue")
+
+      val List(condition) = guardIf.condition.l
+      condition.code shouldBe "i % 2 == 0"
+
+      val List(thenPrint, thenAssign) = guardIf.whenTrue.astChildren.isCall.l
```
maltek, line 48: generally, I think it's better to write such tests like this instead:

```scala
      val List(thenPrint: Call, thenAssign: Call) = guardIf.whenTrue.astChildren.l
```

That way, any potential children which are not calls aren't swept under the rug.

(Depending on the warning levels of the project, `inside` might be required.)

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
maltek, line 223: are these the `pop`s corresponding to the `push`es in the different handlers? Could you add a comment somewhere about this?

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForStmtSyntaxCreator.scala
maltek: this is outdated - should be `(<tmp>0 = foo()) != nil && ...`

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/GuardTests.scala
maltek, line 369: we also need a test with 3 `let`s, to verify that the `&&` operators are nested in the right way.

joern-cli/frontends/swiftsrc2cpg/src/test/scala/io/joern/swiftsrc2cpg/passes/ast/GuardTests.scala
max-leuthaeuser: Yeah, warning level do not allow this implicit type narrowing.
Rewriting everything to `inside` matchers is something thats needs to be addressed separately as the current test style with `.isFoo` traversals is scattered across the whole test base here.

maltek (approved):

PR #6007, opened 2026-05-26
[rust2cpg] lower tuple-like struct calls

xavierpinho:
Rewrites tuple-like struct literals into a block with alloc + field access assignments.

Originally, I was thinking of adding an `<init>` to the structs (TYPE_DECLS) and rewrite the calls to use such method. This would work ok for parsed struct declarations, but not so for externally-defined structs. In rust, a declaration `struct Foo(i32,bool)` (implicitly) introduces a function `Foo(x:i32,y:bool) -> Foo`. So, the original rewrite would create a call with a non-existing methodFullName, viz. `Foo::<init>`.

The alternative approach here is to lower both internal/external struct initializations as a block, in which we allocate a tmp with the appropriate type, and set each field accordingly.

This doesn't handle generic structs yet. (And I couldn't find any non-generic stdlib struct to use to test this.) The current `isStructCtorCall` predicate needs to be updated. I could perhaps do some string trimming to remove generics and compare stripped types, but I believe it's saner to maybe signal these cases from rust_ast_gen properly. So, I would leave generics pending until the next rust_ast_gen release.

### DIFF
--- a/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
+++ b/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
@@ -298,18 +298,69 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
   // CallExpr =
   //  Attr* Expr ArgList
   private def visitCallExpr(callExpr: CallExpr): Ast = {
-    viewExprAsPathExpr(callExpr.expr) match {
-      case Some(nameRefs) =>
-        val name           = code(nameRefs.last)
-        val methodFullName = methodFullNameForCallExpr(callExpr, nameRefs)
-        val typeFullName   = typeFullNameForExpr(callExpr)
-        val dispatch       = DispatchTypes.STATIC_DISPATCH
-        val call =
-          callNode(callExpr, code(callExpr), name, methodFullName, dispatch, None, Some(typeFullName))
-        val args = callExpr.argList.expr.map(visitExpr)
-        callAst(call, args)
-      case None => notHandledYet(callExpr)
+    if (isStructCtorCall(callExpr)) {
+      lowerStructCtorCall(callExpr)
+    } else {
+      viewExprAsPathExpr(callExpr.expr) match {
+        case Some(nameRefs) =>
+          val name           = code(nameRefs.last)
+          val methodFullName = methodFullNameForCallExpr(callExpr, nameRefs)
+          val typeFullName   = typeFullNameForExpr(callExpr)
+          val dispatch       = DispatchTypes.STATIC_DISPATCH
+          val call =
+            callNode(callExpr, code(callExpr), name, methodFullName, dispatch, None, Some(typeFullName))
+          val args = callExpr.argList.expr.map(visitExpr)
+          callAst(call, args)
+        case None => notHandledYet(callExpr)
+      }
     }
+  }
+
+  // Lowers a tuple-like struct call e.g. `Foo(x,y)` into an alloc-followed-by-field-assignments-block:
+  // BLOCK
+  //   LOCAL tmp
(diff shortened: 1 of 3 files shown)


ml86:
I dont quite get the reasoning behind the choice against `<init>` methods for external tuple like structs. You could just as easily synthesize them like you now to with the blocks right? With the here now implemented approach we effectively just inlined all the calls to those `<init>` functions which is fine.

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
```diff
@@ -298,18 +298,69 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
...
+  //   tmp.1 = y
+  //   tmp
+  // We perform this lowering regardless of where the struct is defined (internal/external).
+  // For internal structs, it would be easy to add a fake "<init>" method that performs the assignments.
+  // But to do the same for external structs, we would have to track them all, and visit the file
+  // more than once.
+  private def lowerStructCtorCall(callExpr: CallExpr): Ast = {
+    // LOCAL tmp
+    val typeFullName = typeFullNameForExpr(callExpr)
+    val local        = localNode(callExpr, name = "tmp", code = "tmp", typeFullName)
```
ml86, line 333: Use a string constant for "tmp" here and in the other places.

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
```diff
@@ -298,18 +298,69 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
...
+      val assign    = assignmentNode(callExpr, s"tmp = ${Operators.alloc}")
+      callAst(assign, Seq(tmpAst, Ast(allocCall)))
     }
+
+    // tmp.0 = x; tmp.1 = y; etc.
+    val fieldAssignAsts = callExpr.argList.expr.zipWithIndex.map { case (argExpr, i) =>
+      val argAst      = visitExpr(argExpr)
+      val argType     = typeFullNameForExpr(argExpr)
+      val tmp         = identifierNode(argExpr, "tmp", "tmp", typeFullName)
+      val baseAst     = Ast(tmp).withRefEdge(tmp, local)
```
ml86, line 349: Do not create the REF edge manually like this. I will be a headache to maintain this. You anyway need to have a scoping mechanism that calculated the association between identifiers and locals/parameters. Please use that here as well.

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
```diff
@@ -298,18 +298,69 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
...
+      val fieldAccess = fieldAccessAst(argExpr, argExpr, baseAst, s"tmp.$i", i.toString, argType)
+      val assign      = assignmentNode(argExpr, s"tmp.$i = ${code(argExpr)}")
+      callAst(assign, Seq(fieldAccess, argAst))
+    }
+
+    // tmp
+    val retAst = {
+      val tmp = identifierNode(callExpr, name = "tmp", code = "tmp", typeFullName)
+      Ast(tmp).withRefEdge(tmp, local)
```
ml86, line 358: Same as above. No manual ref edges please.

ml86 (changes requested):

xavierpinho:
@ml86 and I discussed this offline and I will extend rust_ast_gen with more information so that i) we have a more accurate predicate and ii) use rust naming convention for the `<init>` method, so that call site lowering becomes trivial.

PR #6008, opened 2026-05-27
publish linterRules

bbrehm:
We plan to also use the joern linter-rules in downstream closed-source codescience. Hence, this PR would publish them along the other joern subprojects.

@max-leuthaeuser : your linter rule that checks for single-letter identifiers in `case c: Case` uses itself single-letter identifiers, and you @ml86 personally approved that.

I applaud this level of trolling. You made your point, let's allow single-letter case matching, and leave the old code commented-out in order to make the next reader smile.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,7 +27,6 @@ lazy val abap2cpg          = Projects.abap2cpg
 lazy val rust2cpg          = Projects.rust2cpg
 lazy val linterRules       = Projects.linterRules
 
-// aggregate project which doesn't include the helper project `linterRules` - we don't want to include it in any standard task
 lazy val root = project
   .in(file("."))
   .aggregate(
@@ -51,7 +50,8 @@ lazy val root = project
     swiftsrc2cpg,
     csharpsrc2cpg,
     abap2cpg,
-    rust2cpg
+    rust2cpg,
+    linterRules
   )
   .dependsOn(linterRules % ScalafixConfig)
 
--- a/linter-rules/build.sbt
+++ b/linter-rules/build.sbt
@@ -1,4 +1,6 @@
 name := "linter-rules"
+//scalafix otherwise gets unhappy because of circular dependency
+disablePlugins(ScalafixPlugin)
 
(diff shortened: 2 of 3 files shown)


max-leuthaeuser:
Well, this linter itself has more single letter identifier anyway and should be cleaned-up. I would keep the case rule, and do the clean-up right away.

ml86 (approved):

bbrehm:
cf https://github.com/joernio/joern/pull/5962 for cross reference.

@max-leuthaeuser -- I'm not entirely sure what you want regarding the single-letter match/case. We floated 3 variants, and you were not explicit in your preference:
1. Keep SingleLetterIdentifiers.scala‎ unmodified
2. Use multi-letter identifiers in SingleLetterIdentifiers.scala
3. Do the modification of  SingleLetterIdentifiers.scala‎ that this PR proposed: We change the linting rules such that `case c: Case => ...` is no longer a linting violation.

My preference ordering is 3 better than 1 better than 2.

`case c: Case => // issue linting violation that "case c: Case => ..." is unreadable` is beautiful, trolling is a art, and I applaud you for it. It made me smile and will make everyone smile who ever looks at this code. And I think that this piece of code makes a strong argument for (3).

There is no pressing reason against (1) anyways: Due to sbt circular-dependency shennenigans, the linting rule currently cannot lint itself, so this won't cause any linter errors ;)

max-leuthaeuser:
I really do not have a strong opinion on that. I would have done point 2, fixing the single letter identifiers in `SingleLetterIdentifiers.scala`. And then keep the case rule.

The benefits of this case rule leave me with  a very mixed feeling anyway. For simple single line cases, it doesn't make much sense to enforce it. But when it comes to complex logic in the case block where the case variable is used multiple times, it would probably make sense to have a readable and meaningful variable naming in place.

ml86 (changes requested):
@bbrehm Stop trying to open up that discussion again.

ml86 (approved):

PR #6016, opened 2026-06-02
[rust2cpg] basic REF linking for identifiers to locals and parameters.

xavierpinho:
Brings in VariableScopeManager to batch-create REF edges once the visitor has finished.

### DIFF
--- a/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
+++ b/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
@@ -2,6 +2,7 @@ package io.joern.rust2cpg.astcreation
 
 import io.joern.rust2cpg.parser.RustNodeSyntax.*
 import io.joern.x2cpg.datastructures.Stack.*
+import io.joern.x2cpg.datastructures.VariableScopeManager.ScopeType
 import io.joern.x2cpg.{Ast, Defines, ValidationMode}
 import io.shiftleft.codepropertygraph.generated.{
   ControlStructureTypes,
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
ml86, line 183: Can we rename `type` in the generated classes to `typ`, having to use `` ` `` must be annoying.

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
```diff
@@ -184,6 +185,7 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
               case None =>
                 val lhsName = code(identToken)
                 val local   = localNode(identToken, lhsName, lhsName, typeFullName)
+                variableScope.addVariable(local.name, local, local.typeFullName, ScopeType.BlockScope)
```
ml86: I would have assumed that a `let` always introduces a variable, independent from the existence of a RHS initialization expression.

xavierpinho: Of course, my bad. I had not removed an existing `withRefEdge` and that's why tests passed. Fixed now.

In the process also included a basic shadowing test which, at the moment, doesn't introduce a new local, e.g. in `let x = 1; let x = 2` (the latter is only an assignment referencing the local node introduced by the former.)

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
xavierpinho: There's a patch in https://github.com/joernio/astgen-monorepo/pull/52. Once a new version is tagged, I'll bring it and update here as well.

xavierpinho:
After an offline discussion with @ml86, I'm reworking the consecutive `let x` linking - so there are 2 local nodes instead of just one being reused. (VariableScopeManager would also always link to the latter, so it needs to tweaked as well.)

PR #6018, opened 2026-06-03
[rust2cpg] bring TypeNodePass.

xavierpinho:
* replaces the `rustAstGen{MethodFullName,TypeFullName}` helpers by their `RustNode.{methodFullName,typeFullName}` counterparts now that they are auto-generated.
* expands `typeFullNameForType` but there are still a few cases to be worked.
* brings RustTypeNodePass to create the missing `TYPE` nodes.

### DIFF
--- a/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/passes/RustTypeNodePass.scala
+++ b/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/passes/RustTypeNodePass.scala
@@ -4,20 +4,16 @@ import io.joern.x2cpg.passes.frontend.TypeNodePass
 import io.shiftleft.codepropertygraph.generated.Cpg
 import io.joern.rust2cpg.astcreation.RustFullNames
 
-object RustTypeNodePass {
-  def withTypesFromCpg(cpg: Cpg): TypeNodePass = {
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/passes/RustTypeNodePass.scala
```diff
@@ -0,0 +1,23 @@
+package io.joern.rust2cpg.passes
+
+import io.joern.x2cpg.passes.frontend.TypeNodePass
+import io.shiftleft.codepropertygraph.generated.Cpg
+import io.joern.rust2cpg.astcreation.RustFullNames
+
+object RustTypeNodePass {
+  def withTypesFromCpg(cpg: Cpg): TypeNodePass = {
+    new TypeNodePass(Set.empty, cpg, getTypesFromCpg = true) {
+
```
ml86: I know you just used the same pattern as the other code implementing `TypeNodePass` but I dont see the benefit of using a factory method plus an anonymous class here. It will just make it harder to debug since the runtime type will be anonymous.
I suggest you just go for the classic `class RustTypeNodePass(...) extends TypeNodePass(...)`.

ml86 (approved):

ml86 (approved):

PR #6021, opened 2026-06-10
[rust2cpg] support ArrayExpr syntax.

xavierpinho:
Motivated by the `vec!`'s test from #6020.

Lowers `[x, y, ..., z]` and `[x; count]` as an `<operator>.arrayInitializer` of `x, y, ...,z` expressions. In particular, `count` gets dropped.

### DIFF
--- a/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
+++ b/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
@@ -74,7 +74,7 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
 
   @tailrec
   private def visitExpr(expr: Expr): Ast = expr match {
-    case x: ArrayExpr                   => notHandledYet(x)
+    case arrayExpr: ArrayExpr           => visitArrayExpr(arrayExpr)
     case x: AsmExpr                     => notHandledYet(x)
     case awaitExpr: AwaitExpr           => visitAwaitExpr(awaitExpr)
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
```diff
@@ -628,6 +628,23 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
...
+  // ArrayExpr =
+  //  Attr* '[' (
+  //    (Expr (',' Expr)* ','?)?
+  //  | Expr ';' Expr
+  //  )']'
+  private def visitArrayExpr(arrayExpr: ArrayExpr): Ast = {
+    val typeFullName = typeFullNameForExpr(arrayExpr)
+    val callNode     = operatorCallNode(arrayExpr, code(arrayExpr), Operators.arrayInitializer, Some(typeFullName))
+
+    // In `[value; count]`, we drop `count`.
```
maltek: not a must, but we'd lose less information by lowering this as a separate operator (say `<operator>.repeatInArray`)

maltek (approved):

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
xavierpinho: I didn't want to introduce a new `<operator>` if it didn't make sense to do so. But since you proposed it, I've updated accordingly. Thanks!

PR #6027, opened 2026-06-15
Experimental and partial Bazel build system support

ml86:
This PR brings in experimental, partial support for the Bazel build system.
It is setup purely as a secondary build option for development purposes and thus does
not touch the current CI setup or release process.
The idea is to validate if this improves the development process enough to justify
a later phase out of SBT.


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/astgen/AstGenRunner.scala
```diff
@@ -191,7 +219,7 @@ abstract class AstGenRunner(metaData: AstGenProgramMetaData, config: X2CpgConfig
           logger.error(s"""Local ${metaData.name} binary not found at '$localPath' or is not executable!
                |Please make sure to have a compatible ${metaData.name} version installed and available on this system$envHint.
                |""".stripMargin)
-          scala.sys.exit(1)
```
max-leuthaeuser, line 194: We did `exit(1)` on purpose. We had customer issues where the astgen location could not be found and the exception was swallowed by the frontend's ast creator pass (or only logged as debug) resulting in an empty CPG which then passed the CS pipeline just fine but with 0 findings ofc.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/JoernRunfilesLocator.scala
```diff
@@ -0,0 +1,11 @@
+package io.joern.x2cpg.utils
+
+import scala.jdk.OptionConverters.RichOptional
+
+// Use this class inside Joern repository and not in repositories
+// which bring in the Joern repository as dependency.
+object JoernRunfilesLocator {
```
maltek: can we mark it as `private[io.joern]`?

maltek (commented):
How do I run CLI tools, e.g. `jssrc2cpg` outside the IDE? I see a few `jssrc2cpg.sh` scripts (some in bazel directories). But not one that works. (For testing as a daily driver, I'd actually want to have scripts at the old `sbt` location so that existing tests and scripts can find and invoke them. But I guess it's not hard to script some symlinks.)

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/astgen/AstGenRunner.scala
ml86: I will take that out and we can discuss that separately. This anyway does not belong into this PR. Was a leftover from my development of this PR.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/JoernRunfilesLocator.scala
ml86: Good point, this will catch most unintended uses.

ml86:
@maltek Regarding the frontend execution:
I would like to avoid a separate set of starter scripts. Of course we could also adjust them but this might lead to confusion if we have some detection logic in them if people used the same directory for both the bazel and sbt build. 
Thus I add aliases to easily run the frontends via bazel. E.g. `bazel run javasrc2cpg -- /tmp/someJavaDir -o /tmp/my.cpg`.

maltek:
fwiw - a Metals import does not work because the `.bazelbsp/aspects/rules/scala/scala_info.bzl` file it generates is missing an import of `rules_java`. That seems to be an incompatibility with bazel 9 - it worked after the AI downgraded the build to bazel 8.5.1. (This is similar to https://github.com/scalameta/metals/issues/8268 except this is about bazel 9, and we already have the import which is supposed to fix that problem in MODULE.bazel.)

I don't think any of us relies on VS code as a daily driver, so likely not a blocker. Not great though for the occasional open source contributors. I mostly tried because I wanted to see whether the metals MCP server was any less buggy with `bazel` than with `sbt`. (My first impression is that it's just as buggy, but in different ways.)

ml86:
Thats good to know. Ultimately a broken Metals support would not be a blocker. But for now we anyway have both worlds side by side to compare.

maltek (approved):

PR #6030, opened 2026-06-17
[swiftsrc2cpg] Follow-up fix for desugaring for optional binding syntax

max-leuthaeuser:
This resolves the issues with dependent optional bindings.
sptests is happy.


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -256,12 +253,23 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
     val isUnresolved   = variableOption.isEmpty
 
     identifierName match {
+      case _ if optionalBindingNames.contains(identifierName) =>
```
maltek: could you add a comment here explaining what code this is dealing with, and how we are dealing with it?

I would also move it closer to the regular variable case, so that it's easier to compare where the handling of the two cases differs.

maltek (approved):

PR #6031, opened 2026-06-17
[rust2cpg] support `impl X` blocks.

xavierpinho:
Only `impl X` methods are handled here, and they become child methods to the `X` TYPE_DECL. I intend to remove the ugly `storeInDiffGraph` inside RustVisitor as I tackle the `impl X for Y` variant.

### DIFF
--- a/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
+++ b/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
@@ -470,13 +470,15 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
         s"&$mut$enclosingType"
       case None => enclosingType
     }
+    val evaluationStrategy =
+      if (selfParam.ampToken.isDefined) EvaluationStrategies.BY_SHARING else EvaluationStrategies.BY_VALUE
     val paramNode = parameterInNode(
       node = selfParam,
       name = code(selfParam.name),
       code = code(selfParam),
       index = 0,
       isVariadic = false,
-      evaluationStrategy = EvaluationStrategies.BY_SHARING,
+      evaluationStrategy = evaluationStrategy,
       typeFullName = typeFullName
     )
     Ast(paramNode)
--- a/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
+++ b/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
@@ -51,7 +51,7 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
     case x: ExternBlock         => notHandledYet(x) :: Nil
     case x: ExternCrate         => notHandledYet(x) :: Nil
     case fn: Fn                 => visitFn(fn) :: Nil
-    case x: Impl                => notHandledYet(x) :: Nil
+    case impl: Impl             => visitImpl(impl)
     case macroCall: MacroCall   => visitMacroCall(macroCall)
     case macroRules: MacroRules => visitMacroRules(macroRules)
     case macroDef: MacroDef     => visitMacroDef(macroDef)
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
```diff
@@ -288,6 +288,24 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
     returnAst(ret, Seq(exprAst))
   }
 
+  // Impl =
+  //  Attr* Visibility?
+  //  'default'? 'unsafe'?
+  //  'impl' GenericParamList? ('const'? '!'? Type 'for')? Type WhereClause?
+  //  AssocItemList
+  // TODO: support `impl X for Y` and remove the side-effects (storeInDiffGraph).
+  private def visitImpl(impl: Impl): Seq[Ast] = {
```
max-leuthaeuser, line 350: Does rust allow to specify multiple `impl X` decls for the same type?
If so, this implementation here would create multiple TypeDecls with the same fullName.

xavierpinho: It does -- and I don't know why I can no longer paste a proper link, but please see the `"multiple impl blocks for the same type"` in the current diff.

But that correctly reminds me: the TypeDecl creation actually happens outside (see the last paragraph of https://github.com/joernio/joern/pull/6028). Well, not for `structs` declared in the source-code, but for external ones. Anyway, for the `impl X for Y` I was thinking of having an extra pass à la swifsrc2cpg's ExtensionPass so that these kind of changes only happen in one place, rather than potentially scattered inside RustVisitor.

max-leuthaeuser: Oh, just one more thing came into my mind: is it possible to specify multiple `impl X` decls for the same type but spread over different files? Does that lead to multiple TypeDecls with the same fullName?

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
```diff
@@ -383,6 +402,31 @@ trait RustVisitor(implicit withValidationMode: ValidationMode) { this: AstCreato
...
+        s"&$mut$enclosingType"
+      case None => enclosingType
+    }
+    val paramNode = parameterInNode(
+      node = selfParam,
+      name = code(selfParam.name),
+      code = code(selfParam),
+      index = 0,
+      isVariadic = false,
+      evaluationStrategy = EvaluationStrategies.BY_SHARING,
```
ml86: Rust has BY_VALUE parameter passing.

joern-cli/frontends/rust2cpg/src/test/scala/io/joern/rust2cpg/passes/ast/ImplTests.scala
```diff
@@ -0,0 +1,176 @@
...
+  "an inherent method in an impl block" should {
+    val cpg = code("""
+        |struct Foo;
+        |impl Foo {
+        |  fn bar(&self) {}
+        |}
+        |""".stripMargin)
+
+    "have correct fullName" in {
+      cpg.method.nameExact("bar").fullName.l shouldBe List("rust2cpgtest::Foo::bar")
```
ml86, line 22: `rust2cpgtest` is the module name right? If yes please it something like `testModule` or a like. The import part that should come across is that it is not the file name.
This this will affect lots of tests you should likely do that in a separate PR.

ml86 (changes requested):

joern-cli/frontends/rust2cpg/src/test/scala/io/joern/rust2cpg/passes/ast/ImplTests.scala
xavierpinho: The test fixture creates at least 2 files: 

* `Cargo.toml`, which is where `rust2cpgtest` comes from.
* `src/lib.rs` (by default, since `code(..)` accepts a file name, and there's the usual `moreCode` to add more files) where the snippet is written into.

So, `rust2cpgtest` is not coming from the filename. Have I missed your point?

ml86: I was aware of that. I was just suggesting that if you name the module `testModule` it is immediately obvious that this is not a file.

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
xavierpinho: It's also fine (no duplicates), but it's good to have a test for it.

ml86 (approved):

PR #6035, opened 2026-06-19
[swiftsrc2cpg] Implement de-sugaring for var/let tuple pattern assignments

max-leuthaeuser:
- `var/let (a, b) = source` is de-sugared into a temporary variable holding the RHS (evaluated once) followed by individual field-access assignments:

```
var/let (a, b) = source =>
  <tmp>0 = source
  a = <tmp>0.0
  b = <tmp>0.1
```

- Wildcard elements are dropped. No local and no assignment is emitted for them.

- Nested tuple patterns are handled recursively, e.g.:

```
let (a, (b, c)) = source =>  
  <tmp>0 = source
  a = <tmp>0.0
  b = <tmp>0.1.0
  c = <tmp>0.1.1
```

sptests is happy (builders/82/builds/693).

### DIFF
--- a/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForDeclSyntaxCreator.scala
+++ b/joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForDeclSyntaxCreator.scala
@@ -1249,89 +1249,153 @@ trait AstForDeclSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
     else { VariableScopeManager.ScopeType.MethodScope }
 
     val bindingAsts = variableDecl.bindings.children.flatMap { binding =>
-      val namesWithNode = binding.pattern match {
-        case expr: ExpressionPatternSyntax =>
-          notHandledYet(expr)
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstForDeclSyntaxCreator.scala
```diff
@@ -1249,89 +1249,155 @@ trait AstForDeclSyntaxCreator(implicit withSchemaValidation: ValidationMode) {
...
-        case w: WildcardPatternSyntax =>
-          Seq((scopeLocalUniqueName("wildcard"), w))
+      val maybeTuplePattern: Option[TuplePatternSyntax] = binding.pattern match {
+        case tp: TuplePatternSyntax => Some(tp)
+        case vb: ValueBindingPatternSyntax =>
+          vb.pattern match {
+            case tp: TuplePatternSyntax => Some(tp)
+            case _                      => None
+          }
+        case _ => None
```
maltek, line 1262: handle these unhandled cases in some way, or add comments explaining why doing nothing is the right thing

maltek (approved):

PR #6037, opened 2026-06-22
[rubysrc2cpg] Correctly handle access modifier calls

TNSelahle:
This PR addresses a few things in regard to access modifiers. 

### Access modifiers modelled as operator calls
Each access modifier has its own operator call name.
This also allows us to remove the special implicit return handling.

### Incorrect access modifier arguments
1. Normal access modifiers' arguments were previously not captured in the CPG.
2. Method access modifiers' arguments, if any, were limited to the first argument, even when the argument was a list. 

### Access modifiers not placed in class `<body>` method
Access modifiers were not placed as children of the `<body>` method as they were seen as declarative statements and not calls. Since it's been established that these are calls, they have now been moved to be in the class `<body>` method.
However, doing this means the ordering of statements is lost. Ordering is important for access modifier calls that aren't targeted. Example:

```rb
class Test
  def foo; end
  private
  def bar; end
end
```

In the above example, `foo` is public and `bar` is private. Method declarations are not placed in the `<body>` method. Since my change move the access modifier inside the `<body>` method, the ordering is lost. 
This is remedied by using the statement line numbers to re-order the statements before traversing them.

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
@@ -338,7 +338,7 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this: 
           m.body.asInstanceOf[StatementList].statements
         case other => other :: Nil
       }
-      .sortBy(_.line.getOrElse(0))
+      .sortBy(el => (el.line.getOrElse(0), el.column.getOrElse(0)))
 
     var currentModifier = ModifierTypes.PUBLIC
     val visibilityMap   = mutable.LinkedHashMap.empty[String, String]
--- a/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
@@ -349,9 +349,9 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this: 
       case node: AccessModifier =>
         val modifier = accessModifierType(node)
         node.arguments.foreach {
-          case lit: StaticLiteral => visibilityMap.put(lit.span.text.stripPrefix(":"), modifier)
+          case lit: StaticLiteral => visibilityMap.put(symbolName(lit), modifier)
           case arr: ArrayLiteral =>
-            arr.elements.foreach(el => visibilityMap.put(el.span.text.stripPrefix(":"), modifier))
+            arr.elements.foreach(el => visibilityMap.put(symbolName(el), modifier))
           case _ =>
         }
       case node: MethodAccessModifier =>
@@ -361,9 +361,9 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this: 
         }
         node.arguments.foreach {
           case proc: ProcedureDeclaration => visibilityMap.put(proc.methodName, modifier)
-          case lit: StaticLiteral         => visibilityMap.put(lit.span.text.stripPrefix(":"), modifier)
+          case lit: StaticLiteral         => visibilityMap.put(symbolName(lit), modifier)
           case arr: ArrayLiteral =>
-            arr.elements.foreach(el => visibilityMap.put(el.span.text.stripPrefix(":"), modifier))
+            arr.elements.foreach(el => visibilityMap.put(symbolName(el), modifier))
           case _ =>
         }
       case node: ProcedureDeclaration if !visibilityMap.contains(node.methodName) =>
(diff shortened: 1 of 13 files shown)


joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -325,4 +325,52 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this:
...
+      case _: ProtectedModifier => ModifierTypes.PROTECTED
+    }
+
+    val allStatements = classBody.statements
+      .flatMap {
+        case m: MethodDeclaration if m.methodName == Defines.TypeDeclBody =>
+          m.body.asInstanceOf[StatementList].statements
+        case other => other :: Nil
+      }
+      .sortBy(_.line.getOrElse(0))
```
ml86: I dont think sorting by line number is enough. Consider:
```
class User
  public
  private; def helper
    # ...
  end
end
```
In this case your processing order might be: `public`, `def helper...`, `private`.
Why do you sort by line number in the first place? The classBody statementList should anyway be sorted right?

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -325,4 +325,52 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this:
...
+    var currentModifier = ModifierTypes.PUBLIC
+    val visibilityMap   = mutable.LinkedHashMap.empty[String, String]
+
+    allStatements.foreach {
+      case node: AccessModifier if node.arguments.isEmpty =>
+        currentModifier = accessModifierType(node)
+      case node: AccessModifier =>
+        val modifier = accessModifierType(node)
+        node.arguments.foreach {
+          case lit: StaticLiteral => visibilityMap.put(lit.span.text.stripPrefix(":"), modifier)
```
ml86: `lit.span.text.stripPrefix(":")` looks brittle. Can we avoid that?

ml86 (changes requested):
What is the reasoning for changing from the push/pop mechanism to the precomputed visibility map?

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
TNSelahle: I don't think we can avoid that. 
image

If you're concerned about using content from the span, then I can look into also having the `text/code` field one-level up in `StaticLiteral`.
If you're concerned about the `stripPrefix` call, I don't think we can get around that since symbols are all prefixed with `:`. 
An alternative, more effort approach, would be to have a separate node type for `SymbolLiteral` so we know that the `stripPrefix` call is always necessary. Then strings and others remain under `StaticLiteral`.

TNSelahle:
> What is the reasoning for changing from the push/pop mechanism to the precomputed visibility map?

@ml86 the push/pop mechanism is still present. It is used for bare `private/protected/public` statements that don't target specific methods. 
The visibility map is used when there is an argument on the access modifier calls. It keeps a mapping of the method name -> visibility.

joern-cli/frontends/rubysrc2cpg/src/main/scala/io/joern/rubysrc2cpg/astcreation/AstForTypesCreator.scala
TNSelahle: image

The above shows the body statements of a class declaration. Method declarations are not placed as statements under the `<body>` method, but access modifier calls are. Ordering is lost due to this. I had placed a small explanation in the PR description as well.


Regarding your concern, yeah 100%. Thanks for flagging that. I think sorting by (line, column) should do the trick.

ml86: @TNSelahle I thought your comment in the PR description was for the resulting CPG format. I cannot imaging that the Ruby parser gives the method declarations separately. I guess that is something the Ruby AST generator does right? If yes could we change that or would that trigger lots of other changes?

TNSelahle: @ml86 It's definitely not the Ruby parser. It's the logic responsible for converting AST JSON to the intermediate nodes. My suspicion was that changing that would result in a lot of changes since current logic relies on the method declarations being statements of the class. 
I agree that the correct approach is to have them in the same statement list, as that's also a true representation of the code. My initial thought was to have this bundled with the work needed to represent method declarations as expressions with return values.

Let me look a bit into how much effort it would be to not split them.

ml86 (approved):

PR #6039, opened 2026-06-23
[pysrc2cpg] Temp variable naming improvements

SuperUserDone:
https://harness.atlassian.net/browse/QT-887

Temp variables can no longer collide with a source variable name
Temp variable names are scoped by scope, no longer per-file.

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/ContextStack.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/ContextStack.scala
@@ -169,15 +169,16 @@ class ContextStack {
     val context  = stack.head
     val nameBase = prefix.map(prefixValue => s"${prefixValue}_tmp").getOrElse("tmp")
 
-    var result: String = ""
-    while (result == "" || context.reservedNames.contains(result)) {
+    var unusedName = Option.empty[String]
+    while (unusedName.forall(context.reservedNames.contains)) {
       val counter = context.tmpCounters.getOrElse(nameBase, 0)
-      result = s"$nameBase$counter"
+      unusedName = Some(s"$nameBase$counter")
       context.tmpCounters.update(nameBase, counter + 1)
     }
 
-    context.reservedNames.add(result)
-    result
+    val name = unusedName.get
+    context.reservedNames.add(name)
+    name
   }
 
   private def findEnclosingMethodContext(contextStack: List[Context]): MethodContext = {
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/ContextStack.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/ContextStack.scala
@@ -31,6 +31,8 @@ class ContextStack {
     val astParent: nodes.NewNode
     val order: AutoIncIndex
     val variables: mutable.Map[String, nodes.NewNode]
+    val reservedNames: mutable.Set[String]
+    val tmpCounters: mutable.Map[String, Int]
     var lambdaCounter: Int
     val methodCounter: mutable.Map[String, Int]
   }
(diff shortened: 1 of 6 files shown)


joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/memop/MemoryOperationCalculator.scala
```diff
@@ -44,41 +45,74 @@ class MemoryOperationCalculator extends AstVisitor[Unit] {
...
 
   override def visit(functionDef: ast.FunctionDef): Unit = {
+    addName(functionDef.name)
     push(Load)
     accept(functionDef.decorator_list)
     accept(functionDef.args)
     accept(functionDef.returns)
     pop()
+    pushScope(functionDef)
+    functionDef.args.posonlyargs.foreach(arg => addName(arg.arg))
```
ml86: Maybe better to do the handling of args inside the argument visitor handler.

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
```diff
@@ -391,15 +397,16 @@ class PythonAstVisitor(
     isAsync: Boolean,
     methodRefNode: Option[nodes.NewMethodRef],
     returnTypeHint: Option[String],
-    lineAndColumn: LineAndColumn
+    lineAndColumn: LineAndColumn,
+    reservedNames: Iterable[String] = Nil
```
ml86: Please remove the default value here and in the other places.

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/ContextStack.scala
```diff
@@ -147,6 +167,32 @@ class ContextStack {
...
 
+  def getUnusedName(prefix: Option[String] = None): String = {
+    val context = stack.head
+    // Temporary stores in special contexts are materialized as locals in the
+    // enclosing method, so their names must be allocated from that method's
+    // shared counter and reservation set.
+    val enclosingMethodContext = context match {
+      case _: SpecialBlockContext => Some(findEnclosingMethodContext(stack))
+      case _                      => None
+    }
```
ml86: Check if we can avoid that by setting the reservedNames of the SpecialBlockContext according to the outer scope/context.

ml86 (changes requested):

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/ContextStack.scala
ml86: Please do not mark this as resolved if the construct is this in the PR and you have given no explanation as to why we need it.

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/ContextStack.scala
```diff
@@ -147,6 +165,21 @@ class ContextStack {
     result
   }
 
+  def getUnusedName(prefix: Option[String] = None): String = {
+    val context  = stack.head
+    val nameBase = prefix.map(prefixValue => s"${prefixValue}_tmp").getOrElse("tmp")
+
+    var result: String = ""
```
ml86: Please use `Option[String]` instead of empty string to indicate empty and renaming from `result` to `unusedName` would be beneficial.

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/TemporaryVariableNameTests.scala
```diff
@@ -0,0 +1,74 @@
+package io.joern.pysrc2cpg.cpg
+
+import io.joern.pysrc2cpg.testfixtures.PySrc2CpgFixture
+import io.shiftleft.semanticcpg.language.*
+
+class TemporaryVariableNameTests extends PySrc2CpgFixture {
```
ml86, line 6: I think this scenario is not tested and not handled i your code:
```
def foo():
    tmp = 1
    def bar():
        <construct that requires tmp variable for lowering>
        print(tmp)
    bar()

foo()
```

ml86 (changes requested):

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/TemporaryVariableNameTests.scala
SuperUserDone: Yes, this is not tested, but is handled. Added a test for it.

ml86 (approved):

PR #6042, opened 2026-06-23
[bazel] Add rust2cpg support.

ml86:


MODULE.bazel
```diff
@@ -179,3 +179,58 @@ http_file(
...
 )
+
+#rust2cpg astgen binary imports
+http_file(
+    name = "rust2cpg_astgen_macos_arm",
+    executable = True,
+    # TODO add sha256 for now we skip this as the frontend is in early developemnt and we do not want
+    # to update this all the time.
+    sha256 = "",
+    url = "https://github.com/joernio/astgen-monorepo/releases/download/rust-astgen%2Fv0.8.2/rust_ast_gen-macos-arm",
```
maltek: the version number has the same problem that we have to manually update it all the time, in multiple places.

Maybe we want to teach `updateDependencies.sh` to update these definitions automatically? Otherwise I think it will take all but 5 minutes until the Bazel and `sbt` builds diverge.

maltek (commented):
when merging, please fix the commit message to mention bazel

MODULE.bazel
max-leuthaeuser: Does bazel allow for templated build files or global definitions? With that, we could have all versions in one place.

ml86: The one place to manage the versions is MODULE.bazel. I introduced a variable `RUST2CPG_ASTGEN_VERSION` in order to avoid duplication of the version. Will introduce the same for the other astgen download URLs in a separate PR.

xavierpinho (approved):
Just to confirm: safe for me to bump rust_ast_gen's version also in MODULE.bazel going forward? (Today's version is 0.8.6)

Another thing: we do download rust_ast_gen's binary currently in sbt, but there's also an accompanying RustNodeSyntax.scala (e.g. https://github.com/joernio/astgen-monorepo/releases/tag/rust-astgen%2Fv0.8.6), for which there's no automated process (and I just update it manually.) It's not a problem, but if it's something we could also automate with Bazel, the better.

maltek (approved):

ml86:
@xavierpinho Yes it would be much appreciated if you keep the version number in MODUL.bazel in sync. I was not aware of this by hand generation. We could easily automate that, it is just that we then should also automate it in the SBT build to avoid too much divergence.

PR #6048, opened 2026-06-24
[CI] Pin macos-15 runners

max-leuthaeuser:
See: https://github.com/actions/runner-images/issues/14167

With Xcode 26.4.1 being the new default some swiftsrc2cpg test fail. I need time to figure that out and we need to unblock PRs until then.


ml86 (approved):

max-leuthaeuser:
Replaced with https://github.com/joernio/joern/pull/6049

PR #6051, opened 2026-06-24
[x2cpg, kotlin2cpg] Add dependency resolver v2 to resolve source dependencies for source discovery

johannescoetzee:
This PR introduces a new version of the Gradle dependency fetcher which uses the Gradle init script to create a set of `depInfo.json` files (one per subproject in the build) containing the following information:
- **projectName:** the name of the project
- **sourceRoots:** all of the source roots for the given project
- **sourceDependencies:** a non-transitive list of sibling projects this project depends on
- **artifactDependencies:** a non-transitive list of jars/aars this project depends on

This information is collected into the newly added `DependencyGraph`, which can be traversed to get detailed information about the dependency tree (although for now it's only used to get a flattened list of source dependencies and artifact dependencies.

The main benefit of this PR is introducing the collection of source dependency information. On current master, scanning a single subproject in a large project, e.g. a large monorepo, results would be very bad as sources for sibling projects which the subproject depends on would be excluded. With this new approach, we can use the information from the dependency graph to figure out where the source dependencies are located and include those in the CPG as well.

This is currently only supported in kotlin2cpg and is gated behind the `-enable-dependency-resolver-v2` CLI option or the `ENABLE_DEPENDENCY_RESOLVER_V2` environment variable (which may be set to any non-empty value). If this is not set, the original GradleDependencies and init script are used, so this should be a low-risk change.


johannescoetzee:
There are still tests failing on Windows, most likely due to some unresolved path issue, but the PR runner test loop is quite slow. I'll set up a windows VM again on Monday to debug this properly

max-leuthaeuser:
> There are still tests failing on Windows, most likely due to some unresolved path issue, but the PR runner test loop is quite slow. I'll set up a windows VM again on Monday to debug this properly

You could try removing everything for the moment that is not required from the PR workflow and invoke only the exact test you are interested in (sbt "testOnly ...")

johannescoetzee:
I've fixed the windows issue. It was a combination of broken path handling in the gradle script and test cases

max-leuthaeuser:
This is extremly hard to review. Especially things like `joern-cli/frontends/x2cpg/src/main/resources/io/joern/x2cpg/utils/dependency/dependency-fetcher-init-v2.gradle` and `joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/AarExtractor.scala`

Anyone with more experience in that technical space should step in.

As a general remark: is x2cpg the right place for all of these dependency handling utils?
kotlin2cpg is the only consumer right?

johannescoetzee:
> This is extremly hard to review. Especially things like joern-cli/frontends/x2cpg/src/main/resources/io/joern/x2cpg/utils/dependency/dependency-fetcher-init-v2.gradle and joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/AarExtractor.scala
>
> Anyone with more experience in that technical space should step in.

This is true, unfortunately. The original implementation was done by Claudiu and since then I've been the only person to significantly update this (with you having perhaps been the only other person to work on it at all, but I haven't checked the git log). It's also just a lot of code because there seem to be a lot of different cases to handle and defend against. This, combined with the fact that it relies quite heavily on the gradle API, means that it is just difficult to get a clear picture of exactly what's happening. I've done a fair amount of testing on my end and am not aware of any cases or configurations for which this breaks, but since this is a big change, we decided it would be safer to split it into the separate V2 implementation with a fallback available if it fails. For now, we only expect one customer to use it too, so the risk should be fairly low.

> As a general remark: is x2cpg the right place for all of these dependency handling utils?
kotlin2cpg is the only consumer right?

javasrc2cpg uses the same dependency fetcher. This PR doesn't include javasrc2cpg support for V2 yet, as this would be a bunch more code in an already big PR (and the customer asking for this was using Kotlin), but the idea is to follow up with that once this is in

max-leuthaeuser:
This is what Claude gives me. Take it with a grain of salt. Nevertheless, could you check?

1) `joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala`

v2 fallback silently skips dependency download

`gatherJarsOfDependencies` (added else if (config.enableDependencyResolverV2) → Seq() branch) is also called from the case None fallback arm of `createCpg` — i.e., when v2 is enabled but fails. In that situation `config.enableDependencyResolverV2` is true, so the short-circuit fires and `downloadDependencies` is never called, even though `withEnableDependencyResolverV2` explicitly sets `downloadDependencies = true` to avoid requiring a second flag.

Concrete trigger: `--enable-dependency-resolver-v2` set, no jar4import service, Gradle resolution fails (missing wrapper, network error, unsupported structure). Result: zero dependency JARs, worse analysis quality than plain `--download-dependencies` would have produced — with only a debug-level log, invisible at default verbosity.

Fix: in the case None arm, pass `config.copy(enableDependencyResolverV2 = false)` to `gatherJarsOfDependencies`, so the v2 guard doesn't fire on the v2-failed-fallback path.

---

2) `joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/dependency/DependencyResolverV2.scala`

Dead `flagFromConfig` parameter and dead `isContainedIn` method

`resolve(inputPath, flagFromConfig, params)` — `flagFromConfig` is accepted but never read anywhere in the method body; `getDependencyGraph` also does not receive it. The sole call site in `Kotlin2Cpg.scala` hardcodes `true`. Additionally, the `private def isContainedIn` helper is declared but never called anywhere (no internal call site, and private prevents external use).

Both should be removed before the API surface stabilizes.

---

3) `joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/Kotlin2Cpg.scala`

Stale // TODO Aar extraction comment

The comment implies AAR extraction is not yet implemented, but the PR itself adds `AarExtractor.materializeJars` and it is already called inside `DependencyGraph.transitiveDependenciesForProjectsInDir` — so by the time `dependencyJars` reaches the `case Some(...)` branch in `createCpg`, every `.aar` has already been materialized into a `.jar`. The TODO is a false signal.

johannescoetzee:
> This is what Claude gives me. Take it with a grain of salt. Nevertheless, could you check?

These were legitimate issues, so I've pushed a commit to fix them. I also had claude do a big multi-agent review of the PR and it identified a few more issues, so following commits are all there to address the ones that made sense to me

[The full list of comments](https://github.com/user-attachments/files/29506279/feedback.txt)

max-leuthaeuser (approved):

PR #6055, opened 2026-06-29
[x2cpg] Move LanguageFrontend classes from CodeScience

SuperUserDone:
In relation to the support code for https://harness.atlassian.net/browse/QT-858

Will need to be released then I'll open the pr on the CS side.

### DIFF
--- a/linter-rules/src/main/scala/fix/RestrictedImports.scala
+++ b/linter-rules/src/main/scala/fix/RestrictedImports.scala
@@ -8,7 +8,10 @@ object RestrictedImports {
 
   /** If a file is within any of the packages defined below, it will not be checked for restricted imports.
     */
-  private val ignoredPackages: Seq[String] = Seq()
+  private val ignoredPackages: Seq[String] = Seq(
+    // We want long-lived processes for the server classes, the ExternalCommand Import from shiftleft does not allow us to do this.
+    "io.joern.x2cpg.utils.server"
+  )
 
   /** The rules for restricted imports.
     */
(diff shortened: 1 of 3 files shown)


linter-rules/src/main/scala/fix/RestrictedImports.scala
```diff
@@ -8,7 +8,10 @@ object RestrictedImports {
 
   /** If a file is within any of the packages defined below, it will not be checked for restricted imports.
     */
-  private val ignoredPackages: Seq[String] = Seq()
+  private val ignoredPackages: Seq[String] = Seq(
+    // We want long-lived processes for the server classes, the ExternalCommand Import from shiftleft does not allow us to do this.
+    "io.joern.x2cpg.utils.server"
```
ml86, line 13: Please elaborate on this. Why does it not allow for long-lived processes and in which way is this related to the linter rules?

SuperUserDone: `ExternalCommand` only supports one-shot commands, ie start process, wait for it to exit, collect stdout/stderr, return an `ExternalCommandResult`.

Because we needs to:
- start the frontend server & keep the process alive across requests
- watch stdout until the server prints the port number
- destroy it later (via `shutdown()`)

We cannot use the `ExternalCommand` from shiftleft for that (without substantially extending/creating a different `LongLivedExternalCommand` implementation for it)

The original implementation from CodeScience used the scala `Process(...)` api, but joern has a linter rule that disallows the scala `Process(...)` api import in favour of shiftleft `ExternalCommand`.

ml86: Ok, makes sense.

ml86 (approved):

PR #6060, opened 2026-06-29
[rust2cpg] lower `?` expressions.

xavierpinho:
Lower `x?` as `<operator>.tryUnwrap` of `x`.


max-leuthaeuser (approved):

maltek (approved):
At some point we'll want the conditional early return semantics of the operator to be reflected in the CPG - so we'll need some more complex lowering than a simple call. But it's better than a `Unknown` block, so I'm all for merging this now, and improving it later.

PR #6068, opened 2026-07-02
[x2cpg] Refactor Language Frontend Adapter utilities

SuperUserDone:
This PR, together with the corresponding [codescience-side PR](https://github.com/ShiftLeftSecurity/codescience/pull/8988), refactors the `LanguageFrontend` execution model.

On the codescience side, all previous usages of `DirectInvocationLanguageFrontend` except `CSharp` (`go2cpg`, `java`, `python`) already needed bespoke execution logic. Based on that, this PR removes the generic direct-invocation abstraction and splits the remaining responsibilities.

Changes:
- Remove `DirectInvocationLanguageFrontend` completely. Frontends that need direct execution should provide their own `FrontendExecutor` implementation and implement [`execute`](https://github.com/joernio/joern/blob/8336f2be7f5bd1839ef4cca42b19dc561ccc6c5d/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/server/FrontendExecutor.scala#L5).
- Introduce `FrontendExecutor` as the common interface for “given an input, produce a CPG output file”.
- Add [`HttpServerExecutor`](https://github.com/joernio/joern/blob/louis/refactorLanguageFrontends/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/server/HttpServerExecutor.scala#L9), which owns the server process lifecycle and executes requests against it for the frontends using FrontendHTTPServer.
- Add [`ExecutableLocator`](https://github.com/joernio/joern/blob/8336f2be7f5bd1839ef4cca42b19dc561ccc6c5d/joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/server/ExecutableLocator.scala#L5-L22), for executeable resolution
- Keep the codescience `LanguageFrontend` as the higher-level wrapper for `policyPlatform`, validation level, file suffix, and the chosen `FrontendExecutor`.

See the codescience-side PR for concrete usage examples.


joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/server/HttpServerExecutor.scala
```diff
@@ -0,0 +1,87 @@
+package io.joern.x2cpg.utils.server
+
+import java.io.File
+import java.util.concurrent.{CompletableFuture, TimeUnit, TimeoutException}
+import scala.sys.process.{Process, ProcessLogger}
+import scala.util.{Failure, Success}
+import org.slf4j.LoggerFactory
+
+class HttpServerExecutor(
```
ml86: I think this class should be called `HttpClientExecutor` after all it is running `FrontendHttpClient` which then connects to the server.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/server/ExecutableLocator.scala
```diff
@@ -0,0 +1,22 @@
+package io.joern.x2cpg.utils.server
+
+import java.nio.file.{Files, Path}
+
+class ExecutableLocator(
```
ml86: I think this should be an interface and the implementation should go into CS repo.
I think this would make the use case in go2cpg easier were this concrete configuration option set is not needed.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/server/FrontendExecutor.scala
```diff
@@ -0,0 +1,19 @@
+package io.joern.x2cpg.utils.server
+
+import java.io.File
+
+trait FrontendExecutor {
+  def execute(input: File, extraArgs: String*): File
```
ml86: The file related APIs are a mixture of old `File` and new `Path`. Please change everything to `Path` if there are no specific reasons against it.

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/server/HttpServerExecutor.scala
```diff
@@ -0,0 +1,87 @@
...
+        (argument.stripPrefix("--") -> Some(value)) +: parseArguments(rest)
+      case argument :: rest =>
+        (argument.stripPrefix("--") -> None) +: parseArguments(rest)
+    }
+  }
+
+  override def execute(sourceCodeDir: File, extraArgs: String*): File = {
+    start()
+
+    val cpgFile = FrontendExecutor.newTemporaryCpgOutputFile("cpg", cpgFormat.fileSuffix)
```
ml86: I suggest we let the user of `execute` pass in a cpg output file. That way the user can handle the file lifetime and the suffix. Currently the file lifetime is a mixed responsibility where `newTemporaryCpgOutputFile` registers the file for deletion at JVM shutdown, but `LanguageFrontend.createCpg` deletes it after load.

ml86 (changes requested):

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/utils/server/HttpServerExecutor.scala
ml86: Agreed on Zoom to call it `HttpExecutor`.

ml86 (approved):

PR #6071, opened 2026-07-03
[post frontend validator] add check for dangling AST nodes

maltek:


semanticcpg/src/main/scala/io/shiftleft/semanticcpg/validation/validation.scala
max-leuthaeuser: scaladoc still references V2 as highest level.
Tbh.: remove that useless info there - it will be out of date every single time.

maltek:
the frontends not enabled here, are either already behind on the validation version, or fail the new check (c, swift)

max-leuthaeuser (approved):

max-leuthaeuser:
I'll try to fix c2cpg and swiftsrc2cpg next week. Thanks for checking.

PR #6073, opened 2026-07-03
[php2cpg] Added support for named call arguments

max-leuthaeuser:
Ref: QT-954

### DIFF
--- a/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
+++ b/joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
@@ -371,8 +371,9 @@ class CallTests extends PhpCode2CpgFixture {
     val cpg = code("""<?php
         |foo(bar: $x);
         |""".stripMargin)
-
-    inside(cpg.call.nameExact("foo").argument.l) { case List(xArg: Identifier) =>
+    val call = cpg.call.nameExact("foo").loneElement
+    call.code shouldBe "foo(bar: $x)"
+    inside(call.argument.l) { case List(xArg: Identifier) =>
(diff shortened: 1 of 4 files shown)


maltek (commented):
I think this should set the argument name not the label? the name takes precedence over the index just like in all our other supported languages except Swift: https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments

ml86:
Yes, please use ARGUMENT_NAME instead of ARGUMENT_LABLE and add dataflow tests in CS repo for this.

joern-cli/frontends/php2cpg/src/test/scala/io/joern/php2cpg/querying/CallTests.scala
```diff
@@ -367,6 +367,38 @@ class CallTests extends PhpCode2CpgFixture {
...
+    inside(cpg.call.nameExact("foo").argument.l) { case List(xArg: Identifier) =>
+      xArg.name shouldBe "x"
+      xArg.code shouldBe "$x"
+      xArg.argumentName shouldBe Some("bar")
+    }
+  }
+
+  "a call with mixed positional and named arguments should only set argumentName on named ones" in {
+    val cpg = code("""<?php
+        |foo($a, name: $b, $c);
```
ml86: I dont think this is valid PHP code. After the first named argument, positional arguments are not allowed. Please check if this statement holds true for all PHP versions and remove the test if that is the case.

max-leuthaeuser: It's never possible to do that. I adjusted the test (for simply having multiple named arguments - which we should test).

ml86 (approved):

PR #6075, opened 2026-07-03
[rust2cpg] lower `impl X for Y`.

xavierpinho:


joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/AstCreator.scala
```diff
@@ -161,6 +161,20 @@ class AstCreator(val config: Config, val parseResult: ParseResult)(implicit with
     )
   }
 
+  protected def typeDeclForTraitImpl(impl: RustNodeSyntax.Impl): NewTypeDecl = {
+    val implTypeFullName  = typeFullNameForType(impl.typ.last)
+    val traitTypeFullName = typeFullNameForType(impl.typ.head)
```
max-leuthaeuser: .last and .head are unsafe calls.
Maybe wrap this safely and emit an UnknownNode + proper logging?

xavierpinho: You are right, but such situation really shouldn't happen - the syntax node from rust_ast_gen just doesn't distinguish/label each of these `Typ` sub-nodes differently, so they are bundled together in a Seq[Typ], instead of having an accessor for `X` and `Y` specifically. I'm not against adding the checks, but I fear they won't be doing anything. If you still prefer the check after this context, I'm happy to add it though.

max-leuthaeuser: That's right only if rust_ast_gen never changes in this regard. Better safe than sorry.

max-leuthaeuser (approved):

PR #6082, opened 2026-07-07
markus/bazelDecoupleBazelTooling

ml86:
- **[bazel] Decouple Joern and bazel_tooling module versions.**
- **[bazel] Add common runtime deps to further rule macros.**


maltek (commented):
Referencing files from the same repo, but in an old revision is too much for me. I don't like this at all.

What problem does this solve, and is there any other way to solve it?

ml86:
Fair point. I will move it into flatgraph which is at the root of our repo dependency tree. That way we avoid this.

PR #6085, opened 2026-07-07
[c2cpg][jssrc2cpg][swiftsrc2cpg] More validation fixes

max-leuthaeuser:
For the remaining sptest V3 violations.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/ControlStructureTests.scala
+++ b/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/ControlStructureTests.scala
@@ -425,4 +425,26 @@ class ControlStructureTests extends C2CpgSuite(FileDefaults.CppExt) {
     }
   }
 
+  "while loop with const assigment" should {
+    val cpg = code(
+      """
+        |void main() {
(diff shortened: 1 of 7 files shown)


joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/ControlStructureTests.scala
```diff
@@ -425,4 +425,26 @@ class ControlStructureTests extends C2CpgSuite(FileDefaults.CppExt) {
     }
   }
 
+  "while loop with const assigment" should {
+    val cpg = code(
+      """
+        |void main() {
+        |  while (const char* c = read()) {
```
maltek, line 432: how is this lowered exactly? this defines a local in the scope of the loop, and then assigns to it. Where does the local end up in the AST?

maltek (approved):

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/ControlStructureTests.scala
max-leuthaeuser: Added a test for the local position as discussed.

PR #6086, opened 2026-07-08
[rust2cpg] refactor `visitImpl`.

xavierpinho:
The main goal being to remove `storeInDiffGraph` calls during lowering.

### DIFF
--- a/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
+++ b/joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
@@ -355,23 +355,33 @@ trait RustVisitor(implicit withSchemaValidation: ValidationMode) { this: AstCrea
   //  AssocItemList
   private def visitImpl(impl: Impl): Seq[Ast] = {
     if (impl.forKwToken.isDefined) {
-      impl.typ match {
-        case implTrait :: implType :: Nil =>
-          val typeDecl = typeDeclForTraitImpl(impl, implTrait, implType)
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
```diff
@@ -355,25 +355,35 @@ trait RustVisitor(implicit withSchemaValidation: ValidationMode) { this: AstCrea
...
+        val typeDecl = typeDeclForTraitImpl(impl, implTrait, implType)
+        methodAstParentStack.push(typeDecl)
+        val methodAsts = impl.assocItemList.assocItem.collect { case fn: Fn => visitFn(fn) }
+        methodAstParentStack.pop()
+        Ast(typeDecl).withChildren(methodAsts)
+      case _ => notHandledYet(impl)
+    }
+  }
+
+  // There can be multiple `impl X` in the same file/project, so we lower their
```
max-leuthaeuser, line 378: And their fullnames are still unique?

max-leuthaeuser (approved):

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
xavierpinho: Yes, via `typeDeclForImpl`, which either takes rust_ast_gen's resolved name or builds one from the type name  + current crate as a fallback. (But didn't change any logic in this PR, it's really just a refactoring.)

PR #6088, opened 2026-07-08
[jssrc2cpg] JVM/Scala performance & memory fixes

max-leuthaeuser:
Targeted, behavior-preserving performance work, from an Opus 4.8 review of `jssrc2cpg/src/main`. Fixes hot-path allocation, redundant regex compilation, and a couple of O(n²)/O(n) lookups on the per-node and per-file paths. No functional changes.

Findings from the LLM were manually reviewed.

### High impact
- **Memoize array-replacement regex in `typeFor`** — hoist `Pattern.compile` out
  of the per-type hot path.
- **Key function-name memoization by source range** — `functionNodeToNameAndFullName`
  was keyed by `BabelNodeInfo`, whose `ujson.Value` hashes structurally over the
  whole function subtree; now keyed by `start:end`.
- **Compute node offsets once in `createBabelNodeInfo`** — resolve `start`/`end`
  and the line binary-search a single time per node instead of 3–4×.

### Medium
- **`JsTypes` List → Set** (`x2cpg`) — O(1) `isBuiltinType`.
- **Index `seenAliasTypes` by name** — `HashSet` linear scan → `HashMap`, removing
  the O(n²) over alias-heavy files.
- **Decompose class members once per class** — one walk feeds the constructor +
  both decoration helpers instead of ~4 re-walks.
- **Read each candidate file at most once during filtering** (`AstGenRunner`) —
  previously read twice (minified + transpiled checks).
- **Strip file extension without recompiling a pattern per parsed file**
  (`BabelJsonParser`/`AstCreationPass`).
- **Reuse already-built callee in `handleCallNodeArgs`** instead of rebuilding it.
- **Single-pass EJS scan** — replace the O(n·m) per-char `positions.exists` +
  `zipWithIndex` with an advancing pointer; drop a throwaway `\S`→space regex.
- **Build block-statement list once** in `createBlockStatementAsts`.

### Cleanups
- Avoid redundant ujson `ArrayBuffer`→`List` copies across AST creation
  (`astForNodes` takes `collection.Seq`, single template-literal `partition`/zip,
  JSX `charAt`, `addAll` via iterator).
- Reuse the import `Matcher` and short-circuit `TypeReplacements` in `typeFor`.
- Drop duplicate specifier lookups in `astForImportDeclaration`.
- Accumulate dependencies in place (`PackageJsonParser`) and filter `.key` files
  before copying (`PrivateKeyFilePass`).
  
### DIFF
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstCreatorHelper.scala
@@ -18,9 +18,8 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
   protected def nodeTypeOf(json: Value): BabelNode = fromString(json("type").str)
 
   protected def createBabelNodeInfo(json: Value): BabelNodeInfo = {
-    // Resolve start/end offsets and their line numbers once, then derive columns and code from them. Going through the
-    // individual code/line/column/lineEnd/columnEnd accessors would recompute start/end and the line binary search
-    // several times per node, and createBabelNodeInfo runs for every AST node.
+    // Resolve start/end offsets and their line numbers once, then derive columns and code from them. This runs for
+    // every AST node, so we avoid recomputing the offsets and the line binary search per accessor.
     val startOffset   = start(json)
     val endOffset     = end(json)
     val lineOfStart   = startOffset.map(getLineOfSource)
--- a/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstCreatorHelper.scala
@@ -1,12 +1,12 @@
 package io.joern.jssrc2cpg.astcreation
 
 import io.joern.jssrc2cpg.parser.BabelAst.*
-import io.joern.jssrc2cpg.parser.{BabelAst, BabelNodeInfo}
-import io.joern.x2cpg.{Ast, ValidationMode}
+import io.joern.jssrc2cpg.parser.BabelNodeInfo
 import io.joern.x2cpg.frontendspecific.jssrc2cpg.Defines
 import io.joern.x2cpg.utils.IntervalKeyPool
-import io.shiftleft.codepropertygraph.generated.{EdgeTypes, PropertyDefaults, PropertyNames}
+import io.joern.x2cpg.{Ast, ValidationMode}
 import io.shiftleft.codepropertygraph.generated.nodes.*
+import io.shiftleft.codepropertygraph.generated.{EdgeTypes, PropertyDefaults, PropertyNames}
 import ujson.Value
 
 import scala.collection.mutable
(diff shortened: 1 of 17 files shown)


joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/TypeHelper.scala
```diff
@@ -8,9 +8,10 @@ import java.util.regex.Pattern
 
 trait TypeHelper { this: AstCreator =>
 
-  private val TypeAnnotationKey = "typeAnnotation"
-  private val ReturnTypeKey     = "returnType"
-  private val ImportMatcher     = Pattern.compile("(typeof )?import\\([\"'](.*)[\"']\\)")
+  private val TypeAnnotationKey   = "typeAnnotation"
+  private val ReturnTypeKey       = "returnType"
+  private val ImportMatcher       = Pattern.compile("(typeof )?import\\([\"'](.*)[\"']\\)")
+  private val ArrayReplacePattern = Pattern.compile(":[^,.]+\\[]")
```
maltek: this isn't wrong, but scala's `"".r` syntax is nicer

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -18,15 +18,36 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
-    val c     = code(json)
-    val ln    = line(json)
-    val cn    = column(json)
-    val lnEnd = lineEnd(json)
-    val cnEnd = columnEnd(json)
-    val node  = nodeTypeOf(json)
-    BabelNodeInfo(node, json, c, ln, cn, lnEnd, cnEnd)
+    // Resolve start/end offsets and their line numbers once, then derive columns and code from them. Going through the
+    // individual code/line/column/lineEnd/columnEnd accessors would recompute start/end and the line binary search
+    // several times per node, and createBabelNodeInfo runs for every AST node.
```
maltek: LLMs love to add comments talking about how things used to behave before their changes. Scrub them vigorously - this is outdated/useless as soon as this PR is merged.

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForTypesCreator.scala
```diff
@@ -76,20 +76,36 @@ trait AstForTypesCreator(implicit withSchemaValidation: ValidationMode) { this:
     case _                               => safeStr(json, "kind").contains("constructor")
   }
 
-  private def classMembers(clazz: BabelNodeInfo, withConstructor: Boolean = true): Seq[Value] = {
+  /** The members of a class, decomposed once so callers can derive the member sequence (with or without the
+    * constructor) and reuse the located constructor without re-walking the class body.
+    */
```
maltek: I find the comment about doing this once misleading at this location - the class doesn't care how often it is used. The functions creating instances of it don't care either. Only in `createClassConstructor()` does this comment make sense.

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForStatementsCreator.scala
```diff
@@ -36,7 +36,7 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) { t
     }
 
   protected def createBlockStatementAsts(json: Value): List[Ast] = {
-    val blockStmts = sortBlockStatements(json.arr.toList.map(createBabelNodeInfo))
+    val blockStmts = sortBlockStatements(json.arr.iterator.map(createBabelNodeInfo).toList)
```
maltek, line 39: I suppose it's a larger refactor, so not necessarily to be done in this PR. But if `AstCreatorBase` just took `collection.Seq` instead of `List` and `Seq` everywhere, we could avoid a conversion from an efficient data structure to an inefficient one here (and probably in other places, too).

maltek (approved):

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/AstForStatementsCreator.scala
max-leuthaeuser: Yes, that's what I wanted to do initially - but the change would be way to large for now.

joern-cli/frontends/jssrc2cpg/src/main/scala/io/joern/jssrc2cpg/astcreation/TypeHelper.scala
max-leuthaeuser: True for ArrayReplacePattern. But I leave ImportMatcher as-is because we pass it around.

PR #6090, opened 2026-07-09
[swiftsrc2cpg] JVM/Scala performance & memory fixes

max-leuthaeuser:
Targeted, behavior-preserving performance work, from an Opus 4.8 review of `swiftsrc2cpg/src/main`. Two groups: fixes on the per-node hot path, and a set of medium/low-impact wins across visitors, the parser, and the CPG passes.

Findings from the LLM were manually reviewed.

### Hot path (per-AST-node)

- **code(node)**: decode directly from the shared UTF-8 byte array instead of slicing a copy first, and fix a latent charset bug (was using the platform default charset).
- **Parser child accessors**: cache generated single-child accessors and children as lazy val so re-accessing a field on the same wrapper no longer rebuilds a fresh wrapper (and its child map / Seq). Regenerated `SwiftNodeSyntax.scala` via the astgen generator; callers bind repeated .children chains once.
- **FullnameProvider**: memoize the recursive range resolution (per-file `typeMap` is immutable), short-circuit `filterForNodeKind`, and guard the unconditional `<extension>` replace.
- **astForNode dispatch**: match on the base syntax category before testing the 6-way `FunctionDeclLike` cross-category union, so every node no longer runs those 6 `isInstanceOf` tests first.

### Passes (whole-graph, one-time)

- **ObjcCallFullNamePass**: lead with `nameExact("init")` so the NAME index is used, instead of scanning every call via `.isStatic` first.
- **ExtensionsPass**: replace linear `values.find` with Set membership and add early-return guards when the gating maps are empty.
- **SwiftTypesProvider type-map build**: use two bounded thread pools (blocking compiler invocations vs. JSON-line workers) instead of one shared unbounded pool that could over-subscribe and starve workers.

### AST-creation core

- **variableFromStack (x2cpg)**: single hash probe per scope level instead of contains + get.
- **astsForBlockElements**: fast path that skips the `partition/indexWhere` scans when a block has no defer/guard.
- **astForTypeDeclSyntax**: classify members in one pass instead of three filter passes re-running the member predicates.
- **argAstsForCall**: extract the repeated positional-args + trailing-closures construction shared by five call handlers (evaluation order preserved).
- **Removed dead astParentInfo helper**: left-over, not used.

### AST visitors

- **astForCodeBlockItemListSyntax**: stable `partition` instead of `sortBy`.

### Parser & I/O

- **ParseResult**: retain only the UTF-8 byte array and decode the source String on demand (was holding both).
- **GsonTypeInfoReader**: skip the running-range recompute after scalar fields (they never affect `safeRange`), only refreshing after structural fields.
- **SwiftTypesProvider**: collect unique mangled names into a Set up front for batch demangle instead of a per-reference list.
- **removeModifier**: precompute padded modifier variants and early-exit for names with no space.

### Config / misc passes

- **ConfigFileCreationPass**: read each config file once, classify it, and reuse the content for the text branch (was reading the whole file for the binary check, then again).
- **ClosureBindingsPass**: single `nextOption` lookup per closure instead of `isEmpty` + `loneElement`, computing the bound-method arg count only when a stub is built.


joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/utils/FullnameProvider.scala
```diff
@@ -36,6 +37,13 @@ private object FullnameProvider {
   */
 class FullnameProvider(typeMap: SwiftFileLocalTypeMapping) {
 
+  /** Memoizes the (relatively expensive) recursive range resolution. A FullnameProvider is created once per file and
+    * only accessed from that file's single AST-creation thread, and `typeMap` is immutable, so an unsynchronized cache
+    * is safe and its results are stable. The same node is typically resolved several times per occurrence (e.g. once
+    * for the decl name, again for the type, again while building the identifier node).
+    */
+  private val fullNameCache = mutable.HashMap.empty[(Int, Int, FullnameProvider.Kind, String), Option[String]]
```
maltek: define this variable next to `cachedFullName()` since that method is how the variable is supposed to be accessed

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/datastructures/VariableScopeManager.scala
```diff
@@ -553,8 +553,11 @@ class VariableScopeManager {
     variableName: String
   ): Option[(NewNode, String, String)] = {
     stack.flatMap { stackElement =>
-      if (stackElement.nameToVariableNode.contains(variableName)) { stackElement.nameToVariableNode.get(variableName) }
-      else { variableFromStack(stackElement.surroundingScope, variableName) }
+      // Single hash probe per scope level: a `contains` followed by `get` would probe the map twice.
```
maltek: I don't think this needs a comment

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -140,13 +140,22 @@ trait AstCreatorHelper(implicit withSchemaValidation: ValidationMode) { this: As
...
-    val deferElementsAstsOrdered = deferElements.reverse.map(astForNode)
-    val indexOfGuardStmt = otherElements.indexWhere(n =>
-      n.isInstanceOf[CodeBlockItemSyntax] && n.asInstanceOf[CodeBlockItemSyntax].item.isInstanceOf[GuardStmtSyntax]
-    )
+    def isDefer(node: SwiftNode): Boolean =
+      node
+        .isInstanceOf[CodeBlockItemSyntax] && node.asInstanceOf[CodeBlockItemSyntax].item.isInstanceOf[DeferStmtSyntax]
+    def isGuard(node: SwiftNode): Boolean =
+      node
+        .isInstanceOf[CodeBlockItemSyntax] && node.asInstanceOf[CodeBlockItemSyntax].item.isInstanceOf[GuardStmtSyntax]
```
maltek: move the new lines after the `&&`

joern-cli/frontends/x2cpg/src/main/scala/io/joern/x2cpg/frontendspecific/swiftsrc2cpg/Defines.scala
```diff
@@ -51,6 +51,11 @@ object Defines {
       Void
     )
 
+  /** `Set` view over [[SwiftTypes]] for O(1) membership tests on the per-node hot path (the `List` order is kept for
+    * callers that rely on it, e.g. `BuiltinTypesPass` deriving `order` from `zipWithIndex`).
+    */
+  val SwiftTypesSet: Set[String] = SwiftTypes.toSet
```
maltek: if we're micro-optimizing: mutable sets are faster than immutably ones, and we can still use the type system to enforce they are not mutated. (Immutable sets are optimized for building them not in bulk, but for building them one item at a time and still having `O(1)` access - but the constant factors for reading from them is worse than for mutable sets.)

 `val SwiftTypesSet: collection.Set[String] = SwiftTypes.toSetMutable`

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ConfigFileCreationPass.scala
```diff
@@ -25,20 +25,23 @@ class ConfigFileCreationPass(cpg: Cpg, config: Config)
...
-  private def isBinaryPlist(file: Path): Boolean = {
-    if (!file.extension().contains(PlistExt)) return false
-    Try(IOUtils.readLinesInFile(file)) match {
-      case Success(dataBeginning :: _) => dataBeginning.trim.startsWith("bplist")
-      case _                           => false
+  /** Reads the file once and classifies it:
+    *   - `Some(content)` — a text file whose content can be added to the CPG directly.
+    *   - `None` — a binary plist that must be parsed via [[PropertyListParser]] first.
+    */
+  private def textContentOrBinaryPlist(file: Path): Try[Option[String]] = {
```
maltek: The name doesn't match what it does. Instead of returning `None`, call a helper function in here to convert the bplist.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ConfigFileCreationPass.scala
```diff
@@ -25,20 +25,23 @@ class ConfigFileCreationPass(cpg: Cpg, config: Config)
...
-    val parseResult = if (isBinaryPlist(file)) {
-      val sourceComment = s"\n<!--This file was generated from ${file.toAbsolutePath}-->"
-      Try((PropertyListParser.parse(file.toFile).toXMLPropertyList, sourceComment))
-    } else {
-      Try((IOUtils.readEntireFile(file), ""))
+    val parseResult = textContentOrBinaryPlist(file).flatMap {
+      case Some(textContent) => Success((textContent, ""))
+      case None =>
+        val sourceComment = s"\n<!--This file was generated from ${file.toAbsolutePath}-->"
+        Try((PropertyListParser.parse(file.toFile).toXMLPropertyList, sourceComment))
```
maltek: use a single `Try` around the whole block, instead of multiple small ones

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ConfigFileCreationPass.scala
```diff
@@ -25,20 +25,23 @@ class ConfigFileCreationPass(cpg: Cpg, config: Config)
...
-    Try(IOUtils.readLinesInFile(file)) match {
-      case Success(dataBeginning :: _) => dataBeginning.trim.startsWith("bplist")
-      case _                           => false
+  /** Reads the file once and classifies it:
+    *   - `Some(content)` — a text file whose content can be added to the CPG directly.
+    *   - `None` — a binary plist that must be parsed via [[PropertyListParser]] first.
+    */
+  private def textContentOrBinaryPlist(file: Path): Try[Option[String]] = {
+    Try(IOUtils.readEntireFile(file)).map { content =>
+      val isBinaryPlist = file.extension().contains(PlistExt) && content.trim.startsWith("bplist")
```
maltek: Ideally we'd only read the first 6 bytes from a buffered stream here to check for `bplist`, instead of the whole file. But it's already an improvement, so don't feel obligated to do that.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ClosureBindingsPass.scala
```diff
@@ -21,16 +21,16 @@ class ClosureBindingsPass(cpg: Cpg) extends CpgPass(cpg) {
...
-      seenTypeDecls.getOrElseUpdate(
-        fullName, {
-          val typeDeclStub = TypeDeclStubCreator.createTypeDeclStub("Function", fullName)
-          diffGraph.addNode(typeDeclStub)
-          typeDeclStub
-        }
-      )
-    } else {
-      cpg.typeDecl.fullNameExact(fullName).loneElement
+    cpg.typeDecl.fullNameExact(fullName).loneElementOption match {
```
maltek: `loneElementOption` has different semantics: it returns `None` if there is more than one match. So we used to crash loud in case of a fullName clash, now we treat it like there is no such type at all, and add another one.

If this is a performance issue, at least use `headOption` / `nextOption` to not make the situation worse in case of clashes.

joern-cli/frontends/swiftsrc2cpg/src/main/scala/io/joern/swiftsrc2cpg/passes/ClosureBindingsPass.scala
```diff
@@ -40,22 +40,21 @@ class ClosureBindingsPass(cpg: Cpg) extends CpgPass(cpg) {
...
-            closureMethod.signature,
-            DispatchTypes.DYNAMIC_DISPATCH,
-            numArgs,
-            diffGraph
-          )
-        }
-      )
-    } else {
-      cpg.method.fullNameExact(methodFullName).loneElement
+    cpg.method.fullNameExact(methodFullName).loneElementOption match {
```
maltek: same here

maltek (approved):

PR #6091, opened 2026-07-09
[rust2cpg] download RustNodeSyntax with sbt.

xavierpinho:
Every time we update rust_ast_gen, we should also update RustNodeSyntax.scala (bundled in the release assets.)

To update rust_ast_gen, one only has to update `application.conf` and sbt takes care of downloading it. But RustNodeSyntax.scala has always been updated manually so far. Hopefully this automates that seamlessly.


max-leuthaeuser (approved):

max-leuthaeuser:
Note: this will only work if the Scala file downloaded is correctly formatted and does not fail the scalafix linting.

xavierpinho:
> Note: this will only work if the Scala file downloaded is correctly formatted and does not fail the scalafix linting.

Linting-wise it should be fine, but I know for a fact that it isn't formatted like scalafmt would. I  have the habit of running scalafmt before pushing, so didn't even think of it. I suppose one could instruct sbt to run scalafmt after the download as well. (Formatting it upstream would work, but drift as soon as we change our scalafmt settings anyway.)

PR #6096, opened 2026-07-10
[php2cpg] More V3 validation fixes

max-leuthaeuser:
**1. Dangling Local under a named namespace (`AstCreator.astForNamespaceStmt`)** Top-level statements inside `namespace Foo; ...` were attached directly under the namespace block, but `handleVariableOccurrence` only wires locals to `BlockScope`/`MethodScope`. Iinsert an intermediate `Block` (and `BlockScope`) between the namespace block and its statements, giving locals a valid AST parent.

**2. Dangling Local from default-constructor field inits (`AstForFunctionsCreator.defaultConstructorAst`)** Array/new field initializers (`protected $x = [...]`) were evaluated before the constructor's `MethodScope` was pushed, so their tmp locals dangled. Removed the premature evaluation so `astForMethodDecl` builds them inside the method scope — matching the explicit-constructor path.

**3. `NONLOCAL_REF` — cross-method variable resolution (`Scope` + `AstCreatorHelper.handleVariableOccurrence`)** `lookupVariable` walked the whole scope stack, so a function/closure resolved a name to a Local/param in an enclosing method (the "global var without global keyword" case, sibling-method params, plain closures). Added `lookupVariableInCurrentMethod` (stops at the nearest method boundary). and use it for non-arrow-closure lookups — arrow/use captures still work. Also deferred parameter scope-registration and param-attribute AST creation (array literals) until after the MethodScope push.

**4. Closure-name collision regression (`Scope.getScopedClosureName`)** Fix 1's block-on-stack broke closure naming (only inspected `stack.headOption`). Changed to `collectFirst` over the stack, so multiple top-level closures in a namespace get unique names.

**5. Duplicate fullNames for symbols declared across multiple files (new `FullNameUniquenessPass`)** PHP allows the same top-level class or function name to be declared in multiple files (they are never loaded together at runtime, e.g., an admin and a frontend variant of a class). The dedup counter in `ScopeElement` is per-file, so these cross-file collisions survive as duplicate fullNames. Added a `FullNameUniquenessPass` (mirroring the c2cpg/swiftsrc2cpg passes; needs DRY-cleanup later), run after `AstCreationPass`, which appends a deterministic `<duplicate>N` suffix (ordered by file/line/column) to duplicate `Method`/`TypeDecl`/`NamespaceBlock` fullnames and fixes up same-file calls resolved to a renamed method.

**6. `NONLOCAL_REF` — use-closure local leaking into the enclosing scope (`AstForFunctionsCreator.astForClosureExpr`)** When a variable was captured by a `use ($var)` closure, `scope.addToScope(var, local)` inserted the closure's captured Local into the enclosing scope (the closure's own `MethodScope` is not pushed until `astForMethodDecl`). A later identifier for the same name in the outer method then resolved to the closure's Local, producing a `REF` edge across the method boundary. Removed that call — the captured Local is already registered into the closure's own method scope via `astForMethodDecl`'s `bodyPrefixAsts` handling — so outer identifiers now ref the outer Local.

Enables CS V3 validation for both integration and sptest.

### DIFF
--- a/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
+++ b/joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
@@ -150,22 +150,6 @@ trait AstCreatorHelper(disableFileContent: Boolean)(implicit withSchemaValidatio
     modifiers: List[String] = List.empty
   ): NewNode = {
     scope.lookupVariable(name) match {
-      case None =>
-        val localCode = if (name == NameConstants.Self) NameConstants.Self else s"$$$name"
-        val local     = localNode(expr, name, code.getOrElse(localCode), tfn.getOrElse(Defines.Any))
-
-        modifiers.foreach { modifier =>
-          val modNode = modifierNode(expr, modifier)
-          diffGraph.addEdge(local, modNode, EdgeTypes.AST)
-        }
-
-        scope.addToScope(name, local) match {
-          case BlockScope(block, _)                 => diffGraph.addEdge(block, local, EdgeTypes.AST)
-          case MethodScope(_, block, _, _, _, _, _) => diffGraph.addEdge(block, local, EdgeTypes.AST)
-          case _                                    => // do nothing
-        }
-
-        local
       case Some(local: NewLocal)
           if scope.isSurroundedByArrowClosure && local.closureBindingId.exists(_.contains("<lambda>")) =>
         local // the contains check ensures that we can capture global variables into an arrow closure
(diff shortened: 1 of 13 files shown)


joern-cli/frontends/php2cpg/src/main/scala/io/joern/php2cpg/astcreation/AstCreatorHelper.scala
```diff
@@ -174,7 +158,26 @@ trait AstCreatorHelper(disableFileContent: Boolean)(implicit withSchemaValidatio
...
+
+            modifiers.foreach { modifier =>
+              val modNode = modifierNode(expr, modifier)
+              diffGraph.addEdge(local, modNode, EdgeTypes.AST)
+            }
+
+            scope.addToScope(name, local) match {
+              case BlockScope(block, _)                 => diffGraph.addEdge(block, local, EdgeTypes.AST)
+              case MethodScope(_, block, _, _, _, _, _) => diffGraph.addEdge(block, local, EdgeTypes.AST)
+              case _                                    => // do nothing
```
maltek, line 176: don't we get a dangling local in this case? The local is added to the diffgraph in line 170, but it doesn't get a AST parent here.

max-leuthaeuser: That's why it's so important to always have a block or method scope available. That was also the reason for the errors before. The whole scoping / ref / closure bindings handling is interesting to say the least. A rewrite to the existing and proven `VariableScopeManager` in x2cpg is time-wise out of scope for now. And having V3 in first is a perfect foundation anyway.

maltek: I was going to say, then log an error at least. But I suppose the validator already serves that purpose.

maltek (approved):

PR #6098, opened 2026-07-10
[pysrc2cpg] Correctly handle delete on list

ml86:
My assumption that delete on a list is invalid python was wrong.
E.g. `del [a, b]` is perferctly valid and is equivalent to `del a`, `del
b`.

Fix for https://harness.atlassian.net/browse/QT-1147

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
@@ -2041,10 +2041,6 @@ class PythonAstVisitor(
   }
 
   def convert(list: ast.List): nodes.NewNode = {
-    // Must be a List as part of a Load memory operation because a List literal
-    // is not permitted as argument to a Del and List as part of a Store does not
-    // reach here.
-    assert(memOpMap.get(list).get == Load)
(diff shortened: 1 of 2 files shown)


maltek (approved):
this lowering will be hard to support on the dataflow side (it's not just a `NULL` mapping as it would usually be for the `del` operator). But I suppose that could just be seen as another (quite rare) instance of overtainting.

ml86:
@maltek Is `del` considered a security relevant sink? Otherwise I dont quite get what you mean since there is not much lowering going on here. `del [a, b]` is represented as CALL <operator>.delete with a list literal CALL child.

maltek:
delete isn't securit-critical itself. I was thinking of something like this:

```python
In [1]: a = {'a': 'A', 'b': 'B', 'c': 'C'}

In [2]: del [a['a'], a['b']]

In [3]: a
Out[3]: {'c': 'C'}
```

if `a['a']` contained tainted data the before the `del`, then it's gone after that is executed. If this was lowered as `del a['a']; del a['b']` then we could use a very simple dataflow semantic for the delete operator to model the deletion in our dataflow engine. (A map from NULL to param 1 in our policies.) But with the list creation in between we can't model it at all and just have to live with pretending that `a` contains everything it did before.

maltek:
But I don't think `del` is used all that often overall, so I don't see this as a major issue.

ml86:
I see your point. Will address this in a followup PR to unblock a customer issue.

PR #6099, opened 2026-07-13
[kotlin2cpg] Enable V3 validation

max-leuthaeuser:
Bumps the kotlin2cpg test `PostFrontendValidator` from `V1` to `V3`, which makes `DANGLING_AST_NODE` (a node with no AST parent) fatal. This surfaced two genuine AST-malformation bugs that V1 had been hiding.

### Fixes

- **Dangling `BLOCK`** — In `astForWhenAsStatement`, a `when (val x = …)` subject is wrapped in a `BLOCK` (`finalAstForSubject`), but the switch's AST children used the unwrapped `astForSubject`; the block was only referenced by the CONDITION edge and had no AST parent. Now attaches `finalAstForSubject` as the child (identical to `astForSubject` for non-property subjects).

- **Dangling `NAMESPACE_BLOCK`** — `javasrc2cpg`'s `AstCreator.createAst` created the `FILE` node and namespace-block AST as disconnected subgraphs. Standalone javasrc2cpg validates at V0 so never caught it, but kotlin2cpg reuses that pass for Java interop at V3. Now roots the translation unit under the file node (`Ast(fileNode).withChild(...)`).

_Note:_ the second fix touches shared `javasrc2cpg` code. The correct place, since it was emitting a malformed AST.

CS integration tests and sptests for both frontends are green.

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForStatementsCreator.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForStatementsCreator.scala
@@ -352,7 +352,7 @@ trait AstForStatementsCreator(implicit withSchemaValidation: ValidationMode) {
       .map { text => s"${Constants.WhenKeyword}($text)" }
       .getOrElse(Constants.WhenKeyword)
     val switchNode = controlStructureNode(expr, ControlStructureTypes.SWITCH, shortenCode(codeForSwitch))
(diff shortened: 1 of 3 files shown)


joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/ast/AstForStatementsCreator.scala
maltek: `astForSubject` is no longer used now, remove it

maltek (approved):

ml86 (approved):

PR #6106, opened 2026-07-13
[rust2cpg] turn RustNodeSyntax.scala a generated source.

xavierpinho:
The automatic download of RustNodeSyntax.scala introduced in #6091 turned out to be problematic for the release build:

- https://github.com/joernio/joern/actions/runs/29251055173/job/86819399027
- https://github.com/joernio/joern/actions/runs/29233904157/job/86764046584
- https://github.com/joernio/joern/actions/runs/29184617956/job/86628353262

Instead of scalafmt'ing this particular file after download, Claude (Opus 4.8) suggested instead to turn this file into a source-generated one. 

I like that approach: RustNodeSyntax should never be manually changed; we already have its diff upstream when we want to see what changed between versions; and it's one less file under revision control. 

But I'm not proficient enough with sbt to know of any pitfalls this might introduce. Moreover, regarding Bazel - would this approach be problematic?

(Credits to @max-leuthaeuser for reporting this issue.)


max-leuthaeuser (approved):

max-leuthaeuser:
For bazel:

The PR removes RustNodeSyntax.scala from the committed source tree and instead downloads it at sbt build time into sourceManaged (i.e., target/scala-3.x.x/src_managed/main/...).

The Bazel BUILD file uses:
`srcs = glob(["src/main/scala/**/*.scala"]),`

That glob only sees committed sources in `src/main/scala/`. It will never find a file in sbt's `sourceManaged` directory, and Bazel's hermetic build model prohibits arbitrary network downloads during a build action anyway.

Potential fix:

1. `MODULE.bazel` — add an `http_file` for `RustNodeSyntax.scala` alongside the existing binary entries:

```
http_file(
    name = "rust2cpg_node_syntax",
    sha256 = "",  # fill in once stable
    url = "https://github.com/joernio/astgen-monorepo/releases/download/rust-astgen%2Fv{}/RustNodeSyntax.scala".format(RUST2CPG_ASTGEN_VERSION),
)
```

2. `joern-cli/frontends/rust2cpg/BUILD` — extend `srcs` in the `scala_library` target:

```
scala_library(
    name = "rust2cpg",
    srcs = glob(["src/main/scala/**/*.scala"]) + ["@rust2cpg_node_syntax//file"],
    ...
)
```

max-leuthaeuser (approved):

ml86 (approved):

PR #6111, opened 2026-07-14
[pysrc2cpg] mangle duplicate class and function names

johannescoetzee:
This PR introduces the ClassDefDuplicateCalculator to pysrc2cpg which is used to calculate duplicate indices for conflicting class definitions, keeping the last definition unsuffixed since it's the one most likely to be used the most. 

I don't know if `<duplicate>N` is a good choice of suffix since it may be used with a different meaning in other frontends, so I'm open to suggestions for a better one!

sptests and integration tests are fine: https://buildbot.global.sltr.io/#/builders/82/builds/774

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
@@ -47,7 +47,7 @@ class PythonAstVisitor(
   private var memOpMap: AstNodeToMemoryOperationMap = scala.compiletime.uninitialized
   private var scopeNames: ScopeNameCollection       = scala.compiletime.uninitialized
 
-  private val classDefToDuplicateIndex = mutable.Map.empty[ast.ClassDef, Int]
+  private val defToDuplicateIndex = mutable.Map.empty[ClassOrFunctionDef, Int]
 
   private val members = mutable.Map.empty[NewTypeDecl, List[String]]
 
@@ -86,7 +86,7 @@ class PythonAstVisitor(
     memOpMap = memOpCalculator.astNodeToMemOp
     scopeNames = memOpCalculator.scopeNames
 
-    classDefToDuplicateIndex ++= new ClassDefDuplicateCalculator().calculate(module)
+    defToDuplicateIndex ++= new RedefinitionCalculator().calculate(module)
 
     val contentOption = if (enableFileContent) Some(nodeToCode.content) else None
     val fileNode      = nodeBuilder.fileNode(relFileName, contentOption)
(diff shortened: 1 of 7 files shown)


johannescoetzee:
I've updated the PR to include function methodFullName mangling as well. These were already mangled to 
```
f
f$redefinition0
...
```
but that didn't follow the same ordering as the new ClassDef version and required a separate mechanism for keeping track of it.

Taking inspiration from the `redefinition` suffix, I also changed the suffix in this PR from `<duplicate>` to `<redefined>` since these were, in fact, redefined and not duplicated.

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/RedefinitionCalculator.scala
```diff
@@ -4,56 +4,67 @@ import io.joern.pythonparser.ast
...
 
   private def popScopeAndAssignIndices(): Unit = {
     val scope = scopeStack.pop()
-    scope.values.foreach { defs =>
-      if (defs.length > 1) {
-        defs.zipWithIndex.init.foreach { case (cd, idx) =>
-          classDefToDuplicateIndex(cd) = idx
+    scope.values.foreach { mixedDefs =>
+      val (classes, functions) = mixedDefs.partition(_.isInstanceOf[ast.ClassDef])
+      List(classes, functions).foreach { defs =>
```
ml86, line 51: Why do we need to partition here?

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/PythonAstVisitor.scala
```diff
@@ -270,19 +268,21 @@ class PythonAstVisitor(
     body: ast.CollType[istmt],
     returns: Option[iexpr],
     isAsync: Boolean,
-    functionDef: istmt
+    functionDef: ClassOrFunctionDef
```
ml86: Please use `FunctionDef | AysncFunctionDef`. This way it is misleading because no classes are expected to be passed in here.

ml86 (changes requested):
Visiting the the hole AST structure at the start of the module traversal is too expensive

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/RedefinitionCalculator.scala
johannescoetzee: We currently create a list of mixed class and function defs and assign indices later. The partition here is to keep the indices separate between classes and functions. It's not necessary to do and not doing so would just mean suffix indices are shared, so something like:
```
class<redefined>0
function<redefined<1>
class<redefined>2
class
function
```

johannescoetzee:
> Visiting the the hole AST structure at the start of the module traversal is too expensive

I could be missing something obvious, but it's not clear to me that we can do much better. Since we don't know if a class or function def is redefined later, we would, at the latest, have to do this when we encounter a class or function def and will most likely end up traversing the whole AST, or close to, anyways. We could potentially save a bit of memory by doing it per-scope, but since only the redefined classes/functions which need a suffix are added to the map, this will be empty in most cases or very small in others.

joern-cli/frontends/pysrc2cpg/src/main/scala/io/joern/pysrc2cpg/RedefinitionCalculator.scala
ml86: Got it. Makes sense.

ml86 (approved):

PR #6112, opened 2026-07-15
[c2cpg] JVM/Scala performance & memory fixes 

max-leuthaeuser:
## [c2cpg] Performance fixes across AST creation pipeline

Targeted, behavior-preserving performance work, from an Opus 4.8 review of c2cpg/src/main. Addresses algorithmic complexity, redundant allocations, and concurrency issues found via profiling by me. Otherwise, findings from the LLM were manually reviewed.

### Results

Running c2cpg on https://github.com/polarphp/polarphp is down to 28sec (from 3min 20sec) on my machine.

### Changes

**Algorithmic fixes**
- Fix O(arguments × nodes) scan in `argumentTrees` — use pre-built index map instead of linear search per argument
- Fix O(n²) string concatenation in `tokenListToString` — replace `++=` accumulation with `StringBuilder`
- Fix O(nodes × captures) in `setEvaluationStrategyForCaptures` — eliminate repeated full-node scans
- Replace O(n²) `List.init` recursion with a left-fold in `fieldAccesses`
- Make `parameters()` tail-recursive to avoid intermediate sequences at each nesting level

**Allocation / hot-path fixes**
- Fix `fileName(node: IASTNode)` path handling
- Eliminate per-character tuple boxing in `genFileOffsetTable` (`zipWithIndex` → index variable)
- Eliminate per-call string allocations in `cleanType`
- Avoid hot-path `List` allocation in `expandedFromMacro`
- Replace `regex.replaceAll` with literal `replace` in `paramIsVariadic`
- Use direct array access (`apply(index)`) instead of `toList(index)` in `typeForDeclSpecifier`
- Compute `safeGetNodeType` once in `typeForIASTArrayDeclarator`
- Call `nullSafeFileLocation` once per node in `nodeOffsets` (was called twice)
- Skip redundant `normalizeSpace` in `cleanType` when input is already normalized

**Caching / concurrency fixes**
- Cache Levenshtein sort results in `HeaderFileFinder`
- Fix concurrent races in `IncludeAutoDiscovery`

Better review alternative: https://diffshub.com/joernio/joern/pull/6112

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
@@ -411,11 +411,12 @@ trait TypeNameProvider { this: AstCreator =>
   @nowarn
   private def typeForIASTArrayDeclarator(a: IASTArrayDeclarator): String = {
     import org.eclipse.cdt.core.dom.ast.ASTSignatureUtil.getNodeSignature
-    if (safeGetNodeType(a).startsWith("? ")) {
+    val nodeType = safeGetNodeType(a)
+    if (nodeType.startsWith("? ")) {
       val tpe = getNodeSignature(a).replace("[]", "").strip()
-      val arr = safeGetNodeType(a).replace("? ", "")
+      val arr = nodeType.replace("? ", "")
       s"$tpe$arr"
-    } else if (safeGetNodeType(a).contains("} ") || safeGetNodeType(a).contains(" [")) {
+    } else if (nodeType.contains("} ") || nodeType.contains(" [")) {
       val tpe = getNodeSignature(a).replace("[]", "").strip()
       val arr = a.getArrayModifiers.map {
         case m if m.getConstantExpression != null => s"[${nodeSignature(m.getConstantExpression)}]"
@@ -428,7 +429,7 @@ trait TypeNameProvider { this: AstCreator =>
       }.mkString
       s"$tpe$arr"
     } else {
-      safeGetNodeType(a)
+      nodeType
     }
   }
 
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
@@ -108,16 +108,16 @@ trait TypeNameProvider { this: AstCreator =>
         }
     stripTemplateTags(replaceWhitespaceAfterKeyword(tpe)) match {
       // Empty or problematic types
-      case ""                                                                      => Defines.Any
-      case t if t.contains("?")                                                    => Defines.Any
-      case t if t.contains("#")                                                    => Defines.Any
-      case t if t.contains("::{") || t.contains("}::")                             => Defines.Any
-      case t if (t.contains("{") || t.contains("}")) && !isThisLambdaCapture(t)    => Defines.Any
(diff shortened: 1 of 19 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/parser/CustomFileContentProvider.scala
```diff
@@ -30,7 +32,11 @@ class CustomFileContentProvider(
     else { Option(path) }
     maybeFullPath.map { foundPath =>
       updateHeaderFileParserLanguage(foundPath)
-      val content = IOUtils.readLinesInFile(Paths.get(foundPath)).mkString("\n").toArray
+      val content =
+        contentCache.computeIfAbsent(
+          foundPath,
+          filePath => IOUtils.readLinesInFile(Paths.get(filePath)).mkString("\n").toArray
+        )
```
maltek: This increases memory by potentially a lot, and creates an additional synchronization point between threads, just to save a few syscalls (the OS disk cache should give us these file contents without hitting the disk as long as there is enough RAM available). Do you have numbers how much of a difference this individual change is making?

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
```diff
@@ -60,34 +60,34 @@ trait TypeNameProvider { this: AstCreator =>
         val parentDecl = s.getParent.asInstanceOf[IASTFunctionDefinition].getDeclarator
         ASTStringUtil.getReturnTypeString(s, parentDecl)
       case s: IASTSimpleDeclaration if s.getParent.isInstanceOf[ICASTKnRFunctionDeclarator] =>
-        val decl = s.getDeclarators.toList(index)
+        val decl = s.getDeclarators.apply(index)
```
maltek: just write this as `s.getDeclarators(index)` (same with the other new `apply` calls)

maltek (commented):
> cleanType is called on every type string in the frontend (millions of
times on large C++ codebases). The foldLeft over ReservedKeywordsAtTypes
and the guard in replaceWhitespaceAfterKeyword both interpolated pattern
strings like s"$repl " and s" $repl " on every invocation, allocating
up to ~24 short-lived strings per call even when no keyword was present.

short-lived strings (or short-lived anything really) should be similarly cheap on the JVM as stack-allocation in e.g. C. Doing less work is always better for perf, but I wouldn't expect any notable difference from that explanation.

> If CDT ever reuses
internal caches (include guards, template instantiation), results would
be silently corrupted.

Have they started development again? I thought it's dead / on life support. Making changes that make things more complex (thread locals) just in case the upstream will change their internals in exactly the way predicted here seems very speculative to me.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
maltek: I don't understand. Scala lets you index Java arrays using `()`

```
ocular> val x = Array("1", "2", "3")
val x: Array[String] = Array("1", "2", "3")

ocular> x.getClass.getPackageName
val res87: String = "java.lang"

ocular> x(0)
val res88: String = "1"
```

max-leuthaeuser:
> > If CDT ever reuses
> > internal caches (include guards, template instantiation), results would
> > be silently corrupted.
> 
> Have they started development again? I thought it's dead / on life support. Making changes that make things more complex (thread locals) just in case the upstream will change their internals in exactly the way predicted here seems very speculative to me.

Maybe you are right, we anyway stay at our own custom CDT version.

maltek (approved):

PR #6113, opened 2026-07-15
[rust2cpg] add binding nodes.

xavierpinho:
1. Each trait method gets a Binding node, e.g.

```rust
trait MyTrait { fn do_stuff(&self); }
```

becomes a TypeDecl (fullName = crate::MyTrait, inheritsFrom = []) with a Method `crate::MyTrait::do_stuff`, and there's an associated Binds edge to a Binding(name=do_stuff) node, with a Ref edge to the `crate::MyTrait::do_stuff` method.

2. Same transformation is applied to impl trait declarations, e.g.

```rust
impl MyTrait for FooBar { fn do_stuff(&self) {} }
```

becomes a TypeDecl (fullName = `<crate::FooBar as crate::MyTrait>`, inheritsFrom = [`crate::MyTrait`]) with a Method `<crate::FooBar as crate::MyTrait>::do_stuff`, and there's an associated Binds edge to a Binding(name=do_stuff) node, with a Ref edge to the `<crate::FooBar as crate::MyTrait>::do_stuff` method.

3. Struct declarations get an inheritsFrom entry for each impl trait they implement, e.g.

```rust
struct FooBar;
// (assuming the previous trait MyTrait and impl MyTrait for FooBar exist somewhere in the code)
```

becomes a TypeDecl (fullName = `crate::FooBar`, inheritsFrom = [`<crate::FooBar as crate::MyTrait>`]). For this we accumulate all impl trait declarations and update inheritsFrom in a post `RustTraitImplUpdatePass`.


maltek:
> Struct declarations get an inheritsFrom entry for each impl trait they implement, e.g.

What are the current implementations with generics?

- Is there any chance of duplicates/clashes e.g. for `impl Trait for Type<u8>` plus `impl Trait for Type<u16>`? (That's not specialization, just two implementations for disjoint types.)
- what about implementations which don't even mention a concrete type name, since they work for any `T` satisfying some type bounds? `impl<T> Trait for T where ...`

xavierpinho:
> Is there any chance of duplicates/clashes e.g. for impl Trait for Type<u8> plus impl Trait for Type<u16>?

We shall end up with TypeDecl fullNamed e.g. `<crate::Type<u8> as crate::Trait>`. Needs some work because in some places the visitor is using only the name (without any type parameters) to create the nodes. It's nice to at least know that the post-validator would catch duplicate TypeDecls too.

> what about implementations which don't even mention a concrete type name, since they work for any T satisfying some type bounds? impl<T> Trait for T where ...

As it stands today, we end up with:
1. `<T as crate::Trait>`, for `impl<T> Trait for T`
2. `<crate::Type<T> as crate::Trait>`, for `impl<T> Trait for Type<T> where T:Something`

In this PR I have purposely ignored type parameters. Any good option you are thinking of? Also, @ml86 mentioned in passing that it would probably be a good idea to serialize `T:Something` somewhere in the fullName.

----

This PR is currently parked until rust_ast_gen gets some extra features, hence the draft mode.

maltek:
> In this PR I have purposely ignored type parameters. Any good option you are thinking of?

It feels like we'd have to fully monomorphize generic functions just like the compiler does just for static dispatch to always point to the correct implementations.

For binding tables/dynamic dispatch, it all depends on what information can be gotten from rust-analyzer. I guess it would be too much to ask for a direct way to find all trait implementations that are used in any type erased context so that we could just generate the vtables that are actually needed. Does it allow querying every trait implemented by a particular concrete type? Or the inverse direction?

xavierpinho:
Can be reviewed now. This PR, together with a `CONSTRUCTOR` method for struct literals (instead of inlining their construction at the moment) got the intended dataflow tests passing (in codescience.) I'm addressing that (and the case when they are external) in a separate PR to keep commits focused. I will address generic types shortly after (also separately) too.

joern-cli/frontends/rust2cpg/src/test/scala/io/joern/rust2cpg/passes/ast/CallTests.scala
```diff
@@ -296,6 +300,44 @@ class CallTests extends Rust2CpgSuite(noSysRoot = true) {
...
+    "have the trait's methodFullName and signature" in {
+      inside(cpg.call.nameExact("m").l) { case call :: Nil =>
+        call.dispatchType shouldBe DispatchTypes.DYNAMIC_DISPATCH
+        call.methodFullName shouldBe "rust2cpgtest::Tr::m"
+        call.signature shouldBe "rust2cpgtest::Tr"
+      }
+    }
+
+    "match the trait's binding node" in {
+      inside(cpg.typeDecl.fullNameExact("rust2cpgtest::Tr").bindsOut.nameExact("m").l) { case binding :: Nil =>
```
ml86: Use `methodBinding` step instead of generated step.

joern-cli/frontends/rust2cpg/src/test/scala/io/joern/rust2cpg/passes/ast/CallTests.scala
```diff
@@ -296,6 +300,44 @@ class CallTests extends Rust2CpgSuite(noSysRoot = true) {
...
+        binding.signature shouldBe "rust2cpgtest::Tr"
+        binding.methodFullName shouldBe "rust2cpgtest::Tr::m"
+        binding.refOut.fullName.l shouldBe List("rust2cpgtest::Tr::m")
```
ml86: Is the ref edge added by the frontend? If yes please dont do that.

joern-cli/frontends/rust2cpg/src/test/scala/io/joern/rust2cpg/passes/ast/CallTests.scala
```diff
@@ -296,6 +300,44 @@ class CallTests extends Rust2CpgSuite(noSysRoot = true) {
...
+      }
+    }
+
+    "match the impl's binding node" in {
+      inside(cpg.typeDecl.fullNameExact("<rust2cpgtest::S as rust2cpgtest::Tr>").bindsOut.nameExact("m").l) {
```
ml86: We need better fullnames for the `impl foo for bar` TYPE_DECLs. This does not need to be in this PR but that is something we need to tackle sooner then later so that we do not need to adjust too much policy.

ml86 (changes requested):

ml86 (approved):

PR #6115, opened 2026-07-15
[x2cpg] Use SourceFiles.determine to find supported build files

johannescoetzee:
This PR replaces https://github.com/joernio/joern/pull/5959 using `SourceFiles.determine` as suggested by @max-leuthaeuser and filtering the results afterwards. 

It also moves the `JvmDefaultIgnoreRegex` previously used in javasrc2cpg into x2cpg to have it available for the DependencyResolver to avoid traversing these directories when looking for build files.

### DIFF
--- a/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
+++ b/joern-cli/frontends/javasrc2cpg/src/main/scala/io/joern/javasrc2cpg/JavaSrc2Cpg.scala
@@ -1,6 +1,7 @@
 package io.joern.javasrc2cpg
 
 import io.joern.javasrc2cpg.passes.{AstCreationPass, OuterClassRefPass, TypeInferencePass}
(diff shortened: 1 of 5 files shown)


max-leuthaeuser:
W.r.t. the `JvmDefaultIgnoreRegex` location. Maybe the `x2cpg.frontendspecific` package (or a `jvm` sub-package there) is a better place? Having it in `SourceFiles` feels wrong.

max-leuthaeuser (approved):

johannescoetzee:
https://buildbot.global.sltr.io/#/builders/82/builds/780 buildbot is red, so will need to look into that

ml86 (approved):

johannescoetzee:
There was a logic bug in the filter which I've fixed (also added a unit test). 

New buildbot run: https://buildbot.global.sltr.io/#/builders/82/builds/781 

~~I'll update this comment once it's done~~ This time it's green

PR #6117, opened 2026-07-16
scala 3.8.3

mpollmeier:
note: this raises the lower bound for the JDK from 11 to 17


max-leuthaeuser (approved):

maltek:
@ml86 I thought about this one moment too late... the bazel build needs to be updated in the same way

PR #6122, opened 2026-07-21
[c2cpg] Handle C++ while and for loop condition decls properly

max-leuthaeuser:
This _will_ change fingerprint hashes as we now have condition declarations for C++ while/for loops in the CPG (and it does for frepple).

### DIFF
--- a/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/ControlStructureTests.scala
+++ b/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/ControlStructureTests.scala
@@ -441,7 +441,7 @@ class ControlStructureTests extends C2CpgSuite(FileDefaults.CppExt) {
       inside(cpg.controlStructure.controlStructureType(ControlStructureTypes.WHILE).l) {
         case List(whileNode: ControlStructure) =>
           inside(whileNode.condition.l) { case List(expr: Expression) =>
-            expr.code shouldBe "const char* c = read() != 0"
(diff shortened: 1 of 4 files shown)


joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/ControlStructureTests.scala
```diff
@@ -441,7 +441,7 @@ class ControlStructureTests extends C2CpgSuite(FileDefaults.CppExt) {
       inside(cpg.controlStructure.controlStructureType(ControlStructureTypes.WHILE).l) {
         case List(whileNode: ControlStructure) =>
           inside(whileNode.condition.l) { case List(expr: Expression) =>
-            expr.code shouldBe "const char* c = read() != 0"
+            expr.code shouldBe "* c = read()" // assignment calls are no canditates for synthetic loop equal checks
```
maltek: this code field is weird - it looks like `c` gets dereferenced. Is this just the output of CDT, or is this an issue in our own processing?

maltek (approved):
more tests would be appreciated

joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/passes/ast/ControlStructureTests.scala
max-leuthaeuser: It's our own processing and should be fixed.
(It's a Delaration which can contain multiple Declarators. `const char` is not part of the declarator code fields and thus, won't show up in the assignment expression)
I'll see if i can fix that.

PR #6129, opened 2026-07-22
[csharpsrc2cpg] handle expression-bodied properties

### LINKED ISSUE #6128: [Bug] csharpsrc2cpg: expression-bodied property (`T P => expr;`) crashes pre-parse and silently drops the whole file
**Describe the bug**

`csharpsrc2cpg` fails on a **C# expression-bodied property** — `T Name => expr;` (a property whose
getter is an `=> expression` arrow clause). `astForPropertyDeclaration` unconditionally dereferences the
property's `AccessorList`, but an expression-bodied property has **no** `AccessorList` (it has an
`ExpressionBody` / `ArrowExpressionClause` instead), so the read yields `null` and `createDotNetNodeInfo`
throws `ujson.Value$InvalidData: Expected ujson.Obj (data: null)`.

This is worse than a hard crash: it is a **silent partial failure**. The exception is caught per file,
the frontend exits **rc=0**, logs a `WARN`, and **drops the entire containing `.cs` file** from the CPG
(all its `TypeDecl`s, methods, and fields vanish). Other files build normally, so the CPG *looks* valid.
Expression-bodied properties are one of the most common constructs in real C# (DTOs, wrappers, computed
properties), so this silently removes large amounts of code from analysis with no error surfaced to the
caller.

**To Reproduce**

The smallest failing input:

```csharp
// Crash.cs
class C { int N => 0; }
```

1. Run `csharpsrc2cpg /path/to/dir_with_Crash.cs -o /tmp/crash.bin --enable-file-content`
2. The process exits 0 but logs:

```
WARN  ProgramSummaryCreator$    Unable to pre-parse C# file, skipping -
java.util.concurrent.ExecutionException: ujson.Value$InvalidData: Expected ujson.Obj (data: null)
Caused by: ujson.Value$InvalidData: Expected ujson.Obj (data: null)
    at ujson.Value$InvalidData$.apply(Value.scala:202)
    at io.joern.csharpsrc2cpg.astcreation.AstCreatorHelper$.createDotNetNodeInfo(AstCreatorHelper.scala:209)
    at io.joern.csharpsrc2cpg.astcreation.AstForDeclarationsCreator.astForPropertyDeclaration(AstForDeclarationsCreator.scala:554)
    ...
WARN  AstCreationPass           Failed to generate a CPG for: '/src/Crash.cs'
```

3. `crash.bin` is a degraded (~5.9 KB) CPG with `Crash.cs` entirely absent. The semantically equivalent
   block-body getter (`class C { int N { get { return 0; } } }`) builds fine (~23 KB, file present).

Construct sweep (isolates the trigger; each construct in its own single file):

| C# construct | example | crashes? |
|---|---|:--:|
| **expression-bodied property** | `int N => 0;` | **YES** |
| expression-bodied property (field-backed) | `string N => _n;` | **YES** |
| expression-bodied property (in struct) | `struct S(int x){ int X => x; }` | **YES** |
| expression-bodied *method* | `string M(object o) => o.ToString();` | no |
| expression-bodied *accessor* | `int N { get => _n; }` | no |
| block-body getter | `int N { get { return _n; } }` | no |
| expression-bodied *indexer* | `int this[int i] => _a[i];` | no |

The crash is specific to a `PropertyDeclaration` that has an `ExpressionBody`/`ArrowExpressionClause` and
**no** `AccessorList`. Expression-bodied methods, accessors, and indexers are all fine.

**Expected behavior**

The file should build, and the expression-bodied property should be lowered into a `get_*` getter method
(as block/`=>` accessors already are), instead of crashing pre-parse and dropping the whole file.

**Root cause**

`AstForDeclarationsCreator.scala:553-557`:

```scala
protected def astForPropertyDeclaration(propertyDecl: DotNetNodeInfo): Seq[Ast] = {
  val accessorList = createDotNetNodeInfo(propertyDecl.json(ParserKeys.AccessorList)) // AccessorList is null here
  val accessors    = accessorList.json(ParserKeys.Accessors).arr.map(createDotNetNodeInfo)
  accessors.flatMap(astForPropertyAccessor(_, propertyDecl)).toList
}
```

For `int N => 0;` the parser emits `"AccessorList": null` and puts the getter in `ExpressionBody` (an
`ArrowExpressionClause`). `createDotNetNodeInfo` then does `json(ParserKeys.MetaData)` on `null`
(`AstCreatorHelper.scala:209`) → `ujson.Value$InvalidData: Expected ujson.Obj (data: null)`.

By contrast, `astForMethodDeclaration` (same file) already null-guards its body
(`if (!jsonBody.isNull) ...`), which is why expression-bodied *methods* do not crash. Property
declarations lack the equivalent guard and have no `ExpressionBody` branch.

**Suggested fix**

Null-guard the `AccessorList` read and, when it is absent, model the expression-bodied property as a
getter (reusing the existing `astForGetAccessorDeclaration`), mirroring the accessor path. A fix + tests
are prepared following the frontend conventions.

**Environment**

- Frontend: `csharpsrc2cpg`, bundled `dotnetastgen`.
- Reproduced on Joern **4.0.579** (Docker); the offending code path is byte-identical at **v4.0.583**
  (`AstForDeclarationsCreator.scala:553-557`, `AstCreatorHelper.scala:209`).
- `ujson` 4.1.0, Scala 3.7.4.

Lekssays:
Closes #6128

### DIFF
--- a/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/PropertyGetterTests.scala
+++ b/joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/PropertyGetterTests.scala
@@ -274,4 +274,76 @@ class PropertyGetterTests extends CSharpCode2CpgFixture {
       cpg.method.nameExact("get_MyProperty").body.astChildren shouldBe empty
     }
   }
+
+  "expression-bodied property declaration" should {
+    val cpg = code("""
+        |class C
+        |{
+        |  public int MyProperty => 42;
+        |}
+        |""".stripMargin)
+
+    "be lowered into a get_* method" in {
+      inside(cpg.method.nameExact("get_MyProperty").l) { case method :: Nil =>
+        method.fullName shouldBe "C.get_MyProperty:System.Int32(C)"
+        method.signature shouldBe "System.Int32(C)"
+      }
+    }
+
+    "have correct modifiers" in {
+      cpg.method.nameExact("get_MyProperty").modifier.modifierType.sorted.l shouldBe List(ModifierTypes.PUBLIC)
+    }
+
+    "have correct parameters" in {
+      inside(cpg.method.nameExact("get_MyProperty").parameter.l) { case thisParam :: Nil =>
+        thisParam.typeFullName shouldBe "C"
+        thisParam.name shouldBe "this"
+      }
+    }
+  }
+
+  "static expression-bodied property declaration" should {
+    val cpg = code("""
+        |public class C
+        |{
+        |  public static string MyProperty => "x";
+        |}
+        |""".stripMargin)
+
+    "be lowered into a get_* method" in {
+      inside(cpg.method.nameExact("get_MyProperty").l) { case method :: Nil =>
+        method.fullName shouldBe "C.get_MyProperty:System.String()"
+        method.signature shouldBe "System.String()"
+      }
+    }
+
+    "have correct modifiers" in {
+      cpg.method.nameExact("get_MyProperty").modifier.modifierType.sorted.l shouldBe List(
+        ModifierTypes.PUBLIC,
+        ModifierTypes.STATIC
+      )
+    }
+
+    "have no parameters" in {
+      cpg.method.nameExact("get_MyProperty").parameter shouldBe empty
+    }
+  }
+
+  "a type declaring an expression-bodied property alongside other members" should {
+    val cpg = code("""
+        |class C
+        |{
+        |  public int N => 0;
+        |  public void Run() {}
+        |}
+        |""".stripMargin)
+
+    // Regression: an expression-bodied property used to throw `Expected ujson.Obj (data: null)` during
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/csharpsrc2cpg/src/test/scala/io/joern/csharpsrc2cpg/querying/ast/PropertyGetterTests.scala
```diff
@@ -274,4 +274,76 @@ class PropertyGetterTests extends CSharpCode2CpgFixture {
...
+  "a type declaring an expression-bodied property alongside other members" should {
+    val cpg = code("""
+        |class C
+        |{
+        |  public int N => 0;
+        |  public void Run() {}
+        |}
+        |""".stripMargin)
+
+    // Regression: an expression-bodied property used to throw `Expected ujson.Obj (data: null)` during
```
max-leuthaeuser: LLMs like to talk about the previous state in code comments a lot but its not needed here. Just remove it.

max-leuthaeuser (approved):

PR #6130, opened 2026-07-22
Same Scala and JDK target versions for SBT and Bazel build.

ml86:
Recently we bumped the Scala version to 3.8.3 for apparently now urgent
need. I would like to avoid that as long as we have a dual build setup
with SBT and Bazel. Thus reverting back to 3.7.4.
To have the JDK target version identical I bumped it to 17 in the Bazel
build as we decided a while ago to go 17 anyway.

### DIFF
--- a/bazel/java.bzl
+++ b/bazel/java.bzl
@@ -4,7 +4,7 @@ _rules = make_java_rules(
     java_common_opts = [
         "-g",
         "-Xlint",
-        "--release=11",
+        "--release=17",
     ],
     common_java_binary_runtime_deps = [
         # We bring in the linked version of codepropertygraph domain classes as default/common
--- a/bazel/java.bzl
+++ b/bazel/java.bzl
@@ -4,7 +4,7 @@ _rules = make_java_rules(
     java_common_opts = [
         "-g",
         "-Xlint",
-        "--release=17",
+        "--release=11",
     ],
     common_java_binary_runtime_deps = [
         # We bring in the linked version of codepropertygraph domain classes as default/common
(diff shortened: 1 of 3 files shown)


ml86:
We dont download the JDK in production in case none is found and
this JDK is a JDK 11. So we need to stick to that version for now.

maltek:
scala 3.8 finally fixed these stupid warnings:

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by scala.runtime.LazyVals$ (file:/Users/gacevedo/.shiftleft/jssrc2cpg-5.0.107/lib/org.scala-lang.scala3-library_3-3.7.4.jar)
WARNING: Please consider reporting this to the maintainers of class scala.runtime.LazyVals$
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
````

This was backported to 3.3 (LTS release) behind a compiler flag, but I don't think it was to 3.7. So by going back, we will subject our customers to these scary warnings again.


> We dont download the JDK in production in case none is found and
this JDK is a JDK 11. So we need to stick to that version for now.

That's incorrect. We have an automatic download of the right JRE/JDK version, as specified in the manifests in our x2cpg-internal repo.

ml86:
@maltek Thanks for clarifying all of that. For reference the PR that brought in 3.8.3 is https://github.com/joernio/joern/pull/6117.
@mpollmeier Please in the future mention the reason why a change happens in the commit/PR.

Since we already have configured JDK 17 for all our frontends, we dont seem to have a problem, so I will close this PR and just adjust the Bazel build.

ml86:
@maltek What is missing is an update of the documentation. [Here](https://docs.shiftleft.io/sast/analyzing-applications/java#java-requirements) is one example but I guess there are more.

PR #6134, opened 2026-07-23
[rust2cpg] lower `match` expressions.

xavierpinho:


max-leuthaeuser (approved):

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
```diff
@@ -1290,4 +1291,45 @@ trait RustVisitor(implicit withSchemaValidation: ValidationMode) { this: AstCrea
...
+  //    JUMP_TARGET
+  //    BLOCK {
+  //      <lowerPatternMatch(pat, tmp)>
+  //      body
+  //    }
+  //    ...
+  //  }
+  // }
+  private def visitMatchExpr(matchExpr: MatchExpr): Ast = {
+    val tmpName      = "tmp"
```
maltek, line 1312: doesn't have to be fixed immediately... but x2cpg has helpers to generate names for temporary variables that don't clash with each other or with actual variable names of the program (the latter mostly by using illegal characters like `<>` in the variable name)

PR #6139, opened 2026-07-24
[CI] Switch to platform-specific Github releases

max-leuthaeuser:
Our Github releases (`joern-cli.zip`) was sitting slightly above 2GB in total size which is not allowed (https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#storage-and-bandwidth-quotas).

With platform-specific releases we are now a bit below that 2GB limit (e.g., for `joern-cli-macos-arm64.zip` with 1.7GB).


ml86 (approved):
Please also update the readme and docs.

PR #6141, opened 2026-07-24
[CI] Update actions/upload-artifact and actions/download-artifact

max-leuthaeuser:
Old one are Node.js 20 based. That's scheduled for removal.

Also fix querydb paths.


maltek (approved):
this is unrelated to this PR - it just reminds me of it: generally it's better to use `action@<git-commit-hash>` instead of `action@<git-tag>` since tags are mutable, and can be changed to do anything when an upstream's SSH credentials are compromised.

PR #6157, opened 2026-07-29
[joern-console] Fix NPE in importCode caused by REPL classloader

TNSelahle:
Scala 3.8.X's REPL interrupt instrumentation re-defines all classes through its own classloader without a `ProtectionDomain`, causing `getCodeSource().getLocation()` to return null. InstallConfig.rootPath relied on this to find the installation directory, resulting in a `NullPointerException` on `importCode`:

```scala
joern> importCode.ruby("/tmp/some/source")
java.lang.NullPointerException: Cannot invoke "java.net.URL.toURI()" because the return value of "java.security.CodeSource.getLocation()" is null
  at io.joern.console.InstallConfig.rootPath$lzyINIT1(ConsoleConfig.scala:29)
  at io.joern.console.InstallConfig.rootPath(ConsoleConfig.scala:25)
  at io.joern.console.cpgcreation.ImportCode$Frontend.apply(ImportCode.scala:131)
  ... 39 elided

```

This fix walks the classloader chain to find a parent that still has the jar's `CodeSource`.

### DIFF
--- a/console/src/main/scala/io/joern/console/ConsoleConfig.scala
+++ b/console/src/main/scala/io/joern/console/ConsoleConfig.scala
@@ -30,7 +30,7 @@ class InstallConfig(environment: Map[String, String] = sys.env) {
       val pathToLibDir = findCodeSourceLocation(clazz, clazz.getClassLoader).map(_.getParent)
       val cwd          = FileUtil.currentWorkingDirectory
       pathToLibDir.flatMap(findRootDirectory(_)).orElse(findRootDirectory(cwd)).getOrElse {
-        val searchedDirs = pathToLibDir.map(p => s"$p and $cwd").getOrElse(cwd.toString)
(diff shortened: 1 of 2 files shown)


maltek (approved):
I ran into another class-loading problem downstream in ocular (https://github.com/ShiftLeftSecurity/codescience/pull/9035) - I think your investigation results now give an explanation why things broke there :)

max-leuthaeuser:
We should add a test for that in `testDistro.py` so something like this does not slip through again in the future.

TNSelahle:
> We should add a test for that in `testDistro.py` so something like this does not slip through again in the future.

@max-leuthaeuser good idea :+1: . No elegant way to test this but I've added in something. Please take a look

max-leuthaeuser:
Thanks, great!
Let's see what the CI thinks about it.

edit: looks like its working 💯

max-leuthaeuser (approved):

PR #6165, opened 2026-08-03
[rust2cpg] add rust-src to CI and re-enable flaky test.

xavierpinho:
rust-analyzer implicitly installs the `rust-src` component if it's not already installed. The hypothesis is that this flakiness results from a race-condition. Thus, installing it before running any tests.


xavierpinho:
Turns out the issue was with Windows paths. Fixable using https://crates.io/crates/dunce. Closing this. Will bump rust_ast_gen and re-enable the test by then.

PR #6169, opened 2026-08-04
[CI] fix Dockerfile.slim (from platform releases)

max-leuthaeuser:
Fails otherwise: https://github.com/joernio/joern/actions/runs/30885968206/job/91917137806

From: https://github.com/joernio/joern/pull/6139

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -12,7 +12,6 @@ jobs:
   deploy:
     if: github.repository_owner == 'joernio'
     runs-on: ubuntu-latest
-    continue-on-error: true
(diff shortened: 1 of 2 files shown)


xavierpinho (approved):
gnarly, and I'm a bit out of the loop as to why "suddenly" all these changes are needed. Maybe here or a comment somewhere would be great for others too.

max-leuthaeuser:
> gnarly, and I'm a bit out of the loop as to why "suddenly" all these changes are needed. Maybe here or a comment somewhere would be great for others too.

You are right. I added a link to the initial PR.

PR #6172, opened 2026-08-04
[rust2cpg] lower `let-else` statements.

xavierpinho:


joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
```diff
@@ -292,6 +293,40 @@ trait RustVisitor(implicit withSchemaValidation: ValidationMode) { this: AstCrea
     }
   }
 
+  // `let pat = expr else { body };` becomes:
+  // LOCAL tmp
+  // <createLocalsForBindings(pat)>
+  // tmp = expr
+  // IF (UNKNOWN) {
```
max-leuthaeuser, line 300: Shouldn't that be something like `IF (tmp != None)` or better `IF (tmp.is_some())`?

xavierpinho: It's not really a nil check... perhaps rust2cpg should have a dedicated `<operator>.patternMatches` (or something) for pattern matching conditions, like the `<operator>.repeatInArray` we introduced in #6021. In `for x in y` I also lowered the conditional pattern matching of `x` with an item of `y` as an unknown for the time being.

> or better IF (tmp.is_some())?

Only applicable to `Option<T>` types. The general case at play is destructuring an enum over its constructors. `Option<T>` would have `None` and `Some(T)`, and `is_some()` is specific to `Option`. Other enums, other considerations.

max-leuthaeuser (approved):

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/RustVisitor.scala
xavierpinho: I should have used samples other than `Option` in the tests! They make it look like `Option` is the only interesting type. I'm currently working on `if let p = e`, which is similar to this, and shall include test samples with other types.

PR #6175, opened 2026-08-05
[linter-rules] add UnorderedIteration scalafix rule

max-leuthaeuser:
Adds a new semantic Scalafix rule, `UnorderedIteration`, that warns about order-sensitive operations (`.foreach`, `.map`, `.toList`, `.find`, `.head`, `.sum`, `.minBy`, `.take`, ...) on collections with undefined iteration order (`mutable`/`immutable` `HashMap`/`HashSet`, `AnyRefMap`, `LongMap`, `WeakHashMap`, `OpenHashMap`, `CollisionProofHashMap`, `TrieMap`, `immutable.IntMap`/`LongMap`, Java `HashMap`/`HashSet`/`Hashtable`/`IdentityHashMap`/`WeakHashMap`, `ConcurrentHashMap`, and priority queues — `mutable.PriorityQueue`, `java.util.PriorityQueue`, `PriorityBlockingQueue`, `DelayQueue` — whose iterators traverse in heap order). Insertion- or sort-ordered types (`LinkedHashMap`, `TreeMap`, ...) are deliberately excluded, as are order-independent queries (`.contains`, `.size`, `.get`, `.exists`, `.forall`, `.count`).

Detection combines SemanticDB type information (declared types and method return types) with a same-file, symbol-keyed fixpoint taint analysis over `val`s (constructor/companion-call initialization, aliases, projections, and order-preserving derivations like `it.take(3)`), and covers direct receivers, one-hop projections (`.values.foreach`, `.keys.toList`, `.asScala.map`), infix syntax (`set foreach println`), and for-comprehension generators. Symbol-keying prevents cross-scope false positives from same-named vals. Materialized derivatives are reported once at the materialization call (`set.toList`) without re-flagging downstream uses.

Escape hatch: suppression uses the idiomatic Scalafix mechanism `// scalafix:ok UnorderedIteration` on the same line.

The rule is wired into the CI linting step with `--diff-base`, so only PR-changed files are checked and existing violations are grandfathered. This requires enabling SemanticDB generation build-wide (`-Xsemanticdb` on all modules), which adds modest compile-time overhead and makes the CI linting job compile the affected modules; a one-time full-repo audit (`sbt "scalafix UnorderedIteration"` without `--diff-base`) can be run to quantify existing violations.

### Changes

- New semantic rule class `fix.UnorderedIteration` (extends `SemanticRule`, `isLinter = true`), registered in `META-INF/services/scalafix.v1.Rule`
- New `linter-rules/input` sbt project hosting Scalafix-testkit input fixtures, compiled with the same Scala 3 version as the build itself, so the test suite runs on the same SemanticDB flavor the rule sees in production/CI; deliberately neither aggregated into root nor published (`publish / skip := true`) so it can never be released
- Build changes: SemanticDB enabled build-wide (`ThisBuild / semanticdbEnabled`), sbt-scalafix bumped 0.14.2 → 0.14.7 (required for the newer scalameta API used by the rule); the rule depends on `scalafix-core` via `CrossVersion.for3Use2_13` (scalafix-core is not published for Scala 3 — the 2.13 artifact is the canonical rule API) and tests use the Scala 3 build of `scalafix-testkit`
- CI: `UnorderedIteration` added to the scalafix linting step in `.github/workflows/pr.yml`; `.scalafmt.conf` excludes the testkit input directory to protect line-anchored assert comments
- `SingleLetterIdentifiers` diagnostic id renamed to `SingleLetterIdentifierRule` for consistency (no in-repo suppressions reference the old id)

### Tests Added

All tests live in the Scalafix testkit suite `fix.UnorderedIterationTests`, driven by assertions embedded in the input file `UnorderedIterationTestCases.scala` (that's the way the scalafix guys test their stuff themselves):

- Direct order-sensitive calls on concrete unordered receivers are flagged: `directForeach`, `directMap`, `directToList`, `directHead`, `directFind`, `directSum`, `directMinBy`, `directDropWhile`, `setForeach`, `setToSeq`, `setTakeWhile`, `immutableForeach`, `immutableHashSetForeach`
- Inline chains report once at the order-preserving call without a second diagnostic on the chained terminal op: `inlineFilterForeach` (`hashMap.filter(p).foreach(f)` flags only `filter`)
- Type-inferred locals are flagged via SemanticDB: `typeInferred`
- Abstract-typed vals are flagged when initialized with a recognizable companion call in the same file: `abstractValForeach`
- Abstract-typed `var`s are not tracked (reassignment may change the runtime type): `abstractVarForeach` (negative); a `var` with a concrete unordered declared type is still flagged via its static type: `concreteVarForeach`
- Member vals are flagged via their declared type alone when the RHS carries no usable type information: `staticallyTypedForeach`
- Cross-scope name-collision regression: a `List` val must not be flagged merely because an unrelated same-named val elsewhere is HashMap-backed: `ScopeA` / `ScopeB`
- One-hop projections are flagged, including the `MapOps` iterator projections: `valuesProjection`, `keySetProjection`, `keysProjection`, `keysIteratorProjection`, `valuesIteratorChain` (the latter via val-chain taint)
- Java collections via `asScala` are flagged, including raw constructors without type arguments at the new-site: `javaForeach`, `rawJavaForeach`
- Further unordered types are flagged: `collisionProofForeach` (hash-based with treeified bins), `intMapForeach`, `longMapForeach` (unsigned-trie order), `priorityQueueForeach`, `javaPriorityQueueForeach`, `priorityBlockingQueueForeach`, `delayQueueForeach` (heap order)
- For-comprehension generators, including over projections, are flagged: `forComprehension`, `forComprehensionProjection`
- Infix syntax is flagged: `infixForeach`, `infixMap`
- Same-file val-chain taint tracking: projection-derived local `chainedIterator`, alias `aliasChain`, order-preserving derivation `takeChain`, multi-hop derivation `multiHopChain`, infix derivation `infixDerivation`, parameterless derivation `distinctPropagates` (`distinct` is not flagged at the call site — its result contents do not depend on encounter order — but propagates taint into derived vals)
- Materialization is flagged once at the call and not downstream: `materializedNotTracked` (negative on the downstream use)
- Ordered derivations and aliases of ordered vals are not flagged: `orderedChain`, `orderedAlias` (negative)
- Method receivers returning an unordered type are flagged, including parameterless defs (by-name return type in Scala 3 SemanticDB) and parameters: `methodReceiver`, `parameterReceiver`
- Order-independent queries are not flagged: `safeContains`, `safeSize`, `safeExists`, `safeGet` (negative)
- The suppression escape hatch works: `suppressedIteration` (negative)


linter-rules/src/test/scala/fix/UnorderedIterationTests.scala
```diff
@@ -0,0 +1,7 @@
+package fix
+
+import scalafix.testkit.AbstractSemanticRuleSuite
+
+class UnorderedIterationTests extends AbstractSemanticRuleSuite {
+  runAllTests()
+}
```
ml86, line 12: Relation to actual test code is not understandable. I guess test code is in `UnorderedIterationInput.scala`? 
1. Can we rename that file so that it indicate that it contains tests.
2. Can we make the relation to the file explicit?

linter-rules/src/main/scala/fix/UnorderedIteration.scala
```diff
@@ -0,0 +1,345 @@
...
+  private val propagatesOnlyOps: Set[String] = Set("distinct")
+
+  // The call sites flagged by this rule: operations whose results embed encounter order or
+  // carry order-dependent side effects. All order-preserving ops are flagged as well — the
+  // derived contents or their encounter order depend on the source order — minus the
+  // propagate-only exceptions above.
+  // Suppression via `// scalafix:ok UnorderedIteration` is available for documented exceptions.
+  private val flaggedOperations: Set[String] =
+    materializingOps ++ reducingOps ++ selectingOps ++ regroupingOps ++ traversingOps ++
+      (orderPreservingOps -- propagatesOnlyOps)
```
ml86: What is the point of having these individual groups? This just makes it hard in the future if a someone wants to add a function to the blacklist and is now wondering in which bucked do I need to put it. The fact that we then even have subtractions makes it even more complicated.

linter-rules/src/test/scala/fix/UnorderedIterationTests.scala
max-leuthaeuser: Yes, thats the exact same test style as the scalafix guys are using. So I kept it. I check whether we can make this more explicit.

linter-rules/src/main/scala/fix/UnorderedIteration.scala
max-leuthaeuser: Understandable. I'll rework that.

ml86 (approved):

PR #6188, opened 2026-08-11
[x2cpg] AstGenRunner hardening

max-leuthaeuser:
`AstGenRunner`:
- New explicit contract: `runAstGenNative` returns `ExternalCommandResult` (both streams kept separate) instead of a lossy `Try[Seq[String]]`.
- New `SkippedFile(fileName, reason)` case class replaces bare filename strings.
- New hooks: `isSuccess`, `astGenOutputFiles`, `postProcessParsedFiles`, `logUnsuccessfulRun`, `collectSkippedFiles` (defensive, logs uniformly), `toRelativeInputPath` (handles symlinked tmp dirs), and combine for multi-invocation frontends.
- `execute()` is now uniform across all frontends: it always scans the output directory for partial results, even after a failed/thrown run, and always attempts to collect skipped files — six of seven frontends previously discarded everything on failure.

Per-frontend fixes:
- jssrc2cpg: fixed the dead skipped-file parser (was reading stdout, but skip lines are on stderr) — now correctly detects broken files via safe pattern matching.
- swiftsrc2cpg: moved the Windows exit-code workaround into `isSuccess`/`logUnsuccessfulRun`; replaced `substring`/`indexOf` parsing with a safe regex
- gosrc2cpg: fixed a bug where "Failed to generate AST for X" was mis-parsed as filename "Failed"; wired skipped files into Report.
- csharpsrc2cpg: fixed a real misattribution bug where compiler errors got attached to the wrong file under interleaved output — now buffers reasons until the terminating line names the file; wired into Report.
- rubysrc2cpg: same interleaving hardening pattern; `runAstGenNative` now returns a real `ExternalCommandResult` from the in-process JRuby run; added a Report (previously had none).
- abap2cpg / rust2cpg: adapted to the new contract; both gained a Report (abap2cpg had none before) that now records skipped files.

### DIFF
--- a/joern-cli/frontends/abap2cpg/src/main/scala/io/joern/abap2cpg/Abap2Cpg.scala
+++ b/joern-cli/frontends/abap2cpg/src/main/scala/io/joern/abap2cpg/Abap2Cpg.scala
@@ -5,6 +5,7 @@ import io.joern.abap2cpg.passes.{AbapTypeInferencePass, AstCreationPass, RefEdge
 import io.joern.x2cpg.{X2Cpg, X2CpgFrontend}
 import io.joern.x2cpg.passes.base.{ContainsEdgePass, TypeEvalPass}
 import io.joern.x2cpg.passes.frontend.{MetaDataPass, TypeNodePass}
+import io.joern.x2cpg.utils.Report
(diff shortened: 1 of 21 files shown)


max-leuthaeuser:
This is still prone to problems and complicated. Will have to do it differently.

PR #6196, opened 2026-08-13
[SBT] Unify test fork settings

max-leuthaeuser:
~~For whatever reasons these~~ Because of global state in the past three frontends still had fork off.

- For swiftsrc2cpg this brings a full test run down to 80 seconds (was: 4 minutes) locally on my machine. The CI here should also benefit.
- No frontend doing a hard `exit(1)` can crash the sbt shell anymore.
- Now its explicitly documented (in the SBT file) if a frontend is not thread-safe (jimple with soot and gidhra)

The actual PR that brought that in was: https://github.com/joernio/joern/pull/2180
But what was actually missing is:
`ThisBuild / Test / testForkedParallel := true`
With that, forking + testsuite-parallelism works.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -119,7 +119,10 @@ ThisBuild / resolvers ++= Seq(
   "Google Maven" at "https://maven.google.com/"
 )
 
-ThisBuild / Test / fork := true
+ThisBuild / Test / fork               := true
+ThisBuild / Test / testForkedParallel := true
+
+trapExit := false
 
 Global / onChangedBuildSource := ReloadOnSourceChanges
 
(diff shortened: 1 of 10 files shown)


mpollmeier (approved):
I'm not quite sure, it's been a while, but I think the reason was that they had some globally shared cache, and/or some tests failed when being forked. 

Either way, if forked tests are green *and* they are faster (please double check, that's not guaranteed) then yes, go ahead and merge.

max-leuthaeuser:
Right, they had indeed global state which we got rid of using `io.shiftleft.passes.ForkJoinParallelCpgPassWithAccumulator`.

Locally, most tests are (much) faster (swiftsrc2cpg benefits the most), some are the same as before. Lets see what the CI tells us. Which won't help too much as the runner performance is rather jumpy anyway.

mpollmeier:
👍🏻 👍🏻

PR #6197, opened 2026-08-14
[rubysrc2cpg] bump ruby_astgen

TNSelahle:
- Bump ruby_astgen to bring in change that allows fallback to the old parser if the prism parser fails to load


max-leuthaeuser (approved):

TNSelahle:
Closed while we figure out the real issue

PR #6204, opened 2026-08-18
[rust2cpg] support or-patterns.

xavierpinho:
Matching an expression `expr` against a pattern `pat1 | pat2` becomes:

```
LOCAL tmp = expr
IF UNKNOWN(pat1) {
  // lower pat1 against tmp
} 
ELSE { 
  IF UNKNOWN(pat2) {
    // lower pat2 against tmp
  }
}
```


johannescoetzee (approved):
This seems fine as an initial solution, although I'm generally wary of adding unknown nodes to the CPG as they could affect downstream passes.

In javasrc2cpg, we lower pattern matches to `instanceof` calls in most cases, so I'm wondering if something similar could work here. Rust patterns do seem much more complicated than java patterns, however, so it probably won't be as simple (especially considering the `let <pat>` pattern)

xavierpinho:
> In javasrc2cpg, we lower pattern matches to `instanceof` calls in most cases, so I'm wondering if something similar could work here. Rust patterns do seem much more complicated than java patterns, however, so it probably won't be as simple

Indeed, I've been thinking of adding a custom `<operator>.patternMatches` (or something similarly named) to replace those, cf. https://github.com/joernio/joern/pull/6172#discussion_r3711534920. But before even attempting we also need to include the type information inside patterns in the astgen.

PR #6209, opened 2026-08-20
[rust2cpg] enum typeFullName from astgen.

xavierpinho:
The bump disambiguates local enums and includes generic type parameters.


TNSelahle (approved):
some timeout in CI but LGTM

PR #6213, opened 2026-08-21
[pysrc2cpg] Change <metaClassCallHandler> and <fakeNew> lowering.

ml86:
Move call to `__init__` function from `<fakeNew>` to `<metaClassCallHandler>`
to reflect what CPython implementation is doing.
This results in dataflows from and to `__init__` that do not have
`<fakeNew>` on the path anymore which always was an ugly artifact.
It also saves a few expansion steps by not going via
`__init__<metaClassAdapter>` since we now call `__init__` directly on the
new instance obtained by `<fakeNew>`.

### DIFF
--- a/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/ClassCpgTests.scala
+++ b/joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/ClassCpgTests.scala
@@ -1,7 +1,7 @@
 package io.joern.pysrc2cpg.cpg
 
 import io.joern.pysrc2cpg.testfixtures.PySrc2CpgFixture
-import io.shiftleft.codepropertygraph.generated.nodes
+import io.shiftleft.codepropertygraph.generated.{DispatchTypes, nodes}
 import io.shiftleft.semanticcpg.language.*
 
 class ClassCpgTests extends PySrc2CpgFixture(withOssDataflow = false) {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/ClassCpgTests.scala
```diff
@@ -73,6 +73,38 @@ class ClassCpgTests extends PySrc2CpgFixture(withOssDataflow = false) {
...
+      val fakeNewCall = newInstanceAssign.argumentOption(2).isCall.head
+      fakeNewCall.receiver.code.head shouldBe "class Foo<meta>(...).<fakeNew>"
+      fakeNewCall.argument(0).code shouldBe "class Foo<meta>(...)"
+      fakeNewCall.argument(1).code shouldBe "x"
+
+      val initCall = handlerMethod.topLevelExpressions.isCall.l(1)
+      initCall.code shouldBe "__newInstance.__init__(x)"
+      initCall.receiver.code.head shouldBe "__newInstance.__init__"
+      initCall.argument(0).code shouldBe "__newInstance"
+      initCall.argument(1).code shouldBe "x"
```
maltek, line 102: maybe also assert that this is a dynamic call

maltek (approved):

joern-cli/frontends/pysrc2cpg/src/test/scala/io/joern/pysrc2cpg/cpg/ClassCpgTests.scala
ml86: Added test for this.

PR #6218, opened 2026-08-24
[pysrc2cpg] Handle `__call__` operator.

ml86:
For `__call__` operator methods we now add a binding into the
corresponding classes type decl in addition to the already existing
handling.


maven_install.json
maltek, line 1: this looks like a merge conflict or accidental commit?

maltek (approved):

maven_install.json
ml86: Yes, did not want to commit that. Corrected.

PR #6224, opened 2026-08-26
[rust2cpg] fix macro call expr returning multiple statements.

xavierpinho:


joern-cli/frontends/rust2cpg/src/test/scala/io/joern/rust2cpg/passes/ast/MacroTests.scala
```diff
@@ -99,6 +99,45 @@ class MacroTests extends Rust2CpgSuite(noSysRoot = true) {
...
+        sLocal.name shouldBe "s"
+        sLocal.typeFullName shouldBe "i32"
+      }
+    }
+
+    "have correct assignments" in {
+      cpg.method.nameExact("main").block.assignment.sortBy(_.order).code.l shouldBe List(
+        "let x = { three!() };",
+        "let mut s = 1;",
+        "s+=2"
```
TNSelahle, line 123: Is it expected for whitespace to be missing here, considering it's in the source?

xavierpinho: Good catch! Unfortunately, yes (expected). Since it's a macro-expanded AST, the code/text comes from the astgen, which has whitespace removed...

Another instance: https://github.com/joernio/joern/blob/1065fbbbf97459d81e25a914d05845ccce86c88c/joern-cli/frontends/rust2cpg/src/test/scala/io/joern/rust2cpg/passes/ast/MacroTests.scala#L84

TNSelahle (approved):

PR #6228, opened 2026-08-27
docs: fix typo writen -> written

vaibhav8a:
One-word typo fix in `semanticcpg/src/main/scala/io/shiftleft/semanticcpg/accesspath/AccessPathAlgebra.md:25`.


max-leuthaeuser:
Should be done with a more general cleanup of that file.

PR #6232, opened 2026-08-31
[all] standardize per-file processing log messages across all frontends

max-leuthaeuser:
Unifies how all frontends log per-file processing results, which was inconsistent in level, phrasing, and quoting.

**Standard applied across all 13 frontends:**
- Success: `log.debug(s"Processed '$file'")`
- Failure: `log.warn(s"Failed to process '$file'", throwable)`

**Specific changes:**
- Replace inaccurate `"Generated a CPG for"` phrasing (each pass contributes a diffgraph, not a full CPG)
- Downgrade success logs from `info` → `debug` where applicable (gosrc2cpg, csharpsrc2cpg)
- Downgrade per-file failure logs from `error` → `warn` (rubysrc2cpg, php2cpg) — these are all recoverable
- Add missing success/failure log lines where absent (kotlin2cpg, jimple2cpg, rubysrc2cpg, pysrc2cpg, javasrc2cpg, php2cpg)
- Replace backtick-quoted file paths with single-quoted paths (kotlin2cpg)
- Pass throwable as second argument instead of inlining `exception.getMessage` in the string

Partially for QT-1348.

### DIFF
--- a/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/passes/AstCreationPass.scala
+++ b/joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/passes/AstCreationPass.scala
@@ -75,8 +75,13 @@ class AstCreationPass(
         accumulator.samInfoEntriesByImplClass,
         disableFileContent
       )
-    diffGraph.absorb(astCreator.createAst())
-    logger.debug(s"AST created for file at `${fileWithMeta.f.getVirtualFilePath}`.")
+    try {
+      diffGraph.absorb(astCreator.createAst())
+      logger.debug(s"Processed '${fileWithMeta.f.getVirtualFilePath}'")
+    } catch {
+      case e: Exception =>
+        logger.warn(s"Failed to process '${fileWithMeta.f.getVirtualFilePath}'", e)
+    }
   }
 
 }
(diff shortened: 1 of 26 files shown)


mpollmeier (approved):
👍🏻 things start to makes sense now, thank you

joern-cli/frontends/kotlin2cpg/src/main/scala/io/joern/kotlin2cpg/passes/AstCreationPass.scala
```diff
@@ -75,8 +75,13 @@ class AstCreationPass(
         accumulator.samInfoEntriesByImplClass,
         disableFileContent
       )
-    diffGraph.absorb(astCreator.createAst())
-    logger.debug(s"AST created for file at `${fileWithMeta.f.getVirtualFilePath}`.")
+    try {
+      diffGraph.absorb(astCreator.createAst())
+      logger.debug(s"Processed '${fileWithMeta.f.getVirtualFilePath}'")
+    } catch {
+      case e: Exception =>
```
maltek, line 82: always use `NonFatal(e)` over plain `Exception` or `Throwable` (unless the caught exception is rethrown immediately after)

edit: oh, too late

max-leuthaeuser: Right, but I need to do that separately. I think most other frontends have the same issue.

PR #6237, opened 2026-09-01
joern-scan: propagate the JVM exit status instead of the retry test's

Portll:
The script's last command is `if [ $? -eq 2 ]`, so the script exits with the status of the IF,
not of the scan. When the JVM exits 2 the retry runs as intended; every other non-zero exit is
reported to the caller as 0.

That means Joern currently gives a non-zero exit for an unrecognised option, a dead frontend and a
generated script that fails to compile, and surface as success.

A caller in CI cannot distinguish 'scanned, found nothing' from 'never scanned', because both
are exit 0 and the second also produces an artifact (the log banner).

PR approach: Capture the status before the test, refresh it after the retry, and exit with it.

### Reproducer

```sh
printf '#include <string.h>\nvoid f(int c,char**v){char b[8];if(c>1)strcpy(b,v[1]);}\n' > /tmp/src/v.c
joern-scan /tmp/src --overwrite --frontend-args --exclude-regex '.*/\.git/.*'
echo $?     # 0, and the output is the "Writing logs to:" banner alone
```

The scan never ran — the generated script failed to compile (that is a separate defect, see the
companion PR #6238 ) — but the caller is told it succeeded.

### Verified

Against joern 4.0.610 (Homebrew, macOS arm64), patching the installed launcher in place:

| case | before | after |
|---|---|---|
| scan that fails | exit 0, 41-byte artifact | **exit 1** |
| healthy scan | exit 0, 2 findings | **exit 0, 2 findings** |

`sh -n` passes on the patched script.

### DIFF
--- a/joern-cli/src/universal/joern-scan
+++ b/joern-cli/src/universal/joern-scan
@@ -15,6 +15,10 @@ fi
 
 echo "Writing logs to: /tmp/joern-scan-log.txt"
 $SCRIPT -J-XX:+UseG1GC -J-XX:CompressedClassSpaceSize=128m -Dlog4j.configurationFile="$SCRIPT_ABS_DIR"/conf/log4j2.xml -J-XX:+UseStringDeduplication "$@" 2> /tmp/joern-scan-log.txt
-if [ $? -eq 2 ]; then
+status=$?
+if [ $status -eq 2 ]; then
   $SCRIPT -J-XX:+UseG1GC -J-XX:CompressedClassSpaceSize=128m -Dlog4j.configurationFile="$SCRIPT_ABS_DIR"/conf/log4j2.xml -J-XX:+UseStringDeduplication "$@" 2> /tmp/joern-scan-log.txt
+  status=$?
 fi
+
+exit $status
--- a/testDistro.py
+++ b/testDistro.py
@@ -235,6 +235,16 @@ class TestRunner:
             if result_count == 0:
                 raise RuntimeError("Result from joern-scan not found")
 
+            # A scan that failed must not reach the caller as a success. The launcher used to exit
+            # with the status of its retry test rather than the JVM's, so every non-zero exit
+            # except 2 arrived as 0 and "scanned, found nothing" was indistinguishable from
+            # "never scanned".
+            missing_path = Path(tmp_dir) / "no-such-directory"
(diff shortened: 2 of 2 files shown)


max-leuthaeuser:
Thanks for the contribution!

Could you please also add this as a test case to https://github.com/joernio/joern/blob/master/testDistro.py#L204?

Portll:
Added in b424bdb — `scan_test` now scans a path that does not exist and requires a non-zero exit.

Every scan the stage ran was one that succeeds, so the launcher could report any non-zero JVM status except 2 as 0 and the distribution test still passed. A nonexistent input path is a failure that does not depend on a frontend, a query bundle or a plugin, so it should be stable across the matrix.

Verified by running the real `scan_test` against the released distribution (joern 4.0.610, macOS arm64) with `joern-scan` pointed at each launcher in turn:

| launcher | outcome |
|---|---|
| this branch | stage passes |
| master | `Joern-scan exited 0 for a scan that failed` |

Two caveats worth stating rather than leaving you to find:

- I could not exercise this on Windows. `joern-scan.bat` is a different file and this PR does not touch it; it invokes `bin\joern-scan.bat` without `call`, so the child's exit code should become the process's, but CI will be the first to actually check that.
- **The PR description above is wrong on one point.** I wrote that an unrecognised option currently gives a non-zero exit. It does not — `joern-scan --not-an-option` exits 0 at the JVM level too, so this change alone does not cover it. The cause is separate from the launcher: `JoernScan.main` does `optionParser.parse(scanArgs, JoernScanConfig()).foreach { ... }`, and a parse failure returns `None`, which `.foreach` silently discards after scopt has already printed `Error: Unknown option`. A missing source path behaves the same way for a different reason — `src` is `.optional()`, so it is not a parse error, and `runScanPlugin` prints the usage text and returns.

I have deliberately left both of those out of this PR to keep it to the shell script. Happy to open a follow-up for them if you would like.

Portll:
Follow-up on the second caveat above: filed as #6241 with the fix in #6242. #6242 is stacked on this branch, because the test for it asserts that a refused invocation reaches the *caller* as non-zero, which needs this launcher change as well as the Scala one.

max-leuthaeuser (approved):

PR #6238, opened 2026-09-01
console: escape frontend args interpolated into the generated script

Portll:
`argsStringFromConfig` inserts each `--frontend-args` value into a generated Scala source with bare
quotes and no escaping. `loadOrCreateCpg`, in the same file, already escapes `src` with
`StringEscapeUtils.escapeJava` for the same reason.

Any value containing a `\` or a double quote therefore produces an invalid string literal,
and the generated script fails to COMPILE. A regex is the common case:

    joern-scan <src> --frontend-args --exclude-regex '.*/\.git/.*'

`\.` is not a legal Scala escape sequence. The scan never starts and the output is the log
banner alone. The same flag passed straight to the frontend works, so this is not a frontend
limitation:

    c2cpg.sh <src> -o out.cpg --exclude-regex '.*node_modules/.*'   # excludes correctly

`StringEscapeUtils` is already imported in this file.

### Reproducer

Same source, same flag, differing only in whether the value contains a backslash:

```sh
joern-scan /tmp/src --overwrite --tags badfn --frontend-args --exclude-regex 'plainnobackslash'
#   -> 2 results

joern-scan /tmp/src --overwrite --tags badfn --frontend-args --exclude-regex 'has\.backslash'
#   -> 0 results; /tmp/joern-scan-log.txt shows
#      "error during script execution: Error during compilation"
```

### The flag itself is fine

Passed straight to the frontend it works, which is what makes this a `joern-scan` bug.

```sh
c2cpg.sh /tmp/src -o /tmp/out.cpg --exclude-regex '.*node_modules/.*'   # excludes correctly
```

Note the failure is currently invisible: the generated script's non-zero exit is reported as 0 by
`joern-cli/src/universal/joern-scan`. This is independent of the companion PR.

### DIFF
--- a/console/src/main/scala/io/joern/console/BridgeBase.scala
+++ b/console/src/main/scala/io/joern/console/BridgeBase.scala
@@ -439,8 +439,11 @@ trait PluginHandling { this: BridgeBase =>
     config.frontendArgs match {
       case Array() => ""
       case args =>
+        // Escaped as a Scala string literal, exactly as `src` is in loadOrCreateCpg above:
+        // these values are interpolated into a generated source file, so an unescaped backslash
+        // or quote yields an invalid literal and the script fails to compile rather than run.
         val quotedArgs = args.map { arg =>
-          "\"" ++ arg ++ "\""
+          "\"" ++ StringEscapeUtils.escapeJava(arg) ++ "\""
         }
         val argsString = quotedArgs.mkString(", ")
         s", args=List($argsString)"
--- a/console/src/main/scala/io/joern/console/BridgeBase.scala
+++ b/console/src/main/scala/io/joern/console/BridgeBase.scala
@@ -438,7 +438,7 @@ trait PluginHandling { this: BridgeBase =>
   private def argsStringFromConfig(config: Config): String = {
     config.frontendArgs match {
       case Array() => ""
-      case args =>
+      case args    =>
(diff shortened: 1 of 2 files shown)


max-leuthaeuser:
Thanks for the contribution!

Could you please also add this as a test case to https://github.com/joernio/joern/blob/master/testDistro.py#L204?

Portll:
Added in a99edf9 — `scan_test` now repeats the `uaf.c` scan with `--frontend-args --exclude-regex '.*/\.git/.*'` and requires the same result line. The exclusion matches nothing under the input, so the only variable is whether the value survives into a script that compiles.

**Verified in the failing direction only.** There is no sbt on this machine, so I could run the new assertion against the released distribution (which doesn't carry this fix) but not against a build that does. Against joern 4.0.610 (macOS arm64):

- with the backslash: 0 results, and `/tmp/joern-scan-log.txt` shows `error during script execution: Error during compilation`
- the same flag with a backslash-free value (`--exclude-regex nobackslashhere`): 1 result, i.e. the expected line

So the test fails today for the reason this PR fixes, and passes whenever the interpolated value happens to already be a legal literal. I have not watched it go green on a patched build; CI here will be the first to do that. (If it comes back red, that's the real answer.)

One detail I tripped over is that the invocation puts the source path **before** `--frontend-args`. With it after, the separator swallows the path, and joern-scan prints its usage text and exits 0 instead of scanning.

max-leuthaeuser (approved):

PR #6242, opened 2026-09-02
joern-scan: exit non-zero when the arguments are rejected

### LINKED ISSUE #6241: joern-scan exits 0 when it refuses the arguments or is given no source path
## Summary

`joern-scan` exits 0 when it refuses to run. Two separate routes reach the same outcome, and in both the tool has already printed a message saying it will not scan.

**1. An argument list scopt rejects.** [`JoernScan.main`](https://github.com/joernio/joern/blob/master/joern-cli/src/main/scala/io/joern/joerncli/JoernScan.scala#L50-L55) is:

```scala
optionParser.parse(scanArgs, JoernScanConfig()).foreach { config =>
  run(config, frontendArgs)
}
```

`parse` returns `None` on a rejected argument list. `.foreach` on `None` does nothing, `main` returns normally, and the JVM exits 0 — after scopt has printed `Error: Unknown option --x` / `Try --help for more information.`

**2. No source path.** `src` is declared `.optional()`, so its absence is not a parse error. It reaches [`runScanPlugin`](https://github.com/joernio/joern/blob/master/joern-cli/src/main/scala/io/joern/joerncli/JoernScan.scala#L155-L158), which prints the usage text and returns:

```scala
if (config.src == "") {
  println(optionParser.usage)
  return
}
```

This one is easy to hit by accident rather than by typo: `--frontend-args` consumes everything after it, so putting the source path last silently leaves the scan with no input.

```sh
joern-scan --overwrite --tags all --frontend-args --exclude-regex 'x' /tmp/src
#   -> usage text, exit 0. The path became a frontend argument.
```

## Reproducer

Measured against joern 4.0.610 (Homebrew, macOS arm64, JDK 25):

```sh
joern-scan --this-option-does-not-exist /tmp/src ; echo $?
#   stderr (in /tmp/joern-scan-log.txt): Error: Unknown option --this-option-does-not-exist
#   -> 0

joern-scan --tags all ; echo $?
#   stdout: the usage text
#   -> 0
```

Both were run with the launcher already carrying the fix in #6237, which propagates the JVM's status, so the 0 is the JVM's own and not the launcher discarding it. The unknown-option case was additionally confirmed by invoking `libexec/bin/joern-scan` directly, bypassing the launcher entirely. #6237 is a necessary but not sufficient condition for a caller to ever see either of these.

## Why it matters

Same reason as #6237: a caller in CI cannot distinguish "scanned, found nothing" from "never scanned". Route 2 is the worse of the two, because the invocation is well-formed and the only signal is usage text on stdout that a non-interactive caller does not read.

## Suggested fix

`System.exit(1)` in both places — exit 2 is already taken by the query-database install path that the launcher retries on, so 1 is free.

## Disclosure

I do not have sbt on this machine, so **I have not compiled or run the patch** — the diff is written from reading the source, and its behaviour after the change is an expectation, not a measurement. Everything stated above about the *current* behaviour is measured. CI on the PR will be the first real check of the fix.

Portll:
Fixes #6241.

`joern-scan` exits 0 when it refuses to run, by two routes:

- `JoernScan.main` did `optionParser.parse(...).foreach { ... }`. A rejected argument list gives `None`, `.foreach` discards it, and the process exits 0 — after scopt has printed `Error: Unknown option --x`.
- `src` is `.optional()`, so an absent source path is not a parse error. It reaches `runScanPlugin`, which prints the usage text and returns. Also 0.

`System.exit(1)` in both places. Exit 2 is already spoken for by the query-database install path the launcher retries on, so 1 is free.

### Rebased onto master — no longer stacked

This was opened stacked on #6237, which has since merged. It is now rebased onto master and stands alone: **two commits, two files**, and the launcher change that used to appear in the diff is gone because it is upstream.

The test in `testDistro.py` needed #6237 to be meaningful, since it asserts that a refused invocation reaches *the caller* as non-zero. With #6237 merged that condition is satisfied and the assertion is live rather than pending.

### Measured

Against joern 4.0.610 (Homebrew, macOS arm64, JDK 25), with #6237's launcher in place so the JVM's status is not being discarded:

```sh
joern-scan --this-option-does-not-exist /tmp/src ; echo $?   # -> 0
joern-scan --tags all                           ; echo $?   # -> 0, usage text on stdout
```

The first was also confirmed straight against `libexec/bin/joern-scan`, bypassing the launcher, so the 0 is the JVM's own.

The second is reachable by accident rather than by typo: `--frontend-args` consumes everything after it, so a source path written last silently becomes a frontend argument and the scan runs with no input.

### Not measured — please read this before merging

**There is no sbt on this machine, so I have not compiled this change.** The Scala diff is written from reading the source; that it behaves as described afterwards is an expectation, not something I watched happen. Everything asserted about the *current* behaviour is measured on the released distribution.

`scalafmt --test` passes under the version `.scalafmt.conf` pins (3.8.1), run with the real formatter rather than matched by hand — the formatting failure on #6238 is what taught me not to trust the latter.

### DIFF
--- a/joern-cli/src/main/scala/io/joern/joerncli/JoernScan.scala
+++ b/joern-cli/src/main/scala/io/joern/joerncli/JoernScan.scala
@@ -49,8 +49,9 @@ object JoernScan extends BridgeBase {
 
   def main(args: Array[String]): Unit = {
     val (scanArgs, frontendArgs) = CpgBasedTool.splitArgs(args)
-    optionParser.parse(scanArgs, JoernScanConfig()).foreach { config =>
-      run(config, frontendArgs)
+    optionParser.parse(scanArgs, JoernScanConfig()) match {
+      case Some(config) => run(config, frontendArgs)
+      case None         => System.exit(1)
     }
   }
 
@@ -154,7 +155,7 @@ object JoernScan extends BridgeBase {
   private def runScanPlugin(config: JoernScanConfig, frontendArgs: List[String]): Unit = {
     if (config.src == "") {
       println(optionParser.usage)
-      return
+      System.exit(1)
     }
 
     if (queryNames().isEmpty) {
--- a/testDistro.py
+++ b/testDistro.py
@@ -257,6 +257,16 @@ class TestRunner:
             if proc.returncode == 0:
                 raise RuntimeError("Joern-scan exited 0 for a scan that failed")
 
+            proc = self.run_command([self.joern_scan_exe, "--this-option-does-not-exist", str(foo_path)],
+                           "Joern-scan with an unrecognised option", check=False)
+            if proc.returncode == 0:
+                raise RuntimeError("Joern-scan exited 0 for an argument list it rejected")
+
+            proc = self.run_command([self.joern_scan_exe, "--tags", "all"],
+                           "Joern-scan with no source path", check=False)
+            if proc.returncode == 0:
+                raise RuntimeError("Joern-scan exited 0 when given no source path")
+
     @stage
     def slice_test(self):
         """Test slicing functionality"""


joern-cli/src/main/scala/io/joern/joerncli/JoernScan.scala
```diff
@@ -49,8 +49,11 @@ object JoernScan extends BridgeBase {
 
   def main(args: Array[String]): Unit = {
     val (scanArgs, frontendArgs) = CpgBasedTool.splitArgs(args)
-    optionParser.parse(scanArgs, JoernScanConfig()).foreach { config =>
-      run(config, frontendArgs)
+    // A rejected argument list yields None, and discarding it exits 0: scopt has printed the
+    // reason, but the caller is told the scan succeeded.
```
max-leuthaeuser: Do not talk about the past / previous implementation.

joern-cli/src/main/scala/io/joern/joerncli/JoernScan.scala
```diff
@@ -152,9 +155,12 @@ object JoernScan extends BridgeBase {
   }
 
   private def runScanPlugin(config: JoernScanConfig, frontendArgs: List[String]): Unit = {
+    // `src` is optional, so an absent source is not a parse error and lands here instead. Printing
+    // the usage text and returning made "you gave me nothing to scan" indistinguishable from a
+    // clean scan.
```
max-leuthaeuser: Do not talk about the past / previous implementation.

Portll:
Rebased onto master now that #6237 has merged. The stack is gone: this is two commits and two files, and the launcher change that used to show in the diff is upstream. GitHub had it as CONFLICTING; it now reports MERGEABLE.

Also stripped the explanatory comments from both hunks, and restored two comment blocks in testDistro.py that belong to #6237 and #6238 — the strip had caught those as well, which would have had this PR quietly delete four lines it does not claim. testDistro.py is now purely additive, +10/-0.

joern-cli/src/main/scala/io/joern/joerncli/JoernScan.scala
Portll: remediated

joern-cli/src/main/scala/io/joern/joerncli/JoernScan.scala
Portll: remediated

Portll:
updated PR and rebased onto main.

max-leuthaeuser (approved):

PR #6248, opened 2026-09-02
[rubysrc2cpg] use surrounding type for member parent

TNSelahle:
- use surrounding type name for member parent
- fixes a bug where a member doesn't have a TYPE_DECL as the parent

### DIFF
--- a/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
+++ b/joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
@@ -407,6 +407,23 @@ class MethodTests extends RubyCode2CpgFixture {
 
     foo.definingTypeDecl.map(_.name) shouldBe Option("C")
     foo.astParent shouldBe cpg.typeDecl("C").head
+
+    val fooMember = cpg.member("foo").head
+    fooMember.typeDecl.name shouldBe "C"
+  }
+
+  "Singleton method with unresolvable target nested inside a method should have its member under the enclosing type decl" in {
+    val cpg = code("""
+        |class C
+        | def outer
+        |  def something.foo
+        |  end
+        | end
+        |end
+        |""".stripMargin)
+
+    val fooMember = cpg.member("foo").head
+    fooMember.typeDecl.name shouldBe "C"
   }
 
   "A Boolean method" should {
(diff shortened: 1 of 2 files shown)


joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
```diff
@@ -407,6 +407,23 @@ class MethodTests extends RubyCode2CpgFixture {
...
+        |class C
+        | def outer
+        |  def something.foo
+        |  end
+        | end
+        |end
+        |""".stripMargin)
+
+    val fooMember = cpg.member("foo").head
+    fooMember.typeDecl.name shouldBe "C"
```
ml86: This does not seem correct. `foo` is added to whatever `something` is. If we cannot get that from the scope then we should add it as if it were free standing but adding it to `C` is wrong.

ml86 (changes requested):

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
TNSelahle: Thanks @ml86 . These singleton methods are actually quite tricky. For now, the member just has `<unresolvedNamespace>` as the ast parent if it can't be found, which would be the case for singleton methods (excluding class methods)

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
```diff
@@ -407,6 +412,26 @@ class MethodTests extends RubyCode2CpgFixture {
...
+    val fooMember = cpg.member("foo").head
+    fooMember.astParentFullName shouldBe "<unresolvedNamespace>"
+    fooMember.astParentType shouldBe NodeTypes.TYPE_DECL
+  }
+
+  "Singleton method with unresolvable target nested inside a method should have a dangling member" in {
+    val cpg = code("""
+        |class C
+        | def outer
+        |  something = ""
```
ml86, line 445: Please also add a test case where `something` is not defined at all.

joern-cli/frontends/rubysrc2cpg/src/test/scala/io/joern/rubysrc2cpg/querying/MethodTests.scala
```diff
@@ -407,6 +412,26 @@ class MethodTests extends RubyCode2CpgFixture {
...
+        |  def something.foo
+        |  end
+        | end
+        |end
+        |""".stripMargin)
+
+    val fooMember = cpg.member("foo").head
+    fooMember.astParentFullName shouldBe "<unresolvedNamespace>"
+    fooMember.astParentType shouldBe NodeTypes.TYPE_DECL
   }
```
ml86, line 460: Please also test for existence of the `foo` method and where that is in the graph.

ml86 (approved):

PR #6262, opened 2026-09-07
[c2cpg] Harden CDT type resolution: fold expr crash + StackOverflowError

### LINKED ISSUE #6260: c2cpg crashes on C++ fold expression: StringIndexOutOfBoundsException in typeForCPPASTFoldExpression
When indexing LibreOffice core with Joern 4.0.621, c2cpg skips shell/source/win32/spsupp/COMOpenDocuments.cxx.

The file contains a C++ fold expression. While building the type name,
c2cpg calls String.substring with an invalid range [0, -1).

Environment
- Joern 4.0.621 (latest at the time)
- OS: Windows
- Frontend: c2cpg
- JDK: Temurin 21
- Command: c2cpg.bat -J-Xmx12G -o cpg.bin <libreoffice-core>

Source (LibreOffice core)
https://github.com/LibreOffice/core/blob/master/shell/source/win32/spsupp/COMOpenDocuments.cxx

### LINKED ISSUE #6261: c2cpg skips file on bundled CDT StackOverflowError in CPPVisitor.createType
When indexing LibreOffice core with Joern 4.0.621, c2cpg skips
svx/source/dialog/langbox.cxx.

Eclipse CDT (bundled with Joern) hits a StackOverflowError while
resolving types. The log points at range-based for / auto typing.

Environment
- Joern 4.0.621
- Bundled CDT: io.joern.eclipse-cdt-core-9.2.100.202507101054+1
- OS: Windows
- Frontend: c2cpg
- JDK: Temurin 21
- Command: c2cpg.bat -J-Xmx12G -o cpg.bin <libreoffice-core>

Source (LibreOffice core)
https://github.com/LibreOffice/core/blob/master/svx/source/dialog/langbox.cxx

max-leuthaeuser:
Fixes two c2cpg issues reported while indexing LibreOffice core, both causing whole files to be dropped from the CPG:

* **#6260** — `typeForCPPASTFoldExpression` assumed `EvalBinary.toString` contains `": "` (only true when a nested leaf evaluation is an `EvalFixed`, e.g. `bool: true`). For folds over non-constant evaluations (e.g. the lambda calls in `(quote(GetHelperExe()) + ... + (L" " + quote(args)))`) the separator is absent → `StringIndexOutOfBoundsException: Range [0, -1)`. Now guarded, degrading to `ANY`.

* **#6261** — CDT's `auto` type deduction runs into unbounded mutual recursion (`CPPVisitor.createType → createAutoType → getAutoInitClauseForRangeBasedFor → EvalBinding.getType → CPPVariable.getType → …` → `StackOverflowError`) when a structured binding introduced by a range-based for loop is itself used as the range of a nested range-based for loop and the container type cannot be resolved. Minimal reproducer (no includes needed):

```cpp
void f() {
  Map m;
  for (const auto& [k, v] : m)
    for (auto& e : v)
      ;
}
```

  `StackOverflowError` is a `VirtualMachineError` which `scala.util.Try`/`NonFatal` do not catch, so the existing `safeGet*` wrappers did not help. Added `safeCdtCall` (catches SOE explicitly, degrades to `None`) and routed all direct CDT type/binding/evaluation resolution calls through it; `typeFor` now degrades to `ANY` on SOE instead of losing the whole file.

Note: the recursion itself is upstream Eclipse CDT (`io.joern:eclipse-cdt-core` is a vanilla re-publish, not a patchable fork).

Closes #6260
Closes #6261

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
@@ -236,7 +236,7 @@ trait AstForExpressionsCreator { this: AstCreator =>
             astForCppCallExpressionUntyped(call)
         }
       case Some(classType: ICPPClassType) if safeGetEvaluation(call).exists(_.isInstanceOf[EvalFunctionCall]) =>
-        val evaluation        = safeGetEvaluation(call).get.asInstanceOf[EvalFunctionCall]
+        val evaluation        = call.getEvaluation.asInstanceOf[EvalFunctionCall]
         val functionType      = safeCdtCall(evaluation.getOverload.getType)
         val functionSignature = functionType.map(functionTypeToSignature).getOrElse(X2CpgDefines.UnresolvedSignature)
         val name              = Defines.OperatorCall
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
@@ -170,8 +170,8 @@ trait AstForExpressionsCreator { this: AstCreator =>
 
   private def astForCppCallExpression(call: ICPPASTFunctionCallExpression): Ast = {
     val functionNameExpr = call.getFunctionNameExpression
-    Try(functionNameExpr.getExpressionType).toOption match {
-      case Some(_: IPointerType) => createPointerCallAst(call, safeGetType(call.getExpressionType))
+    safeCdtCall(functionNameExpr.getExpressionType) match {
+      case Some(_: IPointerType) => createPointerCallAst(call, safeGetExpressionType(call))
       case Some(functionType: ICPPFunctionType) =>
         functionNameExpr match {
           case idExpr: CPPASTIdExpression if safeGetBinding(idExpr).exists(_.isInstanceOf[ICPPFunction]) =>
@@ -198,7 +198,7 @@ trait AstForExpressionsCreator { this: AstCreator =>
               fullName,
               DispatchTypes.STATIC_DISPATCH,
               Some(signature),
-              Some(registerType(safeGetType(call.getExpressionType)))
+              Some(registerType(safeGetExpressionType(call)))
             )
             val args = call.getArguments.toList.map(a => astForNode(a))
             createCallAst(callCpgNode, args)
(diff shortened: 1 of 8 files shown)


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
```diff
@@ -229,15 +229,15 @@ trait AstForExpressionsCreator { this: AstCreator =>
...
+              Some(registerType(safeGetExpressionType(call)))
             )
             createCallAst(callCpgNode, args, base = Some(instanceAst), receiver)
           case _ =>
             astForCppCallExpressionUntyped(call)
         }
       case Some(classType: ICPPClassType) if safeGetEvaluation(call).exists(_.isInstanceOf[EvalFunctionCall]) =>
-        val evaluation        = call.getEvaluation.asInstanceOf[EvalFunctionCall]
-        val functionType      = Try(evaluation.getOverload.getType).toOption
+        val evaluation        = safeGetEvaluation(call).get.asInstanceOf[EvalFunctionCall]
```
maltek: this `Option.get` doesn't make this any less safe than it was before?

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
```diff
@@ -266,12 +284,13 @@ trait TypeNameProvider { this: AstCreator =>
     val x = safeGetNodeType(name)
     safeGetBinding(name) match {
       case Some(v: IVariable) =>
-        v.getType match {
-          case f: IFunctionType =>
+        safeCdtCall(v.getType) match {
```
maltek: this whole function is wrapped in a `try...catch StackOverFlowError`, so we should just use `v.getType` here directly.

maltek (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/AstForExpressionsCreator.scala
max-leuthaeuser: Right, the guard already established that the evaluation is an EvalFunctionCall, so the Option.get added no safety over the direct call. Reverted to call.getEvaluation directly in c3f1411de.

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/astcreation/TypeNameProvider.scala
max-leuthaeuser: Good point - typeForIASTName is only called from within typeFor, which already catches StackOverflowError. Reverted to using v.getType directly in c3f1411de.

PR #6282, opened 2026-09-11
[c2cpg] HeaderFileFinder: select closest header with minByOption instead of sortBy.headOption

### LINKED ISSUE #6280: [c2cpg] HeaderFileFinder.find full-sorts the candidate list to take one minimum: 8x CPG build time on RIOT-OS
## Describe the bug

`HeaderFileFinder.find` selects the closest-matching header with `sortBy(...).headOption` — a full
sort to take a single minimum — and the Levenshtein key is computed *inside* the comparator, so it is
re-evaluated on every comparison instead of once per candidate.

Building a CPG for RIOT-OS takes **1717.5s** stock and **211.9s** with the three-line change below —
**8.1x**, same input, same machine, two builds of the same commit differing only by that change. A
second pair built with instrumentation gave 1523s and 198s (7.7x), so run-to-run variance is on the
order of 10% and the effect is around 8x rather than exactly 8.1x.

Instrumenting `find` shows why: the sort performs **25,235,778** Levenshtein evaluations over this
tree against **3,114,815** for the replacement, an 8.10x reduction that tracks the wall-clock change.

### This completes #6112 rather than contradicting it

PR #6112 ("[c2cpg] JVM/Scala performance & memory fixes", merged 2026-07-15) already found this line
— its change list includes "Cache Levenshtein sort results in `HeaderFileFinder`" — and added the
`findCache` `ConcurrentHashMap` around the body, leaving the sort itself unchanged.

That memoisation helps repeated lookups of the *same* unresolved include string. It cannot help a
tree that presents many *distinct* ones, because each still pays a full sort on first touch. Release
v4.0.614 shipped 2026-08-28, six weeks after #6112 merged, so this is not a regression against that
work — the cache is present and the cost is still there.

## Environment

- Joern built from source at **v4.0.624** (`844cb4eaf3dbcc8c32c31c1de8aeab389035bd1b`)
- OpenJDK 21, `-J-Xmx8g`, Ubuntu 24.04 LTS, Linux aarch64
- No compilation database, so the `HeaderFileFinder` fallback is the path exercised
- Both binaries built from that commit; verified in the bytecode that stock's `HeaderFileFinder`
  references `sortBy` and not `minByOption`, and patched the reverse, each calling
  `Levenshtein.distance` once

All timings are single runs unless stated.

## The code

`joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/parser/HeaderFileFinder.scala`

```scala
def find(path: String): Option[String] = findCache.computeIfAbsent(
  path,
  unresolvedPath =>
    Paths.get(unresolvedPath).nameOption.flatMap { name =>
      val matches = nameToPathMap.getOrElse(name, List())
      matches.sortBy(candidate => Levenshtein.distance(candidate, unresolvedPath)).headOption
    }
)
```

`nameToPathMap` groups the scanned tree's headers by basename, so `matches` is "every file in the
tree called `foo.h`". The key is `org.jline.utils.Levenshtein.distance` over two absolute filesystem
paths — a dynamic-programming edit distance that allocates a table per call.

`sortBy(f)` is `sorted(ord on f)`, and `Ordering#on` builds a comparator whose `compare(x, y)`
evaluates `f(x)` and `f(y)`, so the key is recomputed at every comparison rather than once per
element. That is visible in the counts below.

### 1. RIOT-OS

```bash
#!/bin/bash
set -u
STOCK=${JOERN_STOCK:-}; PATCHED=${JOERN_PATCHED:-}
[ -n "$STOCK" ] || { echo "set JOERN_STOCK to a joern-parse binary" >&2; exit 1; }
WORK=${WORK:-$PWD/riot-repro}; mkdir -p "$WORK"; SRC="$WORK/RIOT"
[ -d "$SRC/.git" ] || git clone https://github.com/RIOT-OS/RIOT "$SRC" || exit 1
( cd "$SRC" && git checkout -q 1f39b3ea7370cb47d3d2c6d6415a39804f3f3971 )

EXCL='^(tests|examples|dist|fuzzing|pkg)/.*'
run() {
  local label="$1" bin="$2" t0 t1 rc
  [ -x "$bin" ] || { echo "$label: not executable: $bin" >&2; return 1; }
  t0=$(date +%s)
  "$bin" -J-Xmx8g "$SRC" --output "$WORK/riot-$label.cpg.bin" --frontend-args \
      --enable-file-content --exclude-regex "$EXCL" > "$WORK/$label.log" 2>&1
  rc=$?; t1=$(date +%s)
  [ $rc -eq 0 ] || { echo "$label FAILED rc=$rc"; tail -5 "$WORK/$label.log"; return 1; }
  printf '%-8s %6ss   cpg=%s bytes\n' "$label" "$((t1-t0))" "$(wc -c < "$WORK/riot-$label.cpg.bin")"
}
run stock "$STOCK"
[ -n "$PATCHED" ] && run patched "$PATCHED"
```

| build | wall clock | CPG bytes |
|---|---:|---:|
| stock v4.0.624 | **1717.5s** | 69,006,174 |
| patched | **211.9s** | 69,005,011 |

The 1,163-byte difference is not the patch. The CPG file is not byte-reproducible — two consecutive
**stock** runs on the same tree also differ in size and md5 — so I compared the graphs instead. They
are identical:

| | stock | patched |
|---|---:|---:|
| nodes | 1,744,050 | 1,744,050 |
| methods | 43,044 | 43,044 |
| calls | 383,982 | 383,982 |
| files | 6,292 | 6,292 |
| literals | 160,243 | 160,243 |
| identifiers | 257,988 | 257,988 |

```scala
// joern --script, once per CPG, with CPGP set to it
importCpg(sys.env("CPGP"))
println("nodes="      + cpg.all.size      + " methods="   + cpg.method.size +
        " calls="     + cpg.call.size     + " files="     + cpg.file.size +
        " literals="  + cpg.literal.size  + " idents="    + cpg.identifier.size)
```

RIOT-OS triggers this because it carries one configuration header per supported board. Measured over
the file set that command parses — 1,770 `.c` and 4,383 `.h` after the exclusions, since
`--exclude-regex` narrows `nameToPathMap` too:

| basename | candidates C | include directives naming it |
|---|---:|---:|
| `board.h` | 232 | 580 |
| `periph_conf.h` | 270 | 287 |
| `periph_cpu.h` | 58 | 416 |
| `adc.h` | 31 | 345 |
| `pm.h` | 33 | 300 |

Those `.c` and `.h` files issue **28,830 `#include` directives naming 2,877 distinct basenames**, of
which **1,997 are present in the scanned tree** (the rest are system or toolchain headers, which have
no candidates at all). Of the 1,997 present, **78% are unique** — median C = 1, p99 = 24, max = 270.

Weighted by use rather than by name, though, **55.4% of the 28,830 directives name a basename with
C > 1** and 4.5% name one with C > 50. The median candidate list is short; the lookups that actually
happen are concentrated on the long ones.

### 2. Synthetic, controlled: stock vs patched at the same C

Everything is held constant and both binaries run over the *same* generated tree, so both open the
same files and build the same graph; only the selection differs. Run the script once per binary.

| C (headers sharing the requested basename) | stock | patched | speedup |
|---:|---:|---:|---:|
| 1 | 3.7s | 3.7s | **1.00x** |
| 100 | 8.3s | 4.3s | 1.93x |
| 400 | 23.9s | 6.0s | 3.98x |
| 800 | 44.8s | 8.1s | 5.53x |

The C=1 row is the safety check: with the guard in the patch, key evaluations are identical (0 vs 0,
see below) and wall clock is indistinguishable at this precision.

<details><summary>generator — run once with JOERN_PARSE=stock, once with JOERN_PARSE=patched</summary>

```bash
#!/bin/bash
set -u
J=${JOERN_PARSE:-}
[ -n "$J" ] || { echo "set JOERN_PARSE to a joern-parse binary" >&2; exit 1; }
W=${W:-/tmp/hffrepro}; M=${M:-2000}; TOTAL_HDRS=${TOTAL_HDRS:-800}; WANT=shared_config.h

gen() {
  local c="$1" root="$2" i d
  rm -rf "$root"; mkdir -p "$root"
  for i in $(seq 1 "$c"); do
    d=$(printf '%s/include/platform/family_%04d/config' "$root" "$i"); mkdir -p "$d"
    printf '#ifndef SHARED_CONFIG_H\n#define SHARED_CONFIG_H\n#define SHARED_VALUE %d\n#endif\n' \
      "$i" > "$d/$WANT"
  done
  local filler=$(( TOTAL_HDRS - c ))            # `seq 1 0` is not portable; guard the empty case
  if [ "$filler" -gt 0 ]; then
    for i in $(seq 1 "$filler"); do
      d=$(printf '%s/include/support/module_%04d' "$root" "$i"); mkdir -p "$d"
      printf '#ifndef S_%04d\n#define S_%04d\nint support_fn_%04d(int x);\n#endif\n' \
        "$i" "$i" "$i" > "$d/support_$(printf '%04d' "$i").h"
    done
  fi
  for i in $(seq 1 "$M"); do
    d=$(printf '%s/src/subsystem/component_%04d/impl' "$root" "$i"); mkdir -p "$d"
    printf '#include "%s"\nint component_fn_%04d(int x) { return x + SHARED_VALUE; }\n' \
      "$WANT" "$i" > "$d/impl.c"
  done
}

mkdir -p "$W"
printf '%6s %10s\n' C seconds
for C in 1 100 400 800; do
  gen "$C" "$W/t$C"
  t0=$(date +%s)
  "$J" -J-Xmx8g "$W/t$C" --output "$W/out$C.cpg.bin" --frontend-args --enable-file-content \
      > "$W/c$C.log" 2>&1
  rc=$?; t1=$(date +%s)
  [ $rc -eq 0 ] || { echo "C=$C FAILED rc=$rc"; tail -5 "$W/c$C.log"; exit 1; }
  printf '%6s %9ss\n' "$C" "$((t1-t0))"
  rm -rf "$W/t$C" "$W/out$C.cpg.bin"
done
```

</details>

### 3. Key evaluations, counted directly

`joern --script`, with the key wrapped in a counter, using the same `org.jline.utils.Levenshtein`
over absolute-path-shaped strings. Counts only — timings from a single un-warmed REPL pass would be
biased toward whichever construct runs first, so they are not reported here.

| C | `sortBy(f).headOption` | `minByOption(f)` | guarded (the patch) | same result |
|---:|---:|---:|---:|:--|
| 0 | 0 | 0 | 0 | yes |
| 1 | **0** | **1** | **0** | yes |
| 2 | 2 | 2 | 2 | yes |
| 3 | 4 | 3 | 3 | yes |
| 16 | 30 | 16 | 16 | yes |
| 100 | 686 | 100 | 100 | yes |
| 400 | 3,012 | 400 | 400 | yes |
| 800 | 6,160 | 800 | 800 | yes |
| 1600 | 12,518 | 1,600 | 1,600 | yes |

The C=1 row is why the patch carries a length guard: `sorted` currently short-circuits at length 1,
so a unique basename costs zero evaluations today, and a bare `minByOption` would cost one.

<details><summary>script</summary>

```scala
import org.jline.utils.Levenshtein
var calls = 0
def counted(a: String, b: String): Int = { calls += 1; Levenshtein.distance(a, b) }
val target = "/project/src/subsystem/component_0001/impl/shared_config.h"

def row(c: Int): Unit = {
  val cands = (1 to c).map(i => f"/project/include/platform/family_$i%04d/config/shared_config.h").toList
  calls = 0; val viaSort = cands.sortBy(x => counted(x, target)).headOption; val s = calls
  calls = 0; val viaMin  = cands.minByOption(x => counted(x, target));       val m = calls
  calls = 0
  val viaGuarded = cands match {
    case Nil           => None
    case single :: Nil => Some(single)
    case ms            => ms.minByOption(x => counted(x, target))
  }
  println(f"C=$c%-5d sortBy=$s%-6d minByOption=$m%-5d guarded=$calls%-5d " +
          f"agree=${viaSort == viaMin && viaSort == viaGuarded}")
}
List(0, 1, 2, 3, 4, 8, 16, 50, 100, 400, 800, 1600).foreach(row)
```

</details>

### 4. Where the time goes, measured inside `find`

The A/B above shows the change is responsible for the difference, but not that the sort is where the
time went. So I built a third binary from the same commit with counters inside `HeaderFileFinder`
and a runtime switch between the two strategies — both arms therefore carry identical
instrumentation overhead — and ran it over the same RIOT tree.

Identical in both arms, as they must be, since the patch changes how the minimum is found and not
what gets looked up:

| | |
|---|---:|
| `find` calls | 1,289,924 |
| cache misses (bodies executed) | 714,487 |
| cache hits (#6112's `findCache`) | 575,437 (44.6%) |
| sum of candidate-list lengths over misses | 3,583,187 |
| mean candidate list | 5.02 |
| largest candidate list | 270 |
| lookups whose basename is absent from the tree | 13,695 |

Differing between arms:

| | `sortBy` | guarded `minByOption` | ratio |
|---|---:|---:|---:|
| **Levenshtein evaluations** | **25,235,778** | **3,114,815** | **8.10x** |
| wall clock, this instrumented build | 1523s | 198s | 7.69x |

The evaluation counts are the reliable figure: they are exact and unaffected by scheduling. I also
recorded aggregate time inside the body (6,082s vs 933s), but that is the sum of per-invocation
durations across ~20 concurrent ForkJoinPool workers, so it exceeds wall clock and is inflated by
contention — it works out to 241 microseconds per evaluation, which is not the cost of a Levenshtein
over two paths. I am reporting it for completeness, not as an attribution.

Two things worth drawing out.

**The number of lookups is not the number of `#include` directives.** RIOT's sources contain 28,830
include directives, but `find` is called **1,289,924** times, because CDT probes an include from
several candidate locations and does so per translation unit. Any estimate built from directive
counts — including one I made before measuring — is short by more than an order of magnitude.

**`findCache` is working, and it is not sufficient.** 44.6% of calls are served from #6112's cache.
The remaining 714,487 still execute the body, because a tree this size presents that many *distinct*
unresolved paths.

Candidate lists actually encountered, over the 714,487 misses:

| candidates C | lookups |
|---|---:|
| 0 (basename absent from the tree) | 13,695 |
| 1 | 468,372 |
| 2 | 77,342 |
| 3–4 | 68,670 |
| 5–8 | 37,901 |
| 9–16 | 25,977 |
| 17–32 | 6,077 |
| 33–64 | 11,015 |
| 65–128 | 0 |
| 129–256 | 2,343 |
| 257–270 | 3,095 |

**65.6% of misses have exactly one candidate.** That is why the patch needs its length guard: a bare
`minByOption` would have added 468,372 Levenshtein evaluations on this tree alone. The cost lives in
the tail instead — the 5,438 lookups with C > 128 account for between 1.10M and 1.44M of the 3.58M
candidates examined (31–40%), from bucket bounds.

As an internal check on the instrumentation, the guarded arm's evaluation count should be exactly the
sum of candidate-list lengths minus the single-candidate lookups the guard short-circuits:
3,583,187 − 468,372 = 3,114,815, which is the measured figure to the digit.

## Expected behavior

Selecting one minimum should evaluate the key once per candidate rather than once per comparison.

## Suggested fix

```scala
def find(path: String): Option[String] = findCache.computeIfAbsent(
  path,
  unresolvedPath =>
    Paths.get(unresolvedPath).nameOption.flatMap { name =>
      nameToPathMap.getOrElse(name, List()) match {
        case Nil           => None
        case single :: Nil => Some(single)
        case matches       => matches.minByOption(candidate => Levenshtein.distance(candidate, unresolvedPath))
      }
    }
)
```

The `minByOption` is where the speedup comes from. The guard is cheap insurance rather than a source
of speedup: without it, the C=1 case would go from zero key evaluations to one, and unique basenames
are 78% of the header basenames in this tree. With the guard the change is a strict improvement or an
exact tie at every candidate-list length.

On tie-breaking: `minBy`/`minByOption` replace the incumbent only on a strictly smaller key, so they
return the first minimum in iteration order; `sorted` is documented stable, so `headOption` is also
the first minimum. Same element either way, which the `same result` column confirms empirically.

## Verification

- **`sbt c2cpg/test`**, stock and patched, identical output both times:
  `Total number of tests run: 667` / `Tests: succeeded 667, failed 0, canceled 0, ignored 0,
  pending 15` / `All tests passed.` There are no `HeaderFileFinder`-specific tests.
- **CPG equivalence**: identical on RIOT (table above) and on the synthetic C=800 tree, where nodes
  58,822, methods 6,802, calls 4,000, files 2,802 and the set of resolved headers all match. Counts
  rather than bytes, because two consecutive *stock* runs also produce different file sizes and md5s.
- **Header selection is unchanged**: a separate two-arm test that forces the fallback to choose
  between same-named headers returns the same file on stock and on patched, so the patch changes how
  the minimum is found and not which candidate wins.

max-leuthaeuser:
`HeaderFileFinder.find` selected the closest-matching header with `sortBy(...).headOption` — a full sort to take a single minimum — and since Scala's `sortBy(f)` is `sorted(ord.on(f))`, the Levenshtein key was recomputed on every comparison instead of once per candidate. On trees presenting many same-named headers (e.g. RIOT-OS with one config header per board) this dominated CPG build time; the issue reporter measured **1717.5s → 211.9s (8.1x)** on RIOT-OS, with Levenshtein evaluations dropping from 25,235,778 to 3,114,815.

### Change

```scala
nameToPathMap.getOrElse(name, List()) match {
  case Nil           => None
  case single :: Nil => Some(single)
  case matches       => matches.minByOption(candidate => Levenshtein.distance(candidate, unresolvedPath))
}
```

- `minByOption` evaluates the key once per candidate.
- The length guard keeps single-candidate lookups at zero key evaluations, matching `sorted`'s current short-circuit at length 1 — on the RIOT-OS tree 65.6% of cache misses have exactly one candidate, so a bare `minByOption` would have *added* evaluations there.
- Selection behavior is unchanged: `sorted` is stable, so `headOption` returns the first minimum in iteration order; `minByOption` replaces the incumbent only on a strictly smaller key, also returning the first minimum.

This completes #6112, which added the `findCache` memoisation (44.6% hit rate on RIOT-OS) but left the sort itself unchanged.

Credit: analysis, benchmarks, and suggested patch by @iljavs in #6280.

Fixes #6280.

### DIFF
--- a/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/parser/HeaderFileFinder.scala
+++ b/joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/parser/HeaderFileFinder.scala
@@ -31,8 +31,11 @@ class HeaderFileFinder(config: Config) {
     path,
     unresolvedPath =>
       Paths.get(unresolvedPath).nameOption.flatMap { name =>
-        val matches = nameToPathMap.getOrElse(name, List())
-        matches.sortBy(candidate => Levenshtein.distance(candidate, unresolvedPath)).headOption
+        nameToPathMap.getOrElse(name, List()) match {
+          case Nil           => None
+          case single :: Nil => Some(single)
+          case matches       => matches.minByOption(candidate => Levenshtein.distance(candidate, unresolvedPath))
+        }
       }
   )
 
--- a/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/parser/HeaderFileFinderTests.scala
+++ b/joern-cli/frontends/c2cpg/src/test/scala/io/joern/c2cpg/parser/HeaderFileFinderTests.scala
@@ -1,0 +1,54 @@
+package io.joern.c2cpg.parser
+
+import io.joern.c2cpg.Config
+import io.shiftleft.semanticcpg.utils.FileUtil.*
+import io.shiftleft.semanticcpg.utils.FileUtil
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+import java.nio.file.Files
+
+class HeaderFileFinderTests extends AnyWordSpec with Matchers {
+
+  "HeaderFileFinder" should {
+
+    "return None for a basename not present in the scanned tree" in {
+      FileUtil.usingTemporaryDirectory("headerFileFinder") { dir =>
+        Files.writeString(dir / "foo.h", "#pragma once")
+        val finder = new HeaderFileFinder(Config().withInputPath(dir.toString))
+        finder.find("/some/unresolved/path/bar.h") shouldBe None
+      }
+    }
+
+    "return the only candidate for a unique basename" in {
+      FileUtil.usingTemporaryDirectory("headerFileFinder") { dir =>
+        val header = dir / "foo.h"
+        Files.writeString(header, "#pragma once")
+        val finder = new HeaderFileFinder(Config().withInputPath(dir.toString))
+        finder.find("/some/unresolved/path/foo.h") shouldBe Some(header.toString)
+      }
+    }
+
+    "select the closest match among multiple same-named headers" in {
+      FileUtil.usingTemporaryDirectory("headerFileFinder") { dir =>
+        val configDirA = dir / "include" / "platform" / "family_0001" / "config"
+        val configDirB = dir / "include" / "platform" / "family_0002" / "config"
+        Files.createDirectories(configDirA)
+        Files.createDirectories(configDirB)
+        val headerA = configDirA / "shared_config.h"
+        val headerB = configDirB / "shared_config.h"
+        Files.writeString(headerA, "#pragma once")
+        Files.writeString(headerB, "#pragma once")
+
+        val finder = new HeaderFileFinder(Config().withInputPath(dir.toString))
+        // exact path matches resolve to themselves
+        finder.find(headerA.toString) shouldBe Some(headerA.toString)
+        finder.find(headerB.toString) shouldBe Some(headerB.toString)
+        // an unresolved path resolves to the candidate with the smallest edit distance
+        val unresolved = dir / "include" / "platform" / "family_0002" / "shared_config.h"
+        finder.find(unresolved.toString) shouldBe Some(headerB.toString)
+      }
+    }
+  }
+
+}


joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/parser/HeaderFileFinder.scala
```diff
@@ -31,8 +31,11 @@ class HeaderFileFinder(config: Config) {
     path,
     unresolvedPath =>
       Paths.get(unresolvedPath).nameOption.flatMap { name =>
-        val matches = nameToPathMap.getOrElse(name, List())
-        matches.sortBy(candidate => Levenshtein.distance(candidate, unresolvedPath)).headOption
+        nameToPathMap.getOrElse(name, List()) match {
+          case Nil           => None
+          case single :: Nil => Some(single)
+          case matches       => matches.minByOption(candidate => Levenshtein.distance(candidate, unresolvedPath))
+        }
```
maltek, line 38:
```suggestion
        nameToPathMap.getOrElse(name, List()).minByOption(Levenshtein.distance(_, unresolvedPath))
```

maltek (approved):

joern-cli/frontends/c2cpg/src/main/scala/io/joern/c2cpg/parser/HeaderFileFinder.scala
max-leuthaeuser: I do not want to do that. A singleton list would pay one Levenshtein call that the existing fix avoided. That's precisely the case the issue author added the guard for: on the RIOT-OS tree it's 468,372 extra evaluations (3.58M vs 3.11M). In wall-clock terms that's negligible (sub-second against a ~200s build), but the guarded form has the nicer property of being a strict improvement at every candidate-list length which matters for a perf PR on trees where most names are unique.

maltek: A comment to explain that would have been good then. The original issue was TL;DR. (A PR to the scala stdlib to optimize this would also be good, but I can't be arsed either.)

PR #6284, opened 2026-09-11
[joern-cli] Fix NPE in joern-parse --overlaysonly

### LINKED ISSUE #6283: joern-parse --overlaysonly: NullPointerException in applyDefaultOverlays (applyPostProcessingPasses on a generator that was never created)
### Version
Joern 4.0.623 (joern-cli), Linux, JDK from sdkman (Temurin 21).

### What I did
Apply the default overlays to an existing CPG produced by `php2cpg`, without re-running the frontend:

```
php2cpg repo/ -o cpg.bin
cp cpg.bin out.bin
joern-parse --language php --overlaysonly --output out.bin cpg.bin
```

### What happened
The overlay passes run (ReachingDefPass logs its usual warnings), then:

```
java.lang.NullPointerException: Cannot invoke "io.joern.console.cpgcreation.CpgGenerator.applyPostProcessingPasses(io.shiftleft.codepropertygraph.generated.Cpg)" because the return value of "io.joern.joerncli.JoernParse$.applyDefaultOverlays$$anonfun$1(...)" is null
	at io.joern.joerncli.JoernParse$.applyDefaultOverlays$$anonfun$1(JoernParse.scala:159)
	at scala.util.Try$.apply(Try.scala:195)
	at io.joern.joerncli.JoernParse$.applyDefaultOverlays(JoernParse.scala:163)
	at io.joern.joerncli.JoernParse$.run$$anonfun$2$$anonfun$1$$anonfun$1(JoernParse.scala:90)
	...
```

Exit code 1; the output file is left unchanged (same size as the input).

### Expected
With `--overlaysonly` there is no frontend run, so there is no `CpgGenerator` to take post-processing passes from. Either the post-processing step should be skipped (or resolved from `--language`) in that mode, or `--overlaysonly` should say it cannot be used without a generator. As it stands the flag is unusable for the case it documents ("Only apply default overlays").

### Notes
- Without `--language php` the same command fails earlier with `Could not guess language from input path ... cpg.bin`, so the language is known at the point of the NPE.
- Workaround that works: a script doing `importCpg(file); save` and taking `workspace/<name>/cpg.bin`.
- Context: we build with `php2cpg` directly (to pass `--exclude`, see #6281) and want the overlays computed once at build time rather than by every `importCpg`, because on one 9.6 MB PHP graph the reaching-definitions pass runs a 2 GB heap out of memory at every load.

max-leuthaeuser:
`joern-parse --overlaysonly` skipped CPG generation, leaving the `generator` field unset, and `applyDefaultOverlays` then crashed with a `NullPointerException` when invoking `generator.applyPostProcessingPasses(cpg)`. The flag was unusable for its documented purpose.

### Fix

When no generator was created (overlays-only mode), resolve one from the CPG's own `metaData.language` (falling back to `--language`), mirroring `Console.applyPostProcessingPasses`. Language-specific post-processing passes (e.g. PHP type recovery and call linking) still run, so `--overlaysonly` output matches a full `joern-parse` run. If no generator can be resolved at all, post-processing is skipped with a warning instead of crashing.

Fixes #6283.

### DIFF
--- a/joern-cli/src/main/scala/io/joern/joerncli/JoernParse.scala
+++ b/joern-cli/src/main/scala/io/joern/joerncli/JoernParse.scala
@@ -161,13 +161,16 @@ object JoernParse {
         // CPG's language instead, so language-specific post-processing passes are still applied.
         val postProcessingGenerator = Option(generator).orElse {
           cpg.metaData.language.headOption
-            .orElse(Option(config.language).filter(_.nonEmpty).map(_.toUpperCase))
             .flatMap(cpgGeneratorForLanguage(_, FrontendConfig(), installConfig.rootPath, args = Nil))
         }
         postProcessingGenerator match {
           case Some(resolvedGenerator) => resolvedGenerator.applyPostProcessingPasses(cpg)
           case None =>
-            System.err.println("Could not resolve a language frontend; skipping post-processing passes")
+            cpg.close()
+            throw new RuntimeException(
+              s"Could not resolve a language frontend for post-processing passes: the CPG at ${config.outputCpgFile} " +
+                "has no metadata node with a known language and appears to be broken."
+            )
         }
         cpg.close()
       }
--- a/joern-cli/src/main/scala/io/joern/joerncli/JoernParse.scala
+++ b/joern-cli/src/main/scala/io/joern/joerncli/JoernParse.scala
@@ -5,6 +5,7 @@ import io.joern.console.{FrontendConfig, InstallConfig}
 import io.joern.joerncli.CpgBasedTool.newCpgCreatedString
 import io.joern.x2cpg.frontendspecific.FrontendArgsDelimitor
 import io.shiftleft.codepropertygraph.generated.Languages
+import io.shiftleft.semanticcpg.language.*
 
 import java.nio.file.{Files, Paths}
 import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


joern-cli/src/main/scala/io/joern/joerncli/JoernParse.scala
```diff
@@ -151,12 +152,23 @@ object JoernParse {
...
+        // CPG's language instead, so language-specific post-processing passes are still applied.
+        val postProcessingGenerator = Option(generator).orElse {
+          cpg.metaData.language.headOption
+            .orElse(Option(config.language).filter(_.nonEmpty).map(_.toUpperCase))
+            .flatMap(cpgGeneratorForLanguage(_, FrontendConfig(), installConfig.rootPath, args = Nil))
+        }
+        postProcessingGenerator match {
+          case Some(resolvedGenerator) => resolvedGenerator.applyPostProcessingPasses(cpg)
+          case None =>
+            System.err.println("Could not resolve a language frontend; skipping post-processing passes")
```
maltek: exit here with error exit code. if there is no metadata node, something is completely broken.

maltek (approved):

joern-cli/src/main/scala/io/joern/joerncli/JoernParse.scala
max-leuthaeuser: Done in e95ef48f4 — the None case now throws (exit code 1 via main), and the --language fallback is gone since it would have masked exactly those broken CPGs. Added a test covering the no-metadata failure.

PR #6291, opened 2026-09-16
[rust2cpg] add line/column to macro-expanded nodes.

xavierpinho:
Macro-expanded nodes had no line/column information since they were not found in the source file. They now use their originating macro call line/column.


joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/ContextStack.scala
```diff
@@ -134,6 +145,10 @@ class ContextStack {
     stack.collectFirst { case ctx: TypeDeclContext => ctx.typeDecl.fullName }
   }
 
+  def outermostMacroCall: Option[MacroCall] = {
+    stack.collect { case ctx: MacroExpansionContext => ctx.macroCall }.lastOption
```
maltek: Using a stack to find the last element of a kind is not efficient, and I can't stop myself from saying so :). For this use case we could skip the whole stack in favor of a  `Option[(Int, MacroCall)]` variable that is cleared whenever the counter falls to zero. But I don't know if it's actually worth optimizing and will leave that decision to you.

maltek (approved):

joern-cli/frontends/rust2cpg/src/main/scala/io/joern/rust2cpg/astcreation/ContextStack.scala
xavierpinho: You're right. Ended up replacing it with a `withMacroCall` helper that internally keeps track of the outermost macro call only.

Issue #5356, opened 2025-03-09
Error when attempting to export to graphml et al


Bazmundi:
I used pysrc2cpg to parse into a cpg file a collection of blender plugins.  I was able to write a number of scala functions in the console to investigate what had been imported and all seems fine (as fine as I can deduce for a first time play).   I then tried joern-export to get to either a graphml or neo4jcsv, but get the following in message.txt.  I tried all and then only cpg as repr.

```
(base) asterion@MonstaPC:~/dev/rag/data/blender/blender-addons$ joern-export blender-addon-cpg --out ./graphmlout --repr cpg --format graphml
CPG does not have dataflow overlay. Calculating.
2025-02-16 16:45:26.311 ERROR ReachingDefPass: Pass io.joern.dataflowengineoss.passes.reachingdef.ReachingDefPass failed
flatgraph.SchemaViolationException: OUT edge with label REF to an adjacent METHOD is mandatory, but not defined for this METHOD_REF node with id=124554056509, (seq=4925)
at io.shiftleft.codepropertygraph.generated.neighboraccessors.AccessNeighborsForMethodRef$.referencedMethod$extension(MethodRef.scala:456) ~[io.shiftleft.codepropertygraph-domain-classes_3-1.7.23.jar:1.7.23]
at io.shiftleft.codepropertygraph.generated.neighboraccessors.AccessNeighborsForMethodRefTraversal$.referencedMethod$extension$$anonfun$1(MethodRef.scala:933) ~[io.shiftleft.codepropertygraph-domain-classes_3-1.7.23.jar:1.7.23]
at scala.collection.Iterator$$anon$9.next(Iterator.scala:584) ~[org.scala-lang.scala-library-2.13.14.jar:?]
at scala.collection.Iterator$$anon$10.nextCur(Iterator.scala:594) ~[org.scala-lang.scala-library-2.13.14.jar:?]
at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:608) ~[org.scala-lang.scala-library-2.13.14.jar:?]
at scala.collection.Iterator$$anon$6.hasNext(Iterator.scala:477) ~[org.scala-lang.scala-library-2.13.14.jar:?]
at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:619) ~[org.scala-lang.scala-library-2.13.14.jar:?]
[...]

PR #6293, opened 2026-09-19
[dataflowengineoss] Resolve METHOD_REF -> METHOD optionally when walking method refs

shashankkhasare:
Follow-up to #6292, related to #5356.

### Problem

`MethodRef.referencedMethod` is the strict accessor. When a `METHOD_REF` has no `REF` edge it throws

```
flatgraph.SchemaViolationException: OUT edge with label REF to an adjacent METHOD is mandatory,
but not defined for this METHOD_REF node
Caused by: java.util.NoSuchElementException: next on empty iterator
```

and that aborts the entire enclosing pass. For `ReachingDefPass` this means no data flow edges are
produced at all, and for `joern-export` it means no output files are written. A single malformed node
takes down the whole run.

#6292 fixes the case that produced such nodes in practice (`joern-export` running the data flow
overlay before the call graph overlay had created the `REF` edges). This PR is the defensive half: it
makes the traversals themselves tolerant, so that if a malformed `METHOD_REF` ever reaches them again,
from another code path or another frontend, the result is a skipped edge rather than a failed pass.

### Change

Resolve the edge optionally, using the `_refOut.collectAll[Method]` idiom already used in
`queryengine/TaskCreator`. Two call sites:

- `DdgGenerator.addEdgesToCapturedIdentifiersAndParameters`
- `PythonTypeRecovery.prepopulateSymbolTable`

This is hardening rather than a behaviour change: on a well-formed CPG both traversals yield exactly
what they did before, since every `METHOD_REF` resolves.

### Tests

`MethodRefWithoutRefEdgeTests` builds a normal CPG, removes every `REF` edge leaving a `METHOD_REF` to
reproduce the malformed state, and then asserts that `ReachingDefPass` and the Python type recovery
both complete.

Verified that both tests fail without the change, with the `SchemaViolationException` above, and pass
with it.

The Python snippet closes over a *parameter* of the enclosing function on purpose: that is what puts a
parameter into `capturedByMethodRef`, which is the traversal `DdgGenerator` walks. Without it the test
would pass vacuously either way.

The tests live in `pysrc2cpg` rather than next to `DdgGenerator` because `dataflowengineoss` has no
frontend-backed CPG fixture, and reproducing the condition needs a real closure capture. Happy to move
them if you would rather they sat elsewhere.

`sbt dataflowengineoss/test pysrc2cpg/test` passes (36 and 571 tests). `sbt scalafmt Test/scalafmt`
reports no changes.

### Note

The two call sites are independent, so I am happy to split this into one PR per site if that is
easier to review.


max-leuthaeuser:
> MethodRef.referencedMethod is the strict accessor. When a METHOD_REF has no REF edge it throws

Yes, and thats the desired behavior. A ref without a referenced method indicates a frontend issue and needs to be fixed purely at the frontend side (or in the case from your referenced issue by running the correct pass). Your changes would hide that.

